Issue #693, opened 2014-12-19
Context Broker crash in some cases of startup with existing ONINTERVAL subscriptions in database


fgalan:
A dump of the DB that makes CB crash has been distributed among developers.

fgalan:
Fixed in PR #697

PR #697, opened 2014-12-23
Bug/693 ontimeinterval subscription recovery

kzangeli:
Fixed the bug of issue #693.

This fix will probably also solve the mysterious case of 'false leaks' we've been seeing (and trying to fix).

Instead of creating a vector of subscriptions, a callback function (treating the subscription) is sent to the function that loops over the subscriptions.

The bug in the previous implementation was that the vector got corrupted as soon as the function that created the vector finished (symptom telling us we had stuff on the stack - in this case the BSON objects inside the vector).

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -60,6 +60,14 @@ using std::auto_ptr;
 
 /* ****************************************************************************
 *
+* OtisTreatFunction - callback signature for treatOntimeintervalSubscriptions
+*/
+typedef void (*OtisTreatFunction)(std::string tenant, BSONObj* bobjP);
+
+
+
+/* ****************************************************************************
+*
 * Globals
 */
 static DBClientBase*        connection;
@@ -454,13 +462,13 @@ void ensureLocationIndex(std::string tenant) {
 
 /* ****************************************************************************
 *
-* getOntimeIntervalsThreads -
+* treatOnTimeIntervalSubscriptions -
 */
-static void getOntimeIntervalsThreads(std::string tenant, std::vector<BSONObj>& subs)
+static void treatOnTimeIntervalSubscriptions(std::string tenant, OtisTreatFunction treatFunction)
 {
-  /* Look for ONTIMEINTERVAL subscriptions in database */
-  std::string condType= CSUB_CONDITIONS "." CSUB_CONDITIONS_TYPE;
-  BSONObj query = BSON(condType << ON_TIMEINTERVAL_CONDITION);
+  /* Look for ONTIMEINTERVAL subscriptions in the database */
+  std::string condType = CSUB_CONDITIONS "." CSUB_CONDITIONS_TYPE;
+  BSONObj     query    = BSON(condType << ON_TIMEINTERVAL_CONDITION);
 
   DBClientBase* connection = getMongoConnection();
   auto_ptr<DBClientCursor> cursor;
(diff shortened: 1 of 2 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -495,92 +503,87 @@ static void getOntimeIntervalsThreads(std::string tenant, std::vector<BSONObj>&
...
 
-    std::string  subId = idField.OID().str();
+/* ****************************************************************************
+*
+* destroyOnTimeIntervalThread - 
+*/
+static void destroyOnTimeIntervalThread(std::string tenant, BSONObj* subP)
+{
+  BSONElement  idField = subP->getField("_id");
+  std::string  subId   = idField.OID().str();
```
fgalan: Why don't apply "Paranoia check (I)" also in that place for subId?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -60,6 +60,14 @@ using std::auto_ptr;
 
 /* ****************************************************************************
 *
+* OtisTreatFunction - callback signature for treatOntimeintervalSubscriptions
+*/
+typedef void (*OtisTreatFunction)(std::string tenant, BSONObj* bobjP);
```
fgalan: Why don't using references instead pointers? I.e.

```
typedef void (*OtisTreatFunction)(std::string tenant, BSONObj& bobjP);
```
I think this was the common "style" that we are using in general (except in some exceptions) along the code.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Well, those checks were there already and I decided to let them stay.
I can put one in here too if you want :-) (probably not necessary but it doesn't hurt)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I just don't have it in my blood.
Never liked it.
I saw somewhere that C++ references should be used only when 'const XXX&' - when the object passed in is 'read-only'.
Let's look it up.

kzangeli: http://stackoverflow.com/questions/7058339/c-when-to-use-references-vs-pointers

It seems Google agrees with me :-)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in afcf920

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: The answers to that question say:

> Use reference wherever you can, pointers wherever you must.
> Avoid pointers until you can't.
> So the rule of thumb is to use pointers only if there is no other choice. 

In this case, is there any reason that force to use a pointer? Moreover, given that the BSON object passed to the function is not modified (as far as I remember) in the recovery/destroy logic, probably a `conts BSONObj&` could be used.

kzangeli: The crucial point is, is the object modified or not?
This is a callback, and a 'callbacker' shouldn't know whether the object will be modified or not.

However, this is entirely up to us.
I find it confusing (just as they explain) to send what seems to be a copy of an object on the stack, and later find that the 'original' object has been changed.
I can see that you want to use this technique, so I'll just change it to a C++ reference.
No reason to waste any more time on this.

kzangeli: Sorry, I can't.
The parameter is a function (had forgotten about that).
A function is a pointer by definition.

NTC

kzangeli: Fixed in 6aa7d00

fgalan:
LGTM

Issue #690, opened 2014-12-18
context broker crashes with invalid geoquery


fgalan:
Reported by @PabloCGarcia (see https://github.com/telefonicaid/fiware-orion-pep/issues/97):

When we create a syntactically correct query but that is semantically incorrect, the broker crashes. We created the following query that combines a geo restriction of kind polygon and, in the same query, a restriction of kind circle, then the broker crashes.

The query is:
```
    jsonPoligono = '"restriction": {"scopes": [{ "type" : "FIWARE::Location","value" : {"polygon": {'+\
                   '"vertices": [' +\
                   '{"latitude": "-33.603458","longitude": "-70.652788"},' +\
                   '{"latitude": "-33.604231","longitude": "-70.646032"},' +\
                   '{"latitude": "-33.606123","longitude": "-70.646781"},' +\
                   '{"latitude": "-33.606123","longitude": "-70.646781"}' +\
                   ']}}},'
    jsonCirculo = '"{ "type" : "FIWARE::Location","value" : {"circle": {'+\
                  '"centerLatitude": "-33.602391",'+\
                  '"centerLongitude": "-70.661781",'+\
                  '"radius": "200"'+\
                  '}}}]}'
```
    jsonX='{ "entities": [{"type": "","isPattern": "true","id": ".*"}],' + jsonPoligono + jsonCirculo+ '}'

    resp = get_post_results("/v1/queryContext", jsonX)
    for x in resp.iteritems():
        print "Elemento:", x[0]
        print "Atributos:"
        for y in x[1]:
            print "    Name:", y["name"], "Type:", y["type"], "Value:", y["value"]

fgalan:
Fixed in PR #698

PR #698, opened 2014-12-29
Bug/690 geoquery crash

kzangeli:
### Description
Fixes the crash-bug described in issue #690, by simply adding a try/catch. [ I am sure similar protections are necessary in *many* places in mongoBackend - we probably even have an issue about it ]

### NOTE
Before this PR is even looked at, the PR #697 must be merged and this branch (bug/690_geoquery_crash) merged with develop.


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1123,7 +1126,20 @@ bool entitiesQuery
 
         /* Attributes part */
 
-        std::vector<BSONElement> queryAttrV = r.getField(ENT_ATTRS).Array();
+        std::vector<BSONElement> queryAttrV;
```
fgalan: I think this should be solved in a different way. Note that, by definition, any document in the enties collection has an attributes array, thus the try/catch should not be needed if we capture the error in a early place.

In particular, looking to the logs after executing the .test included in this PR I have found this:

```
time=2015-01-02T19:31:16.884CET | lvl=DEBUG | trans=1420223475-825-00000000001 | function=entitiesQuery | comp=Orion | msg=MongoGlobal.cpp[1108]: retrieved document: '{ $err: "Exterior shell of polygon is invalid: { type: "Polygon", coordinates: [ [ [ -70.652788, -33.603458 ], [ -70.64603200000001, -33.604231 ], [ -70.646781...", code: 16693 }'
```
Thus probably a better strategy is to check if the "$err" field exists in the `r` document just after `LM_T(LmtMongo, ("retrieved document ...`) and, in that case, return the proper error response to the user (I'd suggest a 500 Internal Error, using the text in "$err" to fill the details field). I think that is better than a "No context element found" (as in the current version in the code) which may be hidding structural problems in geoquery done at DB.

That way we would avoid Orion to crash. However, we can do even better detecting those "Exterior shell of polygon is invalid" situation before interacting with DB, fixing them if possibe or reporting the user with a 400 Bad Input error about the problem if the situation is not easy to fix. Looking to http://stackoverflow.com/questions/21870088/exterior-shell-of-polygon-is-invalid, it seems they happen when the polygon have duplicated coordinates (this seems to be easy to detect) or is an "intersection polygon" (I'm not sure if al algorithm to reorder coordinates is easy to implement). 

Maybe the 500 Internat Error solution is enough for this PR. If detecting and fixing "Exterior shell of polygon is invalid" situations is difficult, we can defer it to a tech debt issue (nicetohave?).

fgalan: BTW, maybe we should apply the strategy of detecting the "$err" after querying in all the places in which we use `connection->query()`. Probably not as part of this PR (except in the case it doesn't involve to much work), but opening and tech debt issue about it.

test/functionalTest/cases/690_geoquery_crash/two_geoscopes.test
```diff
@@ -0,0 +1,112 @@
...
+              { "latitude": "-33.603458", "longitude": "-70.652788" },
+              { "latitude": "-33.604231", "longitude": "-70.646032" },
+              { "latitude": "-33.606123", "longitude": "-70.646781" },
+              { "latitude": "-33.606123", "longitude": "-70.646781" }
+            ]
+          }
+        }
+      },
+      {
+        "type" : "FIWARE::Location",
```
fgalan: At the end, this second scope with a circle is not relevant for the bug and associated fix. Thus, I'd suggest to remove it.

kzangeli: Fixed in 006b090.
However, I forgot to rename the file ...

kzangeli: Renamed in 6a132fc

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in  09f7218 and 006b090 (two commits as I had to commit to switch to the branch of another PR)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1110,11 +1110,40 @@ bool entitiesQuery
...
+          if (strncmp(err.c_str(), invalidPolygon, strlen(invalidPolygon)) == 0)
+            err = invalidPolygon + std::string(" - not detected at creation time");
+          else
+            err = defaultErrorString;
+
+          //
+          // FIXME P1: It would be nice to fill in the entity but it is
+          //           difficult to do this, I simply fill it in with enV[0] ...
+          //
+          cer->contextElement.entityId.fill(enV[0]);
```
kzangeli: When detecting an error, it is nice to return an entity to add information to the error.
Here the first entity from the vector is returned and if the request had only one entity, this is fine.
However ...
I'm not sure there is a better way to do this.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1110,11 +1110,40 @@ bool entitiesQuery
...
+    while (cursor->more())
+    {
+        BSONObj                 r    = cursor->next();
+        std::string             err  = r.getStringField("$err");
+        ContextElementResponse* cer  = new ContextElementResponse();
+
+        if (err != "")
+        {
+          //
+          // We can't return the error 'as is', as it may contain forbidden characters.
```
fgalan: Good point. However, two things to take into account:

1. We have a rendering logic that scape forbidden characteres, even if they are produced by a "third party" as the DB.
2. The exact message that MongoDB reports as error could vary from version to version (e.g. from MongoDB 2.4 to MongoDB 2.6), which could cause this mechanism change and some confusion.

Having said that, I'm not sure what is the best mechanism (either as it is now in the code or just passing the message returned by MongoDB)...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1110,11 +1110,40 @@ bool entitiesQuery
...
+          // So, we can just match the error and sent a less descriptive text
+          //
+          const char* invalidPolygon      = "Exterior shell of polygon is invalid";
+          const char* defaultErrorString  = "Internal mongo problem";
+
+          LM_W(("Database Error (%s)", err.c_str()));
+
+          if (strncmp(err.c_str(), invalidPolygon, strlen(invalidPolygon)) == 0)
```
fgalan: Remember using `{... }` even for single-line if clauses.

In addition, not sure about the "not detected at creation time" message appending... On the one side is like saying that the CB should have detected something but it has failed in doing so (what is not correct, as the CB is not designed to detect "exterior shell problems" so far). On the other side "creation" is a bit confusing... as the subject of the creation is not clear (an entity? a polygon -note that the polygon itself is not being created, but the structure to query for a polygon, which is not exactly the same).

Whatever we use at the end, I think is a good idea to create a tech debt issue (if you haven't already do so) to enhance the CB to detect these "exterior shell problems"

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: In this case, the problem in a general case is not actually with one entity or the other, but with the query send to mongo, that has "mixed" information of all the entities in the request. Thus, I only see two possible options:

1. Return all the entities
2. Don't return any entity (NGSI response format allows this?)

In my opinion 2 is better (simpler code and shorter response to the client). There is no loss of information, as the client already knows the entities that it included in the requesz producing that response.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1110,11 +1110,40 @@ bool entitiesQuery
...
+        ContextElementResponse* cer  = new ContextElementResponse();
+
+        if (err != "")
+        {
+          //
+          // We can't return the error 'as is', as it may contain forbidden characters.
+          // So, we can just match the error and sent a less descriptive text
+          //
+          const char* invalidPolygon      = "Exterior shell of polygon is invalid";
+          const char* defaultErrorString  = "Internal mongo problem";
```
fgalan: I'd suggest this message:

"Error got querying MongoDB"

or similar (probably the above wording is not good English :)

I don't like "Internal mongo problem" two much due to: 1) "mongo" is ambiguous (it may refer to the mongo shell, which binary names the same, MongoDB is prefereable if we refer to the DB); 2) internal mongo problem could be interpreted as a problem in the MongoDB itself, which is not the case.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1134,7 +1163,20 @@ bool entitiesQuery
 
         /* Attributes part */
 
-        std::vector<BSONElement> queryAttrV = r.getField(ENT_ATTRS).Array();
+        std::vector<BSONElement> queryAttrV;
+
+        try
```
fgalan: The early detection of $err makes this try/catch mostly unnecessary. However, I think we should keep it, as it protects from weird out-of-band manipulation of the DB (e.g. somebody modifying entites documentes removing the `attrs` field) with some modifications:

1. Add a comment explaining why the try/catch is here, as we could easily forget its purpose. Something like this: "This try/catch ward actually is not needed (as every document in the entities collections has an attrs array from creation time) if we assume that out-of-band manipulation of the DB will not happen. However, it adds an extra protection measure in the case of such manitpulations could happen, due to unknown reasons" or something similar.
2. If the problem is raised, it is a serious thing that need to be researched. Thus, I'd add a LM_E in the catch clause to mark this and probably also I'd return a 500 Internal Error instead of SccContextElementNotFound.

kzangeli: Fixed in 837c83b

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: We are inside entitiesQuery, so it is not clear which ngsi response is in question.
We can always return an empty Entity.

However, if there is *only one* entity in the request, we can return that one (as is done right now) and if there is more than one, we return an empty entity.

With an empty entity, I mean an entity with id == "", type == "", isPattern == "".

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: What I looked for with that 'not detected at creation time' was to inform that the error isn't here, but was incorrectly inserted that way earlier. But you are right, it may be misinterpreted in more than one way.
I'll remove it.

About changing strings for different versions etc.
Don't worry about that, we just add another string to check against whenever necessary. If the string is not found, we just return the default value (right now: "Internal mongo problem" - which I will change).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1134,7 +1163,27 @@ bool entitiesQuery
...
+        // Better to be on the safe side ...
+        //
+        try
+        {
+          queryAttrV = r.getField(ENT_ATTRS).Array();
+        }
+        catch (...)
+        {
+          LM_E(("Database Error (no attrs array in document of entities collection)"));
+          cer->statusCode.fill(SccReceiverInternalError);
```
fgalan: Maybe we should use some details string to explain what had happened, e.g. "attrs field missing in entity document".

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I tried to return the value 'as is' and it failed. That's why I invented this mechanism to return something similar, at least.

NTC?

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
 Fix:  Fixed the bug about recovering ONTIMEINTERVAL subscriptions (Issue #693)
+Fix:  Fixed the crash in queryContext with a double geoscope (Issue #690)
```
kzangeli: This text has to be modified also, the problem wasn't (as you detected) having a double geoscope, but that the polygon was not closed.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: What do you mean by "failed"? That the forbidden characteres didn't get escaped?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: OK
Error at querying MongoDB

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in fa12c91

CHANGES_NEXT_RELEASE
fgalan: I'd suggest:

"Fix: Fixed the crash in queryContext due to polygon external shell situations with geoscope (Issue #690)"

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: sure

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I can reproduce it (can't remember the details)

CHANGES_NEXT_RELEASE
kzangeli: Well, the fix is valid for any invalid geoscope, not just this exact failure, right?
The 'exact' info is in the issue.
I changed the text to (simply):

Fixed the crash in queryContext with an invalid geoscope (Issue #690)

If you prefer your text, I have no problem in changing it

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 9c18e71

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 9c18e71

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 9c18e71

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 9c18e71 (awaiting your approval)

kzangeli:
Some func test will fail, due to textual changes at error.
Do not merge just yet!
I have 'make test' running ...

CHANGES_NEXT_RELEASE
fgalan: The new text you are using describes perfectly the change, so it's fine.

fgalan:
LGTM (not merging until @kzangeli 's func test finished rightly)

kzangeli:
With 71d3255, the PR is complete.

PR #700, opened 2015-01-07
Bug/699 context provider responses without payload

kzangeli:
### Description
Made the broker not crash when an empty string is input to the XML/JSON parsing routines.
"OK" is returned on empty payload. Not sure this is the best approach, as the function shouldn't
be called on empty payload and yet there is no real 'generic' error sending a request without payload ...  
Two short unit tests to test this out as it is **very** difficult to provoke this failure 'from the outside', using a functional test.

### DIFF
--- a/src/lib/jsonParse/jsonRequest.cpp
+++ b/src/lib/jsonParse/jsonRequest.cpp
@@ -171,6 +171,13 @@ std::string jsonTreat
   std::string   res   = "OK";
   JsonRequest*  reqP  = jsonRequestGet(request, ciP->method);
 
+
+  //
+  // If the payload is empty, the XML parsing library does an assert
+  // and the broker dies. I don't know about the JSON library, but just in case ...
+  // 
+  // 'OK' is returned as there is no error to send a request without payload.
+  //
   if ((content == NULL) || (*content == 0))
   {
     return "OK";
--- a/src/lib/jsonParse/jsonRequest.cpp
+++ b/src/lib/jsonParse/jsonRequest.cpp
@@ -171,6 +171,11 @@ std::string jsonTreat
   std::string   res   = "OK";
   JsonRequest*  reqP  = jsonRequestGet(request, ciP->method);
 
+  if ((content == NULL) || (*content == 0))
+  {
+    return "OK";
+  }
+
   LM_T(LmtParse, ("Treating a JSON request: '%s'", content));
 
   ciP->parseDataP = parseDataP;
--- a/src/lib/serviceRoutines/postQueryContext.cpp
+++ b/src/lib/serviceRoutines/postQueryContext.cpp
@@ -178,10 +178,12 @@ std::string postQueryContext
   {
     QueryContextResponse qcrs;
 
-    // FIXME P1: What StatusCode should be used here?
-    qcrs.errorCode.fill(SccReceiverInternalError, "invalid context provider response");
+    //
+    // This is really an internal error in the Context Provider
+    // It is not in the orion broker though, so 404 is returned
+    //
+    qcrs.errorCode.fill(SccContextElementNotFound, "invalid context provider response");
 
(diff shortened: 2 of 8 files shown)


src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -173,6 +174,19 @@ std::string postQueryContext
 
   cleanPayload = xmlPayloadClean(out.c_str(), "<queryContextResponse>");
 
+  if ((cleanPayload == NULL) || (cleanPayload[0] == 0))
+  {
+    QueryContextResponse qcrs;
+
+    // FIXME P1: What StatusCode should be used here?
+    qcrs.errorCode.fill(SccReceiverInternalError, "invalid context provider response");
```
fgalan: I don't think this should be a 500 Internal Error. In general, fails in the connection with a CPr should be resolved as if there wasn't any CPr registered for the given context element, i.e. 404 Not Context Element found, from the perspective of the client. Of course, from an "internal" perspective it's perfectly ok (as you do some lines below) logging it as warning message in the log.

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -188,6 +188,19 @@ std::string postUpdateContext
 
     cleanPayload = xmlPayloadClean(out.c_str(), "<updateContextResponse>");
 
+    if ((cleanPayload == NULL) || (cleanPayload[0] == 0))
+    {
+      UpdateContextResponse ucrs;
+
+      // FIXME P1: What StatusCode should be used here?
```
fgalan: I don't think this should be a 500 Internal Error. In general, fails in the connection with a CPr should be resolved as if there wasn't any CPr registered for the given context element, i.e. 404 Not Context Element found, from the perspective of the client. Of course, from an "internal" perspective it's perfectly ok (as you do some lines below) logging it as warning message in the log.

kzangeli: ok, sure.  It is a strange situation. It is a 500, but not in the broker, but in the context provider.
A user might not know about context providers and see (orion + context providers) as a black box.
This reasoning led me to return a 500 (+ the FIXME about which code to use :-))

src/lib/xmlParse/xmlRequest.cpp
```diff
@@ -208,6 +208,11 @@ std::string xmlTreat
   xml_document<>  doc;
   char*           xmlPayload = (char*) content;
 
+  if ((content == NULL) || (*content == 0))
```
fgalan: As discussed by skype, include an explanatory comment here.

src/lib/jsonParse/jsonRequest.cpp
```diff
@@ -171,6 +171,11 @@ std::string jsonTreat
   std::string   res   = "OK";
   JsonRequest*  reqP  = jsonRequestGet(request, ciP->method);
 
+  if ((content == NULL) || (*content == 0))
```
fgalan: As discussed by skype, include an explanatory comment here.

kzangeli: Fixed in 5218a38

src/lib/xmlParse/xmlRequest.cpp
kzangeli: Fixed in 5218a38

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 5218a38

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 5218a38

fgalan:
LGTM

PR #711, opened 2015-01-16
FIX git push command

fgalan:
Fix in a previous PR. I though that `--tag` means "only push the tags" not "push the tags in addition to the branch, as usual".

Thus, the "git push" command shouldn't be commented out compeltely but adjusted.

@dmoranj @alvarolmedo


alvarolmedo:
LGTM

Issue #703, opened 2015-01-08
Request is forwarded with JSON payload and XML MIME Type


dmoranj:
When a Context Provider is created with the following request:
```
POST /NGSI9/registerContext HTTP/1.1
fiware-service: smartGondor
fiware-servicepath: /gardens
host: 192.168.56.101:1026
accept: application/json
content-type: application/json
content-length: 250
Connection: keep-alive

{"contextRegistrations":[{"entities":[{"type":"Light","isPattern":"false","id":"TestClient:Light"}],"attributes":[{"name":"luminescence","type":"Lumens","isDomain":"false"}],"providingApplication":"http://192.168.56.1:4041/NGSI10"}],"duration":"P1M"}
```

and the followign `updateContext` is sent:
```
POST /NGSI10/updateContext HTTP/1.1
fiware-service: smartGondor
fiware-servicepath: /gardens
host: 192.168.56.101:1026
accept: application/json
content-type: application/json
content-length: 175
Connection: keep-alive

{"contextElements":[{"type":"Light","id":"TestClient:Light","isPattern":"false","attributes":[{"name":"luminescence","type":"integer","value":"18"}]}],"updateAction":"UPDATE"}
```

The request forwarded to the context provider arrives with the following headers:
```
    "user-agent": "orion/0.18.0-next libcurl/7.19.7",
    "host": "192.168.56.1:4041",
    "fiware-service": "smartgondor",
    "fiware-servicepath": "/gardens",
    "accept": "application/xml, application/json",
    "content-length": "319",
    "content-type": "application/xml"
```
and the following body:
```
{
  "contextElements" : [
    "contextElement" : {
      "type" : "Light",
      "isPattern" : "false",
      "id" : "TestClient:Light",
      "attributes" : [
        {
          "name" : "luminescence",
          "type" : "integer",
          "value" : "18"
        }
      ]
    }
  },
  "updateAction" : "UPDATE"
}
```

when they should arrive either in XML with `application/xml` MIME Type or in JSON with `application/json` MIME Type.

(The first two requests were sniffed directly from the Context Broker machine with tshark, while the forwarded request was read from the debugger in the client server).

fgalan:
This is related with https://github.com/telefonicaid/fiware-orion/issues/722. Once #722 gets solved, probably the present issue gets solved also.

kzangeli:
The bug of mixing mime type and payload was fixed in PR #721.
Being able to forward in either XML or JSON was fixed in PR #899.

PR #721, opened 2015-01-28
Bug/703 mime types in forwarding requests

kzangeli:
### Description

Forced the rendering of forwarded messages to be in XML.
Fixes issue #703

### DIFF
--- a/src/lib/serviceRoutines/postUpdateContext.cpp
+++ b/src/lib/serviceRoutines/postUpdateContext.cpp
@@ -136,9 +136,15 @@ std::string postUpdateContext
 
 
     //
-    // 3. Render an XML-string of the request we want to forward
+    // 3. Render a string of the request we want to forward, forced to XML
+    //    FIXME P8: The format of this string (XML or JSON) should depend on the
+    //              format of the incoming message, but right now we force it to XML.
     //
-    std::string payloadIn = ucrP->render(ciP, UpdateContext, "");
+    ConnectionInfo ci;
+
+    ci.outFormat = XML;
+
+    std::string payloadIn = ucrP->render(&ci, UpdateContext, "");
 
     LM_T(LmtCtxProviders, ("payloadIn:\n%s", payloadIn.c_str()));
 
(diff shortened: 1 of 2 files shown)


src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -136,9 +136,15 @@ std::string postUpdateContext
 
 
     //
-    // 3. Render an XML-string of the request we want to forward
+    // 3. Render a string of the request we want to forward, forced to XML
+    //    FIXME P8: The format of this string (XML or JSON) should depend on the
+    //              format of the incoming message, but right now we force it to XML.
```
fgalan: In fact, according to our skype discussion it should be:

"The format of this string (XML or JSON) should depend on the CB-CPr content type negotiation, but right now we force it to XML."

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -136,9 +136,15 @@ std::string postUpdateContext
...
     //
-    std::string payloadIn = ucrP->render(ciP, UpdateContext, "");
+    ConnectionInfo ci;
+
+    ci.outFormat = XML;
+
+    std::string payloadIn = ucrP->render(&ci, UpdateContext, "");
```
fgalan: What about forwarded queries? It seems that the PR only solves the updateContext case...

kzangeli: The queries already force XML.  The bug was in Update, using ciP->outFormat from the 'original' request

fgalan: NTC

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yes, that is better (the string is older that the conversation).
Fixed in f4d581f

fgalan:
LGTM

PR #723, opened 2015-01-28
Debian 8.0 support

AlvaroVega:
Tested with g++ 4.8 and g++4-9


src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -100,7 +100,7 @@ HttpStatusCode mongoSubscribeContext
     bool notificationDone = false;
     BSONArray conds = processConditionVector(&requestP->notifyConditionVector,
                                              requestP->entityIdVector,
-                                             requestP->attributeList, oid.str(),
+                                             requestP->attributeList, oid.toString(),
```
fgalan: Just curiosity... what is the diference between using `str()` and `toString()` ?

AlvaroVega:
``` C++
class OID {
...
        std::string str() const { return toHexLower(data, kOIDSize); }
        std::string toString() const { return str(); }
....
```
It's just an alias, but in some cases is more portable because does not matches with str operator.
Some months ago I've got trouble to compile that 4.8.2, but now it work with 4.8.3 :S

fgalan: Ok, I will try to remember that and use `toString()` from now on :)

NTC

fgalan:
LGTM

LeandroGuillen:
LGTM

fgalan:
Wating for jenkins to be idle in the Orion pipeline to merge this PR. First one in queue :)

Issue #725, opened 2015-01-29
CB doesn't honor "-multiservice"


fgalan:
In Orion 0.18.1 (and probably older) the `-multiservice`flag is not taken into account. The CB runs in multiservice mode always, no matter if this flag is used or not.

fgalan:
Fixed in PR #726 (release/0.18.1) and #728 (develop)

PR #726, opened 2015-01-29
Bug/725 multiservice arg ignore

fgalan:
Fixes bug #725

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: always run in multiservice mode, ignoring -multiservice (#725)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,1 @@
-- Fix: always run in multiservice mode, ignoring -multiservice (#725)
+- Fix: The option '-multiservice' was ignored and the broker ran always in multiservice mode. Now the option is taken taken into account (Issue #725)
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -80,6 +80,7 @@ static std::string          subscribeContextCollectionName;
 static std::string          subscribeContextAvailabilityCollectionName;
 static std::string          assocationsCollectionName;
 static Notifier*            notifier;
+static bool                 multitenant;
 
 /* ****************************************************************************
 *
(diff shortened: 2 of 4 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -80,6 +80,7 @@ static std::string          subscribeContextCollectionName;
 static std::string          subscribeContextAvailabilityCollectionName;
 static std::string          assocationsCollectionName;
 static Notifier*            notifier;
+static bool                 multitenant;
```
kzangeli: We already have a global variable 'broker wide', called 'mtenant'. Why not use it?

test/functionalTest/cases/725_multiservice_ignore/service_ignore_without_multiservice.test
```diff
@@ -0,0 +1,103 @@
...
+echo
+echo
+
+echo "2. Check that the number of orion databases is just 1"
+echo "====================================================="
+# In this case mongoCmd helper is not used, due to we don't have
+# a particular database for mongo command and the output could be
+# multiline in a general case
+echo "show dbs" | mongo | grep ${CB_DATABASE_NAME} | wc -l
+
```
kzangeli: So, I imagine that here you want to prove that "orion-foo" hasn't been created.

Is the DB 'orion' created even before any content has been sent?
I mean, there is a possibility that *only orion-foo* exists here. It depends on the implementation ...

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix: always run in multiservice mode, ignoring -multiservice (#725)
```
kzangeli: A quick look at this line (without examining the issue 725), will make our users believe that the broker now always runs in multiservice mode.
Not exactly what you have fixed ...

Better comment:
Fix: The option '-multiservice' was ignored and the broker ran always in multiservice mode. Now the option is taken taken into account (Issue #725)

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Because of encapsulation. I'd prefer all the variables used in mongoBackend to be keep in mongoBackend. Thus, we use `multitenant` in mongoBackend, initialized with the value of `mtenant`.

test/functionalTest/cases/725_multiservice_ignore/service_ignore_without_multiservice.test
fgalan: Not sure if I follow your argumentation...

No "ftest*" database should exists at the begining of the tests. This is a general assumption each time `make ft` is called. And, if that happens, this test will work ok. Or maybe I'm missing something?

CHANGES_NEXT_RELEASE
fgalan: Fixed in 8e78913

src/lib/mongoBackend/MongoGlobal.cpp
NTC

test/functionalTest/cases/725_multiservice_ignore/service_ignore_without_multiservice.test
fgalan: Fixed in 32ac155

kzangeli:
LGTM

PR #729, opened 2015-01-29
FIX explicit default service path in code

fgalan:
This PR doesn't merge in release/0.18.1 as in the IoTP version Fiware-ServicePath is always use, so this hardening is not needed.

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -224,14 +224,6 @@ static int httpHeaderGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, co
   else
     LM_T(LmtHttpUnsupportedHeader, ("'unsupported' HTTP header: '%s', value '%s'", ckey, value));
 
-  /* If servicePath was not provided, then use the default value */
-  if (!headerP->servicePathReceived)
-  {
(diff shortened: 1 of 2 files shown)


src/lib/rest/rest.cpp
```diff
@@ -224,6 +224,13 @@ static int httpHeaderGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, co
   else
     LM_T(LmtHttpUnsupportedHeader, ("'unsupported' HTTP header: '%s', value '%s'", ckey, value));
 
+  /* If servicePath was not provided, then use the default value */
```
fgalan: Not sure if in this case we should add an entry into CHANGES_NEXT_RELEASE. For an user point of view, the change doesn't involves an actual change. Maybe something like this could be useful:

```
- Default service path "/" is set explicitely in the code, which may fix some weird behavours when this header was omitted.
```

What do you think?

kzangeli: I would not mention this in a public place (except here :-))

fgalan: Ok. NTC.

kzangeli:
LGTM

kzangeli:
LGTM II

PR #730, opened 2015-01-30
Hardening/explicit default service path

fgalan:


CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 - Fix: The option '-multiservice' was ignored and the broker ran always in multiservice mode. Now the option is taken taken into account (Issue #725)
 - Fix: Forwarded requests always in XML (Issue #703)
+- Fix: internal substitution of servicePath defaults ("/" for update-like operations, "/#" for query-like operations)
```
fgalan: At the end, I think is good idea to have the entry. I doesn't hurt and may be useful to explain further problems related with this.

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -781,7 +781,7 @@ std::string servicePathSubscriptionRegex(const std::string servicePath, std::vec
 
   if (spathComponents == 0)
   {
-    spathRegex = "^$|^\\/#$|^\\/$";  // FIXME NOW: Including empty service path also, not sure it's correct
```
fgalan: This hasn't to do with this PR but I have spoted looking to code. I think is an old leftover, so I removed it.

NTC

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+   *  registerContext
+   *  updateContext
+   *  queryContext
+   *
+   * Note that in the case of unsubscribeContext and unsubscribeContextAvailability doesn't
+   * actually use a servicePath, but we also provide a default in that case for the sake of
+   * completeness
+   */
+
+  // FIXME P5: this strategy can be improved taking into account full URL. Otherwise, it is
```
fgalan: If you think the function can work in a better way, as suggest this FIXME (you know C-based string comparations techniques better than me :) easy to implemnt, tell me so. Otherwise, mark this with NTC and let it for the future.

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+  // ambigous (e.g. entity which id is "queryContext" and are using a conv op). However, it is
+  // highly improbable that the user uses entities which id (or type) match the name of a
+  // standard operation
+
+  // FIXME P5: unhardwire literals
```
fgalan: Maybe we have defines for "updateContextAvailabilitySubscription" and so in other place in the code? Probably in contextBroker.cpp but not sure if in a .h easy to import here.

src/lib/rest/rest.cpp
```diff
@@ -714,9 +773,12 @@ static int connectionTreat
     }
     LM_T(LmtUriParams, ("notifyFormat: '%s'", ciP->uriParam[URI_PARAM_NOTIFY_FORMAT].c_str()));
 
-    /* In the case an actual Fiware-ServicePath is found, the default will be overwritten */
-    ciP->httpHeaders.servicePath = DEFAULT_SERVICE_PATH;
     MHD_get_connection_values(connection, MHD_HEADER_KIND, httpHeaderGet, &ciP->httpHeaders);
+    if (!ciP->httpHeaders.servicePathReceived)
```
fgalan: Running this after the MHD_get_connection_values() instead of before seems to work and save a function call (and the lot of ifs inside that fucntion). I understand that when MHD_get_connection_values() returns, all the HTTP headers has been processed.

If you agree with this, mark it with NTC.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -78,6 +79,18 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
...
+    // FIXME P8: the stuff about atLeastOneNotDefault was added after PR #729, which makes "/" the default servicePath in
+    // request not having that header. However, this causes as side-effect that a
+    // "Fiware-ServicePath: /" or "Fiware-ServicePath: /,/" header is added in notifications, thus breaking several tests harness.
+    // Given that the "clean" implementation of Fiware-ServicePath propagation will be implemented
+    // soon (it has been scheduled for version 0.19.0, see https://github.com/telefonicaid/fiware-orion/issues/714)
+    // we introduce the atLeastOneNotDefault hack. Once #714 gets implemented,
+    // this FIXME will be removed (and all the test harness adjusted, if needed)
+    if (!atLeastOneNotDefault)
+    {
+      spathList = "";
```
kzangeli: ok, but a lot of negations here. 
It takes some time to understand this.
Let's make it easier to read when the FIXME is removed.

NTC

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+  // FIXME P5: this strategy can be improved taking into account full URL. Otherwise, it is
+  // ambigous (e.g. entity which id is "queryContext" and are using a conv op). However, it is
+  // highly improbable that the user uses entities which id (or type) match the name of a
+  // standard operation
+
+  // FIXME P5: unhardwire literals
+  if      (strcasestr(url, "updateContextAvailabilitySubscription") != NULL) return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "unsubscribeContextAvailability") != NULL)        return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "subscribeContextAvailability") != NULL)          return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "discoverContextAvailability") != NULL)           return DEFAULT_SERVICE_PATH_RECURSIVE;
```
fgalan: I'm using case insensitive comparison with strcasestr(), instead of strstr(). Strictly speaking, the URL includes case, but I don't think we could have in the future different operations for e.g. "discoverContextAvailability" and "discovercontextavailability. Thus, I think is ok doing this way.

If you agree, mark this as NTC.

src/lib/rest/rest.cpp
kzangeli: NTC

src/lib/rest/rest.cpp
kzangeli: NTC

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+  else if (strcasestr(url, "updateContextSubscription") != NULL)             return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "unsubscribeContext") != NULL)                    return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "subscribeContext") != NULL)                      return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "registerContext") != NULL)                       return DEFAULT_SERVICE_PATH;
+  else if (strcasestr(url, "updateContext") != NULL)                         return DEFAULT_SERVICE_PATH;
+  else if (strcasestr(url, "queryContext") != NULL)                          return DEFAULT_SERVICE_PATH_RECURSIVE;
```
kzangeli: Only thing I would change here is the order (1. register, 2. discovery 3-5 ngsi9, 6 updateContext, 7. queryContext, etc).

Actually, one more thing.
Why all 'if - else if' ?
It's enough with 'if' - you do a return after each 'if', so ...

    if (A)   return X;
    if (B)   return X;
    ...

    error (as we're still in the function)

src/lib/rest/rest.cpp
kzangeli: Looks good like this.

NTC

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+
+  /* Look for convenience operation. Subscription-related operations are special, all the other depend on
+   * the method
+   */
+  else if (strcasestr(url, "contextAvailabilitySubscriptions") != NULL)      return DEFAULT_SERVICE_PATH_RECURSIVE;
+  else if (strcasestr(url, "contextSubscriptions") != NULL)                  return DEFAULT_SERVICE_PATH_RECURSIVE;
```
kzangeli: Both of these are probably GET ?

src/lib/rest/rest.cpp
kzangeli: NTC

src/lib/rest/rest.cpp
```diff
@@ -614,6 +614,65 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
+  else if (strcasecmp(method, "POST") == 0)                                  return DEFAULT_SERVICE_PATH;
+  else if (strcasecmp(method, "PUT") == 0)                                   return DEFAULT_SERVICE_PATH;
+  else if (strcasecmp(method, "DELETE") == 0)                                return DEFAULT_SERVICE_PATH;
+  else if (strcasecmp(method, "GET") == 0)                                   return DEFAULT_SERVICE_PATH_RECURSIVE;
+
+  LM_E(("Runtime Error (cannot found default service path for: (%s, %s))", method, url));
```
kzangeli: cannot *find*
Actually, this is more of a 'Bad Method' error ...

src/lib/rest/rest.cpp
kzangeli: Looks good like this. Easier to read without an extra indirection.

src/lib/rest/rest.cpp
fgalan: You are right: if the client send some esoteric HTTP metho such as PATCH, INFO, etc. it is not our fault.

src/lib/rest/rest.cpp
kzangeli: Discussed on skype

NTC

src/lib/rest/rest.cpp
fgalan: Fixed in 5f90fbe

src/lib/rest/rest.cpp
fgalan: Fixed in 5f90fbe

kzangeli:
LGTM

Issue #716, opened 2015-01-21
 ContextProvicer forwarding for attributes not existing in existing entities 


fgalan:
Current version  fowards query/update to ContextProviders in the case the entity doesn't exist. However, if the case the entity exists (e.g. E with atributes A and B) and the query/update is in an attribute that doesn't exits (e.g. update in E-C) but with a registration covering it, the CB doesn't detect this situation and doesn't forward.

fgalan:
Temporal "hack" developed in PR https://github.com/telefonicaid/fiware-orion/pull/732. However, this issues keeps opened given that it is not the definitive solution.

fgalan:
As 0.19.0 will be frozen including this, this issue is being closed.

PR #732, opened 2015-01-30
Feature/716 entity found attribute not found

kzangeli:
### Description
Just a 'hack' to make forward updates work (not queries, only updates) for non-found attributes in entities that exist in the broker.
The code is no good, but it enables this feature for the campus party in Brasil.
After this we need to *really* fix this problem.

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1148,6 +1148,8 @@ static void setResponseMetadata(ContextAttribute* caReq, ContextAttribute* caRes
 * Returns true if entity was actually modified, false otherwise (including fail cases)
 *
 */
+#define ATTRIBUTE_NOT_FOUND 1
+
 static bool processContextAttributeVector (ContextElement*                            ceP,
                                            std::string                                action,
                                            std::map<string, TriggeredSubscription*>&  subsToNotify,
@@ -1158,7 +1160,7 @@ static bool processContextAttributeVector (ContextElement*                      
                                            double&                                    coordLong,
                                            std::string                                tenant,
                                            const std::vector<std::string>&            servicePathV,
-                                           std::string*                               why)
+                                           int*                                       why)
 {
     EntityId*   eP         = &cerP->contextElement.entityId;
     std::string entityId   = cerP->contextElement.entityId.id;
@@ -1192,7 +1194,7 @@ static bool processContextAttributeVector (ContextElement*                      
                                       std::string("action: UPDATE") + 
                                       " - entity: [" + eP->toString() + "]" +
                                       " - offending attribute: " + targetAttr->toString());
-                *why = "Attribute Not Found";
+                *why = ATTRIBUTE_NOT_FOUND;
                 return false;
 
             }
@@ -1740,7 +1742,7 @@ void processContextElement(ContextElement*                      ceP,
             coordLat = loc.getField(ENT_LOCATION_COORDS).Array()[1].Double();
         }
 
-        std::string why;
+        int why = 0;
         if (!processContextAttributeVector(ceP, action, subsToNotify, attrs, cerP, locAttr, coordLat, coordLong, tenant, servicePathV, &why))
         {
             /* The entity wasn't actually modified, so we don't need to update it and we can continue with next one */
@@ -1763,7 +1765,7 @@ void processContextElement(ContextElement*                      ceP,
             //           The real fix will probably be a rewrite of this piece of code,
             //           implemented when fixing 'definitely' the issue #716
             //
-            if ((why != "Attribute Not Found") || (caller != "postUpdateContext"))
+            if ((why != ATTRIBUTE_NOT_FOUND) || (caller != "postUpdateContext"))
             {
               ++docs;
             }
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1763,7 +1763,7 @@ void processContextElement(ContextElement*                      ceP,
             //           The real fix will probably be a rewrite of this piece of code,
             //           implemented when fixing 'definitely' the issue #716
             //
-            if ((why != "Attribute Not Found") || (caller != "postContextUpdate"))
+            if ((why != "Attribute Not Found") || (caller != "postUpdateContext"))
             {
               ++docs;
             }
@@ -1917,7 +1917,7 @@ void processContextElement(ContextElement*                      ceP,
           {
             // Suitable CProvider has been found: creating response and returning
             std::string prApp = crrV[0]->contextRegistration.providingApplication.get();
-            LM_T(LmtCtxProviders, ("context provide found: %s", prApp.c_str()));
+            LM_T(LmtCtxProviders, ("context provider found: %s", prApp.c_str()));
             buildGeneralErrorResponse(ceP, NULL, responseP, SccFound, prApp, &ceP->contextAttributeVector);
           }
           else
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1157,7 +1157,8 @@ static bool processContextAttributeVector (ContextElement*                      
                                            double&                                    coordLat,
                                            double&                                    coordLong,
                                            std::string                                tenant,
-                                           const std::vector<std::string>&            servicePathV)
+                                           const std::vector<std::string>&            servicePathV,
+                                           std::string*                               why)
 {
     EntityId*   eP         = &cerP->contextElement.entityId;
     std::string entityId   = cerP->contextElement.entityId.id;
@@ -1191,6 +1192,7 @@ static bool processContextAttributeVector (ContextElement*                      
                                       std::string("action: UPDATE") + 
                                       " - entity: [" + eP->toString() + "]" +
                                       " - offending attribute: " + targetAttr->toString());
+                *why = "Attribute Not Found";
                 return false;
 
             }
@@ -1547,7 +1549,8 @@ void processContextElement(ContextElement*                      ceP,
                            const std::string&                   tenant,
                            const std::vector<std::string>&      servicePathV,
                            std::map<std::string, std::string>&  uriParams,   // FIXME P7: we need this to implement "restriction-based" filters
-                           const std::string&                   xauthToken
+                           const std::string&                   xauthToken,
+                           const std::string&                   caller
 )
 {
 
@@ -1675,7 +1678,6 @@ void processContextElement(ContextElement*                      ceP,
     while (cursor->more()) {
         BSONObj r = cursor->next();
         LM_T(LmtMongo, ("retrieved document: '%s'", r.toString().c_str()));
-        ++docs;
 
         BSONElement idField = r.getField("_id");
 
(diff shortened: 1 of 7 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1157,7 +1157,8 @@ static bool processContextAttributeVector (ContextElement*
                                            double&                                    coordLat,
```
fgalan: I think we should add an entry about this change in the CHANGES_NEXT_RELEASE file. Although this hack is temporal and hopefully it will be fixed before releasing 0.19.0, it is not completely sure that we achieve that, so it is better to have the "mark" in the CHANGES_NEXT_RELEASE just in case (if we implement a definitive fix for #716 before closing 0.19.0, that entry will be adjusted)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1157,7 +1157,8 @@ static bool processContextAttributeVector (ContextElement*
...
                                            double&                                    coordLong,
                                            std::string                                tenant,
-                                           const std::vector<std::string>&            servicePathV)
+                                           const std::vector<std::string>&            servicePathV,
+                                           std::string*                               why)
```
fgalan: Probably is more efficient to have an integer for that, considering that the number of "why" contitions is finite and small. Moreover, I'd sugget to pass this as a reference (as it is our style in other places, e.g. the `caller` argument added to mongoUpdateContext() function.

```
#define #ATTRIBUTE_NOT_FOUND (in the right .h file)
...
why = ATTRIBUTE_NOT_FOUND
```

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -82,6 +82,40 @@ std::string postUpdateContext
...
+  //
+  // If upcr.contextElementResponseVector contains >= TWO responses and the second of them is a 'SccFound',
+  // then skip the first, the one that isn't 'SccFound'.
+  // Also, if the first of them is 472, then skip the second
+  //
+  // FIXME P6: Temporary hack for 'Entity found but Attribute not found':
+  //           See FIXME P6 'Temporary hack ...' 
+  //           in MongoCommonUpdate.cpp, function processContextElement
+  //
+
```
fgalan: If I'm understanding the hack correctly, the "false double" situation happens when why is "Attribute Not Found" and the operation is updateContext. In this case, we have an SccInvalidParamter and a SccFound. I undestand that this circustance is "captured" by the first `if`, but... why we need the second `if`?

Moreover, I think you are assuming implicitely that the updateContext request have 1 CE. However, the updateContext in general could have N CEs. Could be some case of wrong behaviour in the case of the client uses updateContext with 2 element? E.g. in the case of updateContext with 2 elements, the first one resulting in SccInvalidParamer (but not of type of "Attribute not found", e.g. "location nature of an attribute has to be defined at creation time, with APPEND") and the second one going SccOk, the second if will wroning remove the SccOk, thus the client is sending 2 elements but receiving 1 response, which is inconsistent. This breaks existing functionality, not related with CPr forwarding...

Maybe a solution would be something like this:

// This check is to avoid wrongly removing responses when number of CE in the request is greater than 1
if (&parseDataP->upcr.res.contextElementVector.size() == 1)
{
  if (upcr.contextElementResponseVector.size() >= 2)
  {
  ...
  }
  if (upcr.contextElementResponseVector.size() >= 2)
  {
  ...
  }
}

Probably there would be smarter solutions, but I'm not able to find better ones right now...

test/functionalTest/cases/716_forward_registered_attributes/forward_attribute.test
```diff
@@ -0,0 +1,303 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: What about the similar functionality in the queryContext case (i.e. existing CE Ent1-AttrA and a registration for Ent1-AttrB, doing a query for Ent1-AttrB will forward operation to CPr instead of returning 404 Not CE found)?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 0d998d4

test/functionalTest/cases/716_forward_registered_attributes/forward_attribute.test
kzangeli: Not implemented in queryContext due to time constraints

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: It is an ugly fix, as I've stated.
I had to add the second if to make our functional tests work, that's the only reason it's there.

There are probably **many** cases where this hack doesn't work or even destroys a functionality that would have worked without the fix.

This is merely a 'proposition' for us to either accept or reject.
If rejected, then we wont have the desired functionality until the real solution is implemented.

However, adding the constaint of 'only one incoming CE' seems like a good idea to me.

test/functionalTest/cases/716_forward_registered_attributes/forward_attribute.test
fgalan: Ok, out of the scope of this PR.

NTC

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: I agree there are many cases in which the hack doesn't work. However, if we can assure the limitation of it works when the updateContext includes only 1 CE and *doesn't break existing functionality* (which is what my "nested ifs" proposa is about) I think we can be happy.

fgalan: Regarding the second "if" added for not breaking an existing functional test, I'd suggest to add a comment above the "if" with the test(s) names that such "if" is preventing to fail.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: What we gain in is very close to zero, nothing to worry about.
It should not be passed as reference, as it is an 'output variable' (as we saw not so long ago).
I can change it to an integer, sure, but it will be a lot harder to read the code (especially when you have to go to another file to see the definition).

kzangeli: Fixed in dcbbe54

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: New implementation in dcbbe54

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -82,6 +82,59 @@ std::string postUpdateContext
...
+
+  //
+  // Removing 'garbage' from the response vector (when just *one* incoming contextElement
```
fgalan: I think this comment would be more precise in the following way:

"Removing 'garbage' from the response vector (when then number of incoming contextElement *is different of the number of contextElement responses*)"

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -82,6 +82,59 @@ std::string postUpdateContext
...
+      //
+      // o 1. 472+302:  keep 302
+      //   Out[0]: (472) request parameter is invalid/not allowed
+      //   Out[1]: (302) Found
+      //
+      // o 2. 472+404: keep 472
+      //   Out[0]: (472) request parameter is invalid/not allowed
+      //   Out[1]: (404) No context element found
+      //
+      if (upcr.contextElementResponseVector[ix]->statusCode.code == 472)
```
fgalan: Although it is ok to have 472, 404 and 302 in the comments, I think it should be better to have the corresponding Scc* in the code.

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -82,6 +82,59 @@ std::string postUpdateContext
...
+    for (unsigned int ix = 0; ix < upcr.contextElementResponseVector.size() - 1; ++ix)
+    {
+      //
+      // 2 cases:
+      //
+      // o 1. 472+302:  keep 302
+      //   Out[0]: (472) request parameter is invalid/not allowed
+      //   Out[1]: (302) Found
+      //
+      // o 2. 472+404: keep 472
```
fgalan: To keep track of this aparentely unneded "exception" I'd suggest to add the following comment regarding this case (or similar wording):

"This exception is not related with CPr forwarding functionality, but needed anyway. Otherwise 000_ngsi10_update/update_sequence.test steps 13 and 15 will break"

It doesn't hurt and may provide useful information in the future if we need to dig into the causes of this behaviour before removing the hack.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: References should work also to send information back, but it is ok to use a pointer if you prefer so. Thus, NTC after dcbbe54.

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -82,6 +82,59 @@ std::string postUpdateContext
...
+      //
+      if (upcr.contextElementResponseVector[ix]->statusCode.code == 472)
+      {
+        if (ix + 1 < upcr.contextElementResponseVector.size())
+        {
+          // 472+302
+          if (upcr.contextElementResponseVector[ix + 1]->statusCode.code == 302)
+          {
+            // Remove the 472
+            upcr.contextElementResponseVector.vec.erase(upcr.contextElementResponseVector.vec.begin() + ix);
```
fgalan: Given this new code involves removing information, a valgrind test would be desirable, to ensure that everything keeps ok from the point of view of leaks.

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Actually, it is needed, as a 'biproduct' of the forwarding:

A non-found attribute now gets a 472, but docs it not counted up, so that we'll get into the 'if (docs == 0)' part.  In that part, if the attribute is found in a registration a 302 is added but if not, then a 404 is added.

So, this 404 must be removed and in the case of 'found', the 472 must be removed.

I can add what I just wrote here as a comment in the code ...

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yeah, of course.
I did this as a 'quick test' and then I forgot to change to the enum values.
I didn't even think the C++ compiler would accept it without a warning.

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Of course. Old comment. *Different than*

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yes, there can easily be a leak here

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 2705b4c

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 2705b4c

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 2705b4c

fgalan:
LGTM (from the plane al mobile! :)

PR #734, opened 2015-02-02
fixed a leak

kzangeli:
fixed a leak when skipping context element response for forwarding updates
Possibly fixed a bug as well ... (see the ix = ix  1).


fgalan:
LGTM

PR #736, opened 2015-02-03
Feature/716 entity found attribute not found query

kzangeli:
Func test for "Attribute not found in Entity that is found" (queries forwarded to context providers).


fgalan:
LGTM

Issue #719, opened 2015-01-23
ServicePath for registrations (NGSI9) (not recursive)


fgalan:
To implement the servicePath concept for NGSI9 registrations, in a similar way it works for entities in NGSI10. In detail:

* The registerContext message could include a Fiware-ServicePath header. That servicePath is stored in the database along with the registration (in a similar way we store it along with entities in NGSI10). Only "not recursive" servicePaths are allowed (i.e. the usage of "#" semantics is forbidden in this case).

* The discoverContextAvailability, resembles the same way a queryContext behaves. In that sense, it could include a Fiware-ServicePath header (that could be a list of comma-separated elements) to specify where to search for registrations. In this case, the "#" is allowed, to search in a sub-tree of servicePaths.

* In the case of "locally failed" update, the "search space" for possible registrations to forward the update are the ones in the servicePath used for the update.

* In the case of "locally failed" query, it is more complex, as the Fiware-ServicePath allowed in queryContext can be a comma-separated lists of elements that, in sequence, could include sub-tree search with "#"). Probably the most coherent solution is that the "search space" be the one covered by que queryContext servicePath (taking into acocunt sequences and sub-trees).

As in the case of NGSI10, the lack of Fiware-ServicePath is interpreted implicetly in the following way:

* In the case of registerContext, the lack of Fiware-ServiePath header is implicitely interpreted as "/"
* In the case of discoverContextAvailability, the lack of Fiware-ServicePath header is implicitely interpreted as "/#".
* registrations existing in DB without servicePath (i.e. the ones used in Orion versions previous to the implementation on this issue) implicitly belong to "/". The discoverContextAvailability logic, along with the forwarding logic for "locally failed" NGSI10 request, have to take this posibility into account.

Implementing servicePath in subscribeContextAvaiabilty and notifyContextAvailability is out of the scope of this issue.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/743, but not sure if we should close this issue or wait until after the re-working on the context providers functionality.

fgalan:
At the end, we decide to close it, as it will be part of 0.19.0.

PR #743, opened 2015-02-06
Feature/719 servicepath for registrations

kzangeli:
Issue 719 about service path for registrations and discovery fixed. Pretty covered with four different functional tests.
Tested for:
- normal registry/discovery
- registration in Config Manager and forwarding to context broker
- Context Providers
- More than *one* service path in normal discovery


unittests_that_fail_sporadically.txt
```diff
@@ -4,3 +4,11 @@ mongoOntimeintervalOperations.mongoGetContextElementResponses_ok (kz: Dec 17 201
 mongoOntimeintervalOperations.mongoUpdateCsubNewNotification_dbfail (kz: Dec 19 2014)
```
fgalan: Instead of a log-like list, maybe it is better to have the test ordered by number of times a sporadical fail happens, e.g.:

```
mongoUnsubscribeContext.x: 5 times
mongoUnsubscribeContext.y: 3 times
etc.
```

Thus, each time a new test fails, you add one to the right coner (and eventually reorder the lines if needed).

src/lib/common/string.cpp
```diff
@@ -676,3 +676,41 @@ void strReplace(char* to, int toLen, const char* from, const char* oldString, co
```
fgalan: CHANGES_NEXT_RELEASE corresponding to this PR is missing.

fgalan: Probably in this case, it is better to use several entries, given that the item includes several functional pieces, e.g: NGSI9 basic operation with register/discover and forwarding logic taking into account servicePath for registrations.

src/lib/common/string.cpp
```diff
@@ -676,3 +676,41 @@ void strReplace(char* to, int toLen, const char* from, const char* oldString, co
...
   to[toIx] = 0;
 }
+
+
+
+/* ****************************************************************************
+*
+* servicePathCheck - 
+*/
+std::string servicePathCheck(const char* servicePath)
```
fgalan: A bit surprised of seeing this as new code ("green" in githbu.com)... I though that we were already doing this servicePath checking. Maybe is was moved from another part I haven't seen so far?

kzangeli: We have another function (in rest.cpp) with the same name:

    int servicePathCheck(ConnectionInfo* ciP, const char* servicePath)

The functions don't do the same, however it is probably true I could use the 'new' function inside the old function.

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -124,16 +124,18 @@ static bool processAssociations(MetadataVector mdV, std::string* err, std::strin
...
-                                 map<string, TriggeredSubscription*>& subs,
-                                 std::string&                         err,
-                                 std::string                          tenant)
+static bool processSubscriptions
+(
+  EntityIdVector                        triggerEntitiesV,
+  map<string, TriggeredSubscription*>&  subs,
+  std::string&                          err,
+  const std::string&                    tenant
+)
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -124,16 +124,18 @@ static bool processAssociations(MetadataVector mdV, std::string* err, std::strin
...
 {
-
     /* For each one of the subscriptions in the map, send notification */
     bool ret = true;
-    for (std::map<string, TriggeredSubscription*>::iterator it = subs.begin(); it != subs.end(); ++it) {
-
+    for (std::map<string, TriggeredSubscription*>::iterator it = subs.begin(); it != subs.end(); ++it)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -345,13 +347,20 @@ static bool addTriggeredSubscriptions(ContextRegistration                  cr,
...
 */
-HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, RegisterContextResponse* responseP, OID* id, std::string tenant)
+HttpStatusCode processRegisterContext
+(
+  RegisterContextRequest*   requestP,
+  RegisterContextResponse*  responseP,
+  OID*                      id,
+  const std::string&        tenant,
+  const std::string&        servicePath
+)
```
kzangeli: Style changes + added parameter 'servicePath'

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -345,13 +347,20 @@ static bool addTriggeredSubscriptions(ContextRegistration                  cr,
...
 {
-
     DBClientBase* connection = getMongoConnection();
 
     /* If expiration is not present, then use a default one */
-    if (requestP->duration.isEmpty()) {
+    if (requestP->duration.isEmpty())
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
     EntityIdVector triggerEntitiesV;
 
     BSONArrayBuilder contextRegistration;
-    for (unsigned int ix = 0; ix < requestP->contextRegistrationVector.size(); ++ix) {
-
+    for (unsigned int ix = 0; ix < requestP->contextRegistrationVector.size(); ++ix)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
         ContextRegistration* cr = requestP->contextRegistrationVector.get(ix);
 
         BSONArrayBuilder entities;
-        for (unsigned int jx = 0; jx < cr->entityIdVector.size(); ++jx) {
-
+        for (unsigned int jx = 0; jx < cr->entityIdVector.size(); ++jx)
+        {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
             EntityId* en = cr->entityIdVector.get(jx);
             triggerEntitiesV.push_back(en);
 
-            if (en->type == "") {
+            if (en->type == "")
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
                 entities.append(BSON(REG_ENTITY_ID << en->id));
                 LM_T(LmtMongo, ("Entity registration: {id: %s}", en->id.c_str()));
             }
-            else {
+            else
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1444,12 +1452,21 @@ bool registrationsQuery
...
+    {
         /* If we don't do this checking, the {$in: [] } in the attribute name part will
          * make the query fail*/
         queryBuilder.append(contextRegistrationAttrsNames, BSON("$in" << attrs.arr()));
     }
 
+    //
+    // 'And-in' the service path
+    //
+    const std::string  servicePathString = REG_SERVICE_PATH;
```
fgalan: I understand this is the place were you take servicePath into account for discover (or subscribe availability, which I think also uses registrationQuery()). However, I'm not able to locate the point in which servicePath is taken into account for registrations creation/update... I'm a bit lost in the forest of style-related changes :)

Could you give some hint on where it is?

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
                 LM_T(LmtMongo, ("Entity registration: {id: %s, type: %s}", en->id.c_str(), en->type.c_str()));
             }
-
         }
 
         BSONArrayBuilder attrs;
-        for (unsigned int jx = 0; jx < cr->contextRegistrationAttributeVector.size(); ++jx) {
-
+        for (unsigned int jx = 0; jx < cr->contextRegistrationAttributeVector.size(); ++jx)
+        {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
            ContextRegistrationAttribute* cra = cr->contextRegistrationAttributeVector.get(jx);
             attrs.append(BSON(REG_ATTRS_NAME << cra->name << REG_ATTRS_TYPE << cra->type << "isDomain" << cra->isDomain));
             LM_T(LmtMongo, ("Attribute registration: {name: %s, type: %s, isDomain: %s}",
                                cra->name.c_str(),
                                cra->type.c_str(),
-                               cra->isDomain.c_str())
-            );
+                               cra->isDomain.c_str()));
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -379,39 +389,40 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
...
 
-            for (unsigned int kx = 0; kx < requestP->contextRegistrationVector.get(ix)->contextRegistrationAttributeVector.get(jx)->metadataVector.size(); ++kx) {
+            for (unsigned int kx = 0; kx < requestP->contextRegistrationVector.get(ix)->contextRegistrationAttributeVector.get(jx)->metadataVector.size(); ++kx)
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -441,7 +452,8 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
 
     BSONObj regDoc = reg.obj();
     LM_T(LmtMongo, ("upsert update() in '%s' collection: '%s'", getRegistrationsCollectionName(tenant).c_str(), regDoc.toString().c_str()));
-    try {
+    try
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -450,7 +462,8 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
         mongoSemGive(__FUNCTION__, "update in RegistrationsCollection");
         LM_I(("Database Operation Successful (_id: %s)", oid.toString().c_str()));
     }
-    catch( const DBException &e ) {
+    catch (const DBException& e)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoCommonRegister.h
```diff
@@ -37,7 +37,14 @@ using namespace mongo;
...
 */
-extern HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, RegisterContextResponse* responseP, OID* id, std::string tenant);
+extern HttpStatusCode processRegisterContext
+(
+  RegisterContextRequest*   requestP,
+  RegisterContextResponse*  responseP,
+  OID*                      id,
+  const std::string&        tenant,
+  const std::string&        servicePath
+);
```
kzangeli: Style-changes + added servicePath

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
```diff
@@ -45,15 +45,16 @@ using std::auto_ptr;
...
+  EntityIdVector*                  enV,
+  AttributeList*                   attrL,
+  const std::string&               scope,
+  MetadataVector*                  mdV,
+  std::string*                     err,
+  const std::string&               tenant,
+  int                              offset,
+  int                              limit,
+  bool                             details,
+  const std::vector<std::string>&  servicePathV
```
kzangeli: Added servicePathV as parameter

NTC

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
```diff
@@ -33,10 +33,16 @@
 
 /* ****************************************************************************
 *
-* mongoNofityContextAvailability -
+* mongoNotifyContextAvailability -
 */
-HttpStatusCode mongoNotifyContextAvailability(NotifyContextAvailabilityRequest* requestP, NotifyContextAvailabilityResponse* responseP, const std::string& tenant) {
-
+HttpStatusCode mongoNotifyContextAvailability
```
fgalan: Looking to these modifications to mongoNotifyContextAvailability related files, I wonder if the servicePath in notifications is also supported after this PR. Although it is not part of the functionality described https://github.com/telefonicaid/fiware-orion/issues/719 if we can have it also, of course it would be great :)

src/lib/mongoBackend/mongoRegisterContext.cpp
```diff
@@ -67,8 +80,9 @@ HttpStatusCode mongoRegisterContext(RegisterContextRequest* requestP, RegisterCo
         id = OID(requestP->registrationId.get());
 
         mongoSemTake(__FUNCTION__, "findOne from RegistrationsCollection");
-        reg = connection->findOne(getRegistrationsCollectionName(tenant).c_str(), BSON("_id" << id));
+        reg = connection->findOne(getRegistrationsCollectionName(tenant).c_str(), BSON("_id" << id << "servicePath" << sPath));
```
fgalan: Ok, I think I have located the place where you takes servicePath into account in registrations.

Are you checking that servicePaths with "#" are not used here? (different from discovery or notifyContextAvailability)

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -370,6 +379,7 @@ HttpStatusCode processRegisterContext(RegisterContextRequest* requestP, Register
     }
     reg.append("_id", oid);
     reg.append(REG_EXPIRATION, expiration);
+    reg.append(REG_SERVICE_PATH, servicePath);
```
kzangeli: This is the point where the service-path is 'inserted' into the database

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1405,22 +1405,28 @@ bool registrationsQuery
     BSONArrayBuilder entitiesWithType;
     BSONArrayBuilder entitiesWithoutType;    
 
-    for (unsigned int ix = 0; ix < enV.size(); ++ix) {        
+    for (unsigned int ix = 0; ix < enV.size(); ++ix)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1405,22 +1405,28 @@ bool registrationsQuery
...
         EntityId* en = enV.get(ix);
-        if (isTrue(en->isPattern)) {
+        if (isTrue(en->isPattern))
+        {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1405,22 +1405,28 @@ bool registrationsQuery
...
             BSONObjBuilder b;
             b.appendRegex(contextRegistrationEntitiesId, en->id);
-            if (en->type != "") {
+            if (en->type != "")
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1405,22 +1405,28 @@ bool registrationsQuery
...
                 b.append(contextRegistrationEntitiesType, en->type);
             }
             entityOr.append(b.obj());
         }
-        else { /* isPattern = false */
-            if (en->type == "") {
+        else  /* isPattern = false */
+        {
+            if (en->type == "")
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1405,22 +1405,28 @@ bool registrationsQuery
...
                 entitiesWithoutType.append(en->id);
                 LM_T(LmtMongo, ("Entity discovery without type: id '%s'", en->id.c_str()));
             }
-            else {
+            else
+            {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1429,8 +1435,10 @@ bool registrationsQuery
             }
         }
     }
+
     BSONArrayBuilder attrs;
-    for (unsigned int ix = 0; ix < attrL.size(); ++ix) {
+    for (unsigned int ix = 0; ix < attrL.size(); ++ix)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1444,12 +1452,21 @@ bool registrationsQuery
      * it has no impact on MongoDB query optimizer */
     queryBuilder.append("$or", entityOr.arr());
     queryBuilder.append(REG_EXPIRATION, BSON("$gt" << (long long) getCurrentTime()));
-    if (attrs.arrSize() > 0) {
+
+    if (attrs.arrSize() > 0)
+    {
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1806,12 +1823,13 @@ static HttpStatusCode mongoUpdateCasubNewNotification(std::string subId, std::st
 * This method returns true if the notification was actually send. Otherwise, false
 * is returned.
 */
-bool processAvailabilitySubscription(EntityIdVector enV, AttributeList attrL, std::string subId, std::string notifyUrl, Format format, std::string tenant) {
-
-    std::string err;
+bool processAvailabilitySubscription(EntityIdVector enV, AttributeList attrL, std::string subId, std::string notifyUrl, Format format, std::string tenant)
+{
+    std::string                      err;
```
kzangeli: Only style-changes

NTC

src/lib/mongoBackend/MongoGlobal.h
```diff
@@ -343,10 +344,11 @@ extern bool registrationsQuery
...
-  int                                 offset  = DEFAULT_PAGINATION_OFFSET_INT,
-  int                                 limit   = DEFAULT_PAGINATION_LIMIT_INT,
-  bool                                details = false,
-  long long*                          countP  = NULL
+  const std::vector<std::string>&     servicePathV,
+  int                                 offset       = DEFAULT_PAGINATION_OFFSET_INT,
+  int                                 limit        = DEFAULT_PAGINATION_LIMIT_INT,
+  bool                                details      = false,
+  long long*                          countP       = NULL
 );
```
kzangeli: indentation + added servicePathV

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
```diff
@@ -329,10 +332,11 @@ static HttpStatusCode conventionalDiscoverContextAvailability
...
 (
-  DiscoverContextAvailabilityRequest*        requestP,
-  DiscoverContextAvailabilityResponse*       responseP,
-  const std::string&                         tenant,
-  std::map<std::string, std::string>&        uriParams
+  DiscoverContextAvailabilityRequest*   requestP,
+  DiscoverContextAvailabilityResponse*  responseP,
+  const std::string&                    tenant,
+  std::map<std::string, std::string>&   uriParams,
+  const std::vector<std::string>&       servicePathV
```
kzangeli: Indentation + added servicePathV

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
```diff
@@ -33,10 +33,16 @@
...
+(
+  NotifyContextAvailabilityRequest*   requestP,
+  NotifyContextAvailabilityResponse*  responseP,
+  const std::string&                  tenant,
+  const std::string&                  servicePath
+)
+{
```
kzangeli: style changes + added servicePath

src/lib/mongoBackend/mongoNotifyContextAvailability.h
```diff
@@ -31,8 +31,14 @@
...
+* mongoNotifyContextAvailability -
 */
-extern HttpStatusCode mongoNotifyContextAvailability(NotifyContextAvailabilityRequest* requestP, NotifyContextAvailabilityResponse* responseP, const std::string& tenant = "");
+extern HttpStatusCode mongoNotifyContextAvailability
+(
+  NotifyContextAvailabilityRequest*   requestP,
+  NotifyContextAvailabilityResponse*  responseP,
+  const std::string&                  tenant      = "",
+  const std::string&                  servicePath = ""
+);
```
kzangeli: style changes + added servicePath

NTC

src/lib/mongoBackend/mongoRegisterContext.cpp
```diff
@@ -44,19 +44,32 @@ using namespace mongo;
...
 * mongoRegisterContext - 
 */
-HttpStatusCode mongoRegisterContext(RegisterContextRequest* requestP, RegisterContextResponse* responseP, const std::string& tenant)
+HttpStatusCode mongoRegisterContext
+(
+  RegisterContextRequest*   requestP,
+  RegisterContextResponse*  responseP,
+  const std::string&        tenant,
+  const std::string&        servicePath
+)
```
kzangeli: style changes + added servicePath

NTC

src/lib/mongoBackend/mongoRegisterContext.cpp
```diff
@@ -44,19 +44,32 @@ using namespace mongo;
...
     /* Check if new registration */
-    if (requestP->registrationId.isEmpty()) {
-        HttpStatusCode result = processRegisterContext(requestP, responseP, NULL, tenant);
-        reqSemGive(__FUNCTION__, "ngsi9 register request");
-        return result;
+    if (requestP->registrationId.isEmpty())
+    {
+      HttpStatusCode result = processRegisterContext(requestP, responseP, NULL, tenant, sPath);
+      reqSemGive(__FUNCTION__, "ngsi9 register request");
+      return result;
```
kzangeli: Only style changes

src/lib/mongoBackend/mongoRegisterContext.h
```diff
@@ -36,6 +36,12 @@
...
 * mongoRegisterContext - 
 */
-extern HttpStatusCode mongoRegisterContext(RegisterContextRequest* requestP, RegisterContextResponse* responseP, const std::string& tenant = "");
+extern HttpStatusCode mongoRegisterContext
+(
+  RegisterContextRequest*   requestP,
+  RegisterContextResponse*  responseP,
+  const std::string&        tenant       = "",
+  const std::string&        servicePath  = ""
+);
```
kzangeli: style changes + added servicePath

NTC

unittests_that_fail_sporadically.txt
kzangeli: Fixed in 56e64b6

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
kzangeli: Part of notifications is definitely supported, however, to really support it, func tests must be constructed and it must be tested.
I don't think it will be very hard to make subscriptions/notifications work with servicePath, but it is out of scope for this issue/PR.

NTC

src/lib/mongoBackend/mongoRegisterContext.cpp
kzangeli: The function servicePathCheck in lib/common/ does this check (not accepting '#').
Called by postRegisterContext.cpp and postNotifyContextAvailability.cpp.
So, this should be covered.
However, I will add a func test with bad service path (I have one already, but more would be good).

kzangeli: All good, see the new func test in 407ffb5

src/lib/common/string.cpp
kzangeli: Fixed in bfae851

src/lib/common/string.cpp
fgalan: I will try to unify both them to avoid entropy in the code. Moreover, the comment just above the function signature should describe what each function does so it will be easy to distinguish both them.

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: NTC

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Already spotted. NTC

src/lib/mongoBackend/mongoRegisterContext.cpp
fgalan: NTC

src/lib/mongoBackend/mongoRegisterContext.cpp
fgalan: NTC

src/lib/serviceRoutines/postRegisterContext.cpp
```diff
@@ -163,6 +165,32 @@ std::string postRegisterContext
...
+  if (ciP->servicePathV.size() > 1)
+  {
+    LM_W(("Bad Input (more than one service path for a registration)"));
+    rcr.errorCode.fill(SccBadRequest, "more than one service path for notification");
+    answer = rcr.render(RegisterContext, ciP->outFormat, "");
+    return answer;
+  }
+  else if (ciP->servicePathV.size() == 0)
+  {
+    ciP->servicePathV.push_back("/");
```
fgalan: I think we have a `#define` in some place for default servicePath. I'd suggest to use that.

src/lib/mongoBackend/mongoRegisterContext.cpp
```diff
@@ -44,19 +44,32 @@ using namespace mongo;
...
+    std::string sPath = servicePath;
+
     reqSemTake(__FUNCTION__, "ngsi9 register request");
 
     LM_T(LmtMongo, ("Register Context Request"));
 
     DBClientBase* connection = getMongoConnection();
 
+    // Default value for service-path is "/"
+    if (sPath == "")
```
fgalan: Style fix needed.

In addition, let's use existing `#define" for the "/".

src/lib/serviceRoutines/postNotifyContextAvailability.cpp
```diff
@@ -50,10 +51,35 @@ std::string postNotifyContextAvailability
...
+  if (ciP->servicePathV.size() > 1)
+  {
+    ncar.responseCode.fill(SccBadRequest, "more than one service path for notification");
+    LM_W(("Bad Input (more than one service path for a notification)"));
+    answer = ncar.render(NotifyContextAvailability, ciP->outFormat, "");
+    return answer;
+  }
+  else if (ciP->servicePathV.size() == 0)
+  {
+    ciP->servicePathV.push_back("/");
```
fgalan: Let's use existing `#define` for "/".

src/lib/serviceRoutines/postRegisterContext.cpp
fgalan: I have found it, at rest.h:

```
define  DEFAULT_SERVICE_PATH            "/"
```

Not sure if that is the best place or if it is better to move to a different .h ....

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
```diff
@@ -183,7 +186,8 @@ TEST(mongoRegisterContext_update, updateCase1)
   prepareDatabase();
 
   /* Invoke the function in mongoBackend library */
-  ms = mongoRegisterContext(&req, &res);
+  ms = mongoRegisterContext(&req, &res, "", "/");
+  LM_M(("Invoke mongoRegisterContext in mongoBackend library and EXIT. ms: %d", ms));
```
fgalan: We don't use to have LM_M() traces in unittest code... it is actually needed or a leftover?

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
```diff
@@ -0,0 +1,456 @@
...
+# 07. Check the registration in CB, service-path /cm/cb
+# 08. Verify service-path in CB, test with erroneous service-path /cm/cb2
+# 09. Check that everything is ok in database layer
+#
+
+
+echo "01. Initial registration in CM, service-path /cm/cb"
+echo "==================================================="
+payload='<?xml version="1.0" encoding="UTF-8"?>
+<registerContextRequest>
```
fgalan: Please, remember to use JSON by default in new tests... Although probably it doesn't make sense to change all the .test in this PR, take it into account for next ones.

fgalan: I've just realized that this is a test about ConfMan, so probably XML is intentional in this case. In that case, please forget my previos comment and mark this thread with NTC :$

test/functionalTest/cases/719_servicepath_for_registrations/register_and_discovery_with_servicepath.test
```diff
@@ -0,0 +1,367 @@
...
+    "errorCode": {
+        "code": "404", 
+        "reasonPhrase": "No context element found"
+    }
+}
+
+
+--TEARDOWN--
+brokerStop CB
+#dbDrop CB
```
fgalan: This line should be uncomented?

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
```diff
@@ -0,0 +1,456 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: After looking to all the .test, not sure if we cover this functionality (from issue #719):

"In the case of "locally failed" query, it is more complex, as the Fiware-ServicePath allowed in queryContext can be a comma-separated lists of elements that, in sequence, could include sub-tree search with "#"). Probably the most coherent solution is that the "search space" be the one covered by que queryContext servicePath (taking into acocunt sequences and sub-trees)."

I miss some test with queryContext using:

* A single meaningfull servicePath "#"-based (by meaningfll I mean e.g., registration in "/A/A1" and queryContext in "/A/#")
* A list of 2 regular servicePaths
* A list of 2 elements, one of them a meaningufll "#" and the other a regular servicePath

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
kzangeli: Actually, it is a copy of an old test, made in XML.
I could translate it to JSON, I just don't see the need.
New tests in JSON, yes of course, but that is not really the case here.

NTC

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: Garbage.
We have lots of LM_M in the unittest code so when I search for LM_Ms in the code I do a 'grep -v unittest', that's why I didn't find this line.
I will remove it

src/lib/common/string.cpp
kzangeli: Not sure you should unify these two functions. They are very different.
I would leave them as is.
Comments are always good, of course.

src/lib/serviceRoutines/postRegisterContext.cpp
kzangeli: Let's use that one.

src/lib/mongoBackend/mongoRegisterContext.cpp
kzangeli: ok!

test/functionalTest/cases/719_servicepath_for_registrations/register_and_discovery_with_servicepath.test
kzangeli: Yes!

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
kzangeli: ok, I guess I have been pretty concentrated on register/discovery.
More tests, no problem :-D

src/lib/common/string.cpp
kzangeli: Comments added in 728594e

src/lib/serviceRoutines/postRegisterContext.cpp
kzangeli: Fixed in 728594e

src/lib/mongoBackend/mongoRegisterContext.cpp
kzangeli: Fixed in 728594e

src/lib/serviceRoutines/postNotifyContextAvailability.cpp
kzangeli: Fixed in 728594e

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: Fixed in 728594e

test/functionalTest/cases/719_servicepath_for_registrations/register_and_discovery_with_servicepath.test
kzangeli: Fixed in 728594e

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
kzangeli: Just to synchronize ...

What about this test:

    #
    # 01. Register E1/A1 with service path /c1/c1/e1
    # 02. Register E2/A1 with service path /c1/c2/e2
    # 03. Register E3/A1 with service path /c1/c2/e3
    # 04. Register E4/A1 with service path /c2/c3/e4
    # 05. Register E5/A1 with service path /c2/c3/e4
    # 06. Register E6/A1 with service path /c2/c4/e4
    # 07. Query E.*/A1 with service path /c2/c3/#
    # 08. Query E.*/A1 with service path /c1/c1/e1,/c1/c2/e2
    # 09. Query E.*/A1 with service path /c1/#,/c2/c4/e4
    #

fgalan: * Step 7 will resolve to E4 and E5 in /c2/c3/e4. Not sure if this functionality is implemented, as it result in more than one registration matching in the scope of the same queryContext.
* Step 8 will resolve to E1 (in /c1/c1/e1) and E2 (/c1/c2/e2). Similar case than the one above.
* Step 9 will resolve to several cases also.

I think that in all the cases, only one forwarding result should happen.

kzangeli: Slight change in the test, but 'here' it is (commit 1c0bf18)

src/lib/common/string.cpp
```diff
@@ -676,3 +676,45 @@ void strReplace(char* to, int toLen, const char* from, const char* oldString, co
...
 }
+
+
+
+/* ****************************************************************************
+*
+* servicePathCheck - check one single component of the service-path
+*
+* NOTE
+* This function is used for register/notify only. '#' is not an accepted character,
```
fgalan: From a functional point of view, the constrain that this function enforces also applies to update, given that in that case `#` is not accepted. I understand that adapting the code to use also this function for updates is what the "FIXME P5: updates should also call the other servicePathCheck (in common lib)" in other place of the code is about, right?

src/lib/rest/rest.cpp
```diff
@@ -413,17 +414,26 @@ static int outFormatCheck(ConnectionInfo* ciP)
 
 /* ****************************************************************************
 *
-* servicePathCheck - 
+* servicePathCheck - check vector of service paths
 *
-* Not static just to let unit tests call this function
+* This function is called for ALL requests, when aservice-path URI-parameter is found.
```
fgalan: aservice-path -> a service-path

src/lib/rest/rest.cpp
```diff
@@ -413,17 +414,26 @@ static int outFormatCheck(ConnectionInfo* ciP)
 
 /* ****************************************************************************
 *
-* servicePathCheck - 
+* servicePathCheck - check vector of service paths
```
fgalan: Do we want to have both functions (the one here and the other in string.cpp) with the same name? Of course they are different, as they have different paramters, but having the same name for both is a bit confusing, isn't it?

src/lib/mongoBackend/MongoCommonRegister.h
fgalan: NTC

src/lib/mongoBackend/MongoGlobal.h
fgalan: NTC

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
fgalan: NTC

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
fgalan: I have the feeling that the behaviour regarding step 15 may change after "the big refactoring". Howeve, let's keep the .test as it is now.

NTC.

src/lib/common/string.cpp
kzangeli: Yes, the function should be used for updates also.
I wonder if we let '#' pass as a valid service-path character for updates right now ... ?

src/lib/rest/rest.cpp
kzangeli: This function, as it checks the validity for the URI-parameter should remain here.
The other function with the same name (I don't have any problem at all with that - overloading is a common technique) could be moved to another file, but it doesn't belong here as it isn't used here.
I already put it in the best place I could find ...  [ doesn't mean I'm right - if you find a better place for it, shoot ]

test/functionalTest/cases/719_servicepath_for_registrations/cm_forwards_cb_with_servicePath.test
kzangeli: I'm not so sure 15 will change, the initial service -path for 14 would have, so I changed it (first service path is not found).
However, these things will change and if a functest becomes invalid because the broker is *better*, that is just OK.

src/lib/common/string.cpp
fgalan: We shouldn't let the cases pass. Don't we have any functional test to ensure that?

src/lib/rest/rest.cpp
fgalan: The place of both functions is ok. My only "concertn" is in having the same name. However, if you think is ok with that, then mark this NTC this comment thread and we can let them as they are now.

kzangeli: NTC

src/lib/common/string.cpp
kzangeli: We might.
Adding an issue to check this (#746).

src/lib/rest/rest.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/a91b5469331a9fa51ea89e994fc815f9b7b24072

fgalan:
LGTM

PR #748, opened 2015-02-09
Bug/746 valid service path for updates

kzangeli:
Calling the 'servicePathCheck' for incoming update requests, to not accept service paths with the '#' character and also not multiple service paths.


fgalan:
LGTM

PR #752, opened 2015-02-10
$BROKER_USER omission

jariza:
Init script of Ubuntu is omitting $BROKER_USER. There is also one /dev/null redirection missing.


fgalan:
It seems to be ok, but probably @cquijano could have a look to it, given that the modification is in a file authored by him.

cquijano:
Absolutely agree with @jariza 's commit. 
I forgot these parameters
 It looks good for me.

fgalan:
Thanks for your contribution and feedback! Merging...

Issue #715, opened 2015-01-21
CPr: propagate X-Auth-Token header in forwarded requests


fgalan:
If the queryContext/updateContext (or equivalent conv op) is going to be forwarded to a Context Provider due to some registration match and such queryContext/updateContext includes a X-Auth-Token header, then the forwarded request has to include it as well.

If the originating queryContext/updateContext doesn't include a X-Auth-Token header, then CB doesn't include any X-Auth-Token header.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/753

Issue #714, opened 2015-01-21
CPr: propagate Fiware-ServicePath header in forwarded requests


fgalan:
If the queryContext/updateContext (or equivalent conv op) is going to be forwarded to a Context Provider due to some registration match and such queryContext/updateContext includes a Fiware-ServicePath header, then the forwarded request has to include it as well.

If the originating queryContext/updateContext doesn't include a Fiware-Service header, then CB doesn't include any Fiware-ServicePath header.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/751

kzangeli:
If a query or update is issued without Service-Path, the default Service-Path is set for the request and this 'default Service-Path' is forwarded to the context-provider.

fgalan:
The last issue with this issue :) has been implemented in PR https://github.com/telefonicaid/fiware-orion/pull/753

PR #753, opened 2015-02-10
Feature/715 forward xauth token to cprs

kzangeli:
o Func test for forwarding of X-Auth tokens in HTTP header (Issue #715)
o Made the broker not forward Service-Path if no Service-Path has been issued in the initial request (Issue #714)


scripts/accumulator-server.py
```diff
@@ -115,6 +114,18 @@ def all_done():
...
+    global ac
+    ac += "Got a /v1/updateContext";
+    return Response(status=200)
+
+@app.route("/v1/queryContext", methods=['POST'])
+def record():
+    global ac
+    ac += "Got a /v1/queryContext";
+    return Response(status=200)
+
```
kzangeli: Now tell me why this 'stupid' modification makes the accumulator behave the way I want ...
I expected to see 'Got a /v1/updateContext' and 'Got a /v1/queryContext' in the dump, and after that keep working to get the message in 'ac' instead.
However, in the first test I got already the exact output I wanted from /dump. Magic ... ?

Also, I've seen that in some functions use declare 'ac' as global and in others you don't - confusing.

src/lib/rest/rest.cpp
```diff
@@ -789,7 +789,6 @@ static int connectionTreat
     if (!ciP->httpHeaders.servicePathReceived)
     {
       ciP->httpHeaders.servicePath = defaultServicePath(url, method);
-      ciP->httpHeaders.servicePathReceived = true;
```
kzangeli: This here was a mistake from not so long ago.
I saw it in the PR but I let it pass, didn't think it mattered.
However, for this fix I need to know whether the servicePath is actually path of the request or not, so I had to remove it.

Just informative.
NTC

CHANGES_NEXT_RELEASE
```diff
@@ -7,3 +7,4 @@
 - Fix: Service-Path supported for forwarding of registrations between Config Manager and Context Broker (Issue #719)
 - Fix: Service-Path supported for forwarding to Context Providers (Issue #719)
 - Fix: Hash sign (#) in service-path, or multiple service-paths not valid for updateContext requests (Issue #746)
+- Fix: Not forwarding the default Service-Path to context providers (Issue #714)
```
fgalan: I understand that in the case of Service header everything is ok from time ago, right?

fgalan: Maybe we need a .test (in the 714_* directory) to ensure that service path is not being forwarded in this case?

scripts/accumulator-server.py
fgalan: What is even more weird is  having three times the same function defined, i.e. `record()`... maybe the `@app.route` is taken into account in the function definition and makes signatures different.

It seems that the function that is actually called is the last instance of `record()`. That would explain what it works and you don't have the "Got blabla" messages in the /dump.

Try to remove the first two function instances and set th following in the last one:

```
@app.route("/v1/updateContext", methods=['POST'])
@app.route("/v1/queryContext", methods=['POST'])
@app.route(server_url, methods=['GET', 'POST', 'PUT', 'DELETE'])
def record():
    global ac, t0, times
    ...
```

test/functionalTest/cases/715_forward_xauth_token_to_cprs/propagate_xauth_token_for_update_and_query_standard_ops.test
```diff
@@ -0,0 +1,216 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: Maybe we should add a a .test to ensure that X-Auth-Token is not being forwarded when X-Auth-Token is not included in the request? Or existing tests are covering that functionality?

CHANGES_NEXT_RELEASE
kzangeli: Nope, it works, but it is *not* covered in tests, also not that a query/update without X-Auth token doesn't send an X-Auth token in the forward.

test/functionalTest/cases/715_forward_xauth_token_to_cprs/propagate_xauth_token_for_update_and_query_standard_ops.test
kzangeli: Yes, as I just responded to the previous comment :-)
New func-tests are in the oven.

kzangeli: Fixed in 27668ef

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 5e51d1e

scripts/accumulator-server.py
kzangeli: Fixed in 5e51d1e

fgalan:
LGTM

Issue #117, opened 2013-11-26
Remove entire convenienceMap layer


kzangeli:
### Description
The convenienceMap layer has proven not necessary and should be removed completely.
What is done in the convenience map functions should be done in the corresponding service function (that currently invokes the convenience map function).

These convenience service functions should transform the incoming structures into their corresponding standard structures (e.g. RegisterProviderRequest into RegisterContextRequest) and call the service function for the standard operation.

**Effort: 30 man day**

kzangeli:
Actually, the later shouldn't be removed, but modified.
There are cases with more than one convop translated to one and the same standard op.
So, the translation of the input data to 'standard op input data' should be implemented in the library convenienceMap.  Also the translation of 'standard output data' to 'convop output data'.

fgalan:
In addition, we have been talking about a change name:

> Refactorize convenienceMap library. The new name will be convenienceTranslation and will be used from serviceRoutines. The serviceRoutines functions will keep the “flow control” instead of the current approach in which some operations are passed to convenienceMap (i.e. convenienceMap is able to invoke mongBackend). In the new approach, only serviceRoutines library can invoke mongoBackend (in some cases, one convop-oriented serviceRoutine function will invoke a “sibling” standard-operation oriented serviceRoutine)

kzangeli:
Finally, I believe no library will be necessary (which means the convenienceMap library will disappear).
The new functionality is added to fill methods for the standard request types. More elegant IMHO.

fgalan:
Sound good. I'm eager for seening the first PR about this :)

kzangeli:
Fixed by a number of PRs on branches called **hardening/117_XXX**

PR #754, opened 2015-02-11
Hardening/117 get context entities by entity

kzangeli:
ConvOp /v1/registry/contextEntities/{entityId::id} 'converted' to use Standard Operation.


src/lib/serviceRoutines/getContextEntitiesByEntityId.cpp
```diff
@@ -49,15 +64,23 @@ std::string getContextEntitiesByEntityId
...
+  std::string  answer;
+
+  //
+  // Fill in parseDataP->dcar.res to pass to postDiscoverContextAvailability
+  //
+  EntityId                             eId(entityId, "", "");
+  std::vector<std::string>             attributeV;
+  Restriction                          restriction;
+
+  parseDataP->dcar.res.fill(eId, attributeV, restriction);
```
kzangeli: Here is the 'beauty' in my new idea (that removes the needfor a new library):
1. fill in the standard request using its new 'fill' method.
2. call the standard operation
3. (sometimes) convert the standard response to convop response (fill method in convop-struct).

src/lib/ngsi/AttributeList.h
```diff
@@ -54,6 +54,16 @@ typedef struct AttributeList
                      const std::string&  indent,
```
fgalan: PR #117 will be a long issue to implement... maybe we should include the CHANGES_NEXT_RELEASE entry in this first PR, for not forgetting in the future.

src/lib/serviceRoutines/getContextEntitiesByEntityId.cpp
fgalan: One though...

Should the method name give a hint about its relation with conv ops mapping? `fill()` seems to be very generic... maybe `convFill()`. Or maybe you prefer to have `fill()` actually because is generic :)

Mark this as NTC if you think it is not a good idea to change the name. Appart from that, the idea is nice... a good use case of OOP :)

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
```diff
@@ -0,0 +1,646 @@
...
+ConvOp getContextEntitiesByEntityId: /v1/registry/contextEntities/{entityId::id}
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# This test case verifies that the convop 'getContextEntitiesByEntityId' does everything
```
fgalan: In which part of the sequence are you using URI parameters? In the outline below is not mentioned...

src/lib/serviceRoutines/getContextEntitiesByEntityId.cpp
kzangeli: This fill method doesn't know where it is used. Could be from a convop or from anywhere else.
I think 'fill' is an excellent name :-)

NTC

src/lib/ngsi/AttributeList.h
kzangeli: Sure

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
kzangeli: Yeah, I thought I would have some 'clear' use case with URI params.
But, as all the URI param treatment is inside mongo backend, not inside the service routine postDiscoverXXX, I don't know exactly what to add to the func-test.
I did add service-path, but if you can think of some more test case, that would be great,

We want to make sure that this conv op does *everything* that the standard op does.
That is what the test is for ...

fgalan: Maybe pagination? However I'd move it to a different .test, for not overloading too much the current file.

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
```diff
@@ -0,0 +1,646 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: A thought:

Probably the new tests in 17_convop_using_standard_ops (which are designed to be exahustive) make the ones in 000_ngsi10_convenience and 000_ngsi9_convenience obsolete. However, I'd not remove the 000_*_convenience ones: the only problem in having *too many* tests is the effort to write them. Given that the effort for 000_*_convenience is already payd, let's keep them.

NTC

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
kzangeli: Yeah, pagination is a 'big' thing to test.
I imagine we have a test for it already, perhaps I could add one 'discover via convop' in an existing test ...

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
kzangeli: My thought too :-)

test/functionalTest/cases/117_convop_using_standard_ops/getContextEntitiesByEntityId.test
fgalan: Good point. It is better to start from an existing test for standard op than from the scratch.

I guess that there should be someone in the cases/ dir (`grep` is your friend :))

kzangeli: Finally I've added a brand new test in the '117 directory'.
All convops regarding discovery will use this file.
Another test file for queries, etc.

kzangeli: Fixed in f007fa9

src/lib/ngsi/AttributeList.h
kzangeli: Fixed in 4c15cc8

fgalan:
LGTM

PR #778, opened 2015-02-16
Hardening/migration 0.19.0 script

fgalan:
This PR consolidates into repository a couple of scripts only present at gist at the present moment (https://gist.github.com/fgalan/a048bf48b4d4cd148d76 and https://gist.github.com/fgalan/1922df50f35ab3529b91)

After merging this PR the following actions need to be done:

* Merge the same changes into release/0.19.0 in a new PR (either with a diff or cherry picks)
* Update the 0.19.0-FIWARE-4.2.2 tag to the HEAD of release/0.19.0
* Change the wiki links, from gist to the files at 0.19.0-FIWARE-4.2.2 tag

@jcalderin @iariasleon (as you use Python in your code)

### DIFF
--- a/scripts/managedb/check_db_sanity.py
+++ b/scripts/managedb/check_db_sanity.py
@@ -48,16 +48,17 @@ def warn(s):
 
 def duplicated_attr_name(id, type, sp, list):
     names = {}
+    at_least_one_dup = False
     for attr in list:
-        if (names.has_key(attr['name'])):
+        if (attr['name'] in names):
             warn('<%s, %s, %s> attribute %s is duplicated' % (id, type, sp, attr['name']))
-            return True
+            at_least_one_dup = True 
         else:
             names[attr['name']] = 1
-    return False
+    return at_least_one_dup
 
 def value_or_empty(d, k):
-    if d.has_key(k):
+    if k in d:
        return d[k]
     else:
        return ''
--- a/scripts/managedb/check_db_sanity.py
+++ b/scripts/managedb/check_db_sanity.py
@@ -1,0 +1,74 @@
+#!/usr/bin/python
+# -*- coding: latin-1 -*-
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


scripts/managedb/check_db_sanity.py
```diff
@@ -0,0 +1,74 @@
...
+
+def warn(s):
+    print 'WARNING: %s' % s
+    global need_fix
+    need_fix = True
+
+def duplicated_attr_name(id, type, sp, list):
+    names = {}
+    for attr in list:
+        if (names.has_key(attr['name'])):
```
jcalderin: Try to use "attr['name'] in names" instead of "has_key", is deprecated, see https://docs.python.org/2/library/stdtypes.html#dict.has_key

scripts/managedb/merge_default_sp_dups.py
```diff
@@ -0,0 +1,201 @@
...
+    return False
+
+def has_attr(attr, doc_attrs):
+    for a in doc_attrs:
+        if a['name'] == attr['name']:
+            return True
+    return False
+
+def is_attr_in_doc_older(attr, doc, dup):
+    # This method assumes that attribute is not duplicated in doc or dup (the method is called
```
jcalderin: Use triple quotes for documentation of the functions instead of line-comment '#'

scripts/managedb/merge_default_sp_dups.py
```diff
@@ -0,0 +1,201 @@
...
+    print '%5d: WARNING: %s' % (n, s)
+    global need_fix
+    need_fix = True
+
+def duplicated_attr_name(n, list):
+    names = {}
+    for attr in list:
+        if (names.has_key(attr['name'])):
+            warn(n, 'attribute %s is duplicated' % attr['name'])
+            return True
```
jcalderin: IF there is more than one attr duplicate, the script is not going to warn about more than one

scripts/managedb/check_db_sanity.py
fgalan: Do you know in which Python version it was deprecated? The script should work for Python 2.6 and 2.7, so if the new way only works in 2.7 we cannot change it.

(I had a bad experience with timedelta.total_seconds() so I'm concerned about this kind of things).

fgalan: Fixed in e7bca62

scripts/managedb/merge_default_sp_dups.py
fgalan: Fixed in e7bca62

scripts/managedb/merge_default_sp_dups.py
fgalan: Fixed in e7bca62

jcalderin:
LGTM

PR #783, opened 2015-02-18
Hardening/117 ngsi9 subscription convops

kzangeli:
Last 'nec ngsi9 convops' converted to use standard ops.
Actually, they already did, just cleaned it up a little and added comments.

### DIFF
--- a/test/functionalTest/cases/117_convop_using_standard_ops/putAvailabilitySubscriptionConvOp.test
+++ b/test/functionalTest/cases/117_convop_using_standard_ops/putAvailabilitySubscriptionConvOp.test
@@ -1,0 +1,226 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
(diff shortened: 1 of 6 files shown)


test/functionalTest/cases/117_convop_using_standard_ops/putAvailabilitySubscriptionConvOp.test
```diff
@@ -0,0 +1,226 @@
...
+--NAME--
+ConvOp putAvailabilitySubscriptionConvOp: PUT /v1/registry/contextAvailabilitySubscriptions/{subscriptionId}
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
```
fgalan: Should we involve NGSI9 notifications (and accumulator-server) in this tests? Or maybe it is not necessary? Not sure about that... just a "feeling"...

kzangeli: I was thinking about that too but as the purpose of this test is not to test the subscriptions themselves but the ability to create/modify them. We have to assume that if the broker says "ALL OK", all is ok :-)

What I would have liked here is to ask the broker for a listing of its subscriptions.

fgalan: NTC

fgalan:
LGTM

PR #790, opened 2015-02-19
mongoBackend unit test draft for CPr full functionality

fgalan:
This PR is mainly to "agree" in the internal interface between mongoBackend and serviceRoutines. Have a in-deep look to the unit test, in order to check if everything is ok from your point of view or you foresee some problem (of course, not foreseeing a problem not doesn't mean we face it once implementation start, but try to catch them at design time :)

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,3 @@
 - Fix: Correct rendering of response to convop "POST /v1/contextEntities/{entityId::id}/attributes/{attributeName} (Issue #772)
 - Fix:  All convenience operations to use the service routines of standard operations (Issue #117)
+- Add: context providers full functionality (Issue #787)
--- a/src/lib/ngsi/ContextAttribute.cpp
+++ b/src/lib/ngsi/ContextAttribute.cpp
@@ -48,6 +48,7 @@ ContextAttribute::ContextAttribute()
   value                 = "";
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
+  providingApplication  = "";
 }
 
 
@@ -62,6 +63,7 @@ ContextAttribute::ContextAttribute(ContextAttribute* caP)
   type                  = caP->type;
   value                 = caP->value;
   compoundValueP        = (caP->compoundValueP)? caP->compoundValueP->clone() : NULL;
+  providingApplication  = caP->providingApplication;
   typeFromXmlAttribute  = "";
 
   LM_T(LmtClone, ("Creating a ContextAttribute: compoundValueP at %p for attribute '%s' at %p",
@@ -101,6 +103,7 @@ ContextAttribute::ContextAttribute
   value                 = _value;
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
+  providingApplication  = "";
 }
 
 
@@ -122,6 +125,7 @@ ContextAttribute::ContextAttribute
   type                  = _type;
   compoundValueP        = _compoundValueP;
   typeFromXmlAttribute  = "";
+  providingApplication  = "";
 }
 
 
@@ -326,6 +330,8 @@ void ContextAttribute::present(const std::string& indent, int ix)
   {
     compoundValueP->show(indent + "  ");
   }
+
+  PRINTF("%s  PA:       %s\n", indent.c_str(), providingApplication.c_str());
 
   metadataVector.present("Attribute", indent + "  ");
 }
--- a/src/lib/ngsi/ContextAttribute.cpp
+++ b/src/lib/ngsi/ContextAttribute.cpp
@@ -49,6 +49,7 @@ ContextAttribute::ContextAttribute()
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
   providingApplication  = "";
+  found                 = true;
 }
 
 
@@ -64,6 +65,7 @@ ContextAttribute::ContextAttribute(ContextAttribute* caP)
   value                 = caP->value;
   compoundValueP        = (caP->compoundValueP)? caP->compoundValueP->clone() : NULL;
   providingApplication  = caP->providingApplication;
+  found                 = caP->found;
   typeFromXmlAttribute  = "";
 
   LM_T(LmtClone, ("Creating a ContextAttribute: compoundValueP at %p for attribute '%s' at %p",
@@ -104,6 +106,7 @@ ContextAttribute::ContextAttribute
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
   providingApplication  = "";
+  found                 = true;
 }
 
 
@@ -126,6 +129,7 @@ ContextAttribute::ContextAttribute
   compoundValueP        = _compoundValueP;
   typeFromXmlAttribute  = "";
   providingApplication  = "";
+  found                 = true;
 }
 
 
--- a/src/lib/ngsi/ContextAttribute.cpp
+++ b/src/lib/ngsi/ContextAttribute.cpp
@@ -34,7 +34,10 @@
 #include "rest/ConnectionInfo.h"
 #include "rest/uriParamNames.h"
 
-
+// FIXME P5: we have the same macro in parseArg library. That is not efficient: the macro (along
+// with probably more stuff) should be isolated in a separate library, invoked but all the other
+// libraries which need it)
+#define FT(x) (x == true)? "true" : "false"
 
 /* ****************************************************************************
 *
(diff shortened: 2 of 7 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 - Fix: Correct rendering of response to convop "POST /v1/contextEntities/{entityId::id}/attributes/{attributeName} (Issue #772)
 - Fix:  All convenience operations to use the service routines of standard operations (Issue #117)
+- Add: context providers full functionality (Issue #787)
```
fgalan: First PR to feature/787_cprs_full_functional, so we include this entry for not forgetting it.

NTC (just informative)

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -48,6 +48,8 @@ ContextAttribute::ContextAttribute()
   value                 = "";
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
+  providingApplication  = "";
+  found                 = true;
```
fgalan: By default, `found` is set to true. I think I have touched all constructors, so no "randomness" will happen. However, it woudl be great @kzangeli if you could double-check is ok.

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -327,6 +335,8 @@ void ContextAttribute::present(const std::string& indent, int ix)
     compoundValueP->show(indent + "  ");
   }
 
+  PRINTF("%s  PA:       %s\n", indent.c_str(), providingApplication.c_str());
+
```
fgalan: Probably the `found` field should be also printed. Given that it is `bool` type, could you suggest a way of printint it? (if this were Python, simply `print str(found)` will work :)

src/lib/ngsi/ContextElement.h
```diff
@@ -48,6 +48,10 @@ typedef struct ContextElement
   ContextAttributeVector  contextAttributeVector;  // Optional
   MetadataVector          domainMetadataVector;    // Optional
 
+  std::string             providingApplication;    // Not part of NGSI, used internally for CPr forwarding functionality
+  bool                    found;                   // Not part of NGSI, used internally for CPr forwarding functionality (update case)
+                                                   // FIXME P10 #787: not sure if we need this at the end... it only would be needed in E-<null> updates, which seems not having too much sense
```
fgalan: I cannot think any unit test leading to a CE-level `found` to `false`. So, that probably confirms that `found` is not needed at CE level. If you are not able to find any case in which this would be used, probably we should remove it and leave it only in ContextAttribute class.

src/lib/ngsi/ContextAttribute.cpp
kzangeli: I should have a macro somewhere ...
Doing this:

X(x) (x == true)? "true" : "false"

kzangeli: BA_FT(boolVar) in "parseArgs/baStd.h"

src/lib/ngsi/ContextAttribute.cpp
kzangeli: Just one thing.
Found means found locally or found remotely (SccFound) ?
Perhaps we should use a more descriptive name of the variable ...
For example: 'bool remote;', initialized to false and set to true if a CPr is found for the attribute.

src/lib/ngsi/ContextElement.h
kzangeli: Just remove it then.
If we later find it's necessary, we'll just add it again.

test/unittests/mongoBackend/mongoContextProvidersUpdate_test.cpp
```diff
@@ -84,6 +88,34 @@
...
 *
+* getAttr -
+*
+* We need this function because we can not trust on array index, at mongo will
+* not sort the elements within the array. This function assumes that always will
+* find a result, that is ok for testing code.
+*
+* FIXME P7: duplicated function in mongoUpdateContext_test.cpp. Unify.
+*/
+static BSONObj getAttr(std::vector<BSONElement> attrs, std::string name, std::string type, std::string id = "") {
```
kzangeli: This is just a unit-test function, but try to avoid pushing big objects on the stack.
Especially vectors of big objects ...

    static BSONObj getAttr(const std::vector<BSONElement>& attrs, const std::string& name, const std::string& type, const std::string& id = "")

src/lib/ngsi/ContextAttribute.cpp
kzangeli: Confirming that all constructors contain the setting of 'found' and 'providingApplication'.

fgalan: Fixed in b8f6aeb

src/lib/ngsi/ContextAttribute.cpp
fgalan: Fixed in b8f6aeb

src/lib/ngsi/ContextElement.h
fgalan: Fixed in b8f6aeb

test/unittests/mongoBackend/mongoContextProvidersUpdate_test.cpp
fgalan: Fixed in 98b3bd2

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -1157,3 +1387,193 @@ TEST(mongoContextProvidersQueryRequest, mixPatternAndNotPattern)
...
+  EXPECT_EQ("A5", RES_CER_ATTR(0, 4)->name);
+  EXPECT_EQ("T", RES_CER_ATTR(0, 4)->type);
+  EXPECT_EQ("", RES_CER_ATTR(0, 4)->value);
+  EXPECT_EQ("http://cpr2.com", RES_CER_ATTR(0, 4)->providingApplication);
+
+  EXPECT_EQ("A6", RES_CER_ATTR(0, 5)->name);
+  EXPECT_EQ("T", RES_CER_ATTR(0, 5)->type);
+  EXPECT_EQ("", RES_CER_ATTR(0, 5)->value);
+  EXPECT_EQ("http://cpr3.com", RES_CER_ATTR(0, 5)->providingApplication);
+
```
kzangeli: Shouldn't A7 also go to CPR3?

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -1157,3 +1387,193 @@ TEST(mongoContextProvidersQueryRequest, mixPatternAndNotPattern)
...
+  /* Context Element response # 3 */
+  EXPECT_EQ("E1", RES_CER(2).entityId.id);
+  EXPECT_EQ("T", RES_CER(2).entityId.type);
+  EXPECT_EQ("false", RES_CER(2).entityId.isPattern);
+  EXPECT_EQ("http://cpr3.com", RES_CER(0).providingApplication);
+  ASSERT_EQ(0, RES_CER(2).contextAttributeVector.size());
+  EXPECT_EQ(SccOk, RES_CER_STATUS(2).code);
+  EXPECT_EQ("OK", RES_CER_STATUS(2).reasonPhrase);
+  EXPECT_EQ(0, RES_CER_STATUS(2).details.size());
+
```
kzangeli: OK, I guess this could work - if the attribute vector is empty a forward is sent that way to query for all attributes of E1/T.
I will have to try to notice that the query for E1-A3 is 'included in the query for 'E1-<null>', to avoid sending two requests to cpr2.

It's going to be *fun* to implement this ... :-)

NTC

test/unittests/mongoBackend/mongoContextProvidersUpdate_test.cpp
```diff
@@ -1354,3 +1490,203 @@ TEST(DISABLED_mongoContextProvidersUpdateRequest, mixPatternAndNotPattern)
...
+*         E3 - (A5, A6)
+*         E4 - A7
+* Result: E1 - A1 - Up Ok
+*              A2 - fwd CPR2
+*         E2 - A3 - Up ok
+*              A4 - fwd CPR3
+*         E3 - A5 - fwd CPR2
+*              A6 - Not found
+*         E4 - A7 - fwd CPR1
+*
```
kzangeli: Looks good

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
fgalan: Fixed in f1794ab

kzangeli:
LGTM

PR #792, opened 2015-02-24
Hardening/117 put individual context entity

kzangeli:
Made the convop "PUT /v1/contextEntities/EID" use standard op /v1/updateContext instead of convMap.


src/lib/convenience/UpdateContextElementResponse.cpp
```diff
@@ -119,3 +120,63 @@ void UpdateContextElementResponse::release(void)
...
+
+
+
+/* ****************************************************************************
+*
+* UpdateContextElementResponse::fill -
+*
+* NOTE
+* This method is used in the service routine of 'PUT /v1/contextEntities/{entityId::id}.
+* Only ONE response in the vector contextElementResponseVector of UpdateContextResponse is possible.
```
fgalan: Like in other cases, maybe a FIXMEis worth...

src/lib/serviceRoutines/putIndividualContextEntity.cpp
```diff
@@ -39,6 +40,22 @@
...
+* Corresponding Standard Operation: UpdateContext/UPDATE
+*
+* PUT /v1/contextEntities/{entityId::id}
+* PUT /ngsi10/contextEntities/{entityId::id}
+*
+* Payload In:  UpdateContextElementRequest
+* Payload Out: UpdateContextElementResponse
+*
+* URI parameters:
+*   - attributesFormat=object
```
fgalan: What about the following filters:

- !exist=entity::type
- exist=entity::type
- entity::type=TYPE

Note that to modify entities without type you need the !exist=entity::type in the PUT operation.

What about the same for POST? Not sure if POST logic has been modified yet (i.e. the PR for POST operation has been issued or not).

test/unittests/testData/ngsi10.restReply.updateContextResponse01.ok.valid.xml
```diff
@@ -2,5 +2,6 @@
   <errorCode>
     <code>404</code>
     <reasonPhrase>No context element found</reasonPhrase>
+    <details>detail</details>
```
fgalan: That means a "slight" functional change? I think is ok (it's very minor and it nos "violating" the XSD, so although clients should be prepared for this newly introduced details tag), but just to be sure...

test/functionalTest/cases/117_convop_using_standard_ops/putIndividualContextEntity.test
```diff
@@ -0,0 +1,417 @@
...
+
+#
+# 01. Update/UPDATE E1/A1=1 and see it fail (PUT /v1/contextEntities/E1)
+# 02. Update/APPEND E1/A1=10 (POST /v1/updateContext)
+# 03. Update/UPDATE E1/A1=1 and see it work (PUT /v1/contextEntities/E1)
+# 04. Query CB for E1/A1
+# 05. Update/APPEND E2/A2=1 in CP1 
+# 06. Register E2/A2 in CB with CP1 as providing application
+# 07. Update/UPDATE E2/A2=22 in CB, making it being forwarded to CP1 (PUT /v1/contextEntities/E2)
+# 08. Query CP1 for E2/A2
```
fgalan: What about testing attributeFormat?

In addition, probably more tests

In addition, probably more tests could be added:

* !exist=entity::type
* exist=entity::type
* entity::type=TYPE

test/unittests/testData/ngsi10.restReply.updateContextResponse01.ok.valid.xml
kzangeli: This is adding info. Can't imagine anything will break for adding info. And if it does, it's their own problem, I'd say.
However, this is from a unit_test, creating the structures by hand. Not sure an external user will ever experience any difference

NTC

src/lib/serviceRoutines/putIndividualContextEntity.cpp
kzangeli: Sure, I'll add a few more phases of the func-test.
And no, not for POST, this is a PUT operation. POST will come later.

kzangeli: Wait, this is not about func-tests but the actual implementation ...
And you are right, these filters must be taken care of right here, to create the Standard Op UpdateContextRequest.

test/functionalTest/cases/117_convop_using_standard_ops/putIndividualContextEntity.test
kzangeli: Yeah, more tests to be added.

kzangeli: Fixed in 879d9bc

src/lib/serviceRoutines/putIndividualContextEntity.cpp
kzangeli: Fixed in 879d9bc

src/lib/convenience/UpdateContextElementResponse.cpp
kzangeli: Not sure what to put in the FIXME ...

kzangeli: Discussed over skype ...

NTC

fgalan:
LGTM

PR #798, opened 2015-03-02
hardening/117_getIndividualContextEntityAttribute

kzangeli:
GET /v1/contextentities/EID/attributes/attibuteName now uses Standard Op

### DIFF
--- a/src/lib/serviceRoutines/postQueryContext.cpp
+++ b/src/lib/serviceRoutines/postQueryContext.cpp
@@ -179,7 +179,6 @@ std::string postQueryContext
                        false,
                        true);
 
-
   if ((out == "error") || (out == ""))
   {
     QueryContextResponse qcrs;
@@ -194,7 +193,6 @@ std::string postQueryContext
   //
   // 5. Parse the XML response and fill in a binary QueryContextResponse
   //
-  ParseData    parseData;
   std::string  s;
   std::string  errorMsg;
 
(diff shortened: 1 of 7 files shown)


src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -215,11 +213,21 @@ std::string postQueryContext
     return answer;
   }
 
-  s = xmlTreat(cleanPayload, ciP, &parseData, RtQueryContextResponse, "queryContextResponse", NULL, &errorMsg);
+  //
+  // NOTE
```
fgalan: The same NOTE is include in other GET operations in which the same comment makes sense? Just to maintain "comment coherence" along code...

kzangeli: I haven't needed this before, this is the first time.
However, all GET conv-op (that translate into a ngsi10 query) will get to this std-op, so, this unique comment covers them all ...

If the same thing is needed for some other "conv-op translation to std-op", I will try to add the same 'NOTE'.

fgalan: Good point. I didn't realice that the comment was in the "common" postQueryContext logic. I though that I was in a per-convop specific part.

I don't like too much manipulating ciP as I see it as an object representing the "connection environment", but I think it is the most pragmatic way of doing. However, maybe I'm not objective... you know I don't like too much ciP ;)

NTC

fgalan:
LGTM

PR #802, opened 2015-03-03
Hardening/117 post individual context entity attribute

kzangeli:
Another conv-op using std-op:

POST /v1/contextEntities/{entity::id}/attributes/{attribute::name}


fgalan:
LGTM

PR #803, opened 2015-03-04
Change the way to set properties, giving only a "base" file with default...

jcalderin:
... values


iariasleon:
LGTM

.gitignore
```diff
@@ -42,3 +42,5 @@ rpm/SOURCES/
 rpm/BUILDROOT/
 rpm/src/
 
+*.pyc
+properties.json
```
fgalan: properties.json seems to be a very general file to be ignored in general... I would need to understand why this is neded. Maybe through skype discussion.

PR #806, opened 2015-03-04
hardening/117_deleteIndividualContextEntityAttribute

kzangeli:
DELETE /v1/contextEntities/{entity::id}/attributes/{attribute::name} now uses Std-op


fgalan:
LGTM

PR #808, opened 2015-03-04
Task/787 mongobackend query

fgalan:
Not all test work after merging this PR. The following unit test will fail (to be fixed with the updateContext functionality gets implemented in a further PR):

```
TEST(mongoContextProvidersUpdateRequest, severalCprs)
```

The following test harness will fail (I guess that mostly of them related with CPr redirection fucntionalty, taking into account that mongoBacken is not returning the 302 Found that current serviceRoutines expect to work):

*Edit:* change list of failing tets after 2e65343

```
  o  117_convop_using_standard_ops/postContextEntitiesByEntityId.test
  o  117_convop_using_standard_ops/postContextEntityAttributes.test
  o  117_convop_using_standard_ops/postEntityByIdAttributeByName.test
  o  499_context_providers_basic/query_redirect_operation.test
  o  499_context_providers_basic/query_redirect_three_providers.test
  o  499_context_providers_basic/update_redirect_delete_operation.test
  o  499_context_providers_basic/update_redirect_operation.test
  o  714_propagate_service_path/no_propagation_of_service_path_when_not_present.test
  o  715_forward_xauth_token_to_cprs/propagate_xauth_token_for_update_and_query_standard_ops.test
  o  716_forward_registered_attributes/forward_attribute_query.test
  o  719_servicepath_for_registrations/forward_attribute_with_servicePath.test
  o  719_servicepath_for_registrations/forward_queries.test
```


src/lib/ngsi/ContextElement.cpp
```diff
@@ -154,7 +157,7 @@ void ContextElement::fill(const struct ContextElement& ce)
   attributeDomainName.fill(ce.attributeDomainName);
   contextAttributeVector.fill((ContextAttributeVector*) &ce.contextAttributeVector);
   domainMetadataVector.fill((MetadataVector*) &ce.domainMetadataVector);
-  providingApplication = ce.providingApplication;  
+  providingApplicationList = ce.providingApplicationList;  // FIXME P10: not sure if this is the right way of doing, taking into account how the same is done for contextAttributeVector in its own fill() method...
```
fgalan: Is the `=` operator clever enough to actually copying vector contents? Looking to other similar structures (attributeList) I think you have encapsulated it in a separated class with speciall method for copying (with pass through all the vector, pushing back element from one to other). Maybe I should do the same?

I'm to used to high level languages in which this is not a problem at all that I'm a bit unsure at C++ :)

src/lib/ngsi/ContextElementResponse.cpp
```diff
@@ -31,6 +31,14 @@
 #include "ngsi/ContextElementResponse.h"
 #include "rest/ConnectionInfo.h"
 
+/* ****************************************************************************
+*
+* ContextElementResponse::render -
+*/
+ContextElementResponse::ContextElementResponse()
+{
+  notPrune = true;
```
fgalan: This is the field I needed to create in ContextElementResponse to implement "shadow" entities that need to be prunned at the end if no CPrs is found for them. Although you also use this class, I think is has no effect for you.

src/lib/rest/HttpStatusCode.h
```diff
@@ -36,6 +36,7 @@ typedef enum HttpStatusCode
 {
   SccNone                   = 0,     // Undefined
   SccOk                     = 200,   // Success
+  // FIXME P10: SccFound 302 will be removed at the end once #787 gets implemented
```
fgalan: 302 will be removed when you implement the new serviceRoutines part for postQueryContext and updateQueryContext. By the moment, we need to keep to avoid awful compliation errors.

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -541,8 +541,9 @@ static void prepareDatabaseSeveralCprs2(void)
 *
 * noPatternAttrsAll -
 *
-* Discover:  E3 -no attrs
-* Result:    E3 - (A1, A2, A3) - http://cr1.com
+* Query:  E3 - <null>
+* Result: E3 - (A1, A2, A3) - http://cr1.com
+*
 */
 TEST(mongoContextProvidersQueryRequest, noPatternAttrsAll)
```
fgalan: Appart of the three new tests (severalCprsX) the old test has been modified and adapted. Now, they are not based on 302 Found but they work in a more proper way.

I would be great if you could have a look and check that everything works as you will expectd in the postQueryContext serviceRoutine (as they will play the role than the unit tests is taking now but for real).

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -1451,27 +2065,27 @@ TEST(mongoContextProvidersQueryRequest, severalCprs1)
   EXPECT_EQ("", RES_CER_ATTR(0, 0)->providingApplication);
 
   EXPECT_EQ("A2", RES_CER_ATTR(0, 1)->name);
-  EXPECT_EQ("T", RES_CER_ATTR(0, 1)->type);
+  EXPECT_EQ("", RES_CER_ATTR(0, 1)->type);
```
fgalan: Important to know, this has changed regarding last time you check these tests.

Attribute type is not returned in the case of forwarded attributes but take into account you don't need it. I mean, when you forward a queryContext to a potential CPr, you cannot include the attribute type in the queryContext.

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -1442,7 +2056,7 @@ TEST(mongoContextProvidersQueryRequest, severalCprs1)
   EXPECT_EQ("E1", RES_CER(0).entityId.id);
   EXPECT_EQ("T", RES_CER(0).entityId.type);
   EXPECT_EQ("false", RES_CER(0).entityId.isPattern);
-  EXPECT_EQ("", RES_CER(0).providingApplication);
+  EXPECT_EQ(0, RES_CER(0).providingApplicationList.size());
```
fgalan: As disccussed time ago, "general" entity-level CPr is now a list instead of a single value.

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -1586,8 +2192,8 @@ TEST(mongoContextProvidersQueryRequest, severalCprs2)
 * Result: E1 - A1 - 1
 *              A2 - fwd CPR1
 *              A3 - fwd CPR2
-*              <null> - fwd CPR2
-*              <null> - fwd CPR3
+*
+*         CPR vector: [CPR2, CPR3]
```
fgalan: As disccussed time ago, "general" entity-level CPr is now a list instead of a single value.

src/lib/ngsi/ContextElement.cpp
kzangeli: The class std::vector could implement the operator '=' as to do what you are proposing.
However, how can it know exactly what has to be done for each type of vector?
Imagine a ContextElement-vector, that has a ContextAttribute-vector inside, which in its turn has a Metadata-vector inside ...

Seems impossible to implement,
Perhaps it is possible in an interpreter, but a compiler ... I doubt it.

I think you will have to implement the 'cloning' yourself.

src/lib/ngsi/ContextElementResponse.cpp
kzangeli: No probs

NTC

src/lib/rest/HttpStatusCode.h
kzangeli: Yeah, I guess Sccfound will no longer be necessary,
as that information comes in a different way with the new implementation.

NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: Sure, I will take a close look.
However, an even closer look will 'be taken' once I start to implement my part of this whole change ...

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: Yeah, just the name ...

src/lib/ngsi/ContextElement.cpp
fgalan: For complex cases (vector which elements are objects that could have vectors in sequences) I agree that implementing ourselves will be better. However, in this case `providingApplicationList` is just a vector of std::string.

Looking to http://www.cplusplus.com/reference/vector/vector/operator=/ it seems that vector contents are copied. Thus, I think it should work. However, I'd add a comment in the code about this.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -598,38 +598,56 @@ void destroyAllOntimeIntervalThreads(std::string tenant)
...
+      idMatch = false;
+    }
+    if (regexec(&regex, en1->id.c_str(), 0, NULL, 0) == 0)
+    {
+      idMatch = true;
+    }
+    else
+    {
+      idMatch = false;
+    }
```
kzangeli: If you initialize idMatch to false, you can save five lines of source code.

Also, if regcomp fails, you probably don't want to execute regexec.
I'd propose to do 

    else if (regexec(&re ...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -598,38 +598,56 @@ void destroyAllOntimeIntervalThreads(std::string tenant)
...
-            regfree(&regex);
-        }
-        else {  /* isPattern=false */
-            idMatch = (en2->id == en.id);
-        }
+/* ****************************************************************************
+*
+* includedEntity -
+*/
+bool includedEntity(EntityId en, EntityIdVector* entityIdV) {
```
kzangeli: For new code, I would put the '{' on a new line - following our style guide

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -598,38 +598,56 @@ void destroyAllOntimeIntervalThreads(std::string tenant)
...
-        /* Note that type == "" is like a * wildcard */
-        if (idMatch && (en.type == "" || en2->type == "" || en2->type == en.type)) {
-            return true;
-        }
-    }
-    return false;
+  for (unsigned int ix = 0; ix < entityIdV->size(); ++ix)
+  {
+    if (matchEntity(&en, entityIdV->get(ix)))
+      {
```
kzangeli: Funny indentation

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -598,38 +598,56 @@ void destroyAllOntimeIntervalThreads(std::string tenant)
...
 
-        /* Note that type == "" is like a * wildcard */
-        if (idMatch && (en.type == "" || en2->type == "" || en2->type == en.type)) {
-            return true;
-        }
-    }
-    return false;
+  for (unsigned int ix = 0; ix < entityIdV->size(); ++ix)
+  {
+    if (matchEntity(&en, entityIdV->get(ix)))
```
kzangeli: EntityIdVector implements the operator[].
I would use it:

  if (matchEntity(&en, entityIdV[ix]))

But before doing that you might have to change *EntityIdVector* *entityIdV* for *const EntityIdVector& entityIdV* in the signature of this function (and in the call, of course).

fgalan: Did we added operator[] recen tly to EntityIdVector class? Maybe should get rid off `get()` method (in general for all vectdor-like classes implementing the [] operator)?

kzangeli: Most vectors have the operator, but not all of them.
I add the operator a little 'on the fly', whenever I need them.
When was this one added? I don't know, perhaps a year ago, perhaps six months ...

And yes, it is not a bad idea to stop using the **get** method. The code gets a little bit more readable with the operator.

fgalan: I will implement the change in this place and open a tech debt issue about using operator[] in all our vector-like classes and complete removel of `get()` in all places.

kzangeli: perfect

fgalan: TechDebt issue: https://github.com/telefonicaid/fiware-orion/issues/810

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -598,38 +598,56 @@ void destroyAllOntimeIntervalThreads(std::string tenant)
```
fgalan: Compring the test harness that have failed:

```
```
  o  117_convop_using_standard_ops/postContextEntitiesByEntityId.test
  o  117_convop_using_standard_ops/postContextEntityAttributes.test
  o  117_convop_using_standard_ops/postEntityByIdAttributeByName.test
  o  499_context_providers_basic/query_redirect_operation.test
  o  499_context_providers_basic/query_redirect_three_providers.test
  o  499_context_providers_basic/update_redirect_delete_operation.test
  o  499_context_providers_basic/update_redirect_operation.test
  o  674_servicepath_in_notifications/service_path_in_onchange_notification.test
  o  674_servicepath_in_notifications/service_path_in_ontimeinterval_notification.test
  o  674_servicepath_in_notifications/service_path_in_single_onchange_notification.test
  o  675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_both_cities_with_isPattern_and_deeper_service_path.test
  o  675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_exact_city_with_isPattern_and_deeper_service_path.test
  o  675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_with_isPattern_and_deeper_service_path.test
  o  675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_with_recursive_service_path.test
  o  675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_with_recursive_service_path_and_isPattern.test
  o  714_propagate_service_path/no_propagation_of_service_path_when_not_present.test
  o  715_forward_xauth_token_to_cprs/propagate_xauth_token_for_update_and_query_standard_ops.test
  o  716_forward_registered_attributes/forward_attribute_query.test
  o  719_servicepath_for_registrations/forward_attribute_with_servicePath.test
  o  719_servicepath_for_registrations/forward_queries.test
```
```

and the following check of .test containing the "CP" word:

```
fermin@debvm:~/src/fiware-orion/test/functionalTest/cases$ grep -R -c 'CP' * | grep -v '0$' | grep '.test'
117_convop_using_standard_ops/postContextEntitiesByEntityId.test:12
117_convop_using_standard_ops/postEntityByIdAttributeByName.test:12
117_convop_using_standard_ops/postContextEntityAttributes.test:12
499_context_providers_basic/query_redirect_three_providers.test:105
499_context_providers_basic/update_redirect_delete_operation.test:24
499_context_providers_basic/update_redirect_operation.test:46
499_context_providers_basic/query_redirect_operation.test:29
673_xauth_token/auth_token_three_steps.test:12
713_propagate_tenant/propagate_tenant_for_update_and_query_standard_ops.test:34
714_propagate_service_path/propagate_service_path_for_update_and_query_standard_ops.test:38
716_forward_registered_attributes/forward_attribute_query.test:13
716_forward_registered_attributes/forward_attribute.test:16
719_servicepath_for_registrations/register_and_discovery_with_invalid_service_paths.test:3
719_servicepath_for_registrations/register_and_discovery_with_servicepath.test:2
719_servicepath_for_registrations/register_and_discovery_with_many_servicepaths.test:5
719_servicepath_for_registrations/forward_queries.test:45
719_servicepath_for_registrations/forward_attribute_with_servicePath.test:16
```

fgalan: It is remarkable that case of 674 and 675, as they are not related with CPrs and they are failing (indeed, they seem to be related with subscriptions).

I will look into them before merging this PR.

fgalan: Fixed in 2e65343 (it was easy).

After that, the set of failing tests is: 

```
  o  117_convop_using_standard_ops/postContextEntitiesByEntityId.test
  o  117_convop_using_standard_ops/postContextEntityAttributes.test
  o  117_convop_using_standard_ops/postEntityByIdAttributeByName.test
  o  499_context_providers_basic/query_redirect_operation.test
  o  499_context_providers_basic/query_redirect_three_providers.test
  o  499_context_providers_basic/update_redirect_delete_operation.test
  o  499_context_providers_basic/update_redirect_operation.test
  o  714_propagate_service_path/no_propagation_of_service_path_when_not_present.test
  o  715_forward_xauth_token_to_cprs/propagate_xauth_token_for_update_and_query_standard_ops.test
  o  716_forward_registered_attributes/forward_attribute_query.test
  o  719_servicepath_for_registrations/forward_attribute_with_servicePath.test
  o  719_servicepath_for_registrations/forward_queries.test
```

All them corresponding to the CPs set.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1268,6 +1286,26 @@ bool entitiesQuery
...
+            {
+              found = true;
+              break;
+            }
+          }
+          if (!found)
+          {
+            ContextAttribute* caP = new ContextAttribute(attrName, "", "", false);
+            cer->contextElement.contextAttributeVector.push_back(caP);
+          }
```
kzangeli: There are a few lines to save here also (4) - if you skip the variable 'found' and instead of setting found = true in the 'if-body' you just create and push the caP to cer->contextElement.contextAttributeVector.
The second if is no longer necessary, of course.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
         cerV->push_back(cer);
     }
 
+    /* All the not-patterned entities in the request not in the response are added (without attributes), as they are
+     * used before prunning in the CPr calculation logic */
```
kzangeli: Here we have it - pruning :-)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+            needToAdd = false;
+            break;  /* jx */
+          }
+        }
+        if (needToAdd)
+        {
+          ContextElementResponse* cerP = new ContextElementResponse();
+          cerP->contextElement.entityId.id = enV.get(ix)->id;
+          cerP->contextElement.entityId.type = enV.get(ix)->type;
+          cerP->contextElement.entityId.isPattern = "false";
```
kzangeli: If you align the '='s, the code gets clearer

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+
+          /* This entity has to be prunned if after CPr searching no attribute is "added" to it. The notPrune attribute distinguish this kind of
```
kzangeli: pruned (I think :-))

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+    /* All the not-patterned entities in the request not in the response are added (without attributes), as they are
+     * used before prunning in the CPr calculation logic */
+    for (unsigned int ix = 0; ix < enV.size(); ++ix)
+    {
+      if (enV.get(ix)->isPattern != "true")
+      {
+        bool needToAdd = true;
+        for (unsigned int jx = 0; jx < cerV->size(); ++jx)
+        {
+          if (cerV->get(jx)->contextElement.entityId.id == enV.get(ix)->id && cerV->get(jx)->contextElement.entityId.type == enV.get(ix)->type)
```
kzangeli: I would put 'cerV->get(jx)->contextElement.entityId.id == enV.get(ix)->id' within parenthesis, and also of course, what's to the right of '&&', for better readibility

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+    }
+
     return true;
 }
 
+/* ****************************************************************************
+*
+* pruneNotFoundContextElements -
+*
+* Remove attributes in the vector with 'found' value is 'false'
```
kzangeli: with 'found' == 'false'
OR
where 'found' is 'false'

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+  for (unsigned int ix = 0; ix < oldCerV.size(); ++ix)
+  {
+    ContextElementResponse* cerP = oldCerV.get(ix);
+    ContextElementResponse* newCerP = new ContextElementResponse();
+
+    /* Note we cannot use the ContextElement::fill() method, given that it also copies the ContextAttributeVector. The side-effect
+     * of this is that attributeDomainName and domainMetadataVector and not being copied, but it should not be a problem, given that
+     * domain attributes are not implemented */
+    newCerP->contextElement.entityId.fill(&cerP->contextElement.entityId);
+    newCerP->contextElement.providingApplicationList = cerP->contextElement.providingApplicationList;  // FIXME P10: not sure if this is the right way of doing, maybe we need a fill() method for this
```
kzangeli: I would implement a fill method:

    newCerP->contextElement.providingApplicationList.fill(cerP->contextElement.providingApplicationList);

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+        if (needToAdd)
+        {
+          ContextElementResponse* cerP = new ContextElementResponse();
+          cerP->contextElement.entityId.id = enV.get(ix)->id;
+          cerP->contextElement.entityId.type = enV.get(ix)->type;
+          cerP->contextElement.entityId.isPattern = "false";
+
+          /* This entity has to be prunned if after CPr searching no attribute is "added" to it. The notPrune attribute distinguish this kind of
+           * entities from the "real ones" that are without attributes in the Orion local database (a rare case, but may happen) */
+          cerP->notPrune = false;
```
kzangeli: It is not a great idea to use 'negative varibles'.
Sometimes it gets unnecesssarily complicated.

Just a suggestion: why not use **cerP->prune = true**?

Imagine this (a little exaggerated perhaps):

    if (!cerP->notPrune == false)

It makes my head spin :-)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+
+          for (unsigned int jx = 0; jx < attrL.size(); ++jx)
+          {
+            ContextAttribute* caP = new ContextAttribute(attrL.get(jx), "", "");
+            caP->found = false;
```
kzangeli: If you want to, you could add a parameter to the ContextAttribute constructor ('found' that defaults to false), or even create a new constructor.
Only if you want to ...

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+     * domain attributes are not implemented */
+    newCerP->contextElement.entityId.fill(&cerP->contextElement.entityId);
+    newCerP->contextElement.providingApplicationList = cerP->contextElement.providingApplicationList;  // FIXME P10: not sure if this is the right way of doing, maybe we need a fill() method for this
+    newCerP->statusCode.fill(&cerP->statusCode);
+
+    bool notPruneEntity = cerP->notPrune;
+
+    for (unsigned int jx = 0; jx < cerP->contextElement.contextAttributeVector.size(); ++jx)
+    {
+      ContextAttribute* caP = cerP->contextElement.contextAttributeVector.get(jx);
```
kzangeli: operator[] exists for ContextAttributeVector.
I know there is an issue for this, but at least for new code we should try to use the operator if it already exists.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+      if (caP->found)
+      {
+        ContextAttribute* newCaP = new ContextAttribute(caP);
+        newCerP->contextElement.contextAttributeVector.push_back(newCaP);
+      }
+
+    }
+
+    /* If after prunning the entity has no attribute and no CPr information, then it is not included
```
kzangeli: pruning

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+     * in the output vector, except if "notPrune" is set to true */
+    if (!notPruneEntity && newCerP->contextElement.contextAttributeVector.size() == 0 && newCerP->contextElement.providingApplicationList.size() == 0)
```
kzangeli: I just hate double negations!
Might be a problem with my brain, I don't know.

So much easier to read
    if (pruneEntity)
than
    if (!notPruneEntity)

[ just my opinion ]

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1275,9 +1313,97 @@ bool entitiesQuery
...
+    {
+      newCerP->release();
+      delete newCerP;
+    }
```
kzangeli: This here got me thinking ...
Is providingApplicationList freed in ContextElement::release()?
If not, we have a leak here.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1338,9 +1464,13 @@ static void processContextRegistrationElement (BSONObj cr, EntityIdVector enV, A
...
     if (crr.contextRegistration.entityIdVector.size() > 0) {
-        std::vector<BSONElement> queryAttrV = cr.getField(REG_ATTRS).Array();
-        for (unsigned int ix = 0; ix < queryAttrV.size(); ++ix) {
+        if (cr.hasField(REG_ATTRS)) /* To prevent registration in the E-<null> style */
+        {
+          std::vector<BSONElement> queryAttrV = cr.getField(REG_ATTRS).Array();
+          for (unsigned int ix = 0; ix < queryAttrV.size(); ++ix)
+          {
             processAttribute(&crr, attrL, queryAttrV[ix].embeddedObject());
+          }
```
kzangeli: I don't see what has changed here.
A for-loop over an empty array should be the same (nothing) as a check of emptiness before the loop.
Unless I'm wrong, this change doesn't do anything but making the code a little more complex?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1631,16 +1761,19 @@ bool processOnChangeCondition
...
-    if (!entitiesQuery(enV, attrL, res, &ncr.contextElementResponseVector, &err, false, tenant, servicePathV))
+    ContextElementResponseVector rawCerV;
+    if (!entitiesQuery(enV, attrL, res, &rawCerV, &err, false, tenant, servicePathV))
     {
         ncr.contextElementResponseVector.release();
         return false;
     }
 
+    /* Prune "not found" CERs */
+    pruneNotFoundContextElements(rawCerV, &ncr.contextElementResponseVector);
```
kzangeli: Isn't *pruneContextElements* a good enough name?
I mean, the semantics of 'prune' already implies 'not found', right?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1651,17 +1784,20 @@ bool processOnChangeCondition
         if (condValues != NULL) {
             /* Check if some of the attributes in the NotifyCondition values list are in the entity.
              * Note that in this case we do a query for all the attributes, not restricted to attrV */
-            ContextElementResponseVector allCerV;
+            ContextElementResponseVector rawCerV, allCerV;
```
kzangeli: Please use one line per variable (unspoken style-guide rule).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1651,17 +1784,20 @@ bool processOnChangeCondition
...
             AttributeList emptyList;
             // FIXME P10: we are using dummy scope by the moment, until subscription scopes get implemented
             // FIXME P10: we are using an empty service path vector until serive paths get implemented for subscriptions
-            if (!entitiesQuery(enV, emptyList, res, &allCerV, &err, false, tenant, servicePathV))
+            if (!entitiesQuery(enV, emptyList, res, &rawCerV, &err, false, tenant, servicePathV))
             {
                 allCerV.release();
```
kzangeli: allCerV was just created and is empty.

    rawCerV.release(); ?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: It's a little like "kill, to make dead" :-)

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
```diff
@@ -132,15 +132,19 @@ HttpStatusCode mongoGetContextElementResponses(const EntityIdVector& enV, const
...
     // FIXME P10: we are using dummy scope by the moment, until subscription scopes get implemented
     // FIXME P10: we are using an empty service path vector until service paths get implemented for subscriptions
+    ContextElementResponseVector rawCerV;
     std::vector<std::string> servicePath;
     Restriction res;
-    if (!entitiesQuery(enV, attrL, res, cerV, err, true, tenant, servicePath))
+    if (!entitiesQuery(enV, attrL, res, &rawCerV, err, true, tenant, servicePath))
     {
         reqSemGive(__FUNCTION__, "get context-element responses (no entities found)");
         cerV->release();
```
kzangeli: Is cerV the correct variable here?
Should it be rawCerV.release() ?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
 
 /* ****************************************************************************
 *
+* someContextElementNotFound -
+*
+* Returns true if some attribut with 'found' set to 'false' is found in the CER vector passed
```
kzangeli: attribute

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* as argument
+*
+*/
+bool someContextElementNotFound(ContextElementResponseVector& cerV)
+{
+  for (unsigned int ix = 0; ix < cerV.size(); ++ix)
+  {
+    for (unsigned int jx = 0; jx < cerV.get(ix)->contextElement.contextAttributeVector.size(); ++jx)
+    {
+      if (!cerV.get(ix)->contextElement.contextAttributeVector.get(jx)->found)
```
kzangeli: operator[]

src/lib/ngsi/ContextElement.cpp
fgalan: Fixed in 0e0e492

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in 0e0e492

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in 0e0e492

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in 0e0e492

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in 0e0e492

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+  }
+  return false;
+}
+
+
+/* ****************************************************************************
+*
+* searchCprForAttribute -
+*
+* Search the CPr, given the entity/attribute as argument. Actually, two CPrs can be returned
```
kzangeli: returned:

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I thought of that when I implemented this part, but it get complex... C++ doesn't allow named parameters in functions (AFAIK) thus depending on which fields you want to use in the constructor, it gets complex.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Ok, I'll do that.

Just curiosity... is the "style checker" script detect this kind of style-break?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+  perEntPa  = "";
+  perAttrPa = "";
+  for (unsigned int ix = 0; ix < crrV.size(); ++ix)
+  {
+    ContextRegistrationResponse* crr = crrV.get(ix);
+    /* Is there a matching entity in the CRR? */
+    for (unsigned jx = 0; jx < crr->contextRegistration.entityIdVector.size(); ++jx)
+    {
+      EntityId* regEn = crr->contextRegistration.entityIdVector.get(jx);
+      if (regEn->id == en.id && (regEn->type == en.type || regEn->type == ""))
```
kzangeli: For 'pre-conditions' is it easier to read the code if you just 'continue if not fulfilled':

    if (regEn->id != en.id || ((regEn->type != en.type) && (regEn->type != "")))
        continue;

After this, the entity matches and the loop can continue without extra blocks and indentations.
Very good 'best practise' IMHO.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: This would involve to become providingApplicationList into a class. Do you think is a good idea? A bit overwhelming for something that is basically a vector of strings and nothing more...

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+      {
+        if (crr->contextRegistration.contextRegistrationAttributeVector.size() == 0)
+        {
+          /* CRR without attributes (we need to keep searching in other CRR) */
+          perEntPa = crr->contextRegistration.providingApplication.get();
+          break; /* jx */
+        }
+        else
```
kzangeli: As the previous if breaks if entered, there is no reason for the 'else':

    if (crr->contextRegistration.contextRegistrationAttributeVector.size() == 0)
    {
        /* CRR without attributes (we need to keep searching in other CRR) */
       perEntPa = crr->contextRegistration.providingApplication.get();
       break; /* jx */
    }

    // Is there a matching entity or the abcense of attributes?
    for (unsigned kx = 0; kx < crr->contextRegistration.contextRegistrationAttributeVector.size(); ++kx)

This is another 'pre-condition'.
As always, this best practice saves you lines and intentations (adding to the readibility).

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+              /* We cannot "improve" this result keep searching in CRR vector, so we return */
+              perAttrPa = crr->contextRegistration.providingApplication.get();
+              return;
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
kzangeli: I see now that *perAttrPa* and *perEntPa* are returned from this function.
To make that clear (especially from the caller), these two variables should be passed by reference, I mean, as real pointers, using the '&' in the call, and received as real pointers in this function.

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+
+/* ****************************************************************************
+*
+* fillContextProviders -
+*
+* Looks in the elements of the CER vector passed as argument, searching for a suitable CPr in the CRR
+* vector passes as argument. If a suitable CPr is found, it is added to the CER (and the 'found' field
```
kzangeli: passed

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* changed to true)
```
kzangeli: is changed

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Not sure if this works... note that I create & push if I *don't find* anything (not if I find).

Maybe if you could show me the complete code snipped it would be more clear to me...

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+*/
+void fillContextProviders(ContextElementResponseVector& cerV, ContextRegistrationResponseVector& crrV)
+{
+  for (unsigned int ix = 0; ix < cerV.size(); ++ix)
+  {
+    ContextElementResponse* cer = cerV.get(ix);
+    for (unsigned int jx = 0; jx < cer->contextElement.contextAttributeVector.size(); ++jx)
+    {
+      ContextAttribute* ca = cer->contextElement.contextAttributeVector.get(jx);
+      if (!ca->found)
```
kzangeli: if (!ca->found)
        continue;

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+      cerV.get(ix)->contextElement.providingApplicationList.push_back(pa);
+      return;    /* by construction, no more than one CER with the same entity information should exist in the CERV) */
+    }
+  }
+
+  /* Reached this point, it means that the cerV doesn't contain a proper CER, so we create it */
+  ContextElementResponse* cerP = new ContextElementResponse();
+  cerP->contextElement.entityId.id = enP->id;
+  cerP->contextElement.entityId.type = enP->type;
+  cerP->contextElement.entityId.isPattern = "false";
```
kzangeli: align '='s

If malloc fails (no memory left), the broker will crash right here (cerP is NULL).
Malloc usually don't fail and I'm sure we have this problem is a hundred places.
No reason to add to it though :-)

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+  cerP->contextElement.providingApplicationList.push_back(pa);
+
+  cerP->statusCode.fill(SccOk);
+
+  cerV.push_back(cerP);
+
```
kzangeli: unnecessary empty line

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+*
+* addContextProviderAttribute -
+*
+*
+*/
+void addContextProviderAttribute(ContextElementResponseVector& cerV, EntityId* enP, ContextRegistrationAttribute* craP, std::string pa)
+{
+  for (unsigned int ix = 0; ix < cerV.size(); ++ix)
+  {
+    if (cerV.get(ix)->contextElement.entityId.id == enP->id && cerV.get(ix)->contextElement.entityId.type == enP->type)
```
kzangeli: if (NOT)
      continue;

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+    {
+      for (unsigned int jx = 0; jx < cerV.get(ix)->contextElement.contextAttributeVector.size(); ++jx)
+      {
+        std::string attrName = cerV.get(ix)->contextElement.contextAttributeVector.get(jx)->name;
+        if (attrName == craP->name)
```
kzangeli: crap name? :-D

NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: no, I don't think it does.
We have a lot of style rules that are not detected by the script.
The style check script is in python and you are more than welcome to add to it ... :-)
I would do it but my python knowledge is below zero :-(

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+      cerV.get(ix)->contextElement.contextAttributeVector.push_back(caP);
+      return;
+    }
+  }
+
+  /* Reached this point, it means that the cerV doesn't contain a proper CER, so we create it */
+  ContextElementResponse* cerP = new ContextElementResponse();
+  cerP->contextElement.entityId.id = enP->id;
+  cerP->contextElement.entityId.type = enP->type;
+  cerP->contextElement.entityId.isPattern = "false";
```
kzangeli: align '='s (also not detected by the style guide checker)

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* processGenericEntities() function)
+*
+*/
+void addContextProviders(ContextElementResponseVector& cerV, ContextRegistrationResponseVector& crrV, EntityId* enP = NULL)
+{
+  for (unsigned int ix = 0; ix < crrV.size(); ++ix)
+  {
+    ContextRegistration cr = crrV.get(ix)->contextRegistration;
+
+    /* In the case a "filtering" entity was provided, check that the current CRR match or skip to need CRR */
```
kzangeli: current CRR matches or skip to next CRR ?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+  }
+  return false;
+}
+
+
+/* ****************************************************************************
+*
+* addContextProviders -
+*
+*
```
kzangeli: remove extra line

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* This functions takes a CRR vector and adds the Context Providers in the CER vector
```
kzangeli: function

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* (except the ones corresponding to some locally found attribute, i.e. info already in the
+* CER vector)
+*
+* The enP paramter is optional. If not NULL, then before adding a CPr the function check that the
```
kzangeli: before adding a CPr, the function checks that

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+* containting CRR match the entiy (this is used for funcionality related with "generic queries", see
```
kzangeli: matches
related to

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Ah sorry, I thought it was a class already ...
I was thinking of AttributeList, I guess.
AttributeList is the same case of a vector of strings and I made a class of it.
However, AttributeList is part of payload and it needs its render and check methods, so it is not the same,

If this assignment works (as I think you verified earlier), this good enough, even better than making a class of it.

NTC

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+      /* Registration without attributes */
+      for (unsigned int jx = 0; jx < cr.entityIdVector.size(); ++jx)
+      {
+        addContextProviderEntity(cerV, cr.entityIdVector.get(jx), cr.providingApplication.get());
+      }
+    }
+    else
+    {
+      /* Registration with attributes */
+      for (unsigned int jx = 0; jx < cr.entityIdVector.size(); ++jx)
```
kzangeli: I try to use indexes (indices?) to illustrate a little what they are counting,
For entityIdVector I would use *eIx*, for contextRegistrationAttributeVector *craIx*.
It's a little nicer than jx, kx ...

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Ah, you're right, I didn't see the '!' in the 'if'.
My mistake, sorry

NTC

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Looking closely to this, it seems that constructor is:

`ContextAttribute(const std::string& _name, const std::string& _type, const std::string& _value = "", bool _found = true);`

Thus, I think I can use it.

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -38,6 +38,268 @@
...
+      }
+    }
+  }
+}
+
+/* ****************************************************************************
+*
+* processGenericEntities -
+*
+*
```
kzangeli: extra line

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -85,33 +349,94 @@ HttpStatusCode mongoQueryContext
...
-          LM_T(LmtCtxProviders, ("context provide found: %s", prApp.c_str()));
-          responseP->errorCode.fill(SccFound, prApp);
-          crrV.release();
-          return SccOk;
+          processGenericEntities(requestP->entityIdVector, rawCerV, crrV);
         }
       }
       else
       {
+        /* Different from fails in DB at entitiesQuery(), DB fails at registrationsQuery() are not considered "critical" */
```
kzangeli: fails => errors

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: providingApplicationList is a std::vector. I tend to think that the std::vector destructor will invoke the destructor of each element it contains, and I think that the std::string destructor will release memory properly.

Probably is simpler to run the valgrind test than thinking on this :)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Ah, it's there already?
Surprising ... :-)
And no, there cannot be any named parameters in C/C++.
I would say there cannot be named parameters in any compiled language.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: The change is the `hasField()` checking. With the new logic (I'm not sure why... I did when I started working on this and I dont' remember very well... something related with E-null> I have some problems if I don't include that.

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -109,7 +110,7 @@ ContextAttribute::ContextAttribute
   compoundValueP        = NULL;
   typeFromXmlAttribute  = "";
   providingApplication  = "";
-  found                 = true;
+  found                 = _found;
```
kzangeli: Ah, ok. You added _found. I misunderstood.
Great

NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yes

NTC

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: The parameters is a C++ reference, thus it is valid for returning things. If the problem is the function preamble comment, maybe you could suggest a better wording, please?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: ok, I still don't see how a check of zero could change a loop over zero attributes, but, no problem ...

NTC

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: Yes, valid according to the compiler but not according the style guide.
We talked about thissome time ago and even looked up 'googles opinion' or something like that.

If a function is to change the value of a parameter, it should be visible from the outside, like this:

    int x = 19;
    changerFunction(&x);

That ampersand shows clearly that changerFunction might change the value of x.

Doing it the C++ reference way (horrible idea in my opinion), the fact that changerFunction might change the value of x is hidden.
This is very confusing, Very C++ ...

Now, what we decided was to send as real C-pointers all parameters that might be altered by the function.
If not altered, and 'complex' (like many strings), we use "const XXX& par", to send only the reference instead of copying the entire object on the stack.

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: I'm declaring variables as references in function signatures when they are to be used for passing back results. I think thats enouthg... otherwise the unit test wouldn't work. Maybe we can discuss this at skype.

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: However, our current style is to use ix, jx, etc. in all places... isn't it?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: It's already there, but because I modified the constructor to add it ;)

Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Related with a previous comment about the same topic. Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in b67b271

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: The new line is this:

cr.hasField(REG_ATTRS)

Is not cheking cero-sized... is checking that the field REG_ATTRS exists in the cr BSONObj. We need that, because if the field doesn't exists... segfault.

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in b67b27152a07007204

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: If there is only one for-loop, no problem with using 'ix', whatever it refers to.
It is much better to use eIx, attrIx, or whatever, of course, but, not a problem.
The problem arises when we have two different indices. There is gets confusing with 'non-saying' variable names.

I don't say I always use eIx etc, but I try. Especially older code might be full of 'ix'.
We should try to always use better names for the indices.
As part of our style guide.
If you don't agree to this, let's talk and come to a conclusion.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Ah, ok, now I understand. [sorry]

NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -563,11 +564,37 @@ TEST(mongoContextProvidersQueryRequest, noPatternAttrsAll)
   /* Check response is as expected */
   EXPECT_EQ(SccOk, ms);
 
-  EXPECT_EQ(SccFound, res.errorCode.code);
-  EXPECT_EQ("Found", res.errorCode.reasonPhrase);
-  EXPECT_EQ("http://cr1.com", res.errorCode.details);
+  EXPECT_EQ(SccNone, res.errorCode.code);
```
kzangeli: I never liked this SccNone.

First doubt, should any mongoBackend 'main' functions ever return with SccNone?
[ I understood this should never happen ... to be considered an error if it ever happened ]

If this SccNone is in errorCode.code, should it be treated as an OK?
And the calling layer will always have to test for ((code == SccOk) || (code == SccNone)).

[ In some cases the SccNone is transformed into a 500 - internal error ... ]

[[ I would initialize the errorCode to 200 OK and in case there is any error, that would be changed by mongBackend.
However, I think we've discussed this more times than the ciP :-D. ]]

I think we perhaps need to revisit this whole thing.

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -625,11 +668,14 @@ TEST(mongoContextProvidersQueryRequest, noPatternAttrOneSingle)
...
-* Discover:  E1 - A1
-* Result:    E1 - A1 - http://cr1.com
-*            E1 - A1 - http://cr2.com
+* Query:  E1 - A1
+* Result: E1 - A1 - http://cr1.com
+*         E1 - A1 - http://cr2.com
 *
-* This test also checks that discovering for type (E1) doesn't match with no-typed
+* In this case, htpt://cr2.com also matches, but current implementation of mongoBackend
+* returns only one of them (which one is not specified).
```
kzangeli: ok ...
Is it very difficult to fix this? (to return both) 
Will this give us problems for the unit tests in different machines (and in coming functests)?

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
fgalan: If the value is SccOk by default and SccOk is part of the valid values that mongoBackend may set, you cannot distinguish between this situations: 1) the result was actually SccOk, 2) the result was not SccOk but mongoBackend didn't change the value due to some internal error.

However, if you use as default a value SccNone you can detect case 2)

In the case of re-visiting this, I'd do that process outside the scope of this PR (otherwise, the PR could be delayed while discussion progress).

NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
```diff
@@ -717,9 +800,12 @@ TEST(mongoContextProvidersQueryRequest, noPatternAttrsSubset)
...
 * noPatternSeveralCREs -
 *
-* Discover:  E1 - no attrs
-* Result:    E1 - (A1, A2, A3) - http://cr1.com
-*            E1 - (A1, A4)     - http://cr2.com
+* Query:  E1 - no attrs
+* Result: E1 - A1, A2, A3 - http://cr1.com
+*              A4         - http://cr2.com
+*
+* For A1 two CPrs overlap and mongoBacken takes http://cr1.com
```
kzangeli: Similar problem here, not so much for queries but definitely for updates.
Is an issue created to explain and remember these problems (this and the previous)?

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
fgalan: It is not easy... the logic in mongoBackend will change again.

However, don't worry about different machines. Although we only return one, it is deterministic (it is based on registrationQuery(), which returns results ordered by registration id).

kzangeli: ok
issue?

fgalan: I'd wait for PO to define that requirement as a functional piece of Orion backlog.

fgalan: NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
fgalan: See answer to the other comment.

NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: Yes, I know your opinion about this. I just don't agree ... :-D
Never mind that discussion for now.

Now, SccNone is returned and should be treated as an internal error and the upper layer must return 500 ...
Because here, 'case 2' will be detected.
See my confusion?
Or, have I misunderstood again :-) ?

And of course, we will not discuss this philosophic question as part of this PR.

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: NTC

test/unittests/mongoBackend/mongoContextProvidersQuery_test.cpp
kzangeli: NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Discussed over skype - it should be rawCerV.release();

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
kzangeli: Discussed over skype - should be rawCerV.release()

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in a6d062b

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
fgalan: Fixed in a6d062b

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in 32035a4

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in 32035a4

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed in 850c438

kzangeli:
LGTM

PR #809, opened 2015-03-04
Change the way to kill ContextBroker

jcalderin:
Use better logger
Add configuration to set the pid file and contextbroker log

### DIFF
--- a/test/acceptance/integration/tools/general_utils.py
+++ b/test/acceptance/integration/tools/general_utils.py
@@ -305,7 +305,7 @@ def start_cb(parms):
     else:
         pid_path = config['pid_file']
     getattr(api, runner)('mkdir -p {pid_dir}'.format(pid_dir=pid_path[:pid_path.rfind('/')]))
-    command = '{bin_path} {parms} -logDir {log_path} -pidpath {pid_path} >> /tmp/cb_acceptance_test.log'.\
+    command = '{bin_path} {parms} -logDir {log_path} -pidpath {pid_path}'.\
         format(bin_path=config['bin_path'], parms=parms, log_path=log_path, pid_path=pid_path)
     getattr(api, runner)(command)
     world.cb_pid = get_cb_pid()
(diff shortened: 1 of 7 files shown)


test/acceptance/integration/tools/general_utils.py
```diff
@@ -291,7 +295,19 @@ def start_cb(parms):
...
     if world.cb_pid != '':
         stop_cb()
-    getattr(api, runner)('{bin_path} {parms} >> /tmp/cb_acceptance_test.log'.format(bin_path=config['bin_path'], parms=parms))
+    if config['log_path'] == '':
+        log_path = '/tmp/acceptance'
+    else:
+        log_path = config['log_path']
+    getattr(api, runner)('mkdir -p {log_path}'.format(log_path=log_path))
+    if config['pid_file'] == '':
+        pid_path = '/tmp/acceptance/contextBroker.pid'
```
fgalan: What would happend if two acceptance runs happens at same time? Maybe is safer to use mktemp for this (cleaning up the directory after that).

Same comment for log_path.

test/acceptance/integration/tools/general_utils.py
```diff
@@ -291,7 +295,19 @@ def start_cb(parms):
...
+    else:
+        pid_path = config['pid_file']
+    getattr(api, runner)('mkdir -p {pid_dir}'.format(pid_dir=pid_path[:pid_path.rfind('/')]))
+    command = '{bin_path} {parms} -logDir {log_path} -pidpath {pid_path} >> /tmp/cb_acceptance_test.log'.\
```
fgalan: Same problem here. An overall grep for "/tmp" in the code could help to locate all cases.

test/acceptance/integration/tools/general_utils.py
jcalderin: The directory are in the properties file, if any wants to execute the tests at the same time, they have to change the directories

NTC

test/acceptance/integration/tools/general_utils.py
jcalderin: Same as before

NTC

fgalan:
LGTM (passing the ball to @iariasleon for final LGTM + merge + delete branch)

iariasleon:
LGTM

PR #813, opened 2015-03-06
Hardening/117 get ngsi10 context entity types attribute

kzangeli:
One more convop that uses std-op:
GET /v1/contextEntityTypes/{entity::type}/attributes/{attribute::name}


fgalan:
LGTM

PR #815, opened 2015-03-09
GET /v1/contextEntities/type/{entity::type}/id/{entity::id} 

kzangeli:
One more convop converted to using std-op

### DIFF
--- a/src/lib/common/statistics.cpp
+++ b/src/lib/common/statistics.cpp
@@ -104,6 +104,7 @@ int noOfRtUnsubscribeContextAvailabilityResponse         = -1;
 int noOfRtUnsubscribeContextResponse                     = -1;
 int noOfRtSubscribeResponse                              = -1;
 int noOfRtSubscribeError                                 = -1;
+int noOfContextElementResponse                           = -1;
 
 int noOfEntityTypesRequest                               = -1;
 int noOfEntityTypesResponse                              = -1;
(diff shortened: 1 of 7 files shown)


src/lib/common/statistics.cpp
```diff
@@ -192,7 +193,7 @@ void statisticsUpdate(RequestType request, Format inFormat)
   case RtUnsubscribeContextResponse:                     ++noOfRtUnsubscribeContextResponse; break;
   case RtSubscribeResponse:                              ++noOfRtSubscribeResponse; break;
   case RtSubscribeError:                                 ++noOfRtSubscribeError; break;
-
+  case RtContextElementResponse:                         ++noOfContextElementResponse; break;
```
fgalan: So one counter was missing until now and this is the bugfix? Need to mention in CHANGES_NEXT_RELEASE or in this case is a very tiny bugfix?

src/lib/serviceRoutines/getAllEntitiesWithTypeAndId.cpp
```diff
@@ -39,6 +41,18 @@
...
 * getAllEntitiesWithTypeAndId - 
+*
+* GET /v1/contextEntities/type/{entity::type}/id/{entity::id}
+* GET /ngsi10/contextEntities/type/{entity::type}/id/{entity::id}
+*
+* Payload In:  None
+* Payload Out: ContextElementResponse
+* 
+* URI parameters:
+*   - attributesFormat=object
```
fgalan: In this case limit/offset/details don't make sense, do they? Almost sure at 99% percent, just to be sure...

kzangeli: Yeah, it's only one output from this request (ContextElementResponse - not a vector).
There is nothing to count.

src/lib/common/statistics.cpp
kzangeli: I only needed this constant for a call to ContextElementResponse::render.
As a side-effect it is added as a counter in statistics.
However, as I haven't included it in the response to a statistics request, an external user will notice no difference.
No big deal really. What we should do is to look into the statistics some day soon ...

fgalan: Ok, NTC.

src/lib/serviceRoutines/getAllEntitiesWithTypeAndId.cpp
fgalan: NTC

fgalan:
LGTM

PR #816, opened 2015-03-09
Hardening/117 post all entities with type and id

kzangeli:
One more conv op using std ops:
POST /v1/contextEntities/type/{entity::type}/id/{entity::id}


fgalan:
LGTM

PR #818, opened 2015-03-10
Hardening/117 delete all entities with type and

kzangeli:
New convop to use std-op:
DELETE /v1/contextEntities/type/{entity::type}/id/{entity::id}

### DIFF
--- a/test/functionalTest/harnessFunctions.sh
+++ b/test/functionalTest/harnessFunctions.sh
@@ -103,9 +103,12 @@ function dbDrop()
 {
   db=$1
 
-  if [ "$db" != "" ]
+  if [ "$CB_DB_DROP" != "No" ]
   then
-    dbInit $db
+    if [ "$db" != "" ]
+    then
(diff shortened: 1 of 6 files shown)


test/functionalTest/harnessFunctions.sh
```diff
@@ -103,9 +103,12 @@ function dbDrop()
 {
   db=$1
 
-  if [ "$db" != "" ]
+  if [ "$CB_DB_DROP" != "No" ]
```
fgalan: Where this variable is set? CLI of testHarness.sh script?

kzangeli: You set it, if you want to avoid to drop the database (my comment in the previous PR):
% CB_DB_DROP=No testHarness.sh XXX.test

fgalan: NTC

fgalan:
LGTM

Issue #46, opened 2013-11-04
Malformed unsubscribeContextRequest not detected


LeandroGuillen:
The Context Broker does not detect this request is malformed:

```
<?xml version="1.0" encoding="UTF-8"?>
<unsubscribeContextRequest>
     <subscriptionId>51a325270a98a06f27db0875</subscriptionId>
</unsubscribeContextRequest>
```

Traces:

https://pdihub.hi.inet/gist/18d9ea259b46927bb2ad

The correct xml is:

```
<?xml version="1.0" encoding="UTF-8"?>
<unsubscribeContextRequest>
    <unsubscribeContext>
        <subscriptionId>51a325270a98a06f27db0875</subscriptionId>
    </unsubscribeContext>
</unsubscribeContextRequest>
```

PR #819, opened 2015-03-10
Fix a compilation problem due to struct sockaddr casting

jmcanterafonseca:


jmcanterafonseca:
@kzangeli r?

fgalan:
I have tested in Cent 6.x (make unit_test and make ft) and compilation works.

LGTM (passing the ball to @kzangeli for final LGTM + merge)

kzangeli:
OK, this slight modification doesn't seem to hurt. (I have compiled and make test is running just in case).
Now, the introduction of 'struct sockaddr' instead of 'struct sockaddr_in', is for some other flavor of Linux?

This PR is with very little information, it would just be nice to know what exactly is fixed with it, before merging it into develop ...

kzangeli:
Actually, looking a little deeper, MHD_ConnectionInfo::client_addr is a pointer to struct sockaddr, so, there is no difference in the two implementations (http://libmicrohttpd.sourcearchive.com/ fooled me into believing MHD_ConnectionInfo::client_addr was a sockaddr_in*, which is probably true in a newer implementation of MHD, but looking at the file I include I find it is a sockaddr).

So, the only difference this PR does it so take two steps, via a 'temporary' pointer to a **struct sockaddr**.

Now, could it be that you are using a newer version of libmicrohttpd, where MHD_ConnectionInfo::client_addr is a pointer to **sockaddr_in** ? [ that at least would explain the need for this modification ... ]

jmcanterafonseca:
In the machine I used to compile Orion the libmicrohttpd I've installed is

````
lrwxrwxrwx 1 root root 22 May 10  2010 libmicrohttpd.so -> libmicrohttpd.so.5.2.1
````

kzangeli:
OK, that's seems to be a really old release of MHD ...

[ judging by the date - the release number 5.2.1 doesn't make much sense as current release is 0.9.34 according to the MHD webpage.
To make it even more confusing, I have the lib libmicrohttpd.so.10.16, while my header says it is 1.9.33. ]

Never mind. I'm ok with the change.

LGTM

jmcanterafonseca:
Oh thanks! I was writing this comment when I saw your merge ;) for the sake of completness I left it here anyway

I was compiling Orion in an Ubuntu machine 12.04 part of the FiLab

````
Linux devrels2.novalocal 3.2.0-29-generic #46-Ubuntu SMP Fri Jul 27 17:03:23 UTC 2012 x86_64 x86_64 x86_64 GNU/Linux
`````

If the patch I've proposed is not applied the following error appears at compilation time

````
make[3]: Entering directory `/home/jmcf/orion/fiware-orion/BUILD_RELEASE'
[ 91%] Building CXX object src/lib/rest/CMakeFiles/rest.dir/rest.cpp.o
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp: In function ‘int connectionTreat(void*, MHD_Connection*, const char*, const char*, const char*, const char*, size_t*, void**)’:
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:736:34: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:736:72: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:738:35: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:739:35: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:740:35: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
/home/jmcf/orion/fiware-orion/src/lib/rest/rest.cpp:741:35: error: ‘struct sockaddr_in’ has no member named ‘sa_data’
make[3]: *** [src/lib/rest/CMakeFiles/rest.dir/rest.cpp.o] Error 1

````

kzangeli:
Yeah, in your MHD_ConnectionInfo, client_addr is sockaddr_in, which is more 'detailed' that sockaddr and it lacks sa_data.
In our centoses, client_addr is sockaddr which contains 'sa_data'. Your typecast from sockaddr_in to sockaddr makes it work in both machines, which is good. We might opt for another version of MHD and that new version might have client_addr of the type sockaddr_in and with this fix it will work without any changes.

Thank you for the fix! :-)

PR #820, opened 2015-03-10
Hardening/117 get individual context entity attribute with type and

kzangeli:
One more convop uses std-op:
GET /v1/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}


fgalan:
LGTM

PR #826, opened 2015-03-11
Hardening/117 delete individual context entity attribute with type and id

kzangeli:
One more convop:
DELETE /v1/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}


fgalan:
LGTM

PR #827, opened 2015-03-12
Hardening/117 get context entities by entity id and type

kzangeli:
Another convop:
GET /v1/registry/contextEntities/type/{entity::type}/id/{entity::id}


src/lib/serviceRoutines/getContextEntitiesByEntityIdAndType.cpp
```diff
@@ -41,8 +43,18 @@
...
 * getContextEntitiesByEntityIdAndType - 
 *
-* Invoked by:
-*  GET /v1/registry/contextEntities/type/{ETYPE}/id/{EID}
+* GET /v1/registry/contextEntities/type/{ETYPE}/id/{EID}
+*
+* URI parameters:
+*   - entity::type=XXX     (must coincide with entity::type in URL)
+*   - !exist=entity::type  (if set - error -- entity::type cannot be empty)
+*   - exist=entity::type   (not supported - ok if present, ok if not present ...)
```
fgalan: We don't support attributeFormat in this case as it doesn't make sense for discoverContextAvailability (in that case attributes doesn't have a value, as in NGS10). Is my argument correct?

kzangeli: We don't support attributeFormat=object for the class ContextRegistrationAttribute.
That's why I didn't implement the URI option for the request.
However, we could support it ...
Not really a part of this PR anyway

fgalan: NTC

fgalan:
LGTM

src/lib/serviceRoutines/getContextEntitiesByEntityIdAndType.cpp
kzangeli: I think it would make a lot of sense to support JSON-object-format for absolutely everything, not only ContextAttribute ...

PR #829, opened 2015-03-12
Hardening/117 post context entities by entity id and type

kzangeli:
Another convop ...
POST /v1/registry/contextEntities/type/{entity::type}/id/{entity::id}


fgalan:
LGTM

PR #833, opened 2015-03-16
hardening/117_getEntityByIdAttributeByNameWithTypeAndId

kzangeli:
Another convop converted:
GET /v1/registry/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}


fgalan:
LGTM

PR #841, opened 2015-03-18
Feature/589 mongo driver timeout tunning

### LINKED ISSUE #589: Allow MongoDB driver tuning (e.g. reduce keep-alive time in DB connections)
When runnning CB with replica set configuration (i.e. using `rplSet` CLI option) it may happen that if the VM running the host where mongo primary runs goes down, the contextBroker may still have connections established to the powered-off machine. It does not acknowledge that mongo primary is down. As a result, all the requests to the CB fails.

Note that this situation is different from the one in which the mongod primary process goes down, in which case, the host OS seems to be able to manage to cut all the connetions with contextBroker, so CB gets reconnected to the new primary (or secondary in read-only mode).

This problem is probably caused by the default keepalive period used by MongoDB driver is too large. We should look at how this can be configured at start time as a driver/connection startup parameter.

**Effort: 2 man day**

fgalan:
Fixes #589 

Ready to be merged. QA Int ( @mcarracedo and @XavierVal ) has given ok.


src/lib/mongoBackend/MongoGlobal.h
```diff
@@ -150,9 +150,15 @@ using namespace mongo;
...
-extern bool mongoConnect(const char* host);
+extern bool mongoConnect(const char* host,
+                         const char* db,
+                         const char* rplSet,
+                         const char* username,
+                         const char* passwd,
+                         bool        _multitenant,
+                         double      timeout);
 #ifdef UNIT_TEST
+extern bool mongoConnect(const char* host);
```
fgalan: I have realized that the mongoConnect() variant using just the host name is only used in unit test. Thus, is safer including it under this `#ifdef` in order to avoid bypassing the timeout (and other parameters) in the case of production code.

NTC

src/app/contextBroker/contextBroker.cpp
```diff
@@ -226,6 +227,7 @@ char            rush[256];
 #define DBUSER_DESC         "database user"
 #define DBPASSWORD_DESC     "database password"
 #define DB_DESC             "database name"
+#define TIMEOUT_DESC        "replica set timeout"
```
kzangeli: A bit more detail here would be great.

"timeout in seconds for connection to database" ?

On the other hand, 'replica set' is perhaps too detailed, and might not inform people what it is really about  :-)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -93,12 +93,19 @@ static void compoundObjectResponse(orion::CompoundValueNode* cvP, const BSONElem
...
+                  const char* passwd,
+                  bool        _multitenant,
+                  double      timeout)
+{
 
     std::string err;
     multitenant = _multitenant;
 
-    LM_T(LmtBug, ("dbName: '%s'", db));
+    LM_T(LmtBug, ("dbName: '%s', rplSet: '%s', timeout: %f", db, rplSet, timeout));
```
kzangeli: Strange trace level for a normal trace ...
This tracelevel is probably meant for buggy situations.

Stupid example to illustrate what I mean:

    bool X;

    switch (X)
    {
    case true:  break;
    case false: break;
    default:      LM_T(LmtBug, ("How did we get here?")); break;
    }

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -221,11 +228,10 @@ bool mongoConnect(const char* host, const char* db, const char* rplSet, const ch
 * mongoConnect -
 *
 * Version of the functions that doesn't uses authentication parameters
-*
 */
 bool mongoConnect(const char* host) {
 
-    return mongoConnect(host, "", "", "", "", false);
+    return mongoConnect(host, "", "", "", "", false, 0);
```
kzangeli: If the first mongoConnect used default values, this function would become obsolete

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -142,7 +149,7 @@ bool mongoConnect(const char* host, const char* db, const char* rplSet, const ch
           rplSetHosts.push_back(HostAndPort(hostTokens[ix]));
       }
 
-      connection = new DBClientReplicaSet(rplSet, rplSetHosts, 0);
+      connection = new DBClientReplicaSet(rplSet, rplSetHosts, timeout);
```
kzangeli: So, this line here is the actual change, right?
NTC

src/app/contextBroker/contextBroker.cpp
fgalan: What do you think about:

"timeout in seconds for connections to the replica set (ignored in the case of not using replica set)"

?

We need to introduce the "replica set" concept in the wording, as the timeout is only used in that case. If the user doesn't know what a replica set is, then this parameter is not for him/her :)

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I reused one of your LM_T traces ;)

Probably the proper tracelevel for this is LmMongo. I will use that one.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I prefer not using defaults for the other function. The other function is the only one that will be used by the contextBroker binary and it should provide all the parameters. Note that this `mongoConnect(const char* host)` is only used in unit tests.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Yes, I would say that this is the actual change. All the rest is the PR "plumbing" to bring the `timeout` to this place.

src/app/contextBroker/contextBroker.cpp
kzangeli: ok

src/lib/mongoBackend/MongoGlobal.cpp

src/lib/mongoBackend/MongoGlobal.cpp

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in bd187d2. Some LM_T with LmtBug has been removed in the process (I think they were stale LM_T traces).

src/app/contextBroker/contextBroker.cpp
fgalan: Fixed in bd187d2

kzangeli:
LGTM

Issue #501, opened 2014-08-07
CORS support


fgalan:
It would be interesting to have a CLI switch (e.g. `-cors`) so Orion CB behaves in a "CORS friendly" mode, i.e. including the `Access-Request-XXXX` headers in the HTTP responses.

paramarco:
+1

fgalan:
CORS support to GET requests implemented in https://github.com/telefonicaid/fiware-orion/pull/846. CORS support for PUT/POST requierest preflight requests through the OPTIONS method and it is deferred to a next PR.

fgalan:
Onces preflight request will be implemented, edit the answer to this question in SOF: http://stackoverflow.com/questions/30652867/cannot-get-data-from-orion-context-broker-installed-in-a-virtual-machine

drasko:
+1

Freeboard (https://www.fiware.org/2015/07/13/fiware-orion-data-source-now-available-for-freeboard/) will not work without it on localhost, unless proxied.

cdupont:
Hi there!
What is the status of implementing OPTIONS requests in Orion?

fgalan:
> What is the status of implementing OPTIONS requests in Orion?

The status is "not implemented". If you want to support the implemenation of CORS features in Orion, please use a +1 reaction on this issue.

Thanks!

cdupont:
+1

reitsma:
+1

cdupont:
In the mean-time, here is a work around: putting Orion behind a proxy.
The proxy should reply with the correct headers and respond to pre-flight.
Here is our HTTPD configuration:
https://github.com/Waziup/Platform/blob/master/identity/identityproxy.conf
```
<Location /v2>
   ProxyPass "http://orion:1026/v2"
   ProxyPassReverse "http://orion:1026/v2"

   # CORS headers including Fiware-ServicePath
   Header always set Access-Control-Allow-Origin "*"
   Header always set Access-Control-Allow-Methods "POST, GET, OPTIONS"
   Header always set Access-Control-Allow-Headers "Origin, X-Requested-With, Content-Type, Accept, Authorization, Fiware-ServicePath, Fiware-Service"

   # Reply to OPTIONS request with empty message (only headers)
   RewriteEngine On 
   RewriteCond %{REQUEST_METHOD} OPTIONS 
   RewriteRule ^(.*)$ $1 [R=200,L]
</Location>
```

McMutton:
[...]

PR #846, opened 2015-03-23
Support CORS GET Requests

jmcanterafonseca:
This PR is intended to enable cross-origin GET requests from Web Browsers to Orion. The advantage of supporting CORS is that Orion will be available directly from a Web Browser client without the need of intermmediate backends or proxies. More info on CORS can be found at

https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS
https://annevankesteren.nl/2015/02/same-origin-policy

For the time being, only the ability to perform GET requests have been implemented. For POST / PUT requests it would be needed to implement preflight requests through the OPTIONS method. But that would part of a subsequent PR. 

I guess the PR would need some unit tests as well, but I am a bit lost on how to start with. If they are needed any hint on how to do that would be appreciated.


fgalan:
Thanks for the contribution! This PR could be an effective solution to issue https://github.com/telefonicaid/fiware-orion/issues/501 and I'm sure many ussers will be happy with that ;)

I would need some time to read the references you cite to learn more on CORS and also to discuss the PR with @kzangeli . Regarding testing for this PR, rather than unit test I'd suggest to use test harness (a kind of end-to-end automatic test framework we have for Orion), but don't worry about that too much: @kzangeli or me could manage that.

src/lib/rest/restReply.cpp
```diff
@@ -85,6 +87,15 @@ void restReply(ConnectionInfo* ciP, const std::string& answer)
       MHD_add_response_header(response, "Content-Type", "application/xml");
     else if (ciP->outFormat == JSON)
       MHD_add_response_header(response, "Content-Type", "application/json");
+
+    // If any origin is allowed the header is sent always with "any" as value
+    if(strcmp(allowedOrigin, "*") == 0) {
```
fgalan: Please use this C++ style:

```
if (...)
{
...
}
else
{
...
}
```

jmcanterafonseca:
thanks for the feedback. I have updated the PR to meet the indenting rules and I have improved the design of the condition check.

jmrobles:
+1! :)

nifkuza:
+2

fgalan:
The PR is now ready for review and final merge. LGTM from @kzangeli and @jmcanterafonseca are required.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -246,6 +247,7 @@ int             writeConcern;
 #define HTTPSCERTFILE_DESC  "certificate key file (for https)"
 #define RUSH_DESC           "rush host (IP:port)"
 #define MULTISERVICE_DESC   "service multi tenancy mode"
+#define ALLOWED_ORIGIN_DESC "CORS allowed origin. use '__ALL' for any"
```
fgalan: Allowing `-corsOrigin '*'` in the CLI is dangerour... If the user forget to include `*` as literal, the shell will extand it with the contents of the current directory. @jmcanterafonseca do you see any problem in using an alias ("__ALL" could be changed for a better one if you want)? Of course, this precludes to use literarly __ALL as domain origin is a name weird enough so nobody should be using it as real domain name :)

test/functionalTest/cases/501_cors/get_any_origin.test
```diff
@@ -0,0 +1,134 @@
+# Copyright 2014 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: @jmcanterafonseca: have specially a look to the .test files included in this PR in order to check if the CB, after being started with SHELL-INIT and feed with what is shown in SHELL section should return the output shown in REGEXPECTED. 

I think the syntax of the test specification is quite straighforward, but if you have doubts, don't hesitate to contact me offline.

test/functionalTest/cases/501_cors/get_any_origin.test
```diff
@@ -0,0 +1,134 @@
...
+orionCurl --url /v1/queryContext --payload "${payload}" --json
+echo
+echo
+
+--REGEXPECT--
+01. Update/APPEND E1-T1-A1
+==========================
+HTTP/1.1 200 OK
+Content-Length: 383
+Access-Control-Allow-Origin: *
```
fgalan: @jmcanterafonseca: This is the response to a POST operation and includes the `Access-Control-Allow-Origin`. Taking into account the PR description, I thought that we are implemented CORS only for GETs... Is this correct?

src/lib/rest/rest.cpp
```diff
@@ -1068,6 +1071,8 @@ void restInit
   rushHost      = _rushHost;
   rushPort      = _rushPort;
 
+  strcpy(restAllowedOrigin, _allowedOrigin);
```
kzangeli: Use strncpy instead. The style guide script will complain ...
Wish we had the style guide check running ...

fgalan: Fixed in f54a492

test/functionalTest/cases/501_cors/get_any_origin.test
kzangeli: That's pretty much what was said in the meeting in Madrid, right? Only GET (for now)

kzangeli:
LGTM

test/functionalTest/cases/501_cors/get_any_origin.test
jmcanterafonseca: yup

test/functionalTest/cases/501_cors/get_any_origin.test
jmcanterafonseca: I see, thanks for the hint

src/app/contextBroker/contextBroker.cpp
jmcanterafonseca: yes, you are right. I don't have any particular opinion on what syntax to use for representing any URI, if you are ok with __ALL , I'm fine

thanks!

jmcanterafonseca:
LGTM as well, thanks for the work!

test/functionalTest/cases/501_cors/get_any_origin.test
fgalan: So you mean that `Access-Control-Allow-Origin` shouldn't be included in this case? In that case, I understand that the new code added to restReply.cpp should check that the request method was GET and only add the header in that case (I think you can get the method from the `ciP` object, maybe @kzangeli could provide more info on this).

src/app/contextBroker/contextBroker.cpp
fgalan: Ok, let's keep "__ALL".

NTC

test/functionalTest/cases/501_cors/get_any_origin.test
kzangeli: ciP contains the method, yes

fgalan: Fixed in 46f6cee4d627c4310fb6a470c7917c801023df2f

fgalan:
LGTM (final)

PR #848, opened 2015-03-24
Hardening/117 valgrind

kzangeli:
Removed all the leaks from the "117-modifications".
Found and fixed an ugly bug that crashes the broker when a convop is using compound values for attributes.


CHANGES_NEXT_RELEASE
```diff
@@ -9,3 +9,4 @@
        about the metadata in the update, this error has been fixed.
 - Fix: A registerContext request that uses associations must specify "attributeAssociations" instead of "attributes" to be consistent with DiscoveContextAvailability messages (Issue #823).
 - Add: new CLI parameter -timeout to specify DB connection timeout in the case of using MongoDB replica sets (Issue #589).
+- Fix: Fixed a bug that makes the broker crash when compount attribute values are used in convenience operations (No issue).
```
fgalan: I think that the "(No issue)" part is not needed in the case of not existing issue. However, it is ok as it is now, so NTC.

fgalan:
LGTM

PR #849, opened 2015-03-24
Hardening/117 post subscribe context conv op

kzangeli:
POST /v1/contextSubscriptions now has its very own service routine, supporting a few URI parameters.

Point of discussion: if URI param 'entity::type=XXX' is used, then the all entity types in the entity-vector are overwritten with this 'XXX' ...

As I said, let's discuss how we want this to work.

### DIFF
--- a/src/lib/serviceRoutines/postSubscribeContext.cpp
+++ b/src/lib/serviceRoutines/postSubscribeContext.cpp
@@ -36,6 +36,15 @@
 /* ****************************************************************************
 *
 * postSubscribeContext - 
+*
+* POST /v1/subscribeContext
+* POST /ngsi10/subscribeContext
+*
+* Payload In:  SubscribeContextRequest
+* Payload Out: SubscribeContextResponse
+*
+* URI parameters
+*   - notifyFormat=XXX    (handled by mongoBackend)
 */
 std::string postSubscribeContext
 (
@@ -49,7 +58,7 @@ std::string postSubscribeContext
   std::string               answer;
 
   //
-  // FIXME P6: for the moment, we are assuming that notification will be sent in the same format as the one
+  // FIXME P6: at the moment, we are assuming that notification are sent in the same format as the one
   // used to do the subscription, so we are passing ciP->inFormat. This is just an heuristic, the client could want
   // for example to use XML in the subscription message but wants notifications in JSON. We need a more
   // flexible approach, to be implemented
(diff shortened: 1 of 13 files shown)


src/lib/serviceRoutines/postSubscribeContext.cpp
```diff
@@ -36,6 +36,15 @@
...
 * postSubscribeContext - 
+*
+* POST /v1/subscribeContext
+* POST /ngsi10/subscribeContext
+*
+* Payload In:  SubscribeContextRequest
+* Payload Out: SubscribeContextResponse
+*
+* URI parameters
+*   - notifyFormat=XXX    (handled by mongoBackend)
```
fgalan: We dedided to ignore entity::type... I guess that exist and !exist are also ignored, aren't they?

fgalan: Ah... I have just realize that this is a std-op, so all the URI params for filters are ignored. NTC

test/functionalTest/cases/117_convop_using_standard_ops/postSubscribeContextConvOp.test
```diff
@@ -0,0 +1,265 @@
...
+    }
+  ]
+}
+=======================================
+
+
+--TEARDOWN--
+brokerStop CB
+accumulatorStop
+#dbDrop CB
```
fgalan: Uncomment this?

src/lib/serviceRoutines/postSubscribeContext.cpp
kzangeli: Just to clarify, in the other service routine (postSubscribeContextConvOp) we noi longer treat 'entity::type=XXX', but we do treat 'exist=entity::type' and '!exist=entity::type'.
And I just realized I have no func-test for that :-( :-D

test/functionalTest/cases/117_convop_using_standard_ops/postSubscribeContextConvOp.test
kzangeli: of course!

kzangeli: Fixed in aedc081

fgalan:
LGTM

PR #853, opened 2015-03-25
Hardening/117 put subscription conv op

kzangeli:
Function header comments and functest for:
PUT /v1/contextSubscriptions/{subscriptionId}

### DIFF
--- a/test/functionalTest/cases/117_convop_using_standard_ops/putSubscriptionConvOp.test
+++ b/test/functionalTest/cases/117_convop_using_standard_ops/putSubscriptionConvOp.test
@@ -1,0 +1,763 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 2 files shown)


test/functionalTest/cases/117_convop_using_standard_ops/putSubscriptionConvOp.test
```diff
@@ -0,0 +1,763 @@
...
+
+
+echo "06. Dump accumulator, see E1/T1/A1+A2, in XML"
+echo "============================================="
+curl localhost:${LISTENER_PORT}/dump -s -S 2> /dev/null
+echo
+echo
+
+
+echo "07. PUT /v1/contextSubscriptions/SUB_ID?notifyFormat=JSON, removing A2 from notifyConditions"
```
fgalan: noityFormat makes sense when you want to change the default notification format (which is the one used in the subscription payload). Thus, in this case, given that the payload is in JSON, shouldn't be `?notifyFormat=XML`?

kzangeli: Actually, whenever the notifyFormat parameter is used, it overrides the 'default' which is whatever formats the request was sent in. It pretty much always makes sense. If notifyFormat is present, never mind the format of the request.

fgalan: What I mean is to use set notifyFormat to the opposite format of the request to ensure it is actually working (if you use JSON payload and notifyFormat=JSON to ckeck you get JSON notification, you are not actually checking it works).

If my point is not clear, let's move discussion to skype.

kzangeli: It is pretty clear ... :-)
?notifyFormat=XML 'the opposite way' is already used in 'part 04'.
Here, in 'part 07', I can either use it or not, doesn't really make a difference ...

fgalan: I didn't realize that part 04 works in cross-format way (text: "04. Subscribe to E1/T1/A1+A2, in XML" doesn't mention it and I usually only pay attention to the test outline). Thus, I think is ok as it is.

NTC

fgalan:
LGTM

PR #855, opened 2015-03-26
Hardening/117 delete attribute value instance

kzangeli:
DELETE /v1/contextEntities/{entity::id}/attributes/{attribute::name}/{metaID}


fgalan:
LGTM

PR #857, opened 2015-03-26
Hardening/117 delete attribute value instance with type and

kzangeli:
DELETE /v1/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}/{metaID}


fgalan:
LGTM

PR #859, opened 2015-03-26
Hardening/117 get attribute value instance with type and

kzangeli:
GET /v1/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}/{metaID}


fgalan:
LGTM

PR #860, opened 2015-03-27
Hardening/117 post attribute value instance with type and

kzangeli:
POST /ngsi10/contextEntities/type/{entity::type}/id/{entity::id}/attributes/{attribute::name}/{metaID}


fgalan:
LGTM

PR #862, opened 2015-03-30
Hardening/117 put attribute value instance

kzangeli:
PUT /v1/contextEntities/{entity::id}/attributes/{attribute::name}/{metaID}


fgalan:
LGTM

Issue #779, opened 2015-02-16
Remove references to FIWARE release


fgalan:
FIWARE uses a "paralel" numbering (e.g. 4.2.2) that is not actually needed, as Orion version (e.g. 0.19.0) is enough as long as the association is done at FIWARE Forge (https://forge.fiware.org/frs/?group_id=7).

Some changes to take into account (not exahustive):

* Remove the "virtual" RPM package contextBroker-fiware in the .spec
* Simplify makefile, not needing anymore env variables for FIWARE stuff.
* Release tags will include only the version number (i.e. "0.19.0" instaed of "0.19.0-FIWARE-4.2.2"). Old tags will not be changed, the new numbering applies only to new tags.

**Effort: 2 man day**

fgalan:
Implemented in PR #866

PR #866, opened 2015-03-30
REMOVE contextBroker-fiware virtual RPM stuff

fgalan:
Implements https://github.com/telefonicaid/fiware-orion/issues/779

I have tested this with `make rpm` and it seems to generate a correct contextBroker-0.19.0_next-dev.x86_64.rpm package. However, the "definitive test" will be done when we produce 0.20.0 using the release.sh script.


kzangeli:
LGTM

PR #868, opened 2015-03-31
FIX small issues in release.sh script

fgalan:
Small fix in argument number checking.

In addition, recovering some commented lines to align with current releasing procedure (which includes merging to master).


kzangeli:
LGTM

Issue #722, opened 2015-01-28
Forward ngsi10 Query/Update requests not only in XML


kzangeli:
The forwarding of requests to context providers is right now always in XML.
Lots of clients prefer JSON so the CB should allow to forward requests in JSON as well.

To implement this we need a few modifications:
- The function **sendHttpSocket** eats the HTTP headers from the response.
   The function must return the headers to the caller somehow.
- The broker is unable to parse QueryContextResponse in JSON
- The broker is unable to parse UpdateContextResponse in JSON
- Selection of the format to use for the forward message (XML or JSON).
  The simplest approach would be to just select the same format as the original
  request. A better approach might be to make the context provider inform the 
  CB about what formats are allowed (perhaps copy the 'Accept' header from the registration?)
- The function sendHttpSocket should get the name changed - *httpMsgSend*?
- sendHttpSocket always sends XML/JSON as Accept HTTP header. Is this what we want?

**Effort: 3 man day**

Optionally (phase II)
- If the context provider doesn't accept the format of the forwarded request (XML/JSON), check the HTTP header 'Accept' and resend the forward in a better format, if possible. If not, report error.

fgalan:
In sprint review meeting on February 27th a simpler approach has been decided:

* The CPr will include either "Accept: application/json" or "Accept: application/xml" in the registerContext request.
* The CB will record "XML" or "JSON" in the registrations document in the DB corresponding to that registerContext. A new field in the document will be needed (similar case that in the subscriptions collection).
* The CB will use always the format stored in the database to interact with the CPr. MongoBackend will provide the format along the CPr URL when return this info to serviceRoutines.

fgalan:
The "simple solution" designed in February 27th has been implemented in PR #899 for Orion 0.21.0. However, let this issue opened in order to implement the optimal solution (HTTP negotiation, paying atention to Accept and Content-Type headers) in the future (milestone will be removed).

fgalan:
NGSIv2 only allows application/json notifications, so maybe this issue can be closed.

kzangeli:
Yes, it can be closed

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

fgalan:
(Wrongly reopnened... after XML removel this issue has no sense anymore. Closing again)

PR #872, opened 2015-04-09
Task/787 queries in same thread

kzangeli:
Query forwarding to multiple context providers.
All requests still running in the same thread.
Lots of functests and unittests still failing, but the new ones about this change work just fine.
Try it, you'll like it:
% testHarness.sh test/functionalTest/cases/787_cprs_full_functional


src/lib/logMsg/logMsg.h
```diff
@@ -497,7 +497,7 @@ do                                                                     \
                                                                        \
   if ((text = lmTextGet s) != NULL)                                    \
   {                                                                    \
-    lmOut(text, 'F', __FILE__, __LINE__, "***", 0, NULL);              \
+    lmOut(text, 'F', "ForcedLog", 0, "***", 0, NULL);                  \
```
fgalan: Just curiosirty... why this change?

src/lib/ngsi/AttributeList.cpp
```diff
@@ -110,6 +112,25 @@ void AttributeList::release(void)
 
 /* ****************************************************************************
 *
+* lookup - 
+*/
+bool AttributeList::lookup(const std::string& attributeName)
```
fgalan: Nice method... probably some existing parts of the code (I'm thinking in mongoBackend) are doing this and we can reuse.

NTC (just a thought)

src/lib/ngsi/AttributeDomainName.cpp
```diff
@@ -99,11 +99,11 @@ void AttributeDomainName::present(const std::string& indent)
 {
   if (string != "")
   {
-    PRINTF("%sAttributeDomainName: %s\n", indent.c_str(), string.c_str());
+    LM_F(("%sAttributeDomainName: %s", indent.c_str(), string.c_str()));
```
fgalan: Which is the difference between PRINTF and the LM_F macro?

src/lib/ngsi/EntityIdVector.cpp
```diff
@@ -120,6 +120,41 @@ void EntityIdVector::present(const std::string& indent)
...
+  for (unsigned int ix = 0; ix < vec.size(); ++ix)
+  {
+    std::string isPatternFromVec = vec[ix]->isPattern;
+
+    if (isPatternFromVec == "")
+    {
+      isPatternFromVec = "false";
+    }
+
+    if ((vec[ix]->id == id) && (vec[ix]->type == type) && (isPatternFromVec == isPatternFromParam))
```
fgalan: Are you taking into account the special semantics of empty type as "match any entity type"? Or is it not meaninfull considering the usage of this function (i.e. the expectations of the function caller)?

src/lib/ngsi10/QueryContextRequestVector.cpp
```diff
@@ -0,0 +1,141 @@
+/*
```
fgalan: This is actualy a ngsi10 class? Or a "helper" class maybe better to place in a different directory?

src/lib/ngsi10/QueryContextResponseVector.cpp
```diff
@@ -0,0 +1,182 @@
+/*
```
fgalan: This is actualy a ngsi10 class? Or a "helper" class maybe better to place in a different directory?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -49,115 +52,57 @@ static char* xmlPayloadClean(const char*  payload, const char* payloadWord)
 
 /* ****************************************************************************
 *
-* postQueryContext -
+* queryForward - 
+*
+* An entity/attribute has been found on some context provider.
+* We need to forward the query request to the context provider, indicated in qcrsP->contextProvider
```
fgalan: Instead of using singular (e.g. "to the context provider"), it should be in plural ("to the context providers"). Not only this line, but all the preamble comment.

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -49,115 +52,57 @@ static char* xmlPayloadClean(const char*  payload, const char* payloadWord)
...
-    answer = qcrP->render(ciP, QueryContext, "");
-    return answer;
+    qcrsP->errorCode.fill(SccContextElementNotFound, "");
+    return;
   }
 
 
   //
-  // 2. The exact same QueryContextRequest will be used for the query-forwarding
+  // 2. Render an XML-string of the request we want to forward
```
fgalan: I understand this is now hardwired to XML as this is the only format support right now, but it would be more flexible when also JSON becaomes a possibility (i.e. fixing https://github.com/telefonicaid/fiware-orion/issues/722), right?

src/lib/ngsi/AttributeDomainName.cpp
kzangeli: LM_F makes the output go to the log file instead of to stdout

src/lib/logMsg/logMsg.h
kzangeli: This is to have all the LM_F lines aligned vertically in the logfile.
If the name of the file is printed bofore the 'message' and the different file names don't have the exact same number  of characters, it gets unaligned and really hard to read.

src/lib/ngsi10/QueryContextRequestVector.cpp
kzangeli: Yeah, I was thinking about this also.
I honestly don't know in which directory these four new files belong ...
orionTypes perhaps ... ?

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Of course.
Right now, we only support XML unfortunately

src/lib/ngsi/EntityIdVector.cpp
kzangeli: Good point.
The way it is used (after receiving the output from mongoQueryContext) this function *should* do all that is needed for the purpose. However, with the generic name **lookup**, it should probably be 100% generic ...
I'll try to make it generic (with your help).

src/lib/ngsi/AttributeList.cpp
kzangeli: Yeah, I'm sure we have lots of 'forgotten' functions out there that could be implemented as a method for the ngsi-request classes instead. Making these functions methods it is easier to find and use them later

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Actually, there can't be any plural here.
This function, *queryForward*, forwards only **one** query to **one** context provider.

The plural would be suited from where this function is called, but not from here.

NTC

src/lib/ngsi/EntityIdVector.cpp
kzangeli: I looked into this a bit closer and I added a comment for the EntityIdVector struct, as you will see in commit e3b872e.

I propose to use the method-name 'lookup' to lookup exact matches and 'match' to find matching entities.
I was thinking about calling it lookupExact to avoid confusion but I don't like the name very much ...

If we can agree on using 'match' for matching functions/methods and lookup for **exact** searches, then we are all good ...


What I **really** would need here is a method to compare which EntityId is 'broader' (which one 'eats' the other) and replace the 'smaller' with the 'bigger'. However, this is a pretty complex task and including both entities for the forwarded query is 100% OK - just not as elegant.

Imagine we have a query to be forwarded with { E.*, A1 } and { E3, A2 }.
With the current implementation, the forwarded query will be:  { [ E.*, E3 ], [ A1, A3 ] }.
As E.* 'eats' E3, it would be more elegant to forward { [ E.* ], [ A1, A2 ] }.

Not worth the effort ...

So, with a comment, fixed in e3b872e

src/lib/logMsg/logMsg.h
fgalan: It seems like black magic :), but ok, it's fine :D

NTC

src/lib/ngsi/AttributeDomainName.cpp
fgalan: Is LM_F always printed or only in debug compilation mode?

Is LM_F using the log format as defined in https://forge.fiware.org/plugins/mediawiki/wiki/fiware/index.php/Publish/Subscribe_Broker_-_Orion_Context_Broker_-_Installation_and_Administration_Guide#Log_format? 

Which msg level is used (I understand it should be INFO)?

src/lib/ngsi/EntityIdVector.cpp
fgalan: I think is ok the way it works now... not a big lost of smartness with the current approach.

(I think there is a typo in the example... it should be  { [ E.*, E3 ], [ A1, A2 ] } instead of  { [ E.*, E3 ], [ A1, A3 ] } )

NTC

src/lib/ngsi10/QueryContextRequestVector.cpp
fgalan: orionTypes sounds good

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Ok, I missed that point.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: NTC

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+      qcrsP->errorCode.fill(SccContextElementNotFound);
+    }
+
+    answer = qcrsP->render(ciP, QueryContext, "");
+    return answer;
+  }
+
+
+  //
+  // DEBUG - present the vector prepared for Context Provider Forwarding
```
fgalan: Should this debug code be removed?

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -96,7 +96,17 @@ std::string postUpdateContext
...
+  upcrsP->errorCode.fill(SccContextElementNotFound);
   ciP->httpStatusCode = mongoUpdateContext(&parseDataP->upcr.res, upcrsP, ciP->tenant, ciP->servicePathV, ciP->uriParam, ciP->httpHeaders.xauthToken, "postUpdateContext");
+  if (upcrsP->errorCode.code == SccContextElementNotFound)
+  {
+    if (upcrsP->contextElementResponseVector.size() == 1)
+    {
+      upcrsP->errorCode.details = upcrsP->contextElementResponseVector[0]->contextElement.entityId.id;
+    }
+  }
+  upcrsP->present("");
```
fgalan: Debug code?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+      continue;
+    }
+
+    QueryContextResponse* qP = new QueryContextResponse();
+    queryForward(ciP, requestV[fIx], qP);
+    responseV.push_back(qP);
   }
 
-  answer = qcrsP->render(ciP, QueryContext, "");
+  responseV.present();
```
fgalan: Debug code?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+
+    for (unsigned int aIx = 0 ; aIx < cerP->contextElement.contextAttributeVector.size(); ++aIx)
+    {
+      EntityId*            eP       = &cerP->contextElement.entityId;
+      ContextAttribute*    aP       = cerP->contextElement.contextAttributeVector[aIx];
+      QueryContextRequest* requestP = requestV.lookup(aP->providingApplication, eP);   // No need to lookup if providingApplication == "" ...
+      
+      //
+      // An empty providingApplication means the attribute is local
+      // In such a case, the response is already in our hand, we just need to forward it to responseV
```
fgalan: by "forward it to responseV" you mean pushing back into that vector? In that case, "forward" can be a bit confusing, as it is associated to forwarding *in a new REST call* in this context...

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+        requestP->attributeList.push_back_if_absent(aP->name);
+        requestP->entityIdVector.push_back_if_absent(eP);
+      }
+    }
+  }
+
+  //
+  // If no redirectioning is necessary, just return the result
+  //
+  if (noOfForwards == 0)
```
fgalan: Why don't check the size of the `requestV` vector? I think you will save the `noOffForwards' variable in that case.

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+    QueryContextResponse* qP = new QueryContextResponse();
+    queryForward(ciP, requestV[fIx], qP);
+    responseV.push_back(qP);
   }
 
-  answer = qcrsP->render(ciP, QueryContext, "");
+  responseV.present();
+
+  //
+  // Is qcrsAccumulatedP OK? Without doing anything special here ... ?
```
fgalan: What is qcrsAccumulatedP ? I don't find this variable... Maybe the comment is stale?

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Yes!

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Old stuff. To be removed as well

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yeah, all calls to 'present()' is nothing by debug really.

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+  // Actually, if there is only ONE forward to be done then there is no reason to
+  // do the forward in a separate shell. Better to do it inside the current thread.
+  //
+  // If providingApplication is empty then that part of the query has been performed already, locally.
+  // 
+  //
+  for (unsigned int fIx = 0; fIx < requestV.size(); ++fIx)
   {
-    qcrsP->contextElementResponseVector[0]->statusCode.details = std::string("Redirected to context provider ") + ip + ":" + portV + prefix;
+    if (requestV[fIx]->contextProvider == "")
```
fgalan: I don't understand the need for this check-to-continue fragment... I understand that every request in the `requestV` vector have a context provider. Could you elaborate on this, pls?

src/lib/xmlParse/xmlQueryContextResponse.cpp
```diff
@@ -374,6 +374,5 @@ void qcrsPresent(ParseData* parseDataP)
   if (!lmTraceIsSet(LmtDump))
     return;
```
kzangeli: This piece of code here responds part of a question you had about LM_F.

LM_F is always printed, but in the 'present()' methods I dont continue unless the traceLevel LmtDump is set, so it is some kind of hybrid.

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,159 @@ std::string postQueryContext
...
+      }
+      else if (requestP == NULL)
+      {
+        requestP = new QueryContextRequest(aP->providingApplication, eP, aP->name);
+        requestV.push_back(requestP);
+        ++noOfForwards;
+      }
+      else
+      {
+        requestP->attributeList.push_back_if_absent(aP->name);
```
fgalan: I understand that with this strategy, if you do the following query E-[A, B, C, D] resulting in:

E-A -> CPr1
E-B -> CPr1
E-C -> CPr2
E-D -> CPr2

At the end 2 forward messages are sent (one for CPr1 with E1-[A,B] and other for E-[C,D]) and not 4 messages (2 to each CPr, containing each message 2 attributes).

Is this correct?

test/functionalTest/cases/787_cprs_full_functional/double_forward.test
```diff
@@ -0,0 +1,239 @@
...
+
+
+--TEARDOWN--
+brokerStop CB
+brokerStop CP1
+brokerStop CP2
+dbDrop CB
+dbDrop CP1
+dbDrop CP2
+
```
fgalan: Empty lines to be removed?

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: The queries that are resolved locally have an empty contextProvider.
The result of that query is already in responseV and no forward is required - thus the 'continue'

test/functionalTest/cases/787_cprs_full_functional/five_different_cprs_in_one_query.test
```diff
@@ -0,0 +1,675 @@
...
+        {
+            "contextElement": {
+                "attributes": [
+                    {
+                        "name": "A1", 
+                        "type": "string", 
+                        "value": "A1 in CP1"
+                    }
+                ], 
+                "id": "E1",
```
fgalan: I don't like too much having each attribute in a separate CE, being all the CE with the same entityId and entityType...

Maybe the solution is to apply some "collapse" function to the final responses vector, in order to merge in the same CE attributes sharing same entityId/entityType.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: But you are right, I don't push that part to requestV, only its result to responseV ...
contextProvider cannot ever be empty (which means this is dead code) and if it is, it is a bug.

test/functionalTest/cases/787_cprs_full_functional/forward_depends_on_attr_field.test
```diff
@@ -0,0 +1,542 @@
...
+                    }
+                ], 
+                "id": "E1", 
+                "isPattern": "false", 
+                "type": "T1"
+            }, 
+            "statusCode": {
+                "code": "200", 
+                "reasonPhrase": "OK"
+            }
```
fgalan: What about A7? What we are doing in a normal (i.e. resolved by CB, without involving CPrs) on a non existing attribute?

Just to be sure we are implementing the same behaviour towards clients.

src/lib/ngsi/AttributeDomainName.cpp
kzangeli: DEBUG is used for LM_F.
I don't think there is any protocol to follow for DEBUG messages ...

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Yes, that is correct

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: True

test/functionalTest/cases/787_cprs_full_functional/five_different_cprs_in_one_query.test
kzangeli: Yes, I agree. I actually commented on that somewhere ...
In postQueryRequest I need to do the same thing for responseV as I do for requestV, check whether the Entity already exists and just add the attribute right there.

test/functionalTest/cases/787_cprs_full_functional/two_attributes_in_one_cpr_three_in_another.test
```diff
@@ -0,0 +1,624 @@
...
+}
+
+
+--TEARDOWN--
+brokerStop CB
+brokerStop CP1
+brokerStop CP2
+dbDrop CB
+dbDrop CP1
+dbDrop CP2
```
fgalan: The existing test are pretty good... the cover a good range os cases.

However, I miss the following scenarios:

* Query with empty attributes with attributes in CPrs, e.g. query E-none and get E-A1,A2,A3,A4 (A1-A2 comming from one CPr and A3-A4 comming from another CPrs).
* Query with empty attributes "mixing" attributes in CPrs and CB, e.g. query E-[none] and get E-A1,A2,A3 (A1 comming from CB and A2-A3 comming from a CPr).
* Situations in which CPr doesn't hold the data, although it is registered in that way. Eg. query E-A, being E-A register in CPr, but the  CPr is empty. It should return 404 to the client.
* Similar to the former, but with pieces of data that preclude to return a 404 to the client. Eg. query E-A1,A2, being E-A1 register in CPr, but the  CPr is empty. E1-A2 is locally stored in CB. The CPr will return 404 for the E-A1 forward, but given that E1-A2 has been resolved locally, the client will finally receive E-A2 (not 404)
* Variation of the former in which E1-A2 is in a second CPr instead that in CB.
* Cases in which local resolution "shadows" CPr. E.g. E-A1 is registered in CPr and at the same time stored locally in the CB. A query for E-A1 will return the value in the CB, no tin the CPr.

test/functionalTest/cases/787_cprs_full_functional/forward_depends_on_attr_field.test
kzangeli: A7 doesn't exist. We agreed some time ago that non-found stuff should be silently ignored.
It doesn't make much sense to add a ContextElementResponse just to say "I can't find A7" ...
If A7 is the ONLY attribute, then it is different. Then it should be seen that it is not found (and it is).

If we do a query right now on { [ E1, E2 ], [ A1, A2 ] } and only E1-A1 is found, I'm pretty sure that we don't respond saying that E1-A2, E2-A1 and E2-A2 are not found. Seems stupid to add payload for this.
What if we have 100 entities and 100 attributes in the query and we find only one pair ... 9999 responses of 'not found' ? :-)

test/functionalTest/cases/787_cprs_full_functional/two_attributes_in_one_cpr_three_in_another.test
kzangeli: OK, I will add more tests :-)

src/lib/xmlParse/xmlQueryContextResponse.cpp
fgalan: In the case it gets printed... Is it using the DEBUG level? (https://forge.fiware.org/plugins/mediawiki/wiki/fiware/index.php/Publish/Subscribe_Broker_-_Orion_Context_Broker_-_Installation_and_Administration_Guide#Log_format)

fgalan: Already anserwed here: https://github.com/telefonicaid/fiware-orion/pull/872/files#r28090356

NTC

kzangeli: Yes, all LM_F are translated to DEBUG

src/lib/ngsi/AttributeDomainName.cpp
fgalan: So I understand that PRINTF will not be used any more, in favour of LM_F. Is that correct? In that case, we would need to to an overall replacement of PRINTF for LM_F. We should open a tech debt issue nice2have for this.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Great! I would problably though in hashmaps and such complicated artifact, but you have solved the problem in a smarter way ;)

NTC

test/functionalTest/cases/787_cprs_full_functional/forward_depends_on_attr_field.test
fgalan: NTC

src/lib/ngsi/AttributeDomainName.cpp
kzangeli: Yes, all of them must and will be replaced.
I'll create an issue about it.

I used these 'present()' methods for debugging when the parser was implemented and with the broker running in foreground. It was easier that way.
However, things have changed and now all testing is done via functests and it is now necessary to see the result of 'present' is the log file.

kzangeli: Issue #873

fgalan: NTC

src/lib/ngsi10/QueryContextRequestVector.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/ngsi10/QueryContextResponseVector.cpp
kzangeli: Moved to orionTypes in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 6e3c8f9

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 6e3c8f9

test/functionalTest/cases/787_cprs_full_functional/double_forward.test
kzangeli: Fixed in 6e3c8f9

test/functionalTest/cases/787_cprs_full_functional/five_different_cprs_in_one_query.test
kzangeli: This is still to be implemented.
The rest of the PR I think is OK now.

test/functionalTest/cases/787_cprs_full_functional/two_attributes_in_one_cpr_three_in_another.test
kzangeli: ... and a 'few' new tests, of course :-)

test/functionalTest/cases/787_cprs_full_functional/five_different_cprs_in_one_query.test
kzangeli: Fixed in 1134e87

kzangeli: Fixed in 79078d7

test/functionalTest/cases/787_cprs_full_functional/two_attributes_in_one_cpr_three_in_another.test
kzangeli: Fixed in commits
68dfd08,
2abb36b,
d5ee84d,
6568e3a,
1b0d191, and
3ca46bd

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,198 @@ std::string postQueryContext
...
+    for (unsigned int aIx = 0 ; aIx < cerP->contextElement.contextAttributeVector.size(); ++aIx)
+    {
+      EntityId*            eP  = &cerP->contextElement.entityId;
+      ContextAttribute*    aP  = cerP->contextElement.contextAttributeVector[aIx];
+      
+      //
+      // An empty providingApplication means the attribute is local
+      // In such a case, the response is already in our hand, we just need to copy it to responseV
+      //
+      if (aP->providingApplication == "")
```
fgalan: What about providingApplicationList? This is a field of ContextElement class, that can be filled by mongoBackend in the case of a queryContext for E-[null] when there are registrations for E-[null] (in addition to the ones that could be registered for individual attributes, such E-A1, E-A2, etc.). Have a look to mongoContextProvidersQueryRequest.severalCprs3 unit test could help.

The providingApplicationList shoulld also be taken into account, then tested with the proper .test. Eg.:

* Register E-{} -> CP1 -> this should add an element to the providingApplicationList 
* Register E-{} -> CP2 -> this should add an element to the providingApplicationList 
* Register E-A3 -> CP3
* Register E-A4 -> CP4
* Create E-A0 in CB
* Create E-A1 in CP1
* Create E-A2 in CP2
* Create E-A3 in CP3
* Create E-A4 in CP4
* Query for E-{} -> should return A[0-4]

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,198 @@ std::string postQueryContext
...
+
+  qcrsP->errorCode.fill(SccOk);
+  ciP->httpStatusCode = mongoQueryContext(&parseDataP->qcr.res, qcrsP, ciP->tenant, ciP->servicePathV, ciP->uriParam);
+
+  //
+  // In this loop the output from mongoQueryContext is examines and the attributes are sorted
+  // by their providing application, in requestV, later to be used to forward the queries to their
+  // respective Context Providers.
+  //
+  // The local part of the query is already tken care of by mongoQueryContext and the result must be moved
```
fgalan: tken -> taken

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -241,26 +178,198 @@ std::string postQueryContext
...
+
+    //
+    // Now, each ContextElementResponse of qP should be tested to see whether there
+    // is already an existing ContextElementResponse in responseV
+    responseV.push_back(qP);
+  }
+
+  answer = responseV.render(ciP, "");
+
+  // requestV.release();
```
fgalan: Stale comment? Or should be uncomented?

In general, I see some new "new" statements in this PR, but not so much releases... is everything ok from the point of view of memory management?

kzangeli: Not at all.
After implementing Updates and fixing all functests I will fix all memory leaks.
 There will be quite a few ...

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: ok, let's do one more test :-)

fgalan: Take into account that in this case maybe it more than just another test... As I mention above, I don't see any point in the code in which you process the ContextElement providingApplicationList field (maybe the code is there, but I haven's seen it).

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Anyway, we should try to implement memory management in the right way PR after PR. I mean, not to completely rely on valgrind de-leaking step. At the end, valgrind step is just a tool to improve the code (a good one, by the way) but it could miss some case.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Yes, you are probably right.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Fixed in 7bed43de4e9cc94ce4858d5ddd99d3f5addecb82

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Fixed in 7bed43de4e9cc94ce4858d5ddd99d3f5addecb82 (I guess)

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Fixed in da7bc02f0b61740d9647be7bafbeefdb9de5b35e (I guess)

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -220,47 +159,282 @@ std::string postQueryContext
...
+  //
+  // If the result from mongoBackend is a 'simple' result without any forwarding needed (this is the
+  // normal case), then there is no need to execute the special treatment that forwarding needs.
+  // It's better not to mix the two very different treatments of the output from mongoBackend.
+  //
+  // Now, the request is 'simple' if all providingApplicationLists of the ContextElements are empty and
+  // no ContextAttribute has any providingApplication.
+  //
+  bool requestDone = true;
+  for (unsigned int ix = 0 ; ix < qcrsP->contextElementResponseVector.size(); ++ix)
```
fgalan: Maybe the "noneCpr" block (from 252 to 271) could be moved to a function for clarity (the `break` statement will become `return`).

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Yes, the providing application list is taken into account with the new implementation (commit da7bc02)

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: I was considering that, actually :-)
Let's do it!

kzangeli: Fixed in 3bd9e87

fgalan:
LGTM

Issue #880, opened 2015-04-15
CPr forwarding timeout


fgalan:
Implement a timeout for CPrs forwarding requests (the value could be specified using a CLI parameter).

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/903

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests

fgalan:
Issues in milestone QATestPending are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeles with QATestPending label in the case thinks change and we need to recover it.

PR #883, opened 2015-04-16
Task/787 updates in same thread

kzangeli:
Update forwarding

Traces still in the code, asking for more func-tests.
Removing traces at last step before merging.
Also, valgrind hasn't been run yet to check for leaks.


test/unittests/mongoBackend/mongoUpdateContext_test.cpp
```diff
@@ -5728,7 +5728,7 @@ TEST(mongoUpdateContextRequest, updateAttrWithId)
 *
 * updateAttrWithAndWithoutId -
 */
-TEST(mongoUpdateContextRequest, updateAttrWithAndWithoutId)
```
fgalan: We should do something with this... Any idea?

test/functionalTest/cases/787_cprs_full_functional/update_false_registration.test
```diff
@@ -0,0 +1,140 @@
...
+Content-Type: application/json
+Date: REGEX(.*)
+
+{
+    "duration": "P1M", 
+    "registrationId": "REGEX([0-9a-f]{24})"
+}
+
+
+02. Update/UPDATE E1/T1/A1 in CB
```
fgalan: I understand that in this case CB tries to update forward to CPr but, given that CPr doesn't exist, the CB gets somekind of connection error and returns a 404. Is that correct?

test/functionalTest/cases/787_cprs_full_functional/update_false_registration_and_one_more_value.test
```diff
@@ -0,0 +1,218 @@
...
+brokerStart CB
+brokerStart CP1
+
+--SHELL--
+
+#
+# 01. Register E1/T1/A1-CP1 in CB
+# 02. Update/APPEND E1/T1/A2 in CB
+# 03. Update/UPDATE E1/T1/A1+A2 in CB
+#
```
fgalan: I understand we need two additional steps:

* Query CB E1/T1/A2 to check the value has been actually updated
* Query CP1 E1/T1/A1 to check the value has been actually updated

test/unittests/mongoBackend/mongoUpdateContext_test.cpp
kzangeli: I can go deeper to get the exact line where it fails.
That might give us some more clues

test/functionalTest/cases/787_cprs_full_functional/update_false_registration.test
kzangeli: Actually, the CPr exists (I will test the thing you are proposing when I start with threads).
I call it 'false' as the attribute doesn't exist where the registration says it should exist, but the CP is answering, so no problem with the connection.

test/functionalTest/cases/787_cprs_full_functional/update_false_registration_and_one_more_value.test
kzangeli: Sure, except, A1 doesn't exist in CP1.
The test is to see that E1/T1/A1 gives a 404 after being forwarded

src/lib/ngsi/ContextElementResponseVector.h
```diff
@@ -44,14 +44,16 @@ typedef struct ContextElementResponseVector
...
-                                  bool                comma = false);
+                                  bool                comma               = false,
+                                  bool                omitAttributeValues = false);
 
   void                     present(const std::string& indent);
   void                     push_back(ContextElementResponse* item);
   unsigned int             size(void);
   ContextElementResponse*  get(unsigned int ix);
-  ContextElementResponse*  lookup(EntityId* eP);
+  ContextElementResponse*  lookup(EntityId* eP, HttpStatusCode code = SccNone);
```
fgalan: I understand that after this PR, the SccNotFound variable (in HttpStatusCode.h I guess) is no longer needed. Is has been removed from that file?

src/lib/ngsi10/UpdateContextResponse.cpp
```diff
@@ -163,3 +164,96 @@ void UpdateContextResponse::release(void)
...
+}
+
+
+
+/* ****************************************************************************
+*
+* UpdateContextResponse::foundPush - 
+*
+* 1. Find contextElementResponse in contextElementResponseVector and add the ContextAttribute.
+* 2. If not found: create a new one.
```
fgalan: Not sure if this comment (bullet 2) is correct, taking into account the name of the function...

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -52,6 +53,321 @@ static char* xmlPayloadClean(const char*  payload, const char* payloadWord)
...
+  //
+  Format outFormat = ciP->outFormat;
+  ciP->outFormat = XML;
+  std::string  payload = upcrP->render(ciP, UpdateContext, "");
+  char*        cleanPayload;
+  ciP->outFormat = outFormat;
+
+  LM_M(("KZ: payload: %s", payload.c_str()));
+
+  if ((cleanPayload = xmlPayloadClean(payload.c_str(), "<updateContextRequest>")) == NULL)
```
fgalan: In the future we will have also JSON. May this could be generalized in the context of this PR? Or it is not possible and this would be modified in the future?

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -52,6 +53,321 @@ static char* xmlPayloadClean(const char*  payload, const char* payloadWord)
...
+  //
+  parseData.upcr.res.release();
+  parseData.upcrs.res.release();
+}
+
+
+
+/* ****************************************************************************
+*
+* foundAndNotFound -
```
fgalan: foundAndNotFound is a little confusing name :)

src/lib/ngsi/ContextElementResponseVector.h
kzangeli: I hadn't thought about that.
But you are right, SccFound should no longer exist.

kzangeli: Fixed in e5604a0

test/functionalTest/cases/787_cprs_full_functional/update_false_registration_and_one_more_value.test
fgalan: You are right. My second step doesn't make sense.

test/unittests/mongoBackend/mongoUpdateContext_test.cpp
fgalan: Or defer this for the future with a "FIXME P10" with "We don't know what the hell is happening here but enabling this crashes the unittest binary. Sad but true :)"

;)

test/functionalTest/cases/787_cprs_full_functional/update_false_registration_and_one_more_value.test
kzangeli: Anyway, A2 looked up in b8063f7

test/functionalTest/cases/787_cprs_full_functional/update_false_registration.test
fgalan: Ok, I missed `brokerStart CP1`.

NTC (timeout and dealing with non-existing CPs is another issue: #880)

src/lib/ngsi10/UpdateContextResponse.cpp
kzangeli: It is 'correct', just not very clear.
I refer to creating a new 'container' for the not found attribute in case there is no contaioner already (contextElementResponse). If you  read first (1) and then (2) it might get a little clearer.
However, I will make the comment clear

kzangeli: Fixed in 3b271e6

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yeah, this will be moved to a function that picks the 'cleaning function' according to the format of the payload. No cleaning will probably be necessary for JSON, but still. I will separate this to a function when implementing the 'JSON for forwarding'.
Let's wait until then. This has nothing to do with the current PR; only visible as code has been moved.

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Yeah, perhaps.
To me it is very clear right now, but in a few weeks ...
foundAndNotFoundAttributeSeparation?

It is all about adding a ContextElementResponse for attributes that are Not Found - they need a different statusCode. 

There are three different cases:
1. All attributes are Found - nothing to touch
2. All attributes are Not Found - the statusCode is changed from 200 to 404
3. A mix of attributes Found and Not Found - here I need a new ContextElementResponse and all the Not Found attributes must be moved to the new ContextElementResponse (that has statusCode 404).

If you find a good name ...

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: NTC

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: foundAndNotFoundAttributeSeparation sounds good

test/unittests/mongoBackend/mongoUpdateContext_test.cpp
kzangeli: Added a fixme to the exact line that causes the crash and left the functest in DISABLED mode, in 9b0269c

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: Fixed in 6e7382c

fgalan:
LGTM

PR #888, opened 2015-04-22
Task/valgrind for feature 787

kzangeli:
Fixed the leaks from the 787 feature branch.
Also, some DB-cleanup in a few functests.
And, a retry if SHELL-INIT fails in the functional test harness.

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -1373,7 +1373,7 @@ void pruneContextElements(ContextElementResponseVector& oldCerV, ContextElementR
     ContextElementResponse* newCerP = new ContextElementResponse();
 
     /* Note we cannot use the ContextElement::fill() method, given that it also copies the ContextAttributeVector. The side-effect
-     * of this is that attributeDomainName and domainMetadataVector and not being copied, but it should not be a problem, given that
+     * of this is that attributeDomainName and domainMetadataVector are not being copied, but it should not be a problem, given that
      * domain attributes are not implemented */
     newCerP->contextElement.entityId.fill(&cerP->contextElement.entityId);
     newCerP->contextElement.providingApplicationList = cerP->contextElement.providingApplicationList;  // FIXME P10: not sure if this is the right way of doing, maybe we need a fill() method for this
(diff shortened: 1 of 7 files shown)


test/functionalTest/testHarness.sh
```diff
@@ -638,8 +638,25 @@ function runTest()
 
   if [ "$exitCode" != "0" ]
   then
-    exitFunction 11 "SHELL-INIT exited with code $exitCode" $path "($path)" "" DIE
-    return
+    # 3.2 Run the SHELL-INIT part AGAIN
```
fgalan: I would add some sentence about why re-running SHELL-INIT is need (something like "this step may be needed due to glitches that we suspect are due to one test starts before the former has ended).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1786,26 +1788,28 @@ bool processOnChangeCondition
         if (condValues != NULL) {
             /* Check if some of the attributes in the NotifyCondition values list are in the entity.
              * Note that in this case we do a query for all the attributes, not restricted to attrV */
-            ContextElementResponseVector rawCerV;
+            ContextElementResponseVector rawCerV2;
```
fgalan: Why this has been renamed? Maybe a temporal fix that at the end is not needed? In that case, let's go back to `rawCerV`.

kzangeli: No, rawCerV cannot be used.
That name is already used in this function, a few lines before.
If you dont like rawCerV2, let's invent some other variable name ...

test/functionalTest/testHarness.sh
kzangeli: ok

kzangeli: Fixed in 024a83c

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: We have end of using the `rawCerV` variable at line 1779 with `rawCerV.release()`. Thus, I understand that the variable is "free" and can be used again here. Is that correct?

kzangeli: I added that call to release :-)
It *can* be used. It is just not a very good idea to shadow your own variables.

If you really want to keep using that name, why redeclare it?
Let's just reuse the same variable ...

fgalan: Ah, ok, I think I get the point...

Redeclaring was wrong here. Thus, remove it, avoid shadowing and let's use the same variable (free at this point of code) for this block of code. Makes sense?

kzangeli: I am compiling that solution right now.

kzangeli: Fixed (and tested) in e618f7b

fgalan:
LGTM

PR #900, opened 2015-04-27
Removed mongoQueryContext_test.cpp.bak

kzangeli:
See title :-)


fgalan:
LGTM

I have added an addition commit to this PR, in order to include *.bak in .gitignore, so this won't happen again (it was my fault, it seems that *.bak is used by Qtcreator for not yet saved filed). If you agree in the commit, then LGTM on your side and do the final merge + delete branch.

By the way, there is another similar branch task/722_bak_fix that seems to be the same and I understand it should also deleted for repository cleanness.

kzangeli:
Yes, the other branch I made from the feature branch, not from develop.
I will remove it asap.

kzangeli:
LGTM

PR #902, opened 2015-04-28
Fixed a 'funny' leak

kzangeli:
hasMetadata() was called with ContextAttribute* when it accepted a ContextAttribute.
The C++ compiler decided to create a ContextAttribute out of the ContextAttribute* and that object got lost, of course ...


fgalan:
LGTM

PR #903, opened 2015-04-29
Feature/880 timeout for forward and notifications

kzangeli:
Timeout (default: 5000 milliseconds) for notifications and forwarding of updates/queries.
A new command line parameter '-appTmo' has been added to the broker for the purpose of modifying the timeout time.

Complete with two simple functests, testing one notification, one query forward and one update forward, all of them producing timeouts.


CHANGES_NEXT_RELEASE
```diff
@@ -7,3 +7,6 @@
        a consequence, the 'attributeFormat' URI parameter is no longer needed in the 'GET /v1/contextTypes', so it is ignored.
 - Fix: DELETE updates are resolved locally, previously they were forwarded to CPrs (Issue #755)
 - Fix: The URI parameter 'notifyFormat' is now treated the same way for subscriptions as for registrations.
+- Add: Notifications and forwarded queries/updates are now controlled by a timeout. (Issue #880)
+       The default timeout is set to 5000 milliseconds (5 seconds), including connection, writing the request and reading the response.
+       To change this timeout, a new CLI option '-appTmo' has been added to the broker.
```
fgalan: I think it would be better (i.e. more easy for users) a parameter with "timeout" in the name, e.g. "-notifyTimeout".

In additio, to make thins more homogeneous, I would suggest to do some changes in the current "-timeout" CLI parameters (used for replica set timeout):

* Change the name to `-dbTimeout`
* Change its magnitude to and integer meaning miliseconds (currently `-timeout` use float meaning seconds, accepting decimal values).

The above changes involve a slight modificiation in exsting CLI. Howeve, taking into account that replica set are not widely used and that `-timeout` hasn't been here from a long time (it was introduced in the very previous Orion version, i.e. 0.20.0) I think the impact is controlled.

test/functionalTest/cases/880_timeout_for_forward_and_notifications/notification_that_times_out.test
```diff
@@ -0,0 +1,153 @@
...
+
+--SHELL--
+
+#
+# 01. Register E1/T1/A1
+# 02. Availability-Subscribe to E1/A1 for accumulator, on a non-responding URL
+# 04. Sleep 6 secs (default timeout for notifications is 5 seconds)
+# 05. Grep in brokers logfile to see that the notification failed and that the broker thread continued
+# 06. Dump notifications in accumulator - see nothing
+#
```
fgalan: I would say that 90% of the notification use cases are for NGSI10. Thus, this test probably models a "closer reality" using NGSI10. However, probably doesn't worth the effort to create a new test if the functionality is already tested with this one (in this case, mark this comment as NTC).

test/functionalTest/harnessFunctions.sh
```diff
@@ -450,7 +450,7 @@ function accumulatorStart()
 
   accumulatorStop $port
 
-  accumulator-server.py $port /notify $bindIp 2> /tmp/accumulator_$port &
+  accumulator-server.py $port /notify $bindIp > /tmp/accumulator_${port}_stdout 2> /tmp/accumulator_$port &
```
fgalan: Interesting... why this change was needed?

kzangeli: This was added as an intent to have some debugging from the accelerator.
I had a lot of problems adding the 'blocking' services (as I don't know python).
After finishing, I let it stay. Might be useful in the future ...

test/functionalTest/cases/880_timeout_for_forward_and_notifications/notification_that_times_out.test
kzangeli: I didn't think very much before chosing the method to provoke the timeout.
The exact same function is used for ngsi9 as for ngsi10, so what's tested is tested equally using the ngsi9 approach.

NTC

CHANGES_NEXT_RELEASE
kzangeli: ok to modify the parameter names.
I used 'app' as a concept that covers both notifications and forwards.
We use 'application' in the presentations.
Only 'notify' isn't good enough as it implies that the timeout is only for notifications.
We need something that covers both notifications and forwards to updates and queries ...

test/functionalTest/harnessFunctions.sh
fgalan: However, if you use "_stdout" for the standard output after `>`, shouldn't we use "_stderr" for the error output after `2>`?

CHANGES_NEXT_RELEASE
fgalan: What about `-httpTimeout`?

fgalan: Or `-httpReqTimeout` if we want to be more specific and refer to HTTP requests.

kzangeli: httpTimeout sounds good. Not too long ...

test/functionalTest/harnessFunctions.sh
kzangeli: Sure, why not.
I didn't change this just to not change already existing names.
But, the idea is good. Let's do it ! ;-)

CHANGES_NEXT_RELEASE
kzangeli: Fixed in f6d08be

test/functionalTest/harnessFunctions.sh
kzangeli: Fixed in 66b3a98

CHANGES_NEXT_RELEASE
kzangeli: Also, fixed CHANGES_NEXT_RELEASE in 433c889

fgalan:
LGTM

Issue #898, opened 2015-04-27
notify_format_in_ngsi9_subscriptions.test


kzangeli:
The functest notify_format_in_ngsi9_subscriptions.test has not been completed.
What's missing: cases for update-subscription.
Also, we might need to modify the broker for this to work.

When a registration/subscription is updated, should the notifyFormat be modified as well?

**Effort: 2 man days**

fgalan:
Regarding timing, implement this issue first, after that implement https://github.com/telefonicaid/fiware-orion/issues/893 (given that #893 implies copying the NGSI9 test).

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/904

PR #904, opened 2015-04-30
Hardening/898 notify format and updating subscriptions

kzangeli:
Implements #898 

Made updates of ngsi9 subscriptions use the URI param **notifyFormat** and finished the functest about ngsi9 subscriptions (the part about updating a subscription).


fgalan:
LGTM

Issue #910, opened 2015-05-04
-useSemaphore CLI parameter


fgalan:
Up to 0.21.0, CB ensures than no more than one request is processed at the same time at mongoBackend using a semaphore-based system. However, in Active/Active configurations, that cannot be ensured globally, so why don't relax some request semaphore?

To this end, develop a new CLI parameter "useSemaphore " of type string:

* If the value is "all" then all the requests use the semafore. This would be de default value and the one we have in 0.21.0 and before.
* If the value is "read" then only read-like requests use the semaphore.
* If the value is "write" then only write-like requests use the semaphore.
* If the value is "none" then no request uses the semaphore.

fgalan:
Maybe useSemaphore is not a very good name... Another alternative is "-blockRequest"

fgalan:
To implement this, the simplest approach maybe is to modify reqSemTake() and reqSemGive() functions in sem.cpp, doing actual taking/giving or pass-through depending on the CLI flag value and the `who` argument.

kzangeli:
-dbMutex ?

Now, what if we actually perform two updates on the same data ...
What is the worst case scenario?

fgalan:
> what if we actually perform two updates on the same data

In this case, we are in the same situation that with 2 CBs configured in Active/Active mode, that, if you rememeber the analysis we did last month ;), is a safe case after implementing https://github.com/telefonicaid/fiware-orion/issues/878 in 0.21.0.

I like `-dbMutex` more than `-useSemaphore`. However, I think I can improve it... what about `-dbMutexPolicy`?

fgalan:
Actually, `db` in the name should be removed, as the CLI parameters is about the *request* semaphore (DB semaphores are for driver protection, should they must not be configurable and remain as they are now). 

Thus, my proposal is for `-mutexPolicy`.

kzangeli:
'-dbMutexPolicy' ... I expect the policy to come after, like "-dbMutexPolicy NONE", or something (actually, "-dbMutex" is a little like that also. "-useSemaphore" on the other is clearly a flag)

Let's try to find some clearer name.

-dbMutexProtect?

fgalan:
`-mutexProtect` is also fine.

kzangeli:
There is a value coming after the option, as I was 'wanting' in my last comment before this one.
-dbMutexPolicy is just fine as a name.

[ I guess I didn't really read the first comment on this issue ... ]

fgalan:
[...]

PR #913, opened 2015-05-05
Request semaphore for mongo backend is now optional via CLI parameter

kzangeli:
Adding a new CLI parameter: -mutexPolicy (none/read/write/all) to turn off (all on is default) mutex protection for requests in mongoBackend.
The idea is to save some execution time, see Issue #910 for more info.


src/app/contextBroker/contextBroker.cpp
```diff
@@ -211,6 +211,7 @@ bool            mtenant;
 char            rush[256];
 long            dbTimeout;
 long            httpTimeout;
+char            mutexPolicy[16];
```
fgalan: CHANGEX_NEXT_RELEASE entry about the new CLI parameter should be included.

src/lib/mongoBackend/mongoNotifyContext.cpp
```diff
@@ -43,7 +43,9 @@ HttpStatusCode mongoNotifyContext
   const std::vector<std::string>&  servicePathV
 )
 {
-    reqSemTake(__FUNCTION__, "ngsi10 notification");
+    bool reqSemTaken;
+
+    reqSemTake(__FUNCTION__, "ngsi10 notification", SemReadOp, &reqSemTaken);
```
fgalan: I think mongoNotifyContext refers to the processing of notifyContextRequests. Thus, it is like a updateContext. In that case, SemWriteOp should be used.

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
```diff
@@ -46,10 +46,11 @@ HttpStatusCode mongoNotifyContextAvailability
 )
 {
     const std::string notifyFormat = uriParam[URI_PARAM_NOTIFY_FORMAT];
+    bool              reqSemTaken;
 
     LM_T(LmtMongo, ("Notify Context Availability: '%s' format", notifyFormat.c_str()));
 
-    reqSemTake(__FUNCTION__, "mongo ngsi9 notification");
+    reqSemTake(__FUNCTION__, "mongo ngsi9 notification", SemReadOp, &reqSemTaken);
```
fgalan: I think mongoNotifyContextAvailabiltity refers to the processing of notifyContextRequests. Thus, it is like a registerContext. In that case, SemWriteOp should be used.

src/app/contextBroker/contextBroker.cpp
kzangeli: Yes, now that develop is 'clean' I can add the line in CHANGES_NEXT_RELEASE

src/lib/common/sem.cpp
```diff
@@ -83,14 +85,39 @@ int semInit(int shared, int takenInitially)
 *
 * reqSemTake -
 */
-int reqSemTake(const char* who, const char* what)
+int reqSemTake(const char* who, const char* what, SemRequestType reqType, bool* taken)
 {
   int r;
 
+  if (reqType == SemNoneOp)
```
fgalan: This is just a "theoretical" case, right? I mean, we don't have any SemNoneOp operation...

src/lib/mongoBackend/mongoNotifyContext.cpp
kzangeli: ok, I was thinking about notifying, not about getting notified ...
If processContextElement modifies the database, it should definitely be SemWriteOp.

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
kzangeli: ok

src/lib/common/sem.cpp
kzangeli: I didn't mean to ...
I guess I was thinking about reqPolicy and treated both of them the same way.
I will remove this part, thanks for discovering it :-)

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in 861eb29

src/lib/mongoBackend/mongoNotifyContext.cpp
kzangeli: Fixed in 861eb29

src/lib/mongoBackend/mongoNotifyContextAvailability.cpp
kzangeli: Fixed in 861eb29

src/lib/common/sem.cpp
kzangeli: Fixed in 861eb29

fgalan:
LGTM

Issue #911, opened 2015-05-04
Include the overall time that threads spend waiting in request and driver semaphores in statistics operation


fgalan:
This information would be very useful to analysis performance scenarios in the broker, specially when combined with useSemaphore/blockRequest CLI (#910)

fgalan:
Implemented in PR #915

PR #915, opened 2015-05-11
Feature/911 sem wait time measures

kzangeli:
Implements #911

Added the accumulated waiting time on database semaphores to the statistics information for the REST service /statistics.
However, this new feature is enabled only if the broker is started with the new CLI parameter 
-mutexTimeStat

### DIFF
--- a/src/lib/common/sem.cpp
+++ b/src/lib/common/sem.cpp
@@ -24,6 +24,7 @@
 */
 #include <semaphore.h>
 #include <errno.h>
+#include <time.h>
 
 #include "logMsg/logMsg.h"
 #include "logMsg/traceLevels.h"
@@ -45,6 +46,15 @@ static SemRequestType  reqPolicy;
 
 /* ****************************************************************************
 *
+* Time measuring variables - 
+*/
+static struct timespec accReqSemTime   = { 0, 0 };
+static struct timespec accMongoSemTime = { 0, 0 };
+
+
+
+/* ****************************************************************************
+*
 * semInit -
 *
 *   parameter #2: 0 - the semaphore is to be shared between threads,
@@ -55,7 +65,7 @@ static SemRequestType  reqPolicy;
 *  -1 on failure
 *
 */
-int semInit(SemRequestType _reqPolicy, int shared, int takenInitially)
+int semInit(SemRequestType _reqPolicy, bool semTimeStat, int shared, int takenInitially)
 {
   if (sem_init(&reqSem, shared, takenInitially) == -1)
   {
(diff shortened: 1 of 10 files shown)


src/lib/common/sem.cpp
```diff
@@ -83,6 +96,42 @@ int semInit(SemRequestType _reqPolicy, int shared, int takenInitially)
 
 /* ****************************************************************************
 *
+* difftimeofday - 
+*/
+static void difftimeofday(struct timespec* endTime, struct timespec* startTime, struct timespec* diffTime)
```
fgalan: calmecase should be diffTimeOfDay

src/lib/common/sem.cpp
```diff
@@ -83,6 +96,42 @@ int semInit(SemRequestType _reqPolicy, int shared, int takenInitially)
...
+  }
+}
+
+
+
+/* ****************************************************************************
+*
+* addtimeofday - 
+*/
+static void addtimeofday(struct timespec* accTime, struct timespec* diffTime)
```
fgalan: camelCase should be addTimeOfDay

src/lib/common/sem.cpp
```diff
@@ -45,6 +46,15 @@ static SemRequestType  reqPolicy;
 
 /* ****************************************************************************
 *
+* Time measuring variables - 
+*/
+static struct timespec accReqSemTime   = { 0, 0 };
+static struct timespec accMongoSemTime = { 0, 0 };
+
```
fgalan: I think we should also do the same for the transaction semaphore (I forget about it when I write #911)

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -143,261 +148,272 @@ std::string statisticsTreat
...
   {
-    out += TAG_ADD("discoveryErrors", noOfDiscoveryErrors);
+    out += TAG_ADD_COUNTER("discoveryErrors", noOfDiscoveryErrors);
+  }
+
+  if (semTimeStatistics)
+  {
+    char requestSemaphoreWaitingTime[64];
+    semTimeReqGet(requestSemaphoreWaitingTime, sizeof(requestSemaphoreWaitingTime));
+    out += TAG_ADD_STRING("requestSemaphoreWaitingTime", requestSemaphoreWaitingTime);
```
fgalan: Why in this case we use TAG_ADD_STRING? requestSemaphoreWaitingTime is not a counter? I'm missing some subtle detail...

kzangeli: requestSemaphoreWaitingTime is not a counter. However, it's also not a string :-)

The 'true' counters are initialized to -1 (that way I know that they are untouched. Can't recall the details right now but it is implemented that way. If the counter is -1 it is not included in the response). So, to report the value of the counter, the macro TAG_ADD_COUNTER adds 1 to the counter before adding it to the output. That is why I nneeded a new macro for the semaphore accumulated times.

src/lib/common/sem.cpp
kzangeli: This name was picked this way because of the standard C library function 'gettimeofday'.
The name 'difftimeofday' shows that it is related to 'gettimeofday' ...

However, after creating these functions I realized gettimeofday wasn't good enough as it only gives microsconds. I found another function instead, clock_gettime, and after this change, the name difftimeofday no longer makes any sense.

Changing it for clock_difftime (and clock_addtime)

src/lib/serviceRoutines/statisticsTreat.cpp
kzangeli: Looking closer at the variable used for the semaphore times, they are actually strings.
semTimeReqGet, semTimeMongoGet and semTimeTransGet all give their output in a string.

NTC

src/lib/common/sem.cpp
kzangeli: See previous answer

src/lib/common/sem.cpp
kzangeli: Fixed in 7cccbfb

src/lib/common/sem.cpp
kzangeli: Fixed in 7cccbfb

src/lib/common/sem.cpp
kzangeli: Fixed in 7cccbfb

fgalan:
LGTM

Issue #828, opened 2015-03-12
Behavior of URL with final '/' vs without it


LeandroGuillen:
The usage of '/' could make Orion behave in a way that is not particularly obvious.

Example:
**GET /v1/contextEntityTypes/** queries for the empty type "".

Is that what was expected to happen?

kzangeli:
Nice side effect! :-)

No, this should not happen. Needs to be fixed, I would say.

LeandroGuillen:
And how would we query for an empty type? Maybe a special parameter of the "type=empty" kind?

fgalan:
In my opinion, as a general rule, each URL in the form:

```
GET /foo/bar/ruru/
```

show "route" to the same URL than:

```
GET /foo/bar/ruru
```

kzangeli:
Right now we can't do it (except for your new-found feature :-)).
Doesn't make much sense to accept the URI parameter 'empty=entity::type' for a request with a type in the URI;

    GET /v1/contextEntityTypes/xxx?empty=entity::type

A request like that right now gives an error.
GET /v1/contextEntityTypes is not supported (there ?empty=entity::type would make sense), so, there is no way to query for empty types (with convops).
Doing a standard query with id = pattern(.*) and type = "" should give you EVERYTHING, but the URI option 'empty=entity::type' is not supported by the standard op.

I believe we need to think a little and see how to support this (and whether we want to support it - I would definitely prefer to forbid empty types before encouraging its usage).

Meanwhile, the bug should be fixed. 

**GET /v1/contextEntityTypes/** should behave just like **GET /v1/contextEntityTypes**

I am pretty sure this 'feature' is present for all requests.
It needs to be solved in the central place where the URI is parsed.

fgalan:
Labeled as bug (the change about making the "slash-ended" URL behave as the "slash-not_ended" URL... regarding *"And how would we query for an empty type? Maybe a special parameter of the "type=empty" kind?"*, it should be spinned off this issue).

kzangeli:
Yes, we have two different issues here, one is a bug and the other one is a possible feature

LeandroGuillen:
This bug is confirmed to happen also with **GET /v1/contextEntities/** as reported by a user in stackoverflow: http://stackoverflow.com/questions/29409565/how-can-i-delete-an-id-less-entity-in-orion/

But remarkably not with **DELETE /v1/contextEntities/**

kzangeli:
[...]

PR #922, opened 2015-05-18
Bug/828 slash at end of uri path

kzangeli:
Implemnets #828 

The title says is all.
The path of the URL of incoming request are stripped of trailing '/' before continuing.
This fixes the issue 828.

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -604,7 +604,7 @@ static int contentTypeCheck(ConnectionInfo* ciP)
 
 /* ****************************************************************************
 *
-* urlCheck - 
+* urlCheck - check for forbidden characters and remove trailing slashes
 *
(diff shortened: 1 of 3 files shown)


src/lib/rest/rest.cpp
```diff
@@ -604,7 +604,7 @@ static int contentTypeCheck(ConnectionInfo* ciP)
```
fgalan: CHANGES_NEXT_RELEASE entry missing?

kzangeli: Ah, of course!
When the branch was prepared (all these queued PR-branches), develop wasn't ready and I couldn't touch CHANGES_NEXT_RELEASE. I had completely forgotten ...

kzangeli: Fixed in c4fc504

fgalan:
LGTM

PR #923, opened 2015-05-18
Bug/691 empty type and subscriptions

### LINKED ISSUE #691: Subscriptions with empty or absent entity.type
### Description
Making tests with service-path in subscriptions (PR #689), we discovered two cases that fail with an empty or absent entity.type in the subscription.

Two functional tests are made (to reproduce the problem):

    test/functionalTest/cases/675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscriptions_with_empty_type_and_service_path.test.DISABLED
    test/functionalTest/cases/675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscriptions_with_no_type_and_service_path.test.DISABLED

[ I'm not sure the service-path has anything to do with these failures, but other func-tests, without service-path, should be made to verify this. ]

kzangeli:
Fix #691 

Fixed a bug regarding subscriptions and empty entity-type.

The functest has suffered a lot of changes. Probably easier to view the entire file instead of the diff ...


fgalan:
LGTM

PR #924, opened 2015-05-18
Bug/885 service path and delete

### LINKED ISSUE #885: Deleting entities within a service path deletes entities in sub-paths
An issue has been raised at :
http://stackoverflow.com/questions/29761530/deleting-entity-in-a-fiware-servicepath-deletes-entities-with-the-same-id-in-su

It basically states that you will delete two entities with the same ID even if they are in different service paths, assuming one path is a sub-path of the other.

The way to replay this situation is as follows.

Create **table1** in service path **/user/home**:

```bash
(curl localhost:1026/v1/updateContext -s -S -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Fiware-ServicePath: /user/home' -d @- | python -mjson.tool) <<EOF
{
 "contextElements": [
   {
      "type": "table",
      "isPattern": "false",
      "id": "table1"
    }
  ],
  "updateAction": "APPEND"
}
EOF
```

Create **table1** in service path **/user/home/room**:
 
```bash
(curl localhost:1026/v1/updateContext -s -S -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Fiware-ServicePath: /user/home' -d @- | python -mjson.tool) <<EOF
{
 "contextElements": [
   {
      "type": "table",
      "isPattern": "false",
      "id": "table1"
    }
  ],
  "updateAction": "APPEND"
}
EOF
```

Right now, the situation in the database is:

```mongo
> db.entities.find()
{
    "_id": {
        "id": "table1",
        "type": "table",
        "servicePath": "/user/home"
    },
    "attrs": [],
    "creDate": 1429630324,
    "modDate": 1429630324
}
{
    "_id": {
        "id": "table1",
        "type": "table",
        "servicePath": "/user/home/room"
    },
    "attrs": [],
    "creDate": 1429630325,
    "modDate": 1429630325
}
```

Now we do a deletion on entity **table1** but indicating service path **/user**:

```bash
(curl localhost:1026/v1/updateContext -s -S -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Fiware-ServicePath: /user/home' -d @- | python -mjson.tool) <<EOF
{
 "contextElements": [
   {
      "type": "table",
      "isPattern": "false",
      "id": "table1"
    }
  ],
  "updateAction": "DELETE"
}
EOF
```

Now the database is empty:
```mongo
> db.entities.find()
> 
```

This is not an acceptable behaviour. What should happen is that only entities that reside in the specified service path header should get deleted.

Furthermore, if you do a deletion but on the root service path or ommit it:
```bash
(curl localhost:1026/v1/updateContext -s -S -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Fiware-ServicePath: /' -d @- | python -mjson.tool) <<EOF
{
 "contextElements": [
   {
      "type": "table",
      "isPattern": "false",
      "id": "table1"
    }
  ],
  "updateAction": "DELETE"
}
EOF
```
you get this:
```
{
	"contextResponses": [
	{
		"contextElement": {
			"type": "table",
			"isPattern": "false",
			"id": "table1"
		},
		"statusCode": {
			"code": "404",
			"reasonPhrase": "No context element found",
			"details": "table1"
		}
	}
	]
}
```

So this is another aspect of the same bug.

In short:
 1. Deleting entities with same id in different service paths should not be allow without explicitly using the # character.
 2. The behavior is not consistent in these two cases:
	- Parent service-path is root
	- Parent service-path is _not_ root

kzangeli:
Fix #885 

Fixed a bug about service paths.
Removed three old unit tests that tested that the "bug worked" ... :-)


CHANGES_NEXT_RELEASE
```diff
@@ -4,3 +4,4 @@
        the new CLI parameter -mutexTimeStat is set (Issue #911)
 - Fix: Removing trailing slashes for path of URI before treating the request (Issue #828)
 - Fix: Fixed a bug regarding subscriptions with empty entity-type (Issue #691)
+- Fix: Fixed a bug regarding Service paths and DELETE operation (Issue # 885)
```
fgalan: The bug actually is broader, so probably this entrey would be more accurate:

"Fix: updateContext propragation to Service Path subtree, instead of just scoping to the particular Service Path in the Fiware-Service header (Issue # 885)"

Or a improved English wording :)

test/functionalTest/cases/885_service_path_and_delete/delete_with_service_path.test
```diff
@@ -0,0 +1,320 @@
...
+----SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create E/T/A1, service-path /father
+# 02. Create E/T/A1, service-path /father/son
+# 03. Query  E/T/A1 - see two entities
```
fgalan: Maybe we could be more precise:

"03. Query  E/T/A1 on (implicity) /# - see two entities"

fgalan: And the same for 05

test/unittests/mongoBackend/mongoUpdateContext_test.cpp
```diff
@@ -9199,110 +9199,6 @@ TEST(mongoUpdateContextRequest, servicePathEntityUpdate_3levels)
 
 /* ****************************************************************************
 *
-* servicePathEntityUpdates_2levels -
-*
-*/
-TEST(mongoUpdateContextRequest, servicePathEntityUpdate_2levels)
```
fgalan: I think you told me but I forget... :$

Why these 3 unit tests has been removed? Looking to the removed code, I don't see the situation they are modeling is impossible (from a cliente API perpective)... maybe they have been removed because they are redundant with other existing tests?

kzangeli: They way I saw it they test that the broker works the way it should NOT work - the bug fixed in this PR.
However, you are definitely better than I am 'dechiffering' unit tests.
If you don't have the same feeling that I had then something might be wrong ...

So, please look again and make sure that the three tests are OK (and that they do NOT test the invalid way of working that was just fixed in this PR).

test/functionalTest/cases/885_service_path_and_delete/delete_with_service_path.test
kzangeli: Fixed in 838cd81

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 838cd81

test/unittests/mongoBackend/mongoUpdateContext_test.cpp
kzangeli: prepareDatabaseWithServicePaths creates two entities:
- E1/T1 with service path "/home/kz/01"
- E1/T1 with service path "/home/kz/02"

mongoUpdateContext is called with service path "/home/kz", not "/home/kz/#", but still, both entities are found and updated.

It seems to be this is the exact behaviour we want to avoid (that this PR fixes).
Have I misunderstood something/everything?

fgalan: No, I think you have understood it perfectly. I was missing al the "context" (prepareDatabase, etc.), but after reading your explanation makes fully sense to remove the tests.

NTC

fgalan:
LGTM

PR #926, opened 2015-05-21
Feature/legacy mongo

fgalan:
PR on behalf of @crbrox (he did all the work, I just only fix a couple of things :)

LGTM from @kzangeli and @crbrox are needed.


CHANGES_NEXT_RELEASE
```diff
@@ -6,3 +6,4 @@
 - Fix: Removing trailing slashes for path of URI before treating the request (Issue #828)
 - Fix: Fixed a bug regarding subscriptions with empty entity-type (Issue #691)
 - Fix: updateContext propagated to the entire Service Path subtree, instead of just scoping to the particular Service Path in the Fiware-Service header (Issue # 885)
+- Mongo driver migrated to legacy-1.0.2
```
fgalan: We don't use to include such internal changes in CNR. However, I think the impact of the change could be large enough to deserve this entry. 

NTC (just informative)

README.md
```diff
@@ -104,16 +104,14 @@ sudo yum install make cmake gcc-c++ scons
...
 
 ```
-wget http://downloads.mongodb.org/cxx-driver/mongodb-linux-x86_64-2.2.3.tgz
-# Don't pay attention to the '-linux-x86_64' in the name: it is actually source code for any platform
-tar xfvz mongodb-linux-x86_64-2.2.3.tgz
-cd mongo-cxx-driver-v2.2
-scons                                         # The build/libmongoclient.a library is generated as outcome
-sudo scons install                            # This puts .h files in /usr/local/include and libmongoclient.a in /usr/local/lib
-sudo chmod a+r -R /usr/local/include/mongo    # It seems that scons install breaks permissions
+wget https://github.com/mongodb/mongo-cxx-driver/archive/legacy-1.0.2.tar.gz
```
fgalan: This is a slight variant of the procedure @crbrox sent by email (based on "git checkout legacy"), in order to ensure that we stick to a particular version rather than to the "moving picture" of a git branch.

NTC (just informative)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -187,12 +215,16 @@ bool mongoConnect(const char*         host,
     //
     // WriteConcern
     //
-    int writeConcernCheck;
+    mongo::WriteConcern writeConcernCheck;
```
fgalan: @kzangeli maybe you find interesting this part... it seems that the writeConcern has changed in legacy, so I needed to adapt your code a little bit. Check if it is ok, please.

README.md
```diff
@@ -104,16 +104,14 @@ sudo yum install make cmake gcc-c++ scons
...
+tar xfvz legacy-1.0.2.tar.gz
+cd mongo-cxx-driver-legacy-1.0.2
+scons                                         # The build/linux2/normal/libmongoclient.a library is generated as outcome
+sudo scons install --prefix=/usr/local        # This puts .h files in /usr/local/include/mongo and libmongoclient.a in /usr/local/lib
 ```
```
fgalan: @robertocs have a look to this: is the new procedure you would need to use to install the driver from 0.22.0 on.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Looks good.
NTC

test/unittests/CMakeLists.txt
```diff
@@ -285,6 +285,7 @@ link_directories("/opt/lib/")
 # flags
 # ------------------------------------------------------------
 add_definitions(-DUNIT_TEST)
+add_definitions(-g)
```
kzangeli: Was this necessary?
I've been running unitTest under gdb for a long long time without problems ... strange ...

kzangeli:
So, the big work has been in unittests. In the 'real' code I only see three changes:
1. OID::str() => OID::toString
2. ensureIndex => createIndex
3. That detail about Write Concern

crbrox:
LGTM

test/unittests/CMakeLists.txt
kzangeli: Fixed in 2981ea3

kzangeli:
LGTM

fgalan:
> So, the big work has been in unittests. In the 'real' code I only see three changes:
> 1. OID::str() => OID::toString
> 2. ensureIndex => createIndex
> 3. That detail about Write Concern

There are some other changes, e.g. mongo::client::initialize() and mongo::client::shutdown(), along with new signatures for update(), insert() and remove() adding an optional argument to specify per-operation write concern (I guess that if that argument is not used, then the default set at connection object level is used, otherwise the setWriteConcern() method would be useless :). We don't use that additional argument in our code but it brought some headache adjusting mocks (that need to much the signature of the function taking into account also optional parameters).

fgalan:
LGTM quorum achieved. Merging...

Issue #909, opened 2015-05-04
-dbPoolSize CLI parameter


fgalan:
Develop a new CLI parameter "dbPoolSize" of type integer with the following behaviour:

* If pool size is 0 then behave as in 0.21.0 and before: DB pool is not used
* If pool size is greater than 0 then use a pool of connections at driver level
* Default value is 0 (to keep backward compatibility). We should also think in a reasonable maximum.

fgalan:
I think that IDAS/IoTAgend code uses this kind of driver. @kzangeli , you can contact @agonlucas so he can tell you where to look.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/928

fgalan:
(Sorry for the assigne/milestone "dance")

PR #928, opened 2015-05-22
Feature/909 mongo connection pool

kzangeli:
Implements #909 

Implemented a connection pool for the database.
Removed the 'mongo semaphore'.


CHANGES_NEXT_RELEASE
```diff
@@ -5,5 +5,6 @@
        the new CLI parameter -mutexTimeStat is set (Issue #911)
 - Fix: Removing trailing slashes for path of URI before treating the request (Issue #828)
 - Fix: Fixed a bug regarding subscriptions with empty entity-type (Issue #691)
-- Fix: updateContext propagated to the entire Service Path subtree, instead of just scoping to the particular Service Path in the Fiware-Service header (Issue # 885)
-- Mongo driver migrated to legacy-1.0.2
+- Fix: updateContext propagated to the entire Service Path subtree, instead of just scoping to the particular Service Path in the Fiware-Service header (Issue #885)
+- Fix: Mongo driver migrated to legacy-1.0.2
```
fgalan: Not sure if this is actually a fix of a "evolution" ;) Not very important anyway, let's keep as it is.

NTC

CHANGES_NEXT_RELEASE
```diff
@@ -5,5 +5,6 @@
...
+- Add: Database connection pool for mongo (Issue #909)
```
fgalan: The CLI parameter used for the pool should be mentioned, along with the default value.

src/lib/common/clockFunctions.cpp
```diff
@@ -0,0 +1,63 @@
+/*
```
fgalan: I guess that this functionas are not new, but a refactoring of the ones we aready have from a previous PR. Is that correct?

src/lib/common/sem.cpp
```diff
@@ -74,16 +73,10 @@ int semInit(SemRequestType _reqPolicy, bool semTimeStat, int shared, int takenIn
...
-  {
-    LM_E(("Runtime Error (error initializing 'mongo' semaphore: %s)", strerror(errno)));
-    return 2;
-  }
-
   if (sem_init(&transSem, shared, takenInitially) == -1)
   {
     LM_E(("Runtime Error (error initializing 'transactionId' semaphore: %s)", strerror(errno)));
-    return 3;
+    return 2;
```
fgalan: This is not coherent with the comment at the function preamble:

```
 * RETURN VALUE (of sem_init)
 *   0 on success,
 *  -1 on failure
```

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -106,62 +106,106 @@ static void shutdownClient(void)
...
 {
+  static bool called = false;
 
-    std::string err;
-    multitenant = _multitenant;
-
-    LM_T(LmtMongo, ("Connection info: dbName='%s', rplSet='%s', timeout=%f", db, rplSet, timeout));
+  if (called == true)
+  {
+    LM_W(("mongoStart already called - can only be called once, so please fix!"));
```
fgalan: If this happen is more sever, let's use a LM_E with "Runtime Error" (or something like that, I don't remember pretty well...).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -106,62 +106,106 @@ static void shutdownClient(void)
...
+  }
+  called = true;
 
-    mongoSemTake(__FUNCTION__, "connecting to mongo");
+  multitenant = _multitenant;
 
-    // We trust this function is called once. If not, this call
-    // should be protected against multiple calls
-    if (!clientIsInitialized)
+	if (!clientIsInitialized)
```
fgalan: `called` and `clientIsInitialized` seems to be for the same purpose. Why don't use just one?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -106,62 +106,106 @@ static void shutdownClient(void)
...
     }
+	}
+
+  if (mongoConnectionPoolInit(host, db, rplSet, username, passwd, timeout, writeConcern, poolSize, semTimeStat) != 0)
+  {
+    LM_E(("Database Startup Error (cannot initialize mongo connection pool)"));
+    return false;
+  }
+
+  atexit(shutdownClient);
```
fgalan: Maybe this atexit() should be done before mongoConnectinPoolInit() call? The "return false" there could cause an exit and, given that mongo::client::initialize() has been called, mongo::client::shutdown() should be called (indirectly, through atexit()).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -106,62 +106,106 @@ static void shutdownClient(void)
...
+  clientIsInitialized = true;
+  return true;
+}
+
+
+
+/* ****************************************************************************
+*
+* mongoConnect - move to mongoConnectionPool?
```
fgalan: +1 to the movement

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -264,26 +312,43 @@ bool mongoConnect(const char*         host,
...
 
+
+
 /* ****************************************************************************
 *
 * mongoConnect -
 *
-* Version of the functions that doesn't uses authentication parameters
+* Version of the function that doesn't use authentication parameters.
+* FIXME: no longer used?
```
fgalan: Let's remove it from the file. If the compiler doesn't complaint, the it is useless and should be removed :)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -264,26 +312,43 @@ bool mongoConnect(const char*         host,
...
 */
-bool mongoConnect(const char* host) {
+bool mongoConnect(const char* host)
+{
+  return true;
+}
+
 
-    return mongoConnect(host, "", "", "", "", false, 0);
+#ifdef UNIT_TEST
```
fgalan: Maybe this block should under unit test code? Or it would be hard doing in that way?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -293,12 +358,15 @@ bool mongoConnect(const char* host) {
 * object to be mocked.
 *
 */
-void mongoDisconnect() {
+void mongoDisconnect()
```
fgalan: Maybe "setMongoConnectionForUnitTest(NULL)" makes this function useless?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -343,9 +413,25 @@ void setNotifier(Notifier* n) {
...
 {
-    return connection;
+#ifdef UNIT_TEST
+  return connection;
+#else
+  return mongoPoolConnectionGet();
+#endif
 }
 
+DBClientBase* mongoInitialConnectionGetForUnitTest(void)
```
fgalan: Why is this function for? A preamble would be helpful.

src/lib/common/clockFunctions.cpp
kzangeli: These functions are now used from two different places, so I had to move them to a separate file.

CHANGES_NEXT_RELEASE
kzangeli: True, I'll add it to the comment line

CHANGES_NEXT_RELEASE
kzangeli: Yeah, I also din't know what keyword to pick ...

src/lib/common/sem.cpp
kzangeli: ok

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: ok.
However, as we control it, it is not especially severe.
It would be if we actually continued and executed the function a second time ...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -343,9 +413,25 @@ void setNotifier(Notifier* n) {
...
+{
+  return mongoPoolConnectionGet();
+}
+
+void releaseMongoConnection(DBClientBase* connection)
```
fgalan: This function seems to be a direct pass-through. Why keep it?

Or maybe, for symmetry, it should be:

```
void releaseMongoConnection(DBClientBase* connection)
#ifndef UNIT_TEST
 mongoPoolConnectionRelease(connection)
#endif
 }
```
Although I know that mongoPoolConnectionRelease() doesn't need to be "protected" in the case of unit test code, it would help to read this code.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yeah, this seems like a merge-error ...
Will fix it

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: true

src/lib/mongoBackend/MongoGlobal.cpp

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I will try.
If both broker and unitTest links ok then all is ok

src/lib/mongoBackend/mongoConnectionPool.cpp
```diff
@@ -0,0 +1,244 @@
...
+  double      timeout,
+  int         writeConcern,
+  int         poolSize,
+  bool        semTimeStat
+)
+{
+  //
+  // Checking incoming parameters
+  //
+  if (poolSize == 0)
```
fgalan: Is this possible to have `-dbPoolSize 0` in the CLI?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Ah this one!
Yeah, I remember I removed it by commenting it out.
I'll try to remove this function as well.

src/lib/mongoBackend/mongoConnectionPool.cpp
```diff
@@ -0,0 +1,244 @@
...
+  // Note that this is a counting semaphore, initialized to connectionPoolSize.
+  //
+  r = sem_init(&connectionSem, 0, connectionPoolSize);
+  if (r != 0)
+  {
+    LM_E(("Runtime Error (cannot create connection semaphore-set)"));
+    return -1;
+  }
+
+  // Measure accumulated semaphore waiting time?
```
fgalan: Why the question mark?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Might be a little difficult to follow the code if this function is in another not only file but *library*.
This function is setting the value for 'getMongoConnection' for use under unit-tests.

I prefer to leave it here.

What can be done though is some cleaning-up about #ifdef UNIT_TEST (grouping everything together under one ifdef/endif) - must be at least two - the one inside getMongoConnection cannot be grouped.

src/lib/common/clockFunctions.cpp
fgalan: NTC

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 37cd3b3

src/lib/common/sem.cpp
kzangeli: Fixed in 37cd3b3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 37cd3b3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 37cd3b3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 37cd3b3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 37cd3b3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 374e74c

src/lib/mongoBackend/mongoConnectionPool.cpp
kzangeli: The question mark as measure or not measure depend on the parameter semTimeStat (on the next line).

src/lib/mongoBackend/mongoConnectionPool.cpp
kzangeli: It happened to me and EVERYTHING stopped working.
It was a bug of course, but I prefer to protect the code from bugs like that. It took me 50 LM_M and a few hours to understand the whole thing ...

NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: The function to GET a connection is called getMongoConnection (and it calls internally mongoPoolConnectionGet).
The function to RELEASE that connection needs a corresponding name, such as releaseMongoConnection. 
Right now, this function only calls mongoPoolConnectionRelease, but this might change in the future. 
I see no reason to change anything here (even though it is true that we'd gain some execution time calling mongoPoolConnectionRelease directly instead of calling releaseMongoConnection that calls mongoPoolConnectionRelease.

The code was already **full** of calls to getMongoCollection, and we can't have code like this:

    connection = getMongoCollection();
    useConnection();
    mongoPoolConnectionRelease();

This would be hard to understand.

I really think we should keep it as is. I see no better solution (unless we start to hunt nanoseconds in run-time).

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 374e74c

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 374e74c

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 374e74c

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Ok, let's keep the function. What about making explicit that for unit test no release operation is needed?

```
void releaseMongoConnection(DBClientBase* connection)
#ifndef UNIT_TEST
 mongoPoolConnectionRelease(connection)
#endif
 }
```

src/lib/mongoBackend/mongoConnectionPool.cpp
fgalan: `-dbPoolSize 0` shouldn't be allowed (the broker should check that this parameter is an strictly positive integer): the broker should exit in that case with an error message (maybe the cheking should be in contextBroker.cpp at CLI arguments processing time?).

In that case, the if in this part of the code would be to detect a very werid runtime situation (given that the poo size parameter has been checked before) and should be reported with a LM_E Runtime Error.

src/lib/mongoBackend/mongoConnectionPool.cpp
fgalan: Ah, I forget about that. You are right!

src/lib/mongoBackend/mongoConnectionPool.cpp
```diff
@@ -0,0 +1,445 @@
...
+*
+* mongoPoolConnectionRelease - 
+*/
+void mongoPoolConnectionRelease(DBClientBase* connection)
+{
+  sem_wait(&connectionPoolSem);
+
+  for (int ix = 0; ix < connectionPoolSize; ++ix)
+  {
+    if (connectionPool[ix].connection == connection)
```
fgalan: Maybe a stupid question but... how the DBClientBase implementes the equal (==) operator? It is consistent with the way we need to detect that the connection a given thread is using (the object which pointer is passed as argument in this function) is the right one in the connectionPool[] vector?

kzangeli: DBClientBase luckily doesn't override the '==' operator.
Not sure I understand your doubt here.

mongoPoolConnectionGet() returns a connection pointer and this very pointer is later sent to mongoPoolConnectionRelease to release it.

fgalan: I was an stupid question... I have realized that you do pointer comparison, not object comparison. So, you are not aplying == directly to DBClientBase.

NTC

src/lib/mongoBackend/mongoConnectionPool.cpp
kzangeli: Fixed in 113d84e

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: After looking at it a little more in-depth, the problem lies in that testInit.cpp/setupDatabase gets a connection by calling mongoInitialConnectionGetForUnitTest, but there is no corresponding function to release the connection.

Creating an issue about this problem.

fgalan:
LGTM

PR #934, opened 2015-05-25
FIX some small issues

fgalan:
LGTM from @kzangeli and @raconte needed

### DIFF
--- a/test/functionalTest/harnessFunctions.sh
+++ b/test/functionalTest/harnessFunctions.sh
@@ -145,7 +145,7 @@ function localBrokerStart()
   traceLevels=$2
   ipVersion=$3
 
-  POOL_SIZE=1
+  POOL_SIZE=10
 
   shift
   shift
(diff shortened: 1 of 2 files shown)


test/functionalTest/harnessFunctions.sh
```diff
@@ -145,7 +145,7 @@ function localBrokerStart()
   traceLevels=$2
   ipVersion=$3
 
-  POOL_SIZE=1
+  POOL_SIZE=10
```
kzangeli: I would prefer to do this a little better:

    POOL_SIZE=${POOL_SIZE:-10}

Meaning, if POOL_SIZE already exists, use that value, if not, set it to 10.
Can't say I like the name of the variable. CB_DB_POOL_SIZE would be a lot better ...

fgalan: Fixed in a57ca79

raconte:
LGTM

kzangeli:
LGTM

PR #936, opened 2015-05-26
FIX linking instructions. boost regexp is missing. Ubuntu 12.04.2

jmcanterafonseca:
Hello,

It seems boost regex is missing from the linking instructions. I have experienced such a problem in Ubuntu 12.04.2, but I cannot confirm the same problem might happen in other Ubuntu versions or other distros. That's why I'm proposing this PR

thanks!

### DIFF
--- a/test/unittests/CMakeLists.txt
+++ b/test/unittests/CMakeLists.txt
@@ -293,9 +293,10 @@ add_definitions(-DUNIT_TEST)
 
 ADD_EXECUTABLE(unitTest ${SOURCES} ${HEADERS})
 
-IF ((${DISTRO} STREQUAL "Ubuntu_13.04") OR (${DISTRO} STREQUAL "Ubuntu_12.04.1_LTS"))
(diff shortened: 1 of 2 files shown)


fgalan:
Looks goot but... what about unit test. In order works, does the `make unit_test` target works? Maybe a similar fix in the CMakeList.txt for unit test would be needed.

test/unittests/CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ find_path(GMOCK_INCLUDE_DIR gmock/gmock.h
           PATHS /usr/includes
           PATH_SUFFIXES gmock)
 
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-conversion-null")
+
```
jmcanterafonseca: I've just added this here in order not to lose this checking in the regular code

jmcanterafonseca:
Unit tests now compiling. I had to disable conversion null warnings, otherwise I was not able to compile on my Ubuntu box.

test/unittests/CMakeLists.txt
fgalan: Should be added only in the "branch" in which this is needed (I understand that an specific Ubuntu version), in the `IF ()` part.

fgalan:
LGTM

Issue #843, opened 2015-03-18
weird detail field in 500 Internal Error messages 


fgalan:
Testing MongoDB primary handover the following response is get:

```
{
  "contextResponses" : [
    {
      "contextElement" : {
        "type" : "device",
        "isPattern" : "false",
        "id" : "modelgtw2mysql3.iotasset2"
      },
      "statusCode" : {
        "code" : "500",
        "reasonPhrase" : "Internal Server Error",
        "details" : "\&quot;"
      }
    }
  ]
}
```

The problem is not getting a 500 error (which is normal during the primary handover) but the weird content of the `details` field. It should provide more meaningful information.

kzangeli:
Seems like a single quote ' ...  Strange ... 
:-)

fgalan:
Ideally, we should find a way of reproducing this without involving the complex test setup used for testing MongoDB primary handover (ideally it should be reproducible using stand alone MongoDB instance).

XavierVal:
I hope this info help you to catch it.
This is the log when this error was showed again today: 
time=2015-05-05T11:39:28.716CEST | lvl=ERROR | trans=1430815520-585-00000000003 | function=processContextElement | comp=Orion | msg=MongoCommonUpdate.cpp[1666]: Database Error ('{ _id.id: "device_id1", _id.type: "thing", _id.servicePath: /^\/$|^\/\/.*/ }', 'socket exception [SEND_ERROR] for 10.0.0.xx:27017')

kzangeli:
The version of the broker in the test described by @XavierVal  is a little old.
Looking at a newer version, it seems the error comer from line 1696 of MongoCommonUpdate.cpp (the line with the LM_E():

    catch (const DBException &e)
    {
        mongoSemGive(__FUNCTION__, "query in EntitiesCollection (mongo db exception)");
        buildGeneralErrorResponse(ceP, NULL, responseP, SccReceiverInternalError,
                           std::string("collection: ") + getEntitiesCollectionName(tenant).c_str() +
                              " - query(): " + query.toString() +
                              " - exception: " + e.what());
        LM_E(("Database Error ('%s', '%s')", query.toString().c_str(), e.what()));
        return;
    }

I have a feeling that if a send-error like this to mongod occurs, we should try again ...

fgalan:
[...]

PR #937, opened 2015-05-27
Hardening/orion curl rewrite

kzangeli:
Complete rewrite of the function 'orionCurl'.
Some of the changes:
  - the payload is now always sent to a file and -d @<path> is used for curl (internal change only) - /tmp/orionFuncTestPayload (in case no file was used by the caller)
  - it is now possible to read/understand the function :-)
  - ORION_FT_DEBUG=1 turns of debugging (output to /tmp/orionFuncTestDebug.log)
  - New parameters: -H and --header (to add headers to curl)

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -117,7 +117,6 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   {
     char* cP = (char*) val;
 
-    LM_M(("value for URI_PARAM_PAGINATION_OFFSET: '%s'", cP));
     while (*cP != 0)
     {
       if ((*cP < '0') || (*cP > '9'))
@@ -135,7 +134,6 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   {
     char* cP = (char*) val;
 
-    LM_M(("value for URI_PARAM_PAGINATION_LIMIT: '%s'", cP));
     while (*cP != 0)
     {
       if ((*cP < '0') || (*cP > '9'))
@@ -208,8 +206,6 @@ static int httpHeaderGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, co
 {
   HttpHeaders*  headerP = (HttpHeaders*) cbDataP;
   std::string   key     = ckey;
-
-  LM_M(("Got header %s: %s", ckey, value));
 
   LM_T(LmtHttpHeaders, ("HTTP Header:   %s: %s", key.c_str(), value));
 
@@ -758,8 +754,6 @@ static int connectionTreat
     //
     char            ip[32];
     unsigned short  port = 0;
-
-    LM_M(("New incoming connection: %s", url));
 
     const union MHD_ConnectionInfo* mciP = MHD_get_connection_info(connection, MHD_CONNECTION_INFO_CLIENT_ADDRESS);
     if (mciP != NULL)
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -117,6 +117,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   {
     char* cP = (char*) val;
 
+    LM_M(("value for URI_PARAM_PAGINATION_OFFSET: '%s'", cP));
     while (*cP != 0)
     {
       if ((*cP < '0') || (*cP > '9'))
@@ -134,6 +135,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   {
     char* cP = (char*) val;
 
+    LM_M(("value for URI_PARAM_PAGINATION_LIMIT: '%s'", cP));
     while (*cP != 0)
     {
       if ((*cP < '0') || (*cP > '9'))
@@ -157,7 +159,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
     }
     else if (limit == 0)
     {
-      OrionError error(SccBadRequest, std::string("Bad pagination limit: /") + value + "/ [a value of ZERO is unaccepted]");
+      OrionError error(SccBadRequest, std::string("Bad pagination limit: /") + value + "/ [a value of ZERO is unacceptable]");
       ciP->httpStatusCode = SccBadRequest;
       ciP->answer         = error.render(ciP->outFormat, "");
       return MHD_YES;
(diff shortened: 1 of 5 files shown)


test/functionalTest/cases/672_forbidden_chars_in_url/all_forbidden_chars_in_uri.test
```diff
@@ -58,7 +58,7 @@ echo
 
 echo '03. Test the forbidden character "'
 echo "=================================="
-orionCurl --url '/test/with/forbidden/character/"' --json
+orionCurl --url '/test/with/forbidden/character/\"' --json
```
kzangeli: I had to escape the " as "" is used in the call to curl.

Just informative, NTC

src/lib/rest/rest.cpp
```diff
@@ -518,6 +518,18 @@ static char* removeTrailingSlash(std::string path)
 */
 int servicePathSplit(ConnectionInfo* ciP)
 {
+  //
+  // Special case: empty service-path 
+  //
+  if ((ciP->httpHeaders.servicePathReceived == true) && (ciP->servicePath == ""))
```
fgalan: This is related with the last comments in https://github.com/telefonicaid/fiware-orion/issues/843?

test/functionalTest/cases/000_bad_requests/pagination_error_in_uri_params.test
```diff
@@ -37,7 +37,7 @@ echo
 echo
 
 echo "+++++ 2. bad characters in URI param 'offset'"
-url="/v1/queryContext?limit=0&offset=qwe"
+url="/v1/queryContext?offset=qwe"
```
fgalan: Is this related with the orionCurl refactoring or only cleanup?

test/functionalTest/harnessFunctions.sh
```diff
@@ -595,22 +625,200 @@ function dbInsertEntity()
 #   --host        <host>           (default: localhost)
 #   --port        <port>           (default: $CB_PORT)
 #   --url         <URL>            (default: empty string)
+#   --urlParams   <params>         (URI parameters 'in' URL-string)
```
fgalan: You have defined this new parameter for orionCurl but it seems that there isn't any .test using it?

fgalan: I have realized that it is not a  new parameter (it was in the previous version of the function), but no .test is using, anyway...

test/functionalTest/harnessFunctions.sh
```diff
@@ -595,22 +625,200 @@ function dbInsertEntity()
...
 #   --payload     <payload>        (default: NO PAYLOAD. Possible values: [filename | "${string}"])
 #   --in          (input payload)  (default: xml => application/xml, If 'json': application/json)
 #   --out         (output payload  (default: xml => application/xml, If 'json': application/json)
 #   --json        (in/out JSON)    (if --in/out is used AFTER --json, it overrides) 
 #   --httpTenant  <tenant>         (tenant in HTTP header)
+#   --tenant      <tenant>         (tenant in HTTP header)
```
fgalan: Is this the new parameter actually used?

src/lib/rest/rest.cpp
kzangeli: Might be, but that was not my intention.
I bumped into a few errors during the troubleshooting of the new implementation of orionCurl.
The changes in this PR apart from that function itself are to remedy those errors.

test/functionalTest/cases/000_bad_requests/pagination_error_in_uri_params.test
kzangeli: In this functest the idea is to get a complaint about offset 'awe'.
If I leave the 'offset=0', then it complains about offset instead.
I don't know why this has changed, but the easiest way to get it to complain as 'wanted', was to remove limit=0.
Why didn't the broker complain about limit=0 before?  Beats me ...

test/functionalTest/harnessFunctions.sh
kzangeli: Yeah, we normally just add '?a=1&b=2' to the URL.
Both ways can be used ...

test/functionalTest/harnessFunctions.sh
kzangeli: I would like to migrate from --httpTenant to --tenant.
--httpTenant is an old name, from when we had both --httpTenant and --urlTenant.
It's name should have changed when --urlTenant was removed.
Now orionCurl is prepared for this change.
I need to create an issue about this ...

kzangeli: Issue #938

test/functionalTest/harnessFunctions.sh
```diff
@@ -595,22 +625,200 @@ function dbInsertEntity()
...
 #   --servicePath <path>           (Service Path in HTTP header)
-#   --urlParams   <params>         (URI parameters 'in' URL-string)
 #   --xauthToken  <token>          (X-Auth token value)
 #   --origin      <origin>         (Origin in HTTP header)
-#   --verbose                      (verbose output)
 #   --noPayloadCheck               (don't check the payload)
+#   --header      <HTTP header>    (more headers)
+#   -H            <HTTP header>    (more headers)
```
kzangeli: These two options are new: -H and --header

src/lib/rest/rest.cpp
fgalan: If we have fixed something in the code, I think it should mentioned in CHANGES_NEXT_RELEASE.

fgalan: Discussed by skype: let's try to figure out if removing this block of code is fixing something and, in that case, include a .test harness to cover the case and an issue to be added to CHANGES_NEXT_RELEASE.

kzangeli: Commented with FIXME in b507e8f

test/functionalTest/cases/000_bad_requests/pagination_error_in_uri_params.test
fgalan: NTC

test/functionalTest/harnessFunctions.sh
fgalan: Great! More flexibility...

(In fact, `--origin` could have been implemented using `-H`, but let's keep it as it is now).

NTC

test/functionalTest/testHarness.sh
```diff
@@ -41,6 +41,40 @@ cd - > /dev/null 2>&1
...
+  _debug='on'
+fi
+
+
+
+# ------------------------------------------------------------------------------
+#
+# dMsg2 - debug output
+#
+function dMsg2()
```
fgalan: dMsg2? Should use a more descriptive name...

test/functionalTest/harnessFunctions.sh
fgalan: NTC

test/functionalTest/testHarness.sh
kzangeli: Fixed in 5fef94f

fgalan:
LGTM

PR #939, opened 2015-05-28
Feature/reuse curl

crbrox:


src/lib/rest/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ SET (SOURCES
     orionReply.cpp
     OrionError.cpp
     HttpStatusCode.cpp
+    curl_context.cpp
```
fgalan: Not sure if this is the right style... we have also CamelCase (e.g. OrionError.cpp) and camelcase (e.g. orionReply.cpp). A little messy :) maybe we should talk about what we want (or what the style guide says) about this by skype in the common chat.

src/lib/rest/curl_context.cpp
```diff
@@ -0,0 +1,94 @@
+#include "curl_context.h"
```
fgalan: License header should be included.

src/lib/rest/curl_context.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include <map>
+
+#include <assert.h>
+
+#include "logMsg/logMsg.h"
+#include "logMsg/traceLevels.h"
+
+
+static pthread_mutex_t contexts_mutex = PTHREAD_MUTEX_INITIALIZER;
```
fgalan: I think we have a common library to deal with semaphores/mutex. Maybe should move this (and the mutex management in general) to that library. In addition, we have a feature to capture statistics about how many time threads are waiting in each mutex and probably this new mutext should be also included there.

@kzangeli could provide more detail or particular changes suggestions, he has designed and implemented that part.

src/lib/rest/curl_context.cpp
```diff
@@ -0,0 +1,94 @@
...
+static std::map<std::string, struct curl_context> contexts;
+
+struct curl_context
+get_curl_context(const std::string& key)
+{
+  struct curl_context cc = {NULL, NULL};
+  int s = pthread_mutex_lock(&contexts_mutex);
+  if(s!=0)
+  {
+       LM_X(1,("pthread_mutex_lock"));
```
fgalan: Is this the common pattern when a mutex/semaphore fails? What we are doing with the other mutex/semaphores we already have?

src/lib/rest/curl_context.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <assert.h>
+
+#include "logMsg/logMsg.h"
+#include "logMsg/traceLevels.h"
+
+
+static pthread_mutex_t contexts_mutex = PTHREAD_MUTEX_INITIALIZER;
+static std::map<std::string, struct curl_context> contexts;
+
+struct curl_context
```
fgalan: A matter of style... I think we put the return value of the function in the same line of the function name. @kzangeli could you confirm?

src/lib/rest/curl_context.h
```diff
@@ -0,0 +1,25 @@
+#ifndef CURL_CONTEXT_H
```
fgalan: License file missing

src/lib/rest/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ SET (SOURCES
     orionReply.cpp
```
fgalan: An entry in CNR file should be needed for this change.

src/lib/rest/curl_context.cpp
kzangeli: common/sem.cpp/h - part of lib common

src/lib/rest/curl_context.cpp
kzangeli: Yeah, we should probably not exit …
Escpecially not from inside a library.

src/lib/rest/curl_context.cpp
kzangeli: Yes, the return type and function name we keep on the same line, like this:

    XXX* xxx(void)
    {
      ...
    }

Some functions have A LOT of parameters and we then put each parameter on a separate line:

    int xxx
    (
        int i1,
        int i2
    )
    {
      …
    }

Two spaces used as indentation. Tabs are not used. Ever. etc ...

src/lib/rest/CMakeLists.txt
kzangeli: The files that contain one C++ class normally have the same name as the class.
Classes start with uppercase so those files also start with uppercase.
Other files contain collections of stuff (e.g. sem.cpp - semaphore stuff) so that file starts with lowercase.
Messy? Not really. I think it is pretty good actually. I wouldn't change a thing.

src/lib/rest/curl_context.cpp
crbrox: I think many libraries have assert for "Impossible" conditions and many errors are so serious that there is no other reasonable behavior other than  exiting immediately.  I can't think on a sane case where a mutex can't  be initialized or be locked and we should continue happily.

src/lib/rest/CMakeLists.txt
crbrox: And there is no class CurlContext in the file ... a plain struct and separate functions, but I don't think it is important, I'll change it to your style

src/lib/rest/curl_context.cpp
crbrox: I don`t think this struct including a curl handle will fit in a common file for semaphores. Here, nothing more than usual operations over mutex are used, The rest is specific logic for the "curl context" management. If we would want to reuse the "curl-context" in other application/library/... (what I don't think it's crazy), we would have to copy&paste fragments of common/semaphore. That doesn't seem a very modular approach or a good software engineering practice

src/lib/rest/curl_context.cpp
crbrox: a link to the style guide in use?

src/lib/rest/curl_context.cpp
kzangeli: When I said 'probably' I meant 'definitely'.

A library can NEVER call exit.
NEVER EVER.

A library doesn't know who its users are. So, it can't make any decisions for them. And the decision to kill the process ... that is just the most important decision to be made.

One user may be serving perfectly 100 connections when a problematic connection comes in. Because of this 'problematic' connection the entire process dies and the 101 connections stop?
No, a library can return an error, but it must always be the user of the library (the process) that decides whether the process should die or not.

src/lib/rest/curl_context.cpp
kzangeli: We don't follow the style guide to 100%
The easiest way is probably to just look at the code and follow its example.
Unfortunately not all the code is up-to-date, for example part of mongoBackend uses 4 spaces for indentation. Some day I might have time to fix this ... [ And add a make step to assure that the code follows the style guide - the more implementors we are, the more important it gets ]

src/lib/rest/curl_context.cpp
crbrox: I just read now ...
1. I'll do the same you're doing in the other code, "when in Rome, do as the Romans do"
2. I disagree, "never ever" is too 'much'. I think it depends on the kind of error (recoverable or not, a sane action could be run or not, ...)
3. I think the current way some errors are being handled (or not handled to be precise) is completely WRONG and is FAR WORSE than dying when a mutex cannot be initialized or memory cannot be allocated, what I think it's completely sane (however 1. Romans rules)

fgalan: We are open to introduce changes in our "way of working" if that improves the way in which the code in implemented. Even Romans may change their habits :)

However, I'd suggest all the Romans being at Rome (currently some ones are in other continents ;) in order to discuss the best way all together. By the moment, let's implement this in the "current Roman" style.

CHANGES_NEXT_RELEASE
```diff
@@ -3,3 +3,5 @@ Fix: Add multiservice flag for the default configuration in the RPM. (Issue #964
 Add: New operation "GET /v2/entities" (Issue #947)
 Add: New operation "GET /v2" (Issue #956)
 Fix: ONCHANGE subscription sends notification when value under condition it is not updated with updateContext-APPEND in some cases (Issue #943)
+Add: Reuse curl context for outgoing notifications
```
fgalan: I would suggest to be more "functionality oriented", i.e.

"Add: Reuse curl context for outgoing notifications, so now connections are persistent if the HTTP server doesn't close them"

Or something like that...

src/lib/rest/httpRequestSend.cpp
```diff
@@ -363,7 +365,8 @@ std::string httpRequestSend
 
     // Cleanup curl environment
     curl_slist_free_all(headers);
-    curl_easy_cleanup(curl);
+    //curl_easy_cleanup(curl);
```
fgalan: Leftover comment to be removed?

src/lib/rest/httpRequestSend.cpp
```diff
@@ -427,7 +430,8 @@ std::string httpRequestSend
 
   // Cleanup curl environment
   curl_slist_free_all(headers);
-  curl_easy_cleanup(curl);
+  //curl_easy_cleanup(curl);
```
fgalan: Leftover comment to be removed?

src/lib/rest/CMakeLists.txt
fgalan: At the end curl_context.cpp was "integrated" in existing files.

NTC

src/lib/rest/curl_context.cpp
fgalan: At the end curl_context.cpp was "integrated" in existing files.

NTC

src/lib/rest/curl_context.cpp
fgalan: NTC

src/lib/rest/curl_context.cpp
fgalan: NTC (in this PR... but Romans have to meet at Senate to discuss on this xD)

src/lib/rest/curl_context.cpp
fgalan: I understand this is fixed (or will be detected by checky style scripts and fixed after this PR).

src/lib/rest/curl_context.h
fgalan: At the end curl_context.h was "integrated" in existing files.

NTC

CHANGES_NEXT_RELEASE
crbrox: Fixed 6d8116f58c6887fab6dbb519325070217845d607

src/lib/rest/httpRequestSend.cpp
crbrox: Just for easing PR review, fixed 8b343e2077e7136ece137fc980f3babe40488615

src/lib/rest/httpRequestSend.cpp
crbrox: Just for easing PR review, fixed 8b343e2077e7136ece137fc980f3babe40488615

fgalan:
LGTM

Issue #943, opened 2015-05-29
[BUG] ONCHANGE subscription sends notification when value under condition it is not updated


XavierVal:
Reported by @jframos
I have confirm that the notifications are sent when the attribute subscribed has not been modified.

Data to reproduce it: 
Given this entity append (without Fiware/pahts headers):
```
 <?xml version="1.0" encoding="UTF-8"?>
<updateContextRequest>
    <contextElementList>
        <contextElement>
            <entityId type="region" isPattern="false">
                <id>qa_ok</id>
            </entityId>
            <contextAttributeList>
                <contextAttribute>
                    <name>sanity_status</name>
                    <type>string</type>
                    <contextValue>OK</contextValue>
                </contextAttribute>
                <contextAttribute>
                    <name>_timestamp</name>
                    <type>string</type>
                    <contextValue>333</contextValue>
                </contextAttribute>
            </contextAttributeList>
        </contextElement>
    </contextElementList>
    <updateAction>APPEND</updateAction>
</updateContextRequest>
``` 

And this subscription: 
``` 
 {
     "entities": [
         {
             "type": "region",
             "isPattern": "true",
             "id": ".*"
         }
     ],
     "attributes": [
         "sanity_status"
     ],
     "reference": "http://x.x.x.x:1028",
     "duration": "PT10M",
     "notifyConditions": [
         {
             "type": "ONCHANGE",
             "condValues": [
                 "sanity_status"
             ]
         }
     ]
 }
``` 

When the entity updates the value  "sanity_status" it is notified properly. 
But if the  "sanity_status" does not change but the "timestamp" does, CB sends the notification to the subscriber. CB should not send it if the "condValues" has not been updated, isn't it?

kzangeli:
I tried to reproduce this issue but without any luck.
See PR #944

fgalan:
We have been able to reproduce at the end in the branch o PR #944 (with new commits).

XavierVal:
Hunted bug confirmed and PR merged

PR #944, opened 2015-06-01
Bug/943 subscriptions

kzangeli:
Functest that ~~cannot reproduce~~ at the end reproduces the problem reported in issue #943


XavierVal:
LGTM

Issue #947, opened 2015-06-02
API v2: List all entities


fgalan:
Implement list all entities operation in v2, i.e. 

```
GET /v2/entities?params
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

kzangeli:
Just to add one little detail.

We have said that if **type** is empty for an attribute (+ no metadata), then the attribute is rendered as:

    "NAME": "VALUE"

Now, if the attribute if of type "number" (+ no metadata) then this special type is considered "empty" as well, and rendered as follows:

    "NAME": VALUE

The same is valid for metadata.

fgalan:
Raised in https://github.com/telefonicaid/fiware-orion/pull/958/files#r31917180, right processing of Accept in /v2/entities operations is not part of PR #958 and would be addressed in a next PR (maybe developing an overall solution for all /v2/ operations, which are purely based on JSON).

fgalan:
Support to location using `geo:point` is not part of PR #958 and would be addressed in a next PR.

fgalan:
Finally, it is not part of PR #958 solving the ambiguenty problems for compound attributes e.g.:

What I mean is that for example the following rendering for attribute A1:

```
"A1": {
  "value": "M1", 
  "type": "a4", 
  "m2": {
    "type": "tm2", 
    "value": "M2"
   }, 
   "m3": 3,
}
```
Could be interpreted in two different ways:

* As an attribute without metadata and type and which value is the following JSON object:

```
{
  "value": "M1", 
  "type": "a4", 
  "m2": {
    "type": "tm2", 
    "value": "M2"
   }, 
   "m3": 3,
}
```

* As an attribute of type "a4" and value "M1" with 2 metadata (one named "m2" with type "tm2" and value "M2" and the other named "m3" of type "number" and value 3.

fgalan:
After latest discussion, the `count` option (within the `options` parameter) will be used instead of `count=true`. The code needs to be fixed (PR #958 was based on `count=true`).

fgalan:
The basic operation has been implemented in PR #958 and all "pending aspects" originally included here as comments have been moved to independent issues, to provide a cleaner task separation and enable better scheduling. Thus, this issue can be closed.

Issue #953, opened 2015-06-03
API v2: List all entities of a given ID


fgalan:
Implement list all entities of a given id Ioperation in v2, i.e. 

```
GET /v2/entities?id=<entityId>&extraparams
```

See detailed descriptoin in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
The `id` parameter cannot be empty.

fgalan:
Under discussion:

* List of comma separate ids?
* Use this operation to query entities without id? Probable it doesn't make sense... empty id is an anomaly that shouldn't be allowed.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1139

Issue #952, opened 2015-06-03
API v2: List all entities of a given type


fgalan:
Implement list all entities of a given type operation in v2, i.e. 

```
GET /v2/entities?type=<entityType>&extraparams
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
The `type` parameter cannot be empty.

fgalan:
Under discussion:

* List of comma separate types?
* Use this operation to query entities without type

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1139

PR #958, opened 2015-06-07
Feature/947 v2 entities

kzangeli:
Implements #947 

This PR implements most if not all of "GET /v2/entities", including the addition of two new directories/libraries for types and service routines for version 2.

One almost complete functional test is present, one more is no its way (test for local pagination).


CMakeLists.txt
```diff
@@ -158,10 +158,12 @@ if (error EQUAL 0)
   ADD_SUBDIRECTORY(src/lib/orionTypes)
```
fgalan: CHANGES_NEXT_RELEASE entry is missing.

CMakeLists.txt
```diff
@@ -158,10 +158,12 @@ if (error EQUAL 0)
...
   ADD_SUBDIRECTORY(src/lib/ngsi)
   ADD_SUBDIRECTORY(src/lib/serviceRoutines)
+  ADD_SUBDIRECTORY(src/lib/serviceRoutinesV2)
```
fgalan: We were discussing on using lib/serviceRoutines/v2 for v2 service routines and (to keep thing homogenoeus) move existing lib/serviceRoutines to lib/serviceRoutines/v1...

kzangeli: We were.
I believe it is better adding the version to the directory name instead.
This way all 'lib directories' are on the same level. More order ...

fgalan: Thus, lib/serviceRoutines should be renamed to lib/serviceRoutinesV1

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
```diff
@@ -0,0 +1,426 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: "947_v2_entities" could be more descriptive... In fact, I'd suggest to use something like:

"947_v2_list_all_entities"

That, by the way, I think is the  name of the operation in the apiary doc.

kzangeli: I was about to change the name of this functest, but my idea was to remove 'simple' (it is not simple anymore) and add GET:

947_v2_entities/GET_v2_entities.test.

Much easier to understand that it's about a "GET /v2/entities" that if we use some other name for the operation, don't you think?

CMakeLists.txt
kzangeli: Not really necessary, but sure, no probs

kzangeli: I just found a really good reason to not change the name of /lib/serviceRoutines. Let's see what you think ...

The name is used in hundreds of places, when including headers from the directory:

      #include "serviceRoutines/xxx.h"

kzangeli: After some discussion on skype, we've decided to leave the names 'as is'. At least for now.

NTC

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
kzangeli: Changed to **GET_v2_entities.test** in 7bb8611

kzangeli:
New functest (pagination for GET /v2/entities) in commit 7bb8611

CHANGES_NEXT_RELEASE
```diff
@@ -1 +1,2 @@
 Fix: Fixed a bug about pagination problem due to missing ErrorCode with details in the case of queries with forwarding to context providers.  (Issue #945)
+Add: New convop "GET /v2/entities" (Issue #947)
```
jmcanterafonseca: It was my understanding that in v2 we are no longer using the term convop, aren't we?

fgalan: Good point... I'd suggest to be neutral and just say:

```
Add: New operation "GET /v2/entities" (Issue #947)
```

src/app/contextBroker/contextBroker.cpp
```diff
@@ -325,6 +328,13 @@ PaArgument paArgs[] =
 
 
 //
+// /v2 conv-ops 
+//
+#define ENT                EntitiesRequest
+#define ENT_COMPS_V2       2, { "v2", "entities" }
```
jmcanterafonseca: probably we can define a constant for 'entities' and use it as well in the entryPoints operation as well, WDYT?

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
fgalan: Directory names are per issue number. Take into accoun that several issues may refer to the same operatoin (from a URL path point of view). In fact, the following issues number share the same /v2/entities path: #947, #952, #953 and (not yet filled) the query entities operation and the tests corresponding to each one should be in their separated directory.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -899,6 +915,8 @@ PaArgument paArgs[] =
 */
 RestService restServiceV[] =
 {
+  CONVENIENCE_OPERATIONS_V2,
```
jmcanterafonseca: drop 'CONVENIENCE' ?

src/app/contextBroker/contextBroker.cpp
```diff
@@ -325,6 +328,13 @@ PaArgument paArgs[] =
 
 
 //
+// /v2 conv-ops
```
fgalan: Taking into account @jmcanterafonseca comment, maybe it is better just `/v2 operations`.

CHANGES_NEXT_RELEASE
kzangeli: ok, I wasn't aware of that ...

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
```diff
@@ -0,0 +1,426 @@
...
+echo
+
+
+--REGEXPECT--
+01. GET /v2/entities
+====================
+HTTP/1.1 404 Not Found
+Content-Length: 0
+Date: REGEX(.*)
+
```
jmcanterafonseca: If there are no entities in the system, then a 200 OK and empty array should be returned

src/lib/apiTypesV2/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
+
+SET (SOURCES
+    Entity.cpp
+    EntityVector.cpp
+    Entities.cpp
+)
+
+SET (HEADERS
```
fgalan: Looking to the new 3 classes, Entity seems to be very similar to existing EntityId and EntityVector very similar to existing EntityIdVector. I understand that we need a class for the "outest" data structure for the operation (and that class I undertand is Entities), but why don't base its internal content in already existing classes instead of introducing new classes that are basically identical?

src/app/contextBroker/contextBroker.cpp
kzangeli: API_V2 ?

src/app/contextBroker/contextBroker.cpp
kzangeli: /v2 operations 

or

  /v2 API

?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1074,7 +1073,7 @@ bool entitiesQuery
     connection = getMongoConnection();
     try
     {
-        if ((details == true) && (countP != NULL))
+        if (countP != NULL)
```
fgalan: This part should be executed only when the client has requested details... removing the `details` checking the count is got always, which is inefficient.

src/app/contextBroker/contextBroker.cpp
kzangeli: What do we win with a change like that?
[ Except that the code gets more difficult to read ... ]

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
kzangeli: ok, mea culpa.

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -245,7 +252,8 @@ HttpStatusCode mongoQueryContext
   QueryContextResponse*                responseP,
   const std::string&                   tenant,
   const std::vector<std::string>&      servicePathV,
-  std::map<std::string, std::string>&  uriParams
+  std::map<std::string, std::string>&  uriParams,
+  long long*                           countP
```
fgalan: This `countP` is used to return back the count number in the case of `count=true` (for v2) or `details=on` (for v1) ?

src/lib/serviceRoutinesV2/getEntities.cpp
```diff
@@ -0,0 +1,93 @@
...
+  std::vector<std::string>&  compV,
+  ParseData*                 parseDataP
+)
+{
+  std::string  answer;
+  Entities     entities;
+
+
+  // 01. Fill in QueryContextRequest
+  parseDataP->qcr.res.fill(".*", "", "true", EntityTypeEmptyOrNotEmpty, "");
```
jmcanterafonseca: wouldn't it be better to use the boolean value true instead of "true" as string?

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -351,6 +351,108 @@ std::string ContextAttribute::render
...
+
+    if (metadataVector.size() > 0)
+    {
+      out += "," + metadataVector.renderV2(ciP, request, false);
+    }
+
+    out += "}";
+  }
+
+  if (comma)
```
fgalan: I understand that this comma is needed when attributes are printed in a vector and the attribute is not the last one. Maybe we should use a name for the parameter closer to that meaning... e.g. "lastAttributeInVector".

src/lib/rest/rest.cpp
```diff
@@ -719,6 +744,27 @@ std::string defaultServicePath(const char* url, const char* method)
...
+* apiVersionGet - 
+*
+* This function returns the version of the API for the incoming message,
+* based on the URL.
+* If the URL starts with "/v2" then the request is considered API version 2.
+* Otherwise, API version 1.
+*/
+static std::string apiVersionGet(const char* path)
+{
+  if ((path[1] == 'v') && (path[2] == '2'))
```
jmcanterafonseca: wouldn't it be better to test for the exact match of "v2" ? I cannot imagine v21 being different but ... who knows ...

src/lib/ngsi/ContextAttributeVector.cpp
```diff
@@ -71,6 +71,74 @@ static std::string addedLookup(const std::vector<std::string>& added, std::strin
 
 /* ****************************************************************************
 *
+* ContextAttributeVector::renderV2 - 
+*
+* Attributes named 'id' or 'type' are not rendered in API version 2, due to the 
+* compact way in which API v2 is rendered. Attributes named 'id' or 'type' would simply
+* collide with the 'id' and 'type' of the entity itself (holder of the attribute).
+*
+* If anybody (stupid enough) needs an attribute named 'id' or 'type', then API v1
```
fgalan: I think is better (and will survive the API v1 removal :) to say: "... then standard operation API will have to be used to retreive that information".

src/lib/rest/rest.cpp
```diff
@@ -616,6 +630,17 @@ static int contentTypeCheck(ConnectionInfo* ciP)
     return 1;
   }
 
+
+  // Case 5
+  if ((ciP->apiVersion == "v2") && (ciP->httpHeaders.contentType != "application/json"))
+  {
```
jmcanterafonseca: maybe we can define a global constant called "v2" ... same for "application/json" ... i do not have a strong opinion on that

src/lib/parse/CompoundValueNode.h
```diff
@@ -127,6 +126,7 @@ class CompoundValueNode
   void                check(void);
   std::string         finish(void);
   std::string         render(Format format, const std::string& indent);
+  std::string         renderV2(void);
```
jmcanterafonseca: I don't if we could call this renderSimple instead of renderV2 ... don't know if it is a good idea to create methods that refer to an specific version rather than to a functionality

src/lib/ngsi/Metadata.cpp
```diff
@@ -174,3 +174,46 @@ void Metadata::fill(const struct Metadata& md)
...
   association = md.association;
 }
+
+
+
+/* ****************************************************************************
+*
+* renderV2 - 
+*/
+std::string Metadata::renderV2(ConnectionInfo* ciP, RequestType request, bool comma)
```
fgalan: Are ciP and request actually used? In negative case, I'd suggest to remove them, in order to keep the function simpler. Probably ciP and request are a leftover of v1 way of working (the XML compataibility probably need them) and we can get rid of it for v2.

src/lib/ngsi/StatusCode.cpp
```diff
@@ -124,6 +124,51 @@ std::string StatusCode::render(Format format, const std::string& indent, bool co
 
 /* ****************************************************************************
 *
+* StatusCode::renderV2 -
+*
+* For version 2 of the API, the unnecessary 'reasonPhrase' is removed.
+*/
+std::string StatusCode::renderV2(bool comma)
```
jmcanterafonseca: same comment above V2

src/lib/ngsi/Metadata.cpp
```diff
@@ -174,3 +174,46 @@ void Metadata::fill(const struct Metadata& md)
...
+  }
+  else
+  {
+    out = "\"" + name +"\":{";
+    out += "\"value\":\"" + value + "\"";
+    out += ",\"type\":\"" + type +"\"";
+    out += "}";
+  }
+
+  if (comma)
```
fgalan: Same comment about using a name better than `comma`... Maybe "lastElement".

src/lib/ngsi/Request.h
```diff
@@ -101,7 +101,12 @@ typedef enum RequestType
   ContextEntitiesByEntityIdAndType,
   EntityByIdAttributeByNameIdAndType,
 
-  InvalidRequest
+  // Version 2 ConvOps: /v2/*
```
jmcanterafonseca: ConvOps ...

src/lib/ngsi/MetadataVector.cpp
```diff
@@ -87,6 +87,75 @@ std::string MetadataVector::render(Format format, const std::string& indent, boo
...
+  // This gives us a small problem in the logic here, about knowing whether the
+  // comma should be rendered or not.
+  //
+  // To fix this problem we need to do two passes over the vector, the first pass to
+  // count the number of valid metadatas and the second to do the work.
+  // In the second pass, if the number of rendered metadatas "so far" is less than the total
+  // number of valid metadatas, then the comma must be rendered.
+  //
+  int validMetadatas = 0;
+  for (unsigned int ix = 0; ix < vec.size(); ++ix)
```
fgalan: It's a pitty having to do a double-pass for the elements, but I guess that it isn't a better option (I cannot imagine another one, at least).

src/lib/ngsi/ContextAttributeVector.cpp
fgalan: In addition "stupid enough" is a bit unpolite ;)

src/lib/ngsi/MetadataVector.cpp
```diff
@@ -87,6 +87,75 @@ std::string MetadataVector::render(Format format, const std::string& indent, boo
 
 /* ****************************************************************************
 *
+* MetadataVector::renderV2 -
+*
+* Metadatas named 'value' or 'type' are not rendered in API version 2, due to the 
+* compact way in which API v2 is rendered. Metadatas named 'value' or 'type' would simply
+* collide with the 'value' and 'type' of the attribute itself (holder of the metadata).
+*
+* If anybody (stupid enough) needs a metadata named 'value' or 'type', then API v1
```
fgalan: "stupid enough" is a bit unpolite ;)

src/lib/apiTypesV2/CMakeLists.txt
kzangeli: First of all, they aren't exactly identical.
Separating them into different classes makes the code easier to read and maintain (talking about EntityId vs Entity).
We have a number of vectors in the system and they are pretty much identical, in that I agree 100%.
A nice refactor would be to replace ALL **Vector with ONE class (or at least make them inherit from a base-vector-class.

However, that's not really part of this PR ...

src/lib/ngsi/StatusCode.cpp
fgalan: This render is used when HTTP response is different from 200 in order to generate the payload, right?

src/lib/ngsi/StatusCode.cpp
```diff
@@ -124,6 +124,51 @@ std::string StatusCode::render(Format format, const std::string& indent, bool co
...
+{
+  std::string  out  = "";
+
+  if (strstr(details.c_str(), "\"") != NULL)
+  {
+    int    len = details.length() * 2;
+    char*  s2    = (char*) calloc(1, len + 1);
```
fgalan: I understand that we are using `s2` here because `s` is already taken... However, I don't find any `s`...

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: It is only done if countP != NULL.
countP is NULL unless '?count=true' is set.
It works just like before.

NTC

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: Yes.
For v2 I need that number to be able to add the HTTP header X-Total-Count.

NTC

src/lib/ngsi/StatusCode.cpp
```diff
@@ -124,6 +124,51 @@ std::string StatusCode::render(Format format, const std::string& indent, bool co
...
+    free(s2);
+  }
+
+  char codeV[16];
+
+  snprintf(codeV, sizeof(codeV), "%d", code);
+
+  out += "{";
+
+  out += std::string("\"code\":\"") + codeV + "\"";
```
fgalan: According to https://github.com/telefonicaid/fiware-orion/issues/947, the payload is `{ "error: "...", "details": "..." }`. If we are mapping an StatusCoce into that structure, I guess that reasonPhrase would be used to fill the error field.

src/lib/ngsi/ContextAttribute.cpp
kzangeli: That is true if the attribute is a part of a vector.
But, the attribute can be *not inside* of a vector also, with some other item after it, thus a comma is needed.

src/lib/rest/rest.cpp
```diff
@@ -616,6 +630,17 @@ static int contentTypeCheck(ConnectionInfo* ciP)
     return 1;
   }
 
+
+  // Case 5
+  if ((ciP->apiVersion == "v2") && (ciP->httpHeaders.contentType != "application/json"))
```
fgalan: Being stricts, this is not part of #947 (as the operation described in that issue number is a GET without payload). I understand that you are preparing the code in advance for the operations that are to come and that use payload. Is that correct?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -297,13 +297,49 @@ std::string postQueryContext
...
+
+
+  //
+  // 00. Count or not count? That is the question ...
+  //
+  // For API version 1, if the URI parameter 'details' is set to 'on', then the total of local
+  // entities is returned in the errorCode of the payload.
+  //
+  // In API version 2, this has changed completely. Here, the total count of local entities is returned
+  // if the URI parameter 'count' is set to 'on', and it is returned in the HTTP header X-Total-Count.
```
fgalan: Actually, is "true" and not "on" in this case (see https://github.com/telefonicaid/fiware-orion/issues/947)

CMakeLists.txt
kzangeli: Fixed in 7bb8611

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -297,13 +297,49 @@ std::string postQueryContext
...
+  //
+  if ((ciP->apiVersion == "v2") && (ciP->uriParam["count"] == "true"))
```
fgalan: Probably I'm missing something... but it seems that the result of both if is the same `countP = &count`. Maybe they could be joined with an OR (`||`) ?

src/lib/serviceRoutinesV2/getEntities.cpp
kzangeli: You will see this in more than one place ...
It was implemented like that initially but as the payload comes in as ASCII it is a lot easier to work this way. The code got full of 'string to bool conversions' and got difficult to read.

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -297,13 +297,49 @@ std::string postQueryContext
...
+  //
+  if ((ciP->apiVersion == "v2") && (countP != NULL))
+  {
+    char cV[32];
+
+    snprintf(cV, sizeof(cV), "%llu", *countP);
+    ciP->httpHeader.push_back("X-Total-Count");
+    ciP->httpHeaderValue.push_back(cV);
+  }
+
```
fgalan: How do we avoid in the case of v2 that the erroCode with `Count: X` gets generated and included in the response?

src/lib/apiTypesV2/Entities.cpp
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  for (unsigned int ix = 0; ix < qcrsP->contextElementResponseVector.size(); ++ix)
+  {
+    ContextElement* ceP = &qcrsP->contextElementResponseVector[ix]->contextElement;
+    Entity*         eP  = new Entity();
+
+    eP->id        = ceP->entityId.id;
+    eP->type      = ceP->entityId.type;
+    eP->isPattern = ceP->entityId.isPattern;
```
jmcanterafonseca: do we need isPattern, as we are not going to render it in the Simple API?

src/lib/rest/rest.cpp
kzangeli: This way of doing it is *many* times faster than to call strcmp.
It is true that when we get up to version 20 one more check is needed (path[3] == '/').
I don't think we will ever get there ...

src/lib/rest/rest.cpp
kzangeli: As I commented before, it only makes the code more difficult to read (Fermin agrees with you on this however :-)).

src/app/contextBroker/contextBroker.cpp
jmcanterafonseca: SIMPLE_API?

src/lib/ngsi/Metadata.cpp
kzangeli: Actually, ciP is there to make all render() have the same parameters.
This exact function doesn't use it but other renders do.
My idea is to someday have all render methods using the exact same set of params (these three).
Now, if you have an opinion on that idea, then let's talk.

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
```diff
@@ -0,0 +1,426 @@
...
+            }, 
+            "m3": 3, 
+            "type": "a4", 
+            "value": "E1/T1/A4/M1-3"
+        }, 
+        "id": "E1", 
+        "type": "T1"
+    }, 
+    {
+        "A0": {
```
fgalan: What happen if one of the fields of the compound is named `value`, `type` or `metadata` ?

src/lib/ngsi/StatusCode.cpp
jmcanterafonseca: I refined a bit this in the APIary doc, see for instance

http://docs.fiwarengsiv2.apiary.io/#reference/create-a-new-entity.-error

src/lib/ngsi/StatusCode.cpp
kzangeli: It's simply the render used for version 2.
Version 1 supports XML and JSON and adds beautifying stuff.
Version 2 is pure JSON and no beautifying.
The higher level render method will choose whether to call it or not, depending on **code**.

src/lib/ngsi/MetadataVector.cpp
kzangeli: Yeah ... JSON and the commas ...
I can't find another way of doing it, now that 'name', 'id', 'value' etc make the count useless

src/lib/ngsi/StatusCode.cpp
kzangeli: We haven't discussed this, but 'reasonPhrase' is a little useless. It adds no information, only payload.
v2 should not use 'reasonPhrase' IMO.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: of course ...

src/lib/rest/rest.cpp
kzangeli: Yes, just a thing I found and I decided to fix it.

NTC

CMakeLists.txt
fgalan: Any change related with this has been deferred. An issue has been created for this: https://github.com/telefonicaid/fiware-orion/issues/959

src/app/contextBroker/contextBroker.cpp
fgalan: OPERATIONS_V2 ?

src/app/contextBroker/contextBroker.cpp
fgalan: Both looks fine :)

src/lib/ngsi/MetadataVector.cpp
jmcanterafonseca: oh absolutely and this is for a case that in the future is not going to happen ...

src/lib/ngsi/ContextAttribute.cpp
fgalan: What about `printComma` ?

src/lib/ngsi/Metadata.cpp
fgalan: My opinion is based on the understanding that ciP and RequestType is needed when you need to know tag names and this kind of stuff in the case of XML. But, in the case of a pure JSON render, are they actually needed? If they are needed, they should be keep; if they are not needed, they should be removed.

src/lib/parse/CompoundValueNode.h
kzangeli: What if we call it 'toJson()' ?
Later on, if changes are needed from v2 to v3, a parameter can be added (version).

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Of course, but that would lead to a VERY log line that would be more difficult to understand.
Let's not program that way ...

src/lib/ngsi/Metadata.cpp
fgalan: Or `printComma`.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Actually it isn't included.
And if in some case in the future it is, then we'll deal with it.

src/lib/ngsi/Metadata.cpp
kzangeli: It's not printed.
Before I used 'commaNeeded' but I decided to shorten it to just 'comma'.
To me it is clear enough ...

'lastElement' is much better that 'printComma'.

src/lib/parse/CompoundValueNode.h
fgalan: In the future there would be only one "family"... defining such renderV2() is just a transitory situation (although the transition may long months :;). See https://github.com/telefonicaid/fiware-orion/issues/959

src/lib/ngsi/StatusCode.cpp
kzangeli: Can't say I remember exactly why I chose that name.
s2 is *twice* as long as the original string, that is probably the reason.
This is a copy from REALLY old stuff.

src/lib/ngsi/StatusCode.cpp
fgalan: reasonPhrase (if used to fill the `error` field) wil add information for humans reading the message.

Let's try to solve in the opossite way... if reasonPhrase is not use to fill `error`, how `error` will be filled?

src/lib/ngsi/StatusCode.cpp
fgalan: In which point (of the v2 logic) is this render being used? I have tried to figure out looking to the .test, but I haven't been able...

src/lib/ngsi/StatusCode.cpp
fgalan: Looking to the structure proposed by @jmcanterafonseca :

```
{
  "code": 1234,
  "message": "Already exists",
  "description": "An entity with the same id and type is already present"
}
```

the semantics seems to be close to the ones used in NGSI, so, why don't use the same names than in StatusCode, i.e.:

```
{
  "code": 1234,
  "reason": "Already exists",
  "details": "An entity with the same id and type is already present"
}
```
That way, existing user would find "closer" the new payload.

fgalan: Looking to the structure proposed by @jmcanterafonseca :

```
{
  "code": 1234,
  "message": "Already exists",
  "description": "An entity with the same id and type is already present"
}
```

the semantics seems to be close to the ones used in NGSI, so, why don't use the same names than in StatusCode, i.e.:

```
{
  "code": 1234,
  "reason": "Already exists",
  "details": "An entity with the same id and type is already present"
}
```
That way, existing user would find "closer" the new payload.

src/lib/parse/CompoundValueNode.h
kzangeli: So, what do we do here?
I thing toJson is a better name than 'render' :-)

fgalan: I like toJson :)

src/app/contextBroker/contextBroker.cpp
kzangeli: I vote for **/v2 API**  (here and in the name of the definition).

src/lib/rest/rest.cpp
fgalan: I'm a fan of using `#define` for all the constant tokens in the token :) However, in this particular case, using APPLICATION_JSON instead of "application/json" could be a bit confusing... I don't have a strong opinion on any side.

src/lib/rest/rest.cpp
fgalan: As long as "GET /v20/entities" returns an error (which I think is not going to happend, taking into account where this checking is) I think is fine doing as it is now.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Another argument in favour of having them separated is that in the future it would be easier to remove the ah hoc v1 logic.

Thus, NTC.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: I will have a look to the part of mongoBacken in which this errorCode is generated, in order to see if this is actually a problem or not.

src/lib/ngsi/StatusCode.cpp
kzangeli: If we put in something in 'reason' that actually adds info, ok.
[ Only translating 200 to OK and 404 to 'not found' doesn't make much sense. ]
However, we want to minimize payload ...
Do we really need 'reason' ?
Is it not enough with 'details' ?

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
kzangeli: Each issue will have it's own directory, of course.
We've been working like that for over a year ...
But that's not really the point here, we're talking about the name of the file, not the directory.
Perhaps we should include the issue number in the name of the files as well?

Right now the test file is named GET_v2_entities.test.

Just let me know if it needs a change.

Detail:

    % testHarness.sh GET_v2_entities.test

Here the directory name is not taken into account. If we have more than one test file named GET_v2_entities.test then both/all will be executed.
If we include the issue number in the test-file, this will never happen

CHANGES_NEXT_RELEASE
kzangeli: Fixed in bb66c46

src/app/contextBroker/contextBroker.cpp
kzangeli: API_V2 in bb66c46

Let's decide a name ... :-D

src/app/contextBroker/contextBroker.cpp
kzangeli: /v2 API in bb66c46

Let's decide ...

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
kzangeli: Fixed in bb66c46

src/lib/rest/rest.cpp
kzangeli: GET /v20/entities gives an error, of course.  The path doesn't match any path of the vector in contextBroker.cpp

src/lib/ngsi/ContextAttributeVector.cpp
kzangeli: Fixed in bb66c46

src/lib/rest/rest.cpp
kzangeli: So, let's not make the code more complicated than it is already ... :-)

src/lib/parse/CompoundValueNode.h
kzangeli: So, I change all 'renderV2' to 'toJson' and the 'render' method might disappear ...

src/lib/ngsi/Metadata.cpp
kzangeli: ok, let's do the following:
I will try to reimplement this 'renderV2', using the method name toJson and with one single parameter (bool isLastElement)   (for the 'comma problem').

src/lib/ngsi/StatusCode.cpp
kzangeli: kz@dellie:fiware-orion>  find . -name "*.cpp" -exec grep "renderV2" {} /dev/null
    ./src/lib/apiTypesV2/Entities.cpp:  return errorCode.renderV2();

src/lib/ngsi/Metadata.cpp
kzangeli: So:

    std::string toJson(bool isLastElement)

ok?

src/lib/ngsi/MetadataVector.cpp
kzangeli: I don't get this.
Will not happen in the future because ... ?

src/lib/ngsi/MetadataVector.cpp
kzangeli: Fixed in bb66c46

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in bb66c46

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Remember, one thing is if mongoBackend adds the text in details, another thing is whether the service routine actually decides to use that information.

This is not 100% clear, it might give a **minor problem** in the future.

I say, let's deal with the problem when it's visible.

src/lib/apiTypesV2/Entities.cpp
kzangeli: The idea is for this method to be generic.
In the case of rendering a GET, we don't want isPattern, but in the case of filling the Entity after parsing payload from a POST, isPattern might have to be there.

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
kzangeli: I responded to this ...

There is no special treatments for any of the content of the compound.
Do you think something like that might be necessary?
The compound value of an attribute will always be one level done, it cannot ever collide with anything ...  Right?

src/lib/ngsi/StatusCode.cpp
kzangeli: 's2' changed to 's' in 5c5e28d

src/lib/ngsi/Request.h
kzangeli: Fixed in 5c5e28d

src/lib/apiTypesV2/CMakeLists.txt
kzangeli: Issue #960

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: Just to clarify, **countP** is both input and output.
If not NULL, then the count is asked for (mongo operation) and returned in *countP.

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
fgalan: Probably I haven't expressed correctly... I was referring all the time to the directory name. It is ok to keep the .test file name as it is now (GET_v2_entities.test).

I men, what I'm proposing to change is:

cases/947_v2_entities/

to 

cases/947_v2_list_all_entities/

src/app/contextBroker/contextBroker.cpp
fgalan: For my API_V2 is fine. If @jmcanterafonseca agress, then NTC.

src/app/contextBroker/contextBroker.cpp
fgalan: For my /v2 is fine. If @jmcanterafonseca agress, then NTC.

src/lib/apiTypesV2/CMakeLists.txt
fgalan: Ok, EntityVector is covered by #960. Regarding, Entity I understand that the main difference comparing with EntityId is that the former includes a ContextAttribute vector while the later doesn't include it.

src/lib/parse/CompoundValueNode.h
fgalan: For me is ok. I mean, to use "toJson" in the "new renders" and let old render disappear along with all the v1 stuff when time comes.

src/lib/ngsi/Metadata.cpp
fgalan: Great!

src/lib/ngsi/StatusCode.cpp
fgalan: Maybe we should have a .test exercising it?

src/lib/ngsi/Metadata.cpp
fgalan: Agreed!

src/lib/rest/rest.cpp
fgalan: What about the opossite? I mean, checking that the Accept includes "application/json" or "*/*" otherwise return a "Media not supported" (or similar... I don't remember exactly right now). I mean, from https://github.com/telefonicaid/fiware-orion/issues/947

> This operation only works with Accept "application/json". Attems of using other Accepts (inclusing application/xml) would result in an error.

By the way, I understand that if the request doesn't include an Accept header, the default Accept is "*/*" (from http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html):

> If no Accept header field is present, then it is assumed that the client accepts all media types

src/lib/apiTypesV2/CMakeLists.txt
kzangeli: Yes, that is the main difference.
But, also, as it is used solely for V2 we can use toJson directly (without checking version number). The code gets cleaner and easier to understand/maintain.

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
fgalan: What I mean is that for example the following rendering for attribute A1:

```
"A1": {
  "value": "M1", 
  "type": "a4", 
  "m2": {
    "type": "tm2", 
    "value": "M2"
   }, 
   "m3": 3,
}
```
Could be interpreted in two different ways:

1. As an attribute without metadata and type and which value is the following JSON object:

{
  "value": "M1", 
  "type": "a4", 
  "m2": {
    "type": "tm2", 
    "value": "M2"
   }, 
   "m3": 3,
}

2. As an attribute of type "a4" and value "M1" with 2 metadata (one named "m2" with type "tm2" and value "M2" and the other named "m3" of type "number" and value 3.

src/lib/apiTypesV2/CMakeLists.txt
fgalan: Ok. NTC.

src/lib/rest/rest.cpp
kzangeli: True. We should do something there also, about accepted output.
In rest.cpp there is a function (outFormatCheck) doing just this, but it accepts JSON and XML.
If the request is v2 then we should only accept JSON ...

Tricky ...
XML is still the default format.
If we change that ... things may stop working for some people ...

However, this PR is big enough already; let's not add to it ...
Create as issue, pls

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Looking to it in detail, it seems this fragment in Entities::fill() is doing the magic:

```
  if (qcrsP->errorCode.code == SccContextElementNotFound)
  {
    errorCode.fill(SccOk);
    return;
  }
  else if (qcrsP->errorCode.code != SccOk)
  {
    errorCode.fill(qcrsP->errorCode);
    return;
}
```
It happens that the errorCode that mongoBackend generated to "transport" the "Count: XX" is a 200 OK, thus the above code causes it to be removed.

Probably it is worth a comment in this part of the code explaining that. It could be complex to understand in the future...

test/functionalTest/cases/947_v2_entities/GET_v2_entities.test
kzangeli: OK, I understand.
Not sure we need to worry about this.
We **could** add a special field saying that the value is a compound and then let the users treat that info the way they see fit:

    {
      "isCompound": "true",
      "value": "M1", 
      "type": "a4", 
      "m2": {
        "type": "tm2", 
        "value": "M2"
       }, 
       "m3": 3,
    }

We could also change 'value' for '_value' etc but ... not sure I like that approach

src/lib/apiTypesV2/Entities.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+/* ****************************************************************************
+*
+* Entities::fill - 
+*/
+void Entities::fill(QueryContextResponse* qcrsP)
+{
+  if (qcrsP->errorCode.code == SccContextElementNotFound)
+  {
+    errorCode.fill(SccOk);
```
fgalan: I undertand that this "transformation" from SccContextElementNotFound to SccOk is related with the GET /v2/entities operation returns an empty arrayt `[]` in this case and not an error. I'd suggest to add a comment about it... it could be complex to understand in the future.

src/lib/rest/rest.cpp
fgalan: Actually, we don't need to create an issue: is part of the #947 (I understand that the present PR doesn't implement completely #947 so we continue the work in next PRs).

I'd add a comment in #947 about it. By the moment, deffered. NTC.

fgalan: > XML is still the default format.
> If we change that ... things may stop working for some people ...

What about make the default depend on the API version. I mean, in the case of /v2 the default is JSON. In the case of not-v2, the default is XML to mantain backward compatibility.

kzangeli: I guess we could do that.
If the path of the url starts with '/v2/', then default output format is json.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: sure

test/functionalTest/cases/947_v2_entities/v2_entities_simple.test
kzangeli: I changed the name of the directory yesterday.
Commit: d579ab

src/lib/serviceRoutinesV2/getEntities.cpp
kzangeli: NTC

src/lib/ngsi/ContextAttribute.cpp
kzangeli: isLastElement is used finally.
Commit: https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/lib/parse/CompoundValueNode.h
kzangeli: renderV2 => toJson in https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/lib/ngsi/Metadata.cpp
kzangeli: Fixed in https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/lib/ngsi/StatusCode.cpp
kzangeli: Propose a way of provoking an error down there and let's do it ... :-)
Remember 404 Not Found is not an option - it is 200 OK and empty vector now

src/lib/ngsi/Metadata.cpp
kzangeli: Fixed in https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/lib/apiTypesV2/Entities.cpp
kzangeli: NTC

src/lib/apiTypesV2/Entities.cpp
kzangeli: Fixed in https://github.com/telefonicaid/fiware-orion/commit/0bc1f0199f2bc6055a2c24ae3ab24a71a422674c

src/app/contextBroker/contextBroker.cpp
kzangeli: NTC

src/lib/rest/rest.cpp
fgalan: Nobody have a strong opinion on this :) so NTC

src/lib/ngsi/StatusCode.cpp
fgalan: Forbidden characters? Media Not Supported? ...

src/lib/ngsi/MetadataVector.cpp
fgalan: I understand that NTC

src/lib/rest/rest.cpp
kzangeli: Oh, I have a **strong** opinion on this :-D

KISS

src/lib/apiTypesV2/Entities.cpp
```diff
@@ -0,0 +1,150 @@
...
+    // and an empty vector of entities ( [] )
+    //
+
+    errorCode.fill(SccOk);
+    return;
+  }
+  else if (qcrsP->errorCode.code != SccOk)
+  {
+    //
+    // If any other error - use the error for the response
```
fgalan: I would add also something like this:

"Taking into account that this way of working avoids that the errorCode generated by API v1 logic for count be included in the response to the user of the API v2 operation"

kzangeli: I don't know where exactly this piece of comment fits in ...
The phrase is true, but ... where do I put it?

kzangeli: Fixed in d24d235

src/lib/ngsi/StatusCode.cpp
kzangeli: My guess is that these two errors (as most others) are captured before we get too far.
I think we would need an errorCode::code != 404 (and 200) returned from mongo backend.

fgalan: Yes, it seems dificult. Let's skip by the moment (I guess that would "reapear" when we implement operations that can generate errors at mongoBackend).

NTC

src/lib/apiTypesV2/Entities.cpp
jmcanterafonseca: NTC

src/lib/ngsi/MetadataVector.cpp
jmcanterafonseca: type and id will not be used as attributes by entity creators ... won't they?

src/lib/rest/rest.cpp
jmcanterafonseca: NTC on my side, I'm happy either way

src/lib/rest/rest.cpp
jmcanterafonseca: NTC

jmcanterafonseca:
congrats @kzangeli good work!

src/lib/ngsi/MetadataVector.cpp
kzangeli: Let's hope not.
And if the use those names for attributes, we'll just have to ask them not to ...

Perhaps we should open an issue about this. Perhaps we should issue an error if anybody tries to create an attribute with those names ... ?
Argument against prohibition: with /v1 they can actually use those attribute names.
However, we'd like to encourage people to leave /v1 and user /v2 ...

Not for this PR, but we might need to discuss this matter further.

fgalan:
LGTM

PR #961, opened 2015-06-08
ADD a set of macros for JSON rendering

jmcanterafonseca:
A few macros for JSON rendering

### DIFF
--- a/src/lib/common/tag.h
+++ b/src/lib/common/tag.h
@@ -30,6 +30,18 @@
 #include "common/Format.h"
 
 
+// Macros to be used for JSON rendering
+
+#define JSON_PROP(name)                std::string("\"" + std::string(name) + "\":")
+
+#define JSON_STR(value)                std::string("\"" + std::string(value) + "\"")
+
+#define JSON_NUMBER(value)             std::string(value)
+
+#define JSON_VALUE(name, value)        std::string(JSON_PROP(name) + JSON_STR(value))
+
+#define JSON_VALUE_NUMBER(name, value) std::string(JSON_PROP(name) + JSON_NUMBER(value))
+
 
 /* ****************************************************************************
 *


src/lib/common/tag.h
```diff
@@ -30,6 +30,22 @@
...
+
+#define JSON_PROP(name) std::string("\"" + std::string(name) + "\":")
+
+#define JSON_STR_LAST(value) std::string("\"" + std::string(value) + "\"")
+#define JSON_STR(value) std::string(JSON_STR_LAST(value) + ",")
+
+#define JSON_NUMBER_LAST(value) std::string(value)
+#define JSON_NUMBER(value) std::string(JSON_NUMBER_LAST(value) + ",")
+
+#define JSON_VALUE_LAST(name,value) std::string(JSON_PROP(name) + JSON_STR_LAST(value))
```
fgalan: I think that our style needs a whitespace just after ",", I mean:

```
JSON_VALUE_LAST(name, value)
```

src/lib/common/tag.h
```diff
@@ -30,6 +30,22 @@
 #include "common/Format.h"
 
 
+// Macros to be used for JSON rendering
+
+#define JSON_PROP(name) std::string("\"" + std::string(name) + "\":")
```
kzangeli: It looks very funny, not closing the first std::string ...
This should work as well;

    #define JSON_PROP(name) std::string("\"") + name + "\":"

src/lib/common/tag.h
kzangeli: Yes, that is correct

src/lib/common/tag.h
```diff
@@ -30,6 +30,22 @@
...
+
+#define JSON_NUMBER_LAST(value) std::string(value)
+#define JSON_NUMBER(value) std::string(JSON_NUMBER_LAST(value) + ",")
+
+#define JSON_VALUE_LAST(name,value) std::string(JSON_PROP(name) + JSON_STR_LAST(value))
+#define JSON_VALUE(name,value) std::string(JSON_VALUE_LAST(name,value) + ",")
+
+#define JSON_VALUE_NUMBER_LAST(name,value) std::string(JSON_PROP(name) + JSON_NUMBER_LAST(value))
+#define JSON_VALUE_NUMBER(name,value) std::string(JSON_VALUE_NUMBER_LAST(name,value) + ",")
+
```
kzangeli: I would not encourage to use the macros with the comma incorporated.
It only adds complexity ... (IMHO)

src/lib/common/tag.h
jmcanterafonseca: but then we may have problems concatenating the macro result with other strings. I have suffered it :(

kzangeli: Of course. The '+' operator of std::string accepts char* as input (among other types).
That's why I started with std::string in my example (I have suffered it too ... believe me).

I wonder how 

    "\"" + std::string(name)

works ...

This is 'char*' + 'std::string' ... Did you compile this?

src/lib/common/tag.h
```diff
@@ -30,6 +30,22 @@
...
 
 
+// Macros to be used for JSON rendering
+
+#define JSON_PROP(name) std::string("\"" + std::string(name) + "\":")
+
+#define JSON_STR_LAST(value) std::string("\"" + std::string(value) + "\"")
+#define JSON_STR(value) std::string(JSON_STR_LAST(value) + ",")
+
+#define JSON_NUMBER_LAST(value) std::string(value)
```
kzangeli: This here will probably not work (JSON_NUMBER*).
It would be great if it did but ... 'value' is supposed to be an integer and we'd need a constructor of std::string accepting integers.
I don't think there is one (not that I have investigated it too much).

I normally end up doing something like this:

    int x = 12;
    char cV[12];

    snprintf(cV, sizeof(cV), "%d", x);

src/lib/common/tag.h
kzangeli: As it seems, std::string accepts char* not only on its right but also to its left :-)

NTC

kzangeli:
LGTM

fgalan:
LGTM

Issue #929, opened 2015-05-22
Remove function mongoDisconnect


kzangeli:
The function mongoDisconnect is no longer needed (after the inclusion of the connection pool for mongo).
The function should be removed from MongoGlobal.cpp/h and wherever it is called (especially in unit-tests).

**Effort: 1 man day**

fgalan:
I understand that is a matter of replacing mongoDisconnect() by setMongoConnectionForUnitTest(NULL) in the unit test, right?

fgalan:
Fixed in PR #977

PR #962, opened 2015-06-09
929 remove mongo disconnect

raconte:
Implements #929 

MongoDisconnect () function removed from ;
/src/lib/mongoBackend/MongoGlobal.cpp and MongoGlobal.h
/src/lib/serviceRoutines/leakTreat.cpp
/src/app/contextBroker/contextBroker.cpp
And from the corresponding tests.


src/app/contextBroker/contextBroker.cpp
```diff
@@ -1083,7 +1083,6 @@ void orionExit(int code, const std::string& reason)
     destroyAllOntimeIntervalThreads(dbs[ix]);
   }
 
-  mongoDisconnect();
```
fgalan: Reading #929 issues is not a matter of only remove `mongoDisconnect()` but replacing the stament by `setMongoConnectionForUnitTest(NULL)`

fgalan: Maybe is a good idea to revert commit https://github.com/telefonicaid/fiware-orion/commit/7eec274497c49dd094df02f8717d0c87636244a9 (I can tell you how to do it if you don't know), then use string replacement in the IDE:

`mongoDisconnect();` -> `setMongoConnectionForUnitTest(NULL);`

Only for files under test/unittes. In src/, just removing the function is ok.

raconte:
Closed because it  has occured an error during the git push and has been switched to another branch.

Issue #945, opened 2015-06-02
Local entiteis pagination is broken


fgalan:
Dicussed with @anabelengp, somehow related with https://github.com/telefonicaid/fiware-orion/issues/847

CB pagination doesn't work for local entities in the case some registration "shadows" the local entity. E.g. 

```
> db.registrations.find();
{ "_id" : ObjectId("556c158b3a1b8459f4f1ad32"), "expiration" : NumberLong(1433147445), "servicePath" : "/Basuras", "format" : "JSON", "contextRegistration" : [     {     "entities" : [     {  "id" : "device_de_ana",     "type" : "device" } ],     "attrs" : [     {     "name" : "PING",     "type" : "command",     "isDomain" : "false" } ],     "providingApplication" : "http://10.0.0.63:8080/iot/ngsi/d" } ] }
{ "_id" : ObjectId("556c18343a1b8459f4f1ad33"), "expiration" : NumberLong(1433233844), "servicePath" : "/Basuras", "format" : "JSON", "contextRegistration" : [     {     "entities" : [     {  "id" : "device_de_ana",     "type" : "device" } ],     "attrs" : [     {     "name" : "PING",     "type" : "command",     "isDomain" : "false" } ],     "providingApplication" : "http://10.0.0.63:8080/iot/ngsi/d" } ] }
```

```
> db.entities.find();
{ "_id" : { "id" : "entity_local", "type" : "local", "servicePath" : "/Basuras" }, "attrNames" : [  "a" ], "attrs" : { "a" : { "type" : "a", "creDate" : 1432296515, "modDate" : 1432296515, "value" : "a" } }, "creDate" : 1432296515, "modDate" : 1432296515 }
{ "_id" : { "id" : "qa_entity2", "servicePath" : "/Electricidad" }, "attrNames" : [ ], "attrs" : {  }, "creDate" : 1432632637, "modDate" : 1432632637 }
[...]

PR #1003, opened 2015-06-12
Fixes related with paginatinon and CPrs

fgalan:
This PR cover an unkwown aspect of #945, detected by @anabelengp team.

This PR will remain open while ensure that it mets all the requirements from the Portal team. Once merged, it has to be ported to develop (conflicts may occur due to code style cheks recently implemented in develop for mongoBackend).

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,2 @@
 Fix: Fixed a bug about pagination problem due to missing ErrorCode with details in the case of queries with forwarding to context providers.  (Issue #945)
+Fix: taking into account the pagination limit before adding "potential" entities to be queried in CPrs. (Issue #945)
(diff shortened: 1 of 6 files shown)


anabelengp:
This solution solves almost the whole issue but there is still a problem with the last page: 
if the last page contains fewer elements than the requested ones, the registrations related to entities not present on the page are added as trailing entities. This is:

create entity1, entity2, entity3, entity_with_registration4, entity5
ask for pages of 2 entities: limit=2
P1: limit=2, offset=0 --> entity1, entity2--------------------------------------------- OK
P2: limit=2, offset=2 --> entity3, entity_with_registration4----------------------- OK
P3: limit=2, offset=4 --> entity5, entity_with_registration4----------------------- WRONG. Only 'entity5' should be returned.

But if the last page contains exactly the number of elements that were requested, it works ok.

anabelengp:
LGTM

PR #1004, opened 2015-06-15
Apiary v2 documentation. Remove example operations

jmcanterafonseca:


doc/apiary/v2/fiware-ngsi-simple.apib
```diff
@@ -229,13 +193,13 @@ to keep your client decoupled from implementation details.
             "temperature": 21.7,
             "humidity": 60,
             "location": {
-                "value":  "41.3763726, 2.1864475,14",
+                "value":  "41.3763726, 2.1864475, 14",
```
fgalan: 3D coordinates are not supported right now (by the moment, we have been talking about it as a remote possibility). Thus, I'm not sure if we should use them in examples. It seems to be better to use just 2 elements.

fgalan:
LGTM

PR #1008, opened 2015-06-16
FIX Issue 1005 API v2: Strings are not properly rendered

jmcanterafonseca:
Work in progress patch

### DIFF
--- a/src/lib/ngsi/ContextAttribute.cpp
+++ b/src/lib/ngsi/ContextAttribute.cpp
@@ -351,23 +351,37 @@ std::string ContextAttribute::render
 
 /* ****************************************************************************
 *
-* toJson - 
+* toJson -
+*
+* FIXME: Refactor this method in order to simplify
+*        the code paths of the rendering process
+*
 */
 std::string ContextAttribute::toJson(bool isLastElement)
 {
   std::string  out;
   bool         isNumber = false;
+  bool         isString = false;
 
   if (type == "number")
   {
     isNumber = true;
   }
 
-  if (((type == "") || (isNumber == true)) && (metadataVector.size() == 0))
+  if (type == "string")
+  {
+    isString = true;
+  }
+
+  if (((type == "") || (isNumber == true || isString == true)) && (metadataVector.size() == 0))
   {
     if (isNumber == true)
     {
       out = JSON_VALUE_NUMBER(name, value);
+    }
+    else if (isString == true)
+    {
+      out = JSON_VALUE(name, value);
     }
     else
     {
--- a/test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
+++ b/test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
@@ -1,0 +1,161 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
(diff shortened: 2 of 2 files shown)


src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -357,18 +357,28 @@ std::string ContextAttribute::toJson(bool isLastElement)
...
   bool         isNumber = false;
+  bool         isString = true;
 
   if (type == "number")
   {
     isNumber = true;
   }
 
-  if (((type == "") || (isNumber == true)) && (metadataVector.size() == 0))
+  if(type == "string")
```
fgalan: Do we really need special treatment for "string" type? As far as I undersrtand, each type that is not a "number" (or a "date") is by default considered string type from a rendering point of view.

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -357,18 +357,28 @@ std::string ContextAttribute::toJson(bool isLastElement)
 {
   std::string  out;
   bool         isNumber = false;
+  bool         isString = true;
 
   if (type == "number")
```
fgalan: Thus, the type for number is exactly "number" (I mean, "Number" is not a valid numeric type). It is fine... just to check that in our documentation we make clear that "number" is case-sensitive.

test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
```diff
@@ -0,0 +1,137 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: I should  have though in four digits numbers when we start numbering functional test cases.... Probably an overall renaming (e,g. 766_XXX to 0766_X) now would make sense (although not as part of this PR, of course).

Or maybe I'm just delaying the problem and in a couple of month we would reach issue number #10000 :D

NTC

test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
```diff
@@ -0,0 +1,137 @@
...
+      "type": "Car",
+      "attributes" : [
+        {
+          "name" : "speed",
+          "type": "number",
+          "value" : 150
+        },
+        {
+          "name" : "brand",
+          "type": "string",
```
fgalan: Maybe some of the attributes should use an "implicit string" type, in order to check that it is also correctly rendered. E.g: `"type": "myString"`.

test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
fgalan: Actually deferred to tech debt issue: https://github.com/telefonicaid/fiware-orion/issues/1010

jmcanterafonseca: cool

test/functionalTest/cases/1005_GET_v2_entities_string/1005_GET_v2_entities_string.test
jmcanterafonseca: Fixed in 002cd1a2337d77ca1131cfbdd09f70952486039a

src/lib/ngsi/ContextAttribute.cpp
jmcanterafonseca: We are going to have booleans as well. 

I think the current rendering code could be simplified a lot , even we could consider using switch case avoiding too many ifs but that could be part of another PR IMHO and when Ken is here 

NTC

src/lib/ngsi/ContextAttribute.cpp
jmcanterafonseca: yup, I would prefer to be consistent with JSON types

src/lib/ngsi/ContextAttribute.cpp
fgalan: If you think that part is fixable, I'd recommend to hadd a comment over these lines:

```
//FIXME P10: rendering code could be simplified a lot. To discuss with Ken, once he is back
```

or something like that. Otherwise, we are in risk of forgetting thinks (it had happended in the past ;)

src/lib/ngsi/ContextAttribute.cpp
fgalan: NTC

fgalan:
LGTM

Issue #766, opened 2015-02-12
Attribute's metadata update fails when no value is given for such an attribute


frbattid:
Orion warns that updating an attribute's metadata when no value is given for such attribute is an offending operation:

```
$ cat updateContext.sh
(curl localhost:1026/v1/updateContext -s -S --header 'Content-Type: application/json' --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
{
    "contextElements": [
        {
            "type": "Room",
            "isPattern": "false",
            "id": "Room1",
            "attributes": [
            {
                "name": "temperature",
                "type": "float",
                "metadatas": [
                {
                        "name": "max",
                        "type": "double",
                        "value": "50"
                }
                ]
            },
            {
                "name": "pressure",
                "type": "integer",
                "value": "763"
            }
            ]
        }
    ],
    "updateAction": "UPDATE"
}
EOF
```

```
$ ./updateContext.sh 
{
    "contextResponses": [
        {
            "contextElement": {
                "attributes": [
                    {
                        "name": "temperature", 
                        "type": "float", 
                        "value": ""
                    }
                ], 
                "id": "Room1", 
                "isPattern": "false", 
                "type": "Room"
            }, 
            "statusCode": {
                "code": "472", 
                "details": "action: UPDATE - entity: [Room1, Room, false] - offending attribute: temperature - empty attribute not allowed in APPEND or UPDATE", 
                "reasonPhrase": "request parameter is invalid/not allowed"
            }
        }
    ]
}
```

Why is it necessary to provide a value for the attribute? I mean, I'm updating the attribute's metadata, not the attribute itself. This kind of behavior avoids this data lifecycle I was implementing through Cygnus:

1. An attribute's change is notified by Orion to Cygnus.
2. Cygnus uses the notified value in order to update certain statistics about the attribute such as maximum and minimum value, or average value.
3. Cygnus updates the attribute's metadata by pushing the calculated max, min, average, etc, but not updating the attribute itself.

**Effort: 2 man day**

fgalan:
Addressed in PR #1009

PR #1009, opened 2015-06-16
Feature/766 only metadata update

fgalan:
Implements #766


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -294,10 +294,29 @@ static bool mergeAttrInfo(BSONObj& attr, ContextAttribute* caP, BSONObj* mergedA
 {
   BSONObjBuilder ab;
 
-  /* 1. Add value */
-  valueBson(caP, ab);
+  /* 1. Add value, if present in the resquest (it could be omitted in the case of updating only metadata)  */
```
kzangeli: typo: resquest

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -294,10 +294,29 @@ static bool mergeAttrInfo(BSONObj& attr, ContextAttribute* caP, BSONObj* mergedA
...
+  if (caP->value != "" || caP->compoundValueP != NULL)
+  {
+    valueBson(caP, ab);
+  }
+  else
+  {
+    /* Slightly different treatment, depending if the attribute value in DB is compound or not */
```
kzangeli: depending on whether the attribute value in DB is compound

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -294,10 +294,29 @@ static bool mergeAttrInfo(BSONObj& attr, ContextAttribute* caP, BSONObj* mergedA
...
   BSONObjBuilder ab;
 
-  /* 1. Add value */
-  valueBson(caP, ab);
+  /* 1. Add value, if present in the resquest (it could be omitted in the case of updating only metadata)  */
+  if (caP->value != "" || caP->compoundValueP != NULL)
+  {
+    valueBson(caP, ab);
+  }
+  else
```
kzangeli: If I understand this correctly, when the value of the attribute is empty (no update needed/wanted),  then the value of the attribute is 'copied' from DB to the variable 'ab' and sent back to mongo, to not destroy the value ?

An explaining comment would be good :-)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in 5dc4402

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in 5dc4402

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in 5dc4402

kzangeli:
LGTM

PR #1018, opened 2015-06-19
ADD apiary reference for /v2/entities/{id}, /v2/entities/type/{type}id/{id} and placeholders for standard operations

fgalan:


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1,5 +1,5 @@
 FORMAT: 1A
-HOST: http://fiware-ngsi2.apiblueprint.org/
+HOST: http://orion.lab.fiware.org/
```
fgalan: Actually, what I would like to have it the operations rendered in the right panel as "GET /v2/entities" etc, without any host. However, I haven't been able (please tell me if you know how :) so I'm using the public Orion instance as reference (although currently it doesn't support v2, of course).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -27,16 +27,6 @@ HOST: http://fiware-ngsi2.apiblueprint.org/
 
 FIWARE-NGSI Simple API.
 
-# FIWARE-NGSI-Simple API Root [/v2]
```
fgalan: Just moving sections around...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+<tr><td colspan="2"><strong>Required Parameters</strong></td></tr>
+<tr><td colspan="2">none</td></tr>
+<tr><td colspan="2"><strong>Optional Parameters</strong></td></tr>
+<tr>
+  <td style="padding-right: 40px; width: 140px;">op</td>
+  <td>Operation type. Allowed values: <code>append</code>.</td>
+</tr>
+</table>
+
+The request payload is an object representing the attributes to append or update. The object follows
```
fgalan: Not sure if a good way of describing the kind of payload this operation, I mean to refer to the JSON entity representation then ammend with "except that `id` and `type` are not allowed". Maybe it is a bit dirty, but it saves adding a whole new section for "JSON attributes list representation" which basically would have the same text than the section in "JSON entity representation format".

What do you think?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
++ Response 204
+
+### Remove entity atributes [DELETE /v2/entities/{entityId}?{attrs}]
+
+<table style="width: 100%; margin: 12px 0 0 0;">
+<tr><td colspan="2"><strong>Required Parameters</strong></td></tr>
+<tr><td colspan="2">none</td></tr>
+<tr><td colspan="2"><strong>Optional Parameters</strong></td></tr>
+<tr>
+  <td style="padding-right: 40px; width: 140px;">attrs</td>
```
fgalan: Maybe this operation is a bit dangerous? E.g. a client that is used to delete attributes using it but that in some moments forget to include `?attrs=myAttr` and deletes the entity wrontly...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -607,344 +797,50 @@ Response code:
 
 + Response 204
 
+## Standard queryContext [/v2/queryContext]
 
-## Context Entity Subscriptions [/v2/entities/{entityId}/subscriptions]
```
fgalan: By the moment, I have removed the not-yet-discussed operations in the reference document. We already have these examples in examples .apib anyway.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -607,344 +797,50 @@ Response code:
...
-    + subscriptionId: abcdefg (required, string) - ID of the suscription
-
-+ Response 204
-
-
-### Create a new subscription [POST /v2/subscriptions]
+### queryContext [POST /v2/queryContest]
 
-+ Request (application/json)
+## Standard updateContext [/v2/updateContext]
```
fgalan: I have created some placeholders for standard operations. However, I don't like too much the result: by the way of working of Apiary, given that each std op uses a different URL, we have to created a lot of 2-level sections which only containg one 3-level section each one (the corresponding to POST). The result is a left index very large (it would be great to have just one item in that index, all the std ops grouped together there).

Not sure if what I want to achieve is not possible in APIary...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -283,7 +281,7 @@ on the shape. In particular, the following geometries can be defined so far:</p>
 
 </td></tr>
 <tr><td style="padding-right: 40px; width: 140px;">attrs</td>
-    <td>List of comma separated attributes to include in the entities of the response. If this parameter is not
+    <td>Comma-separated subset of the attributes to be included in the response. If this parameter is not
```
jmcanterafonseca: Comma-separated list of attribute names which values will be included in the response.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -364,30 +371,37 @@ Response code:
...
+### Retrieve entity [GET /v2/entities/{entityId}{?attrs}]
 
-+ Response 422 (application/json)
+<table style="width: 100%; margin: 12px 0 0 0;">
+<tr><td colspan="2"><strong>Required Parameters</strong></td></tr>
+<tr><td colspan="2">none</td></tr>
+<tr><td colspan="2"><strong>Optional Parameters</strong></td></tr>
+<tr>
+  <td style="padding-right: 40px; width: 140px;">attrs</td>
+  <td>Comma-separated subset of the attributes to be included in the response. If this parameter is not
```
jmcanterafonseca: see comment above

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: yes, the shorter the better.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+
+The entity attributes are updated with the ones in the payload. In particular, depending on
+whether `op=append` is used:
+
+* If `op=append` is not used: the entity attributes are updated (if they previously exist) or appended
+  (in they don't previously exists) with the ones in the payload.
+* If `op=append` is used (i.e. strict append semantics): all the attributes in the payload not previously
+  existing in the entity are appended. In addition to that, in the case some of the attributes in the
+  payload already exist in the entity, an error if returned.
+
```
jmcanterafonseca: instead of op=append we may have options=append , but that would be part of another discussion

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
++ Response 204
+
+### Update existing entity attributes [PATCH /v2/entities/{entityId}]
+
+The request payload is an object representing the attributes to update. The object follows
+the JSON entity representation format (described in a section above), except that `id` and `type`
+are not allowed.
+
+The entity attributes are updated with the ones in the payload. In particular, the entity attributes
+are updated with the ones in the payload.In addition to that, in the case some of the attributes in the
```
jmcanterafonseca: duplicated phrase

jmcanterafonseca: If one or more attributes do not exist an error is returned

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+            "temperature": 25.5,
+            "seatsNumber": 6
         }
 
-### Retrieve entity data by id and type [GET /v2/entities/type/{entityType}/id/{entityId}]
++ Response 204
+
+### Replace entity attributes [PUT /v2/entities/{entityId}]
+
+The request payload is an object representing the attributes to replace. The object follows
```
jmcanterafonseca: representing the new entity attributes

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: well, that would be a separate issue to be considered outside this PR

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+</tr>
+</table>
+
+The response is an object representing the entity identified by the type and ID. The object
+follows the JSON entity representation format (described in a section above).
+
+This operation must return only one entity element, but it may happen that there are more
+than one entity with the same ID and type (e.g. entities with same ID and type in different
+service paths). In those cases  an error mesage is returned, specifying in the description the
+URL that could be used to get the list of conflicting entities.
```
jmcanterafonseca: in this case as we are providing type and id, only one entity can match, can't it?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -468,11 +631,24 @@ Response code:
 
 + Response 204
 
+### Update existing entity attributes [PATCH /v2/entities/type/{entityType}/id/{entityId}]
+
+The request payload is an object representing the attributes to update. The object follows
+the JSON entity representation format (described in a section above), except that `id` and `type`
+are not allowed.
```
jmcanterafonseca: duplicated block?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: well done

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -364,30 +371,37 @@ Response code:
...
-            "error": "AlreadyExists",
-            "description": "An entity with the same id and type is already present"
-        }
+The response is an object representing the entity identified by the ID. The object follows
+the JSON entity representation format (described in a section above).
 
+This operation must return only one entity element, but it may happen that there are more
+than one entity with the same ID (e.g. entities with same ID but different type). In those cases
+an error mesage is returned, specifying in the description the URL that could be used to get
+the list of conflicting entities.
```
jmcanterafonseca: all the entities with such an ID

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
-            "error": "NotFound",
-            "description": "There is no entity with such an id"
+            "temperature": 25.5,
+            "seatsNumber": 6
         }
 
-### Retrieve entity data by id and type [GET /v2/entities/type/{entityType}/id/{entityId}]
++ Response 204
+
+### Replace entity attributes [PUT /v2/entities/{entityId}]
```
jmcanterafonseca: Replace all entity attributes

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+
+### Remove entity atributes [DELETE /v2/entities/{entityId}?{attrs}]
+
+<table style="width: 100%; margin: 12px 0 0 0;">
+<tr><td colspan="2"><strong>Required Parameters</strong></td></tr>
+<tr><td colspan="2">none</td></tr>
+<tr><td colspan="2"><strong>Optional Parameters</strong></td></tr>
+<tr>
+  <td style="padding-right: 40px; width: 140px;">attrs</td>
+  <td>Comma-separated subset of the attributes to be removed. If this parameter is not
```
jmcanterafonseca: same comment as above

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
+### Retrieve entity [GET /v2/entities/type/{entityType}/id/{entityId}?{attrs}]
+
+<table style="width: 100%; margin: 12px 0 0 0;">
+<tr><td colspan="2"><strong>Required Parameters</strong></td></tr>
+<tr><td colspan="2">none</td></tr>
+<tr><td colspan="2"><strong>Optional Parameters</strong></td></tr>
+<tr>
+  <td style="padding-right: 40px; width: 140px;">attrs</td>
+  <td>Comma-separated subset of the attributes to be included in the response. If this parameter is not
+  included, all the attributes are retrieved.</td>
```
jmcanterafonseca: same comment as above. Wondering if the attrs parameter could be described one time only

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -403,60 +417,209 @@ Response code:
...
             "humidity": 60,
+            "location": {
+                "value":  "41.3763726, 2.1864475",
+                "type": "geo:point",
+                "crs": "WGS84"
+            }
         }
 
-### Get entity attributes. Not present [GET /v2/entities/{entityId}{?attrs}]
+### Update or append entity attributes [POST /v2/entities/type/{entityType}/id/{entityId}?{op}]
```
jmcanterafonseca: can we say just that this resource supports the same operations as the one with the entity and save many spec lines?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: ok

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I think standard operations are already described by v1 apiary document. So I think a link between the 2 documents might suffice

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: No comment so far, so I understand that NTC.

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Ok. NTC (in this PR).

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Three things to take into account:

* There isn't a "formal reference" for v1 operations in apiary (as fiware-ngsiv2-reference.apib is or will be). There is an apiary document for v1 operations, but is closer to a cookbook than to a formal reference (have a look to https://github.com/telefonicaid/fiware-orion/blob/develop/doc/apiary/orioncontextbroker.apib)

* Asumming there were such reference (we can create it), the operations could have slight diferences that make sense to have them also in this API v2 document. At least, the URL will change, e.g. from "/v1/queryContext" to "/v2/queryContext" given that each NGSI API version at URL level (i.e. "/v2") should be self-contained.

However, we don't need to close this in this PR. By the moment I would suggest to leave the placeholder (given that it is easier to remove things in the future than removing them now and adding them back if we change our mind :)

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I would use:

"Comma-separated list of attribute names which ~~values~~ will be included in the response. "

Given that is the attribute iself (including not only value but also potentially metadata and type) the piece of information that is included, i.e. not only the attribute value.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I think is a good idea and quite straighforward. Let's change it for this PR.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Not really... if we have the same entity (with same entity and type) in two different service paths and the Fiware-ServicePath used in this REST request uses `/#` we  have ambiguety.

Thus, I think the paragraph make sense to explain what happen in that case.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The introduction to PUT/PATCH/POST is very similar, but not exaclty the same block. There is some repetition but I haven't find a best way of describing this...

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The description is brief (just one line). Thus, I think is better to have it as it is now (the alternative would be to move the description to a common block use a "See definition in X", which I think is more confusing in this case).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: You mean in general to all the operations which use `/type/<type>/id/<id>" ? Could make sense...

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in dfb2263

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: if you don't want to use the word 'values' I would use the word 'data' .

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: Oh I thought there were a default path ... but probably that's part of a different discussion

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: well, that's part of a separated discussion. I'm still in doubt we need a separate resource for  

/type/myType/id/myId

 as we can do the same with /entities/myId?type=xxx 

and that would save us from complexity in the spec

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: well I would prefer a spec which is not redundant and which does not duplicate stuff otherwise we will face maintainance problems as it happens with code

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -364,30 +371,37 @@ Response code:
...
-        {
-            "error": "AlreadyExists",
-            "description": "An entity with the same id and type is already present"
-        }
+The response is an object representing the entity identified by the ID. The object follows
+the JSON entity representation format (described in a section above).
 
+This operation must return only one entity element, but it may happen that there are more
+than one entity with the same ID (e.g. entities with same ID but different type). In those cases
+an error mesage is returned, specifying in the description the URL that could be used to get
```
jmcanterafonseca: specifying in the error payload 

I'm wondering if the payload could include in that case a xx_url field but that would be part of separate discussion

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I remark that I would prefer to describe standard operations in one place only and keep them as they are as much as possible and ideally make them the same

but that would be part of a separate discussion I think :)

jmcanterafonseca:
LGTM provided the latest tiny comments are adressed. I don't know if we are capturing the to be discussed issues in a separate document or issue but it would help tracking

great work Fermin!

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Added to the new issue about open discussions. NTC in this PR.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I like data, ok.

fgalan: Fixed in ffc461f.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Actually, there is a *default value* for the Fiware-Service path, which depends whether the operation is query or update:

* For update: Fiware-ServicePath default is `/`, i.e. entities are created in the root of the path.
* For query: Fiware-ServicePath default is `/#`, i.e. the query search in all the service path tree.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Moved to the issue about open discussions.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I'm still thinking that for one or two lines long descriptions it is clearer if we avoid duplication. Anyway, let's review this at the final stage of the document, when it gets complete, to see how it looks like globally.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: This is related with the open discussion on `/type/<type>/id/<id>` URLs. Let's hold on the text as it is now until that discussion reach a conclusion.

NTC (in this PR)

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: That would "break" the general error payload form consisting on just "error" (always) and "description" (some times). But, let's move it to the open discussions issue.

NTC (in this PR).

PR #1026, opened 2015-06-24
DB Cleanup in functional tests

### LINKED ISSUE #651: DB cleanup in functional tests
All functional tests should/must make sure the databases they are to use are empty (or dropped), and they should also cleanup after themselves in the teardown part.

There are two functions to help:
* dbInit, to be called in SHELL-INIT
* dbDrop, to be called in TEARDOWN

Both these functions do a drop of the database (first and only argument).
I prefer to keep the two functions though, in case we need to modify the behaviour of any of them in the future.

**Effort: 2 man day**

raconte:
Fixes #651 

It has been added the pair functions dbInit/dbDrop.


fgalan:
LGTM

PR #1031, opened 2015-06-24
Rename db unittests

### LINKED ISSUE #628: Rename db unittest to utest
The basename of the database used for functional tests is about to change name from harnesstest to ftest. We should do the same with the db for unit tests: 'unittest' -> 'utest'.

However, 'unittest' is used in ~50 unit tests, instead of using the define already existing in testInit.h.

So, the definition in testInit.h should be changed from 'unittest' to 'utest' and those direct usages of 'unittest' should be changed to use the definition instead.

**Effort: 1 man day**

raconte:
Fixes #628 
The definition in testInit.h has been changed from 'unittest' to 'utest' and from those direct usages of 'unittest' stored in test/unittests/mongoBackend/


raconte:
Fixed in 1beee47

fgalan:
LGTM

PR #1036, opened 2015-06-25
ADD comment about NGSI list archives

fgalan:
In addition, I have removed the licence header in the cookbook .apib (which was defferred due to a hurry in a past PR).


jmcanterafonseca:
thanks LGTM

PR #1055, opened 2015-07-03
Hardening/939 leaks

kzangeli:
Fixed a few leaks

### DIFF
--- a/src/lib/common/sem.cpp
+++ b/src/lib/common/sem.cpp
@@ -294,55 +294,86 @@ int transSemGive(const char* who, const char* what)
 static pthread_mutex_t contexts_mutex = PTHREAD_MUTEX_INITIALIZER;
 static std::map<std::string, struct curl_context> contexts;
 
-//Statistics
+// Statistics
 static struct timespec accCCMutexTime = { 0, 0 };
 
 
-int get_curl_context(const std::string& key, struct curl_context *pcc)
+/* ****************************************************************************
+*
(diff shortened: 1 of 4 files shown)


src/lib/common/sem.cpp
```diff
@@ -294,55 +294,86 @@ int transSemGive(const char* who, const char* what)
...
+// Statistics
 static struct timespec accCCMutexTime = { 0, 0 };
 
 
-int get_curl_context(const std::string& key, struct curl_context *pcc)
+/* ****************************************************************************
+*
+* curl_context_cleanup - 
+*/
+void curl_context_cleanup(void)
```
fgalan: I wonder if curl related functions should be under the curl library (appart from the ones about locks management, that are ok in this file).

kzangeli: I also don't like the curl functions in this file.
Strictly sem (the name is sem.cpp ...)
All the curl functions should be in some other file, IMO.
Not really part of this PR though

fgalan: Maybe we should create a tech debt issue about sem.cpp refactoring to put these functions out. In addition, probably a better name for sem.cpp/sem.h is locks.cpp/looks.h (given that not all locks are based on semaphores, some are based on pthread_mutex, if I'm remembering correctly).

I'd create the issue on Monday. If you have the opportunity to do it before, please go ahead.

NTC (in this PR).

fgalan:
LGTM

PR #1057, opened 2015-07-05
ADD Dockerfile to enable dev with the latest Ubuntu and Docker Containers

jmcanterafonseca:
This is a work in progress. Please do not review yet. Early feedback welcome.

### DIFF
--- a/docker/dev/ubuntu/Dockerfile
+++ b/docker/dev/ubuntu/Dockerfile
@@ -4,42 +4,54 @@ FROM ubuntu:latest
 MAINTAINER FIWARE Orion Context Broker Team. Telefónica I+D
 
 RUN \
-    apt-get update && \
-    apt-get install -y \
-    build-essential \
-    cmake \
-    curl \
-    git \
-    libboost-dev \
-    libboost-regex-dev \
-    libboost-thread-dev \
-    libboost-filesystem-dev \
-    libcurl4-gnutls-dev \
-    libmicrohttpd-dev \
+    apt-get update &&          \
+    apt-get install -y         \
+    build-essential            \
+    cmake                      \
+    curl                       \
+    git                        \
+    libboost-dev               \
+    libboost-regex-dev         \
+    libboost-thread-dev        \
+    libboost-filesystem-dev    \
+    libcurl4-gnutls-dev        \
+    libmicrohttpd-dev          \
     libssl-dev libcrypto++-dev \
-    libxml2-utils \
-    make \
-    mongodb-clients \
-    netcat \
-    scons  \
-    wget && \
+    libxml2-utils              \
+    make                       \
+    # For functional tests
+    mongodb-clients            \
+    # For functional tests
+    netcat                     \
+    # For compiling Mongo Driver    
+    scons                      \
+    # For downloading extra utilities
+    wget                     && \  
     cd /root && \
+    #
+    # MongoDB Driver download and compile
+    # 
     wget https://github.com/mongodb/mongo-cxx-driver/archive/legacy-1.0.2.tar.gz && \
-    tar xfvz legacy-1.0.2.tar.gz && \
-    cd mongo-cxx-driver-legacy-1.0.2 && \
+    tar xfvz legacy-1.0.2.tar.gz                && \
+    cd mongo-cxx-driver-legacy-1.0.2            && \
     scons &&  scons install --prefix=/usr/local && \
-    cd .. && \
-    rm -R mongo-cxx-driver-legacy-1.0.2 && \
+    cd ..                                       && \
+    rm -R mongo-cxx-driver-legacy-1.0.2         && \
+    #
+    # Mocks for unit testing download and compile
+    #
     wget http://googlemock.googlecode.com/files/gmock-1.5.0.tar.bz2 && \
     tar xfvj gmock-1.5.0.tar.bz2 && \
-    cd gmock-1.5.0 && \
-    ./configure && \
-    make && make install && \
-    cd .. && \
-    rm -R gmock-1.5.0 && \
+    cd gmock-1.5.0               && \
+    ./configure                  && \
+    make && make install         && \
+    cd ..                        && \
+    rm -R gmock-1.5.0            && \
+    #
+    # Install Flask for functional testing (accumulator)
+    #
     wget https://bootstrap.pypa.io/get-pip.py && \
-    python get-pip.py && \
-    pip install Flask && \
-    echo cd /root/src >> /root/.bashrc
+    python get-pip.py                         && \
+    pip install Flask                         && \
 
 EXPOSE 1026
--- a/docker/dev/ubuntu/Dockerfile
+++ b/docker/dev/ubuntu/Dockerfile
@@ -20,16 +20,26 @@ RUN \
     libxml2-utils \
     make \
     mongodb-clients \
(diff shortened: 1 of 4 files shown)


docker/dev/ubuntu/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Develop Orion Context Broker Ubuntu-Latest
+FROM ubuntu:latest
```
LeandroGuillen: Why Ubuntu and not CentOS?

docker/dev/ubuntu/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    wget https://github.com/mongodb/mongo-cxx-driver/archive/legacy-1.0.2.tar.gz && \
+    tar xfvz legacy-1.0.2.tar.gz && \
+    cd mongo-cxx-driver-legacy-1.0.2 && \
+    scons && \
+    scons install --prefix=/usr/local && \
+    cd .. && \
+    rm -R mongo-cxx-driver-legacy-1.0.2 && \
+    echo cd /root/src >> /root/.bashrc
+
+EXPOSE 1026
```
LeandroGuillen: I understand this docker is for developing Orion. You will need to create a data volume in this dockerfile pointing to the root of the project tree.

docker/dev/ubuntu/scripts/examples/mongo.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+docker --tlsverify=false run -d --name mongoOrion mongo:2.6.10
```
LeandroGuillen: I woud suggest a docker compose file instead of this one. Or at least in addition to this one.

docker/dev/ubuntu/Dockerfile
jmcanterafonseca: yes, we are going to need the CentOS one as well. Maybe you can contribute it, taking as input the work done at https://registry.hub.docker.com/u/bitergia/fiware-orion/dockerfile/

docker/dev/ubuntu/Dockerfile
jmcanterafonseca: yes, that's the idea to keep the codebase outside the container in order to be able to edit source code locally

docker/dev/ubuntu/scripts/examples/mongo.sh
jmcanterafonseca: oh yes, we need the composer as well

thanks for the feedback!

docker/dev/ubuntu/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Develop Orion Context Broker Ubuntu-Latest
+FROM ubuntu:latest
+
+MAINTAINER FIWARE Orion Context Broker Team. Telefónica I+D
+
+RUN \
+    apt-get update && \
```
fgalan: This script looks a bit obscure for people not used to the Orion build process. Maybe introducing comments in the proper places could help, e.g:

```
# Installing Orion build dependencies
...
# Building mongo driver
...
```

test/functionalTest/harnessFunctions.sh
```diff
@@ -94,32 +94,45 @@ function dMsg()
 #
 function dbInit()
 {
-  role=$1  
-  db=$1
+  role=$1
+  
+  host="${CB_DATABASE_HOST}"
```
fgalan: Why did you need to change the way in which some harness functions are implemented? Or, in other words, which aspect in the current way of working make them incompatible with the docker stuff.

docker/dev/ubuntu/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# Develop Orion Context Broker Ubuntu-Latest
```
fgalan: As general comment, I think that some piece of documentation (i.e. an .md file) explaining how to use the dokcer stuff would be needed, in a similar way a .md file explaining the valgrand stuff is already in the repository  (see https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/vagrant.md)

docker/dev/ubuntu/Dockerfile
fgalan: I agree in that we should provide also CentOS 6.x (it would be a bit weird to include Ubuntu, which is not official, before including the official one). How complex is to create a docker file for CentOS 6.x?

test/functionalTest/harnessFunctions.sh
LeandroGuillen: It seems to me that this is so that the MongoDB instance can be on a host different than localhost for the harness tests. Which is the case when running docker.

docker/dev/ubuntu/Dockerfile
dmoranj: Change `FROM ubuntu:latest` to `FROM centos:6` and it's done. I don't think we should provide different versions of Context Broker for different SOs, as any Docker image should work on any Host SO without changes so, what do we have to gain having multiple images? Docker images should be black boxes of the applications, so there's no point, IMHO, in creating multiple flavours.

fgalan: Not sure if it so simple... note that for instance the `apt-get update` in `RUN`will not work in CentOS.

LeandroGuillen: I agree with Dani. But this image I understand is for development of Orion, not to deploy an Orion to use. I guess this is for folks who want to use other OS than CentOS for development. That's why I think the CentOS one should be preferred.

Then, for deploying Orion or just trying it this image wouldn't make much sense.

LeandroGuillen: Orion docker based on CentOS 6 in https://github.com/telefonicaid/fiware-orion/pull/1059

docker/dev/ubuntu/Dockerfile
fgalan: PR https://github.com/telefonicaid/fiware-orion/pull/1059 contains a very good .md document that could be used as template for the Ubuntu case.

docker/dev/ubuntu/Dockerfile
fgalan: NTC from my side.

docker/dev/ubuntu/scripts/examples/orion-dev.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+source $HOME/.bashrc
+docker --tlsverify=false run -p 1026:1026 --name orionDev --link mongoOrion:mongoOrion -v /Users/jmcf/work/develop/fiware-orion:/root/src -t -i orion-dev:latest /bin/bash
```
fgalan: `/Users/jmcf/work/develop/fiware-orion` is user-specific. Should be replaced in the documention, e.g. using some generic like  `/path/to/fiware-orion` or something like that.

fgalan:
Taking into account how the docker requirements have evolved (and the merging of PR https://github.com/telefonicaid/fiware-orion/pull/1069 implementing such requirements), maybe this PR should be closed?

jmcanterafonseca:
closing this. Non-official docker corresponding to this PR can be found at 

https://registry.hub.docker.com/u/jmcanterafonseca/fiware-orion-dev/

PR #1058, opened 2015-07-06
modified a leak fix so that the broker doesn't crash ... :-)

kzangeli:
Modified a leakfix so that it doesn't make the broker die ...


fgalan:
LGTM

PR #1065, opened 2015-07-10
REMOVE empty version_via_rest.test (leftover of a past PR refactoring)

fgalan:
This is a leftover of the PR in which we change from 3-digit to 4-digit functional test cases directories. The actual version_via_rest.test is actually inside cases/0000_version_operation directory.

Upon merging, the problem with jenkins file compliance job should be fixed.


fgalan:
It should also fix the problem with test harness job:

```
000_version_operation/version_via_rest.test .................................................................. (ERROR 2 ...)
```

kzangeli:
LGTM

PR #1071, opened 2015-07-15
Feature/970 create entity

kzangeli:
Implementation of POST /v2/entities (issue #970)

Also, new JSON parser, introducing 'json types' (String, Boolean and Number).
Lots and lots of changes ...

LM_Ms for Compounds (very complex changes) still in the code. The idea is to maintain these LM_Ms until the compounds are 100% verified to work. Removing them is a simple task and can be done in a future PR.

8 functests to verify the operation, but more to be implemented.

PR #1074, opened 2015-07-16
Added the CMakeLists.txt to jsonParseV2

kzangeli:
Added a missing file to last PR (#1071)

### DIFF
--- a/src/lib/jsonParseV2/CMakeLists.txt
+++ b/src/lib/jsonParseV2/CMakeLists.txt
@@ -43,7 +43,6 @@ SET (HEADERS
 # Include directories
 # -----------------------------------------------------------------
 include_directories("${PROJECT_SOURCE_DIR}/src/lib")
-include_directories("/opt/rapidjson/rapidjson-1.0.2/include")
 
 # Library declaration
 # -----------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/lib/jsonParseV2/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+    parseMetadata.h
+    jsonParseTypeNames.h
+)
+
+
+
+# Include directories
+# -----------------------------------------------------------------
+include_directories("${PROJECT_SOURCE_DIR}/src/lib")
+include_directories("/opt/rapidjson/rapidjson-1.0.2/include")
```
fgalan: This is very environment specific... The usual place of installing libraries is `/usr/lib` and `/usr/local/lib`.

kzangeli: That could be discussed.
/usr/lib should never be used.
/usr/local/lib was invented many years ago when hard disks were expensive and we had NFS mounted operating system (if the disk server didn't answer (mounted on /usr/lib), then the local HD was used as backup, in /usr/local/lib).

Libraries that you add manually to your system should be put in /opt (that way (and if you have /opt as a partition) you can replace the whole linux distro without having to touch what you installed by hand).

This is my opinion, it is no absolute truth ...

kzangeli: Fixed in 9dfe6e3

fgalan:
LGTM

PR #1082, opened 2015-07-17
3.2999999999998 => 3.3 - new python version fixes the problem

kzangeli:
I installed a newer version of python (2.7.10) and my CentOS now works just like Jenkins.


fgalan:
LGTM

Issue #1083, opened 2015-07-20
RPM package contextBroker-tests missing httpsPrepare.sh script


LeandroGuillen:
What the title says.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1084

PR #1084, opened 2015-07-21
ADD missing scripts to RPMs

fgalan:
It solves #1083. In additon, I have take this opportunity to solve #2.


rpm/SPECS/contextBroker.spec
```diff
@@ -117,8 +117,10 @@ rm $RPM_BUILD_ROOT/usr/share/contextBroker/tests/cases/CMakeLists.txt
 cp LICENSE $RPM_BUILD_ROOT/usr/share/doc/contextBroker
 cp scripts/testEnv.sh test/functionalTest/testHarness.sh test/functionalTest/testDiff.py test/functionalTest/harnessFunctions.sh $RPM_BUILD_ROOT/usr/share/contextBroker/tests
 cp scripts/accumulator-server.py $RPM_BUILD_ROOT/usr/share/contextBroker/tests 
+cp scripts/httpsPrepare.sh $RPM_BUILD_ROOT/usr/share/contextBroker/tests
```
LeandroGuillen: We have the script, but I don't think it's reachable yet.

For example, take this line in `0000_https_support/https.test`:
    
    ${SCRIPT_HOME}/../../scripts/httpsPrepare.sh --keyFileName /tmp/harnessTest.key --certFileName /tmp/harnessTest.pem

If the test is run from `/usr/share/contextBroker/tests` then this points to `/usr/scripts/httpsPrepare.sh`.

kzangeli: I guess the easiest way to fix this is to call httpsPrepare.sh using $PATH.
We'll all need to add "~/git/fiware-orion/scripts" (or wherever we have the repo) to our PATH.
The problem arises when we use more than one repo (not that httpsPrepare.sh will change between repos, but still).

Another way would be to create an ENV VAR pointing to the file.

No good way of solving this, really

LeandroGuillen: how about putting httpsPrepare.sh in the same directory as harnessFunctions and testHarness? It is only used there.

kzangeli: Sure, let's do that

fgalan: I'll do that in this PR.

fgalan: As in 8eb1e70d1521a9fc3ad9fb122beaa14346d3a8cc this has been fixed.

LeandroGuillen:
LGTM

PR #1085, opened 2015-07-21
RENAME 766 cases -> 0766 cases

fgalan:
I have found that this case directory was using 3 digits, so I have rename to 4 digits. Easy fix.


kzangeli:
LGTM

Issue #1087, opened 2015-07-21
[BUG] error message v1, is not the expected in v2


iariasleon:
### Request:
```
http://<host>:<port>/v2/entities
Fiware-Service: sdsdfs.et
Fiware-ServicePath: /test
Content-Type: application/json

{"type":"room", "id": "room_0", "timestamp_0": {"type": "date", "value": "017-06-17T07:21:24.238Z", "very_hot_1": {"type": "alarm", "value": "Afe1BuMTcf"}, "very_hot_0": {"type": "alarm", "value": "Afe1BuMTcf"}}}
```
### Response:
```
Status: 201 Created

{
  "orionError" : {
    "code" : "400",
    "reasonPhrase" : "Bad Request",
    "details" : "tenant name not accepted - a tenant string must not be longer than 50 characters and may only contain underscores and alphanumeric characters"
  }
}
```

### Expected:
```
Status: 400 Bad Request

{
  "orionError" : {
    "error" : "Bad Request",
    "description" : "tenant name not accepted - a tenant string must not be longer than 50 characters and may only contain underscores and alphanumeric characters"
  }
}
```

kzangeli:
Perhaps even:

### Expected:

    {
      "error": "Bad Request",
      "description": "tenant name not accepted - a tenant string must not be longer than 50 characters and may only contain underscores and alphanumeric characters"
    }

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1088. @iariasleon please close this issue after checking that everything is ok from QA side.

iariasleon:
LGTM

Issue #970, opened 2015-06-10
API v2: Create entity


fgalan:
Implement the operation to create an entity in v2, i.e. 

```
POST /v2/entities
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
This pseudo-test harness should be taken into account when implementing this issue (only the part not related with canonical model, which is still unders discsussion and, any way, is part of another implementation issue):

https://gist.github.com/fgalan/d39b016be45d2591a6e7

fgalan:
Implemented in PR #1071

PR #1088, opened 2015-07-21
Fixed a bug about error handling for API version 2

kzangeli:
Fixed a bug (Issue #1087)

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,4 @@
 Add:  POST /v2/entities - entity creation under version 2 (Issue #970)
 Add:  API version 2 only supports JSON so we can now distinguish between STRINGS, NUMBERS (all treated as floats for now),
       and BOOLEANS (true, false), for the VALUE for ContextAttribute and Metadata (No Issue)
+Fix:  Fixed a bug about error handling for API version 2 (Issue #1087)
--- a/src/lib/serviceRoutinesV2/postEntities.cpp
+++ b/src/lib/serviceRoutinesV2/postEntities.cpp
@@ -48,7 +48,8 @@
 *
 * 01. Fill in UpdateContextRequest
 * 02. Call standard op postUpdateContext
-* 03. Cleanup and return result
+* 03. Prepare HTTP headers
+* 04. Cleanup and return result
 */
 std::string postEntities
 (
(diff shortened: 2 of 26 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
 Add:  POST /v2/entities - entity creation under version 2 (Issue #970)
 Add:  API version 2 only supports JSON so we can now distinguish between STRINGS, NUMBERS (all treated as floats for now),
       and BOOLEANS (true, false), for the VALUE for ContextAttribute and Metadata (No Issue)
+Fix:  Fixed a bug about error handling for API version 2 (Issue #1087)
```
fgalan: The bug was already in 0.22.0? I though that it was related with #970 (part of 0.23.0), thus it shouldn't be included in the changelog (as there is no change regarding with 0.22.0). But maybe I'm wrong...

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -60,24 +61,24 @@ std::string postEntities
 {
   Entity*  eP = &parseDataP->ent.res;
 
-
-  // 00. Temp debug
-  // LM_M(("Presenting"));
```
fgalan: A little off-topic, but... are is the LM_M/LM_F stuff "controlled"? Is there any other remaining LM_M/LM_F? I'm a bit nervous since I discovered a forgotten LM_M/LM_F was consuming 25% of contextBroker CPU self-time during recent performance tests...

test/functionalTest/cases/1087_error_handling/error_messages_for_invalid_service_path_in_v2.test
```diff
@@ -0,0 +1,79 @@
...
+01. POST /v2/entities with invalid tenant
+=========================================
+HTTP/1.1 400 Bad Request
+Content-Length: 181
+Content-Type: application/json
+Date: REGEX(.*)
+
+{
+    "description": "tenant name not accepted - a tenant string must not be longer than 50 characters and may only contain underscores and alphanumeric characters", 
+    "error": "Bad Request"
```
fgalan: Bad Request -> BadRequest

In addition, this error has not been added tot the .apib file.

Please, remember both things whenever a new error is introduced.

kzangeli: This missing space ... can't say I'm a fan.
I will need a converting function for these, as the generic function works both for v1 and v2 ...
Not difficult, just a little annoying converting a good string into a bad one.

src/lib/serviceRoutinesV2/postEntities.cpp
kzangeli: As I said, don't worry, I have it under control.

CHANGES_NEXT_RELEASE
kzangeli: Little confused here about the versions ...
We are creating 0.24.0 now, so all fixes between 0.23.0 and 0.24.0 should be included, right?
This bug has been there since the introduction of v2.

test/functionalTest/cases/1087_error_handling/error_messages_for_invalid_service_path_in_v2.test
kzangeli: Bad Request -> BadRequest fixed in 6302f42

kzangeli: Documentation fixed in 1139e1b

CHANGES_NEXT_RELEASE
fgalan: If it has been there since the introducion of v2, then it is ok to have it in the CNR file. I though that is was related only  with "POST /v2/entities".

NTC

kzangeli: Any error with 'invalid tenant', 'invalid service path', etc   would produce this 'ex bug'

test/functionalTest/cases/1087_error_handling/error_messages_for_invalid_service_path_in_v2.test
fgalan: Using "BadRequest", "ParserError", etc. is based on some spec (which I don't remember) which uses the same style. @jmcanterafonseca knows better, he could provide the details.

fgalan:
LGTM

PR #1091, opened 2015-07-22
Fixed update context request json

joaonvfigueiredo:
Fixed what seemed like a copy-paste error, where the first example for an update context request was actually the JSON of the response.


raconte:
LGTM

PR #1095, opened 2015-07-22
�fixed issue 1092

kzangeli:
Put the apiVersion check a little earlier (that's where we know whether we're in v1 no v2).
Fixed a few cosmetic stuff as well ...


src/lib/rest/OrionError.cpp
```diff
@@ -98,6 +98,11 @@ std::string OrionError::render(ConnectionInfo* ciP, const std::string& _indent)
   //
```
fgalan: CNR entry missing.

kzangeli: Fixed in 1172c48

fgalan:
LGTM

PR #1097, opened 2015-07-23
Setting httpStatusCode to 201 Created only if not set

kzangeli:
Setting httpStatusCode and Location HTTP header only if all is OK, avoiding to overwrite any error in httpStatusCode.

### DIFF
--- a/src/lib/serviceRoutinesV2/postEntities.cpp
+++ b/src/lib/serviceRoutinesV2/postEntities.cpp
@@ -70,11 +70,14 @@ std::string postEntities
 
 
   // 03. Prepare HTTP headers
-  std::string location = "/v2/entities/" + eP->id;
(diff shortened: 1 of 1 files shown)


src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -70,11 +70,14 @@ std::string postEntities
```
fgalan: Maybe line in CNR is missing? Not sure... maybe not.

kzangeli: No, nothing missing.
As you know, we don't really touch ciP->httpStatusCode inside mongoBackend.
This is just like a security measuremente. No known bugs are fixed + 'make test' worked just fine ...

NTC

fgalan:
LGTM

PR #1098, opened 2015-07-23
Neclab/spec structure

tobiasjacobs:
I added placeholders for further sections in the NGSI introduction, so that the sections can then be easier be worked on in independent pull requests.

I furthermore propose to describe in the NGSI specs also different roles (context provider, context producer, etc.)


fgalan:
LGTM

PR #1081, opened 2015-07-17
APIv2 Apiary spec: complete regs and subs operations + updateContext + other fixes

fgalan:
RESTfull-based operation for manage registration and operations has been added (PATCH/GET/DELETE), along with updateContext RPC-based operation. Apart from that, the following changes/fixed have been implemented:

* RESTfull-based subscription and registration doesn't allow a list of entities.
* URL schema `http` is the only mandatory for notifications, others are taken into account but in an optional way
* `id` for exact entity ID matching, `idPattern` for entity ID patterns.
* Removing of `options` boilerplate, progressing `duration` to top level.
* Fixed `conditions` description.
* Removing `attrs` from `subject` (given that now is part of `conditions`).

PR #1099, opened 2015-07-23
FIX issues not actually included in PR #1081

fgalan:
Some fixes that weren't actually included in PR #1082 (I commited a version of the file with pending changes to save in my IDE, sorry...).


kzangeli:
LGTM

Issue #983, opened 2015-06-11
API v2: Strictly append attributes (by entity ID) 


fgalan:
Implement the following operation in in v2:

```
POST /v2/entities/<entityId>?op=append
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
Implemented in PR #1124

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
LGTM
See 
```
test/acceptance/behave/components/ngsiv2/entities/update_or_append_entity_attributes/update_or_append_entity_attributes_part4.feature
```

Issue #981, opened 2015-06-11
API v2: Append or update attributes (by entity ID)


fgalan:
Implement the following operation in in v2:

```
POST /v2/entities/<entityId>
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
Implemented in PR #1124

PR #1124, opened 2015-08-06
Feature/981 append or update attributes

kzangeli:
Implements #981 and #983 

Implementation of POST /v2/entities/<entity::id>


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1985,7 +1985,7 @@ void searchContextProviders
 * 1. Preconditions
 * 2. Get the complete list of entities from mongo
 */
```
fgalan: Given that return type has changed in this function (void -> int), the function preamble comment should be edit to explain what is returned.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2404,7 +2416,7 @@ void processContextElement
     /* All the attributes existing in the request are added to the response with 'found' set to false
      * in the of UPDATE/DELETE and true in the case of APPEND
      */
-    bool foundValue = (strcasecmp(action.c_str(), "append") == 0);
+    bool foundValue = (strcasecmp(action.c_str(), "append") == 0) || (strcasecmp(action.c_str(), "appendonly") == 0);
```
fgalan: As side effect, "appendonly" becomes a new actionType for the standard operation updateContext (in NGSI v1). Have we take this into consideration?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2140,6 +2140,17 @@ void processContextElement
 
   while (cursor->more())
   {
+    if (strcasecmp(action.c_str(), "appendonly") == 0)
```
fgalan: Should the attribute be checked also? I mean, from my point of view, if the entity is E and have attributes A, B, C an updateContext on E-D is a legal "appendonly", given that E doesn't have the D attribute. However, this if clause precludes that usage...

CHANGES_NEXT_RELEASE
```diff
@@ -6,3 +6,4 @@ Add: httpsPrepare.sh script to contextBroker-tests RPM (Issue #2)
 Fix:  Fixed a bug about generic error handling for API version 2 (Issue #1087)
 Fix:  Fixed a bug about error handling of invalid Service-Path for API version 2 (Issue #1092)
 Add:  Implemented a RAM-cache for patterned ngsi10 subscriptions (Issue #1048)
+Add:  POST /v2/entities/<entityId> entity creation under version 2 (Issue #981)
```
fgalan: Actually, it is not entity creating but append/update attributes in an existing entity, isn't it?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: True, what if we check the API version, and let v1 function as before?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 2b2c559

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: In fact, the operation "POST /v2/entities/<entityId>" assumes that the entity already exists... (in order to create a new entity, "POST /v2/entities" is used).

kzangeli: I don't see it really like that ... We're talking entities here, and either it exists of it doesn't, regardless of its attributes.
However, this is a definition that must be decided upon.

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
```diff
@@ -0,0 +1,171 @@
...
+
+
+echo "06. Make sure it worked"
+echo "======================="
+orionCurl --url /v2/entities/E1 --json
+echo
+echo
+
+
+echo "07. Try some other verb"
```
fgalan: I'd remove this step, as it will fail when we implement #987

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 4e83243

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
```diff
@@ -0,0 +1,171 @@
...
+
+--SHELL--
+
+#
+# 01. Create E1
+# 02. Make sure E1 exists
+# 03. Try to create E1 again, using POST /v2/entities/E1?op=append
+# 04. Make sure it didn't work
+# 05. Now update E1, using POST /v2/entities/E1
+# 06. Make sure it worked
```
fgalan: Asumming that addding new attributes to existing entities is legal, I'd suggest to add the following steps:

* POST /v2/entities/E1 atribute 'attr2' (and check it is there)
* POST /v2/entities/E1?op=append atribute 'attr3' (and check it is there)

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
kzangeli: Why remove it?
It will fail, and it will simply have to be modified ...

CHANGES_NEXT_RELEASE
```diff
@@ -6,3 +6,4 @@ Add: httpsPrepare.sh script to contextBroker-tests RPM (Issue #2)
 Fix:  Fixed a bug about generic error handling for API version 2 (Issue #1087)
 Fix:  Fixed a bug about error handling of invalid Service-Path for API version 2 (Issue #1092)
 Add:  Implemented a RAM-cache for patterned ngsi10 subscriptions (Issue #1048)
+Add:  POST /v2/entities/<entityId> - append or update attributes (by entity ID) under version 2 (Issue #981)
```
fgalan: In addition, I understand that the PR also implements #983 (i.e. ?op=append). In that case, it also should be mentioned in the CNR (maybe in a separate line).

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Or expand that v1 functionality for the same price to make happy all our existing users that have requested it :).

In the later case, we should mention it in the CNR as a v1 feature and also modify .md documentation to explain the new funcionality (I can do the .md part).

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
fgalan: Appart of the "pragmatic" reason of avoiding modifying a step that we know for sure that will fail in a soon future, there are a couple of additional reasons:

* It is not part of the functionality we want to test. Following the same reasoning, why don't test "other URL param"? Or "other (not valid) payload"?
* It is not a common style in other test. I mean, the same step could be applied to many of the .test that are focused on testing a particular verb + URL. E.g. why don't add this "other verb" test to "POST /v2/entities" .test?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1984,8 +1984,13 @@ void searchContextProviders
 * 0. Preparations
 * 1. Preconditions
 * 2. Get the complete list of entities from mongo
+*
+* RETURN VALUE
```
fgalan: If the return value is just to distinguish between "success" or "fail", maybe is better to use a bool for this? I think the style in mongoBackend is using bool for sucess/fail returns.

src/lib/mongoBackend/mongoUpdateContext.cpp
```diff
@@ -68,16 +68,26 @@ HttpStatusCode mongoUpdateContext
...
+          //
+          int r = processContextElement(requestP->contextElementVector.get(ix),
+                                        responseP,
+                                        requestP->updateActionType.get(),
+                                        tenant,
+                                        servicePathV,
+                                        uriParams,
+                                        xauthToken,
+                                        caller);
+          if ((r == 1) && (strcasecmp(requestP->updateActionType.get().c_str(), "appendonly") == 0))
```
fgalan: Is this logic compatible with a "direct" usage of the appendonly functionality in v1 standar operations? The `return SccNoContent` makes me wonder if this piece of code relies to much in the caller (v2) serviceRoutine.

But probalby this piece of code is going to change after our last chat at skype, so let's hold on this comment.

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
kzangeli: Yeah, we should have all these 'bad uri param', 'bad verb', etc in all of our functests really.
However, I can see you don't like this step too much :-), so I'll just remove it.

kzangeli: Fixed in ca58a7e

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 4b1bf4b

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I'm not a big fan about using bool, except for functions that are *truly* boolean, e.g.

    bool isBlack(int color) { return (color == 0)? true : false; }

However, if you really want it to be bool, no probs.

fgalan: I think it would be a good idea in this case, to keep the same style in the same library (mongoBackend). If the library were using most of the cases 0/1, then I would prefer int.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: What if we do this (expand to v1) in a separate PR?
That and more functests ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 57eae82

src/lib/mongoBackend/mongoUpdateContext.cpp
kzangeli: Yes, processContextElement goes back to 'void' and all this disappears

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 7ce2555   (back to void function)

src/lib/mongoBackend/mongoUpdateContext.cpp
kzangeli: Fixed in 7ce2555

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
kzangeli: So, let's do this, but in another PR.
Issue to be created about this ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2466,4 +2500,11 @@ void processContextElement
       responseP->contextElementResponseVector.push_back(cerP);
     }
   }
+
+  if (attributeAlreadyExistsError == true)
+  {
+    buildGeneralErrorResponse(ceP, NULL, responseP, SccBadRequest, "one or more of the attributes in the request already exist");
```
fgalan: What about providing the list of attributes with problems?

src/lib/mongoBackend/mongoUpdateContext.cpp
```diff
@@ -68,8 +68,13 @@ HttpStatusCode mongoUpdateContext
     else
     {
         /* Process each ContextElement */
-        for (unsigned int ix= 0; ix < requestP->contextElementVector.size(); ++ix)
+        for (unsigned int ix = 0; ix < requestP->contextElementVector.size(); ++ix)
         {
+          //
```
fgalan: Obsolete comment?

src/lib/ngsi/ContextAttribute.h
```diff
@@ -63,6 +63,7 @@ typedef struct ContextAttribute
   bool                       found;                   // Not part of NGSI, used internally for CPr forwarding functionality (update case)
                                                       // It means attribute found either locally or remotely in providing application
 
+  bool                       skip;                    // In case of 'op=append' and the attribute already exists
```
fgalan: Maybe is better:

"Internal usage in mongoBackend: in case of 'op=append' and the attribute already exists"

test/functionalTest/cases/0981_POST_v2_attributes/append_only_error_if_exists.test
kzangeli: Issue: 1127

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: See issue #1127

src/lib/ngsi/ContextAttribute.h
kzangeli: sure

src/lib/mongoBackend/mongoUpdateContext.cpp
kzangeli: Nope, I find it 'nice' to start and end a longer comment like that. Surely, you must have seen this before ... ?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: We could do that, I guess.
Do we want this?

kzangeli: Fixed in 692ac8d

src/lib/ngsi/ContextAttribute.h
kzangeli: Fixed in 692ac8d

src/lib/mongoBackend/mongoUpdateContext.cpp
kzangeli: Misunderstanding - fixed in 1c46a93

fgalan:
LGTM

PR #1130, opened 2015-08-07
ADD FIWARE::StringQuery filtering

fgalan:
Core filtering functionality at mogoBackend for q= NGSIv2 filter.


src/lib/common/Format.cpp
```diff
@@ -70,6 +71,19 @@ Format stringToFormat(const std::string& s)
...
+*/
+bool isFloat(const std::string& s)
+{
+  // Thanks to http://stackoverflow.com/questions/447206/c-isfloat-function
+  std::istringstream iss(s);
+  float f;
+  iss >> std::noskipws >> f; // noskipws considers leading whitespace invalid
+  // Check the entire string was consumed and if either failbit or badbit is set
+  return iss.eof() && !iss.fail();
+}
```
kzangeli: Not one single empty line in this function ...
It is very short, but I still get dizzy watching it ... :-D

The first comment I would move to the function-comment.
After aligned local variables, am empty line, please.

If you want you could describe what the function does in the comment-header and have the function itself free of comments (a matter of taste).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1010,6 +1012,348 @@ static bool processAreaScope(Scope* scoP, BSONObj &areaQuery)
...
 
+/* ****************************************************************************
+*
+* qStringFilters -
+*
+* FIXME P9: this function currently abuses of the std::string() wrapping for char* in order to
+* make comparisons work. An smarter solution could be developed.
+*
+*/
+static void qStringFilters(std::string& in, std::vector<BSONObj> &filters)
```
kzangeli: Why not put qStringFilters in its own file (qStringFilters,cpp/h)  ?
It is true that you can 'hide' the function in MongoGlobal (making it static) as it is only used in MongoGlobal ...
However, it wouldn't be a bad thing to have it in its own file, I think.

This is just a matter of taste, really.

CHANGES_NEXT_RELEASE
```diff
@@ -5,3 +5,4 @@ Add: monit_log_processing.py script to contextBroker RPM (Issue #1083)
 Add: httpsPrepare.sh script to contextBroker-tests RPM (Issue #2)
 Fix:  Fixed a bug about generic error handling for API version 2 (Issue #1087)
 Fix:  Fixed a bug about error handling of invalid Service-Path for API version 2 (Issue #1092)
+Add:  FIWARE::StringQuery scope, implementing filtering capabilities (equal, unqual, greater/lesser than, ranges, existence) (Issue #864)
```
kzangeli: I would change *greater/lesser than* for *greater/less than*.

Why?
Well, we use "less than" for expressing "<".
Lesser than is correct English but it isn't used for expressing this concept.

doc/manuals/user/filtering.md
```diff
@@ -117,3 +117,34 @@ described in detail in [the following section](geolocation.md#geo-located-querie
...
+                      "type": "FIWARE::StringFilter",
+                      "value": "temperature<24;humidity==75..90;status=running"
+                  }
+              ]
+          }
+      }
+      ...
+
+This scope allows to express filtering conditions such as equality, unequality,
+greater/lesser than, range or existence.
```
kzangeli: greater/lesser again

src/lib/common/Format.cpp
fgalan: Ok. In addition, I'd probably become this into a inline function (for short functions it use to be a good idea).

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: It is too "tied" with the query functionality to live in a file of it own. What probably is a better idea is to package all the functions related with query (basically, the functions that entityQuery() calls in a separate file). In general, MongoGlobal is too large and there are functions groups that could be moved outside.

Do you think is a good moment to do it know? Any potential conflict in another queued PR?

kzangeli: For something that big, it's better to wait.
We already have an issue about 'mongoBackend refactoring', right?

NTC (except for issue-creation if necessary)

src/lib/common/Format.cpp
kzangeli: Sounds good

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1010,6 +1012,348 @@ static bool processAreaScope(Scope* scoP, BSONObj &areaQuery)
...
+    char*               right;
+    char*               rangeFrom = (char*) "";
+    char*               rangeTo   = (char*) "";
+    std::vector<char*>  valVector;
+
+    s = wsStrip(s);
+
+#if 0
+    printf("statement %d: %s\n", statementNo, s);
+#endif
```
kzangeli: These two '#if 0/#endif's ...
If you thing this trace is useful, use LM_T, if not, just remove

fgalan: This is a way of preserving your original implementation. The idea is to keep them until feature/967 gets finished, then remove them as las task before merging feature/967 into develop.

kzangeli: ok, NTC then :-)

CHANGES_NEXT_RELEASE
fgalan: Fixed in ee6eeb1

doc/manuals/user/filtering.md
fgalan: Fixed in ee6eeb1

src/lib/common/Format.cpp
fgalan: Fixed in ee6eeb1 (although I haven't convert into inline at the end)

kzangeli:
LGTM

PR #1132, opened 2015-08-07
Feature/0987 put attributes

kzangeli:
Implements #987

Added PUT /v2/entities/<entity-id>

I made a few changes to unit tests - replacing a few "200 OK"s for "400 Not Found"s.
@fgalan please take a very close look ...


src/lib/serviceRoutines/badVerbAllFive.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "rest/restReply.h"
+#include "serviceRoutines/badVerbAllFive.h"
+
+
+
+/* ****************************************************************************
+*
+* badVerbAllFive - 
+*/
+std::string badVerbAllFive
```
kzangeli: I added this function, thinking that I would need it ...
I didn't need it as this was a PUT and PATCH is still missing.
However, now that I have it, I leave it. Soon enough we will need it ...

NTC

test/functionalTest/cases/0987_put_attributes/put_entity_error.test
```diff
@@ -0,0 +1,153 @@
...
+
+echo "04. Make sure E1 is as created"
+echo "=============================="
+orionCurl --url /v2/entities/E1 --json
+echo
+echo
+
+
+echo "05. Modify E1"
+echo "============="
```
fgalan: Not sure if we are "capturing" the nature of this operation. #987 is about replacing all atributes in the entity by a different set. Thus, I will not use the same (attr1) in the update. We can do the following:

* Create the entity with: attr1, attr2
* Update (replace) with: attr3, attr4

We could even do a test (in a separate .test) testing the overlaping, i.e.:


* Create the entity with: attr1, attr2
* Update (replace) with: attr2, attr3

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2460,6 +2487,14 @@ void processContextElement
       searchContextProviders(tenant, servicePathV, *enP, ceP->contextAttributeVector, cerP);
       cerP->statusCode.fill(SccOk);
       responseP->contextElementResponseVector.push_back(cerP);
+
+      //
+      // If no context providers found, then the UPDATE was simply for a non-found entity and an error should be returned
+      //
+      if (forwardsPending(responseP) == false)
```
fgalan: I understand that this is the change in mongoBackend that is causing that some 200 OK test pass to 404 Not found. I'd need some more of context of why this is needed to make an opinion...

kzangeli: Yes, this is the change.
The 'why' is pretty clear ...
This is an UPDATE => could be forwarded to context providers.
If forwarded (found in ngsi9 registration) then it cannot be a 404 Not Found, but if there are no forwards to be made (and nothing local found) then we have a 404 to report back to the caller.
I believe this was an old bug and it is now fixed ...

test/functionalTest/cases/0987_put_attributes/put_entity_error.test
kzangeli: True, guess I went too fast ...
I will add a new functest (even better - two functests) for this

kzangeli: And not only functests ... this doesn't work. The broker doesn't accept 'unknown' attributes, as it seems. A bigger change will be necessary in mongoBackend, I'm afraid ... :-(
Also, we need to not break API V1 at the same time. Might get complicated ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: I understand that at the end we disabled v2 forwarding in the upper layer (serviceRoutine), thus this check probably is not needed at mongoBackend and we could go back to the situation in which unit test worked without modifications.

test/functionalTest/cases/0987_put_attributes/put_entity_partially_replace_attributes.test
```diff
@@ -0,0 +1,117 @@
...
+{
+    "A2": 2.0, 
+    "A3": 3.0, 
+    "id": "E1"
+}
+
+
+--TEARDOWN--
+brokerStop CB
+#dbDrop CB
```
fgalan: The `#` has to be removed.

test/functionalTest/cases/0987_put_attributes/put_entity_error.test
kzangeli: Fixed in f32b106

test/functionalTest/cases/0987_put_attributes/put_entity_partially_replace_attributes.test
kzangeli: Fixed in a0ff69b

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Not so sure about this.
That way, mongoBackend returns OK when really we are at case of a NOT FOUND ...
Are the unittests correct?
I would say, lets take a good look at the unittests. If they are 100% ok the way they were, then lets go back.
But if not, then lets keep this the way it is.

fgalan: Ok, let's have a look and sync by skype.

test/unittests/mongoBackend/mongoContextProvidersUpdate_test.cpp
```diff
@@ -734,6 +732,8 @@ TEST(mongoContextProvidersUpdateRequest, noPatternNoAttribute)
   /* Invoke the function in mongoBackend library */
   ms = mongoUpdateContext(&req, &res, "", servicePathVector, uriParams, "");
 
+  res.present("TEST: ");
```
fgalan: Leftover

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Reviewed. NTC

test/unittests/mongoBackend/mongoContextProvidersUpdate_test.cpp
kzangeli: Fixed in 20c81d9

fgalan:

kzangeli:
LGTM

PR #1134, opened 2015-08-07
ADD /op/query, /op/update /op/discover, /op/register

fgalan:
FIX RESTful subscriptions and registrations to allow list of entities

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -225,6 +225,8 @@ In the case of being present, the error payload is JSON object including the fol
 
 * `error` (mandatory): a textual description of the error.
 * `description` (optional): additional information about the error.
+* `affectedItems` (optional): a list of elements affected by the error. Depending on the operation, it may
+  refer to entities, registrations or subscriptions.
 
 Error list (HTTP response code in paranthesis):
 
@@ -235,6 +237,8 @@ Error list (HTTP response code in paranthesis):
   in the request. This typically the case of requesting an entity with not enough information and the
   solution is to enhance entity identification adding more information, e.g. adding entity type
   and/or service path.
+
+# Group NGSIv2
 
 ## API Entry Point [/v2]
 
@@ -1158,20 +1162,19 @@ Response code:
 A subscription is represented by a JSON object with the following fields:
 
 * `id`: Subscription unique identifier. Automatically created at creation time.
-* `subject` : It s an object that describes the subject of the subscription.
-* `notification` : It is an object that describes the notification received by the subscriber.
+* `subject`: It is an object that describes the subject of the subscription.
+* `notification`: It is an object that describes the notification received by the subscriber.
 * `duration`: Duration of the subscription in ISO8601 format. Infinite if not specified.
 
 A `subject` contains the following subfields:
 
-* `id` or `idPattern` : Id or pattern of the affected entities (optional). Both cannot be used at the same time.
-* `type` : Type of the affected entities (optional).
+* `entities`: A list of objects, each one composed of the following subfields (`id`/`idPattern` or `type` must be present):
+ * `id` or `idPattern`: Id or pattern of the affected entities (optional). Both cannot be used at the same time.
+ * `type`: Type of the affected entities (optional).
 * `condition`: Condition that will trigger the notification. It can have two optional properties:
  * `attributes`: array of attribute names
  * `expression`: an expression composed of `q`, `geometry` and `coords` (see "List entities" operation above
-about this field):
-
-`id`/`idPattern` or `type` must be present.
+   about this field).
 
 A `notification` object contains the following subfields:
 
@@ -1206,9 +1209,13 @@ Response:
         [
             {
                 "id": "abcdefg",
-                "subject": {
-                    "id": "Bcn_Welt",
-                    "type": "Room",
+                "subject": {                    
+                    "entities": [
+                        {
+                            "id": "Bcn_Welt",
+                            "type": "Room"
+                        }
+                    ],
                     "condition": {
                        "attrs:" [ "temperature " ],
                        "expression:" {
@@ -1239,7 +1246,11 @@ Response:
 
         {
             "subject": {
-                "type": "Room",
+                "entities": [
+                    {
+                        "type": "Room"
+                    }
+                ],
                 "condition": {
                     "attributes:" [ "temperature " ],
                     "expression:" {
@@ -1278,7 +1289,11 @@ Response:
         {
             "id": "abcdef",
             "subject": {
-                "type": "Room",
+                "entities": [
+                    {
+                        "type": "Room"
+                    }
+                ],
                 "condition": {
                     "attributes:" [ "temperature " ],
                     "expression:" {
@@ -1343,12 +1358,10 @@ corresponds to the URL of the provider service.
 
 A `subject` contains the following subfields:
 
-* `id` or `idPattern` : Id or pattern of the affected entities (optional). Both cannot be used at the same time.
-* `entity` : Id or pattern id which identifies the affected entities (optional).
-* `type` : Type of the affected entities (optional).
+* `entities`: A list of objects, each one composed of the following subfields (`id`/`idPattern` or type must be present):
+ * `id` or `idPattern`: Id or pattern of the affected entities (optional). Both cannot be used at the same time.
+ * `type`: Type of the affected entities (optional).
 * `attributes`: List of attributes to be provided (if not specified it would mean all).
-
-`id`/`idPattern` or `type` must be present.
 
 ### List registrations [GET /v2/registrations]
 
@@ -1365,8 +1378,12 @@ Response:
             {
                 "id": "abcdefg",
                 "subject": {
-                    "entity": "Bcn_Welt",
-                    "type": "Room",
+                    "entities": [
+                        {
+                            "id": "Bcn_Welt",
+                            "type": "Room"
+                        }
+                    ],
                     "attributes": [
                         "temperature"
                     ]
@@ -1391,7 +1408,11 @@ Response:
 
         {
             "subject": {
-                "type": "Room",
+                "entities": [
+                    {
+                        "type": "Room"
+                    }
+                ],
                 "attributes": [
                       "humidity"
                 ]
@@ -1423,7 +1444,11 @@ Response:
         {
             "id": "abcde",
             "subject": {
-                "type": "Room",
+                "entities": [
+                    {
+                        "type": "Room"
+                    }
+                ],
                 "attributes": [
                       "humidity"
                 ]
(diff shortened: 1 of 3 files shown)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1158,20 +1158,29 @@ Response code:
...
 * `id` or `idPattern` : Id or pattern of the affected entities (optional). Both cannot be used at the same time.
 * `type` : Type of the affected entities (optional).
 * `condition`: Condition that will trigger the notification. It can have two optional properties:
  * `attributes`: array of attribute names
  * `expression`: an expression composed of `q`, `geometry` and `coords` (see "List entities" operation above
-about this field):
+about this field).
 
-`id`/`idPattern` or `type` must be present.
+A `subject` for subscribing a list of ids/patterns has the following subfields (`entities` must be present and not empty):
```
jmcanterafonseca: I think we can have a more terse specification if we allow id to be both an string and an array. Same with idPattern

(Note I was not expecting this change as part of this Pull Request)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1341,14 +1350,19 @@ A context registration is represented by a JSON object with the following fields
...
+* `type`: Type of the affected entities (optional).
 * `attributes`: List of attributes to be provided (if not specified it would mean all).
 
-`id`/`idPattern` or `type` must be present.
+A `subject` for registering a list of ids/patterns has the following subfields (`entities` must be present and not empty):
+
+* `entities`: A list of objects, each one composed of the following subfields (`id`/`idPattern` or type must be present):
+ * `id` or `idPattern`: Id or pattern of the affected entities (optional). Both cannot be used at the same time.
+ * `type`: Type of the affected entities (optional).
+* `attributes`: List of attributes to be provided (if not specified it would mean all).
```
jmcanterafonseca: I think these changes should not be part of this PR ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1466,12 +1480,46 @@ Response:
...
+## POJ RPC Operations [/v2/op]
 
-For the sake of completeness here is an enumeration of the OMA-NGSI (9 & 10) operations. These
-operations will be supported under the 'v2' resource as well.
+### Update [POST /v2/op/update]
+
+This operation allows to create, update and/or delete several entities in a single batch operation.
+The payload is an object with two properties:
+
+* `actionType`, to specify the kind of update action to do: either APPEND, APPENDONLY, UPDATE, DELETE.
```
jmcanterafonseca: I'm wondering whether FORCEAPPEND would be a better name than APPENDONLY

fgalan: What about STRICTAPPEND?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1466,12 +1480,46 @@ Response:
...
+* `entities`, an array of entities, each one specified using the JSON entity representation format (described
+  in a section above).
+
+Response:
+
+* Successful operation uses 204 No Content.
+* Errors use a non-200 and (optionally) an error payload. See subsection on Error Responses for more details.
+
```
jmcanterafonseca: I think here notifying errors can be a bit more complex ... for instance what happens if only one of the affected entities was updated and the rest not ... Are these operations expected to work as if they were transactions?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1533,18 +1581,8 @@ Response code:
...
-                {
-                    "type": "FIWARE::Filter::Location",
-                    "value": {
-                        "circle": {
-                            "centerLatitude": "40.418889",
-                            "centerLongitude": "-3.691944",
-                            "radius": "13500"
-                        }
-                    }
+                    "type": "FIWARE::...",
```
jmcanterafonseca: here the example seems to be flawed ... I think we need to be more explicit here

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1577,62 +1615,188 @@ Response code:
...
+### Register [POST /v2/op/register]
 
+This operation allows to create, update and/or delete several registrations in a single batch operation.
 The payload is an object with two properties:
 
-* `action`, to specify the kind of updateContext action to do (see OMA NGSI specification for the semantics),
-   either UPDATE, APPEND or DELETE
-* `entities`, an array of entities, each one specified using the the JSON entity representation format
-  (described in a section above).
+* `actionType`, to specify the kind of register action to do: either APPEND, UPDATE, DELETE.
```
jmcanterafonseca: In registrations, do we really need to distinguish between APPEND or UPDATE? I would prefer not to do that if possible

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1577,62 +1615,188 @@ Response code:
...
+* `registrations`, an array of registration, each one specified using the JSON registration representation
+  format (described in a section above). In the case of APPEND operation, the registration `id` must not be
+  included.
 
 Response:
 
-* Successful operation uses 201 Created.
+* Successful UPDATE or DELETE operation uses 204 No Content.
+* Successful APPEND operation uses 201 Created and returns a list of IDs, each one corresponding to the ID of the
+  element in the request payload in the same order.
```
jmcanterafonseca: where such a list of IDs would be provided in the location header?

jmcanterafonseca: I think we need to determine whether it would be a better idea to have a response status code of 200 for RPC operations

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1577,62 +1615,188 @@ Response code:
...
 
-    + Headers
+        [
+          "abcd",
+          "efgh"
+        ]
 
-            Location: /v2/entities/Bcn-Welt
+### Discover [POST /v2/op/discover]
+
```
jmcanterafonseca: Maybe we need a similar operation in order to discover subscriptions as well. Probably this could be done through /v2/subscriptions as well

jmcanterafonseca: We can have a similar operation in the RESTful interfaces /v2/registrations?q=<Query> right?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The problem with that solution is that is not valid as long as each entity id could have a different entity type, e.g.:

```
entities: [
   {
      "id": "E1",
      "type": "T1
   },
   {
      "id": "E2",
      "type": "T2"
   }
   {
      "id": "E_withoutType"
   },
]
```

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: They are indirectly needed as the closure in the scope definition of the RPC POJ operations related with subscriptions/registrations needs that the RESTful oepration for subscription/registration allow also the posibility of registering/subscribing to list of entities.

fgalan: There is one more reason (and very practical :) to keep this in the PR: it is already done. Removing this modifications to return them back in a different next PR will be an unnecesary edition overhead.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The situation is not so different that RESTful operations. For example, we have operations to stringly append N attributes to an entity and the situation in which some of the N could be appended and the some of the N cannot be appended (due to they already exists) is taken into account. The situation here is similar, but at entity level.

It is not supposed to work as a transaction but in a "do the best you can" way (in this particular case, appending some of them and reporting the error for the others).

Regarding retunring always 200 OK, in the case of error, what would happend? 200 OK + error payload?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The specific scopes to take into account in NGSIv2 (i.e. list with specifid type such as "FIWARE::StringQuery", etc) has not been "consolidated" yet. Thus, maybe a good idea is to leave this with TBD, e.g.:

 ```
{
   "type": "FIWARE::TBD",
   "value": "TBF"
}
```

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 6 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: So you mean to use the same "actionType" keyword for both creating new registrations and updating existing ones? In that case, the only way of distinguishing between both situations is the existence or not of registrationId in the payload (as in OMA NGSI registerContext).

I think we have decided by email that it is better to state explicitly the action (such as in OMA NGIS updateContext) instead of relaying in the content of the payload.

Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Regarding APPEND vs CREATE, the idea was to keep keywords similar to the ones used by updateContext, but maybe is a good idea to use different ones. I don't have a strong opinion in that really...

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 7 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Doest the location header multi-values? E.g.

```
Location: /v2/regsitrations/abcd, /v2/registrations/efgh
```

In that case, bettter to use Location that payload.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Yes, q= could also be used for "GET /v2/registrations" and "GET /v2/subscriptions" in the RESTful part of the interface. Maybe we should create two issues in github about in order not forgetting it.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After skype discussion, we will use only `entities`. The old case for only one id/idPattern will be covered with the `entities` vector having just 1 element.

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: At the end, we will STRICT_APPEND (to be checked).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After skype discussion:

* Use only 200 OK for sucess in RPC operations (i.e. not 201, not 204, etc.)
* Include an optional field `entities` in the error payload, meaning the affected entities in the operations in which that makes sense.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: In addition, include a note in the text explaining that scopes are under definition (as we have donde in similar cases).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: At the end, we will use CREATE, UPDATE, DELETE for registrations.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: From http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html, Location cannot be multivaluated. So, we will use 200 OK with a payload (as in the original proposal, but using 200 instead of 201).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: At the end, we will use APPEND_STRICT.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Issues created: https://github.com/telefonicaid/fiware-orion/issues/1145 and https://github.com/telefonicaid/fiware-orion/issues/1146

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in c2dd0bf8a510a858639732e33ead073b6ed8ff55

kzangeli:
LGTM

PR #1140, opened 2015-08-10
checking illegal characters in URI params, except for the URI param 'q'

kzangeli:
URI param names and values checked, except for the URI param 'q'


src/lib/rest/rest.cpp
```diff
@@ -193,6 +193,21 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
 
   LM_T(LmtUriParams, ("URI parameter:   %s: %s", key.c_str(), ciP->uriParam[key].c_str()));
 
+  //
+  // Now check the URI param has no invalid characters
+  // Except for the URI param 'q' that is not to be checked for invalid characters
+  //
+  if (key != "q")
```
fgalan: Not only in `q` parameter, also in the case of `value` in scopes of `type` SCOPE_TYPE_SINGLE_FILTER (or something like that, I don't remember the `#define` used right now...).

src/lib/rest/rest.cpp
```diff
@@ -193,6 +193,21 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
...
+  {
+    if (forbiddenChars(ckey) || forbiddenChars(val))
```
fgalan: This PR implement ilegal chracters checking in URI params in general. Thus, I think it deserves a line in CNR file (not tied to any specific issue).

kzangeli: Yes, I was surprised to find out we didn't have any check of illegal chars in URI params ...

src/lib/rest/rest.cpp
kzangeli: This PR is only about the q parameter, as its title suggests.
Another PR about the scopes in the payload coming soon ...

NTC

src/lib/rest/rest.cpp
kzangeli: Issue #1141 created

kzangeli: CNR fixed in 095fa6c

fgalan:
LGTM

Issue #1093, opened 2015-07-22
[BUG] Forbidden characters are allowed in create entities fields - NGSI v2


iariasleon:
Failing Fields:

* "type"
* "id"
* "attribute name"
* "attribute value"
* "attribute type"
* "attribute metadata name"
* "attribute metadata value"
* "attribute metadata type"

### Request:
```
http://<host>:<port>/v2/entities
Fiware-Service: sdsd_1
Fiware-ServicePath: /service
Accept: application/json

- payload to type field:
         {"temperature_0": "34", "type": "house;flat", "id": "room2_0"}
- payload toid field:
         {"temperature_0": "34", "id": "house;flat", "type": "room"}
- payload to attribute name field:
         {"house;flat": "34", "type": "room", "id": "room2_0"}
- payload to attribute value field:
         {"temperature": "house;flat", "type": "room", "id": "room2_0"}
- payload to attribute type field:
         {"temperature": {"type": "house;flat" "value": " 34"}, "type": "room", "id": "room2_0"}
- payload to attribute metadata name field:
         {"temperature": {"type": "data", "value": " 34", "house;flat": "danger"}, "type": "room", "id": "room2_0"}
- payload to attribute metadata value field:
         {"temperature": {"type": "data", "value": " 34", "meta": "house;flat"}, "type": "room", "id": "room2_0"}
- payload to attribute metadata type field:
         {"temperature": {"type": "data", "value": " 34", "meta": {"value":"45","type": "house;flat"}}, "type": "room", "id": "room2_0"}
```

### Response:
```
status code: 201 Created
```

### Examples in mongo
```
{ "_id" : { "id" : "room2_0", "type" : "house<flat>", "servicePath" : "/test" }, "attrNames" : [ "temperature_0" ], "attrs" : { "temperature_0" : { "type" : "", "creDate" : 1437571371, "modDate" : 1437571371, "value" : "34" } }, "creDate" : 1437571371, "modDate" : 1437571371 }
{ "_id" : { "id" : "room2_0", "type" : "house=flat", "servicePath" : "/test" }, "attrNames" : [ "temperature_0" ], "attrs" : { "temperature_0" : { "type" : "", "creDate" : 1437571372, "modDate" : 1437571372, "value" : "34" } }, "creDate" : 1437571372, "modDate" : 1437571372 }
[...]

PR #1143, opened 2015-08-10
Bug/1093 forbidden chars in many fields

kzangeli:
Implements #1093 

Added check of invalid characters to
Entity, ContextAttribute and Metadata, for API v2


test/functionalTest/cases/0000_xml_parse/xmlParsePostNotifyContext.test
```diff
@@ -52,9 +52,9 @@ payload='<?xml version="1.0" encoding="UTF-8"?>
                 <value>1110</value>
               </contextMetadata>
               <contextMetadata>
-                <name></name>
-                <type></type>
-                <value></value>
+                <name>id2</name>
```
fgalan: XML is now a deprecated feature, so modifications in XML-based .test are not very relevant. However, I wonder why this has changed... there was any bug in the XML metadata rendering?

kzangeli: These files that have changed (they had empty fields in metadata) are both input and output data for unit tests. The files used for input had to be changed, as with the new checks (the check method is the same for XML, JSONv1 and JSONv2) the unit test fails if the fields are empty.

fgalan: NTC

fgalan:
LGTM

Issue #1108, opened 2015-07-29
[BUG] if "type" or "id" in v2 is not string, return two message in several values 


iariasleon:
### Request:
```
http://<host>:<port>/v2/entities
Fiware-Service: dfgdf
Fiware-ServicePath: /ewe
Content-Type: application/json
```
#### Payload without attribute type
```
{"type": false, "id": "room2", "temperature": true}
{"type": "room", "id": false, "temperature": true}
```
#### Payload with attribute type
```
{"type": false, "id": "room2", "temperature": {"type": "nothing", "value": true}}
{"type": "room", "id":false, "temperature": {"type": "nothing", "value": true}}
```
```
      | type value        |     error message            
      -------------------------------------------------------------------------------
      | rewrewr           |     Errors found in incoming JSON buffer
      | SDFSDFSDF         |     Errors found in incoming JSON buffer
      | false             |     invalid JSON type for entity type
      | true              |     invalid JSON type for entity type
      | 34                |     invalid JSON type for entity type
      | {"a":34}          |     invalid JSON type for entity type
      | ["34", "a", 45]   |     invalid JSON type for entity type
      | null              |     invalid JSON type for entity type
```
Test: 
```
Feature: create_entities.feature
Tags:  
  -  entities_type_no_string_error  
  -  entities_type_no_string_with_type_error  
  -  entities_id_no_string_error
  - entities_id_no_string_with_type_error
Scenarios: 
   -  try to create an entity in NGSI v2 with several wrong entities_type (integer, boolean, no-string, etc)
   -  try to create an entity in NGSI v2 with several wrong entities_type with attribute type (integer, boolean, no-string, etc)
   -  try to create an entity in NGSI v2 with several wrong entities id without attribute type (integer, boolean, no-string, etc)
   -  try to create an entity in NGSI v2 with several wrong entities id with attribute type (integer, boolean, no-string, etc)
```

fgalan:
Not sure of understanding this issue... do you mean that "GET http://<host>:<port>/v2/entities" returns two values? I think we would need to discuss this by skype to understand what is going on... by the moment, this issue is scheduled to 0.25.0.

fgalan:
Fixed in https://github.com/telefonicaid/fiware-orion/pull/1149. Waiting for @iariasleon to check everything is ok and to close this issue.

iariasleon:
These scenarios are divided in two:
 - Errors found in incoming JSON buffer
 - invalid JSON type

Issue #1007, opened 2015-06-15
Normalized [previously canonical] format


fgalan:
In order to ease the programming of clients that need a uniform format for attributes, an option could be passed to the GET operations that render that information, so attribute and metadata will be using always an structure with value, type and metadata (in the case of attributes) or value and type (in the case of metadata).

E.g. non-canonical:

```
         {
            "type": "Car",
            "id": "P-9873-K",
            "colour": "red",
            "speed": {
                "value": 100,
                "type": "number",
                "accuracy": 2,
                "timestamp": {
                    "value": "2015-06-04T07:20:27.378Z",
                    "type": "date"
                }
            }
         }
```

would be:

```
         {
            "type": "Car",
            "id": "P-9873-K",
            "colour": {
              "value": "red",
              "type": ""
            }
            "speed": {
                "value": 100,
                "type": "number",
                "accuracy": {
                   "value": 2,
                   "type": "number",
                "timestamp": {
                    "value": "2015-06-04T07:20:27.378Z",
                    "type": "date"
                }
            }
         }
```

fgalan:
One possiblity so clients can specify they want the info in cannonical way is to use the `cannonical` option, e.g.:

```
GET /v2/entities?options=cannonical
```

fgalan:
This pseudo-test harness should be taken into account when implementing this issue (only the part related with canonical model):

https://gist.github.com/fgalan/d39b016be45d2591a6e7

fgalan:
Another posibility for the options parameter would be `oma`, as the structures returned are close to the one used by the OMA spec, e.g.:

```
GET /v2/entities?options=oma
```

fgalan:
Better terms has been suggested:

* *normalized* (instead of canonical)
* *compact*, for meaning the opossed to normalized.

fgalan:
The changes mentioned in https://github.com/telefonicaid/fiware-orion/issues/1259 will implement this in a different way. Closing.

PR #1149, opened 2015-08-11
Bug/1108 non string fields

kzangeli:
Implements #1108 

Better error handling for non-string fields.


test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
```diff
@@ -0,0 +1,194 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: I understand that the restriction is in entity id and type, so probably we can do this renaming:

1108_non_string_fields/non_string_fields.test

to

1108_entity_id_and_type_non_string_fields/entity_id_and_type_non_string_fields.test

or something like that...

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
```diff
@@ -0,0 +1,194 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+POST /v2/entities with invalid service-path
```
fgalan: I seems that this NAME doesn't correspond with the purpose of the test...

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
```diff
@@ -0,0 +1,194 @@
...
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. POST /v2/entities with non-string in entity::id
+# 02. POST /v2/entities with non-string in entity::type
+# 03. POST /v2/entities with non-string in attribute::type
+# 04. POST /v2/entities with non-string in metadata::type
+# 05. POST /v2/entities with JSON object in metadata::value
```
fgalan: This is going to change in the future when we allow JSON objects/vector for metadata value. Thus, I'd add a FIXME here about this step should be removed when time comes.

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
fgalan: Looking closer to the .test, it seems that is not only a matter of entity::id and entity::type. Maybe my previous comment doesn't make sense, or the suggested name is not the right one.

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
```diff
@@ -0,0 +1,194 @@
...
+POST /v2/entities with invalid service-path
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. POST /v2/entities with non-string in entity::id
```
fgalan: I think that attribute::name and metadata::name should also should have similar restrictions. They are similar in nature to an entity::id, aren't they?

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
```diff
@@ -0,0 +1,64 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+POST /v2/entities with invalid service-path
```
fgalan: Wrong NAME?

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
```diff
@@ -0,0 +1,64 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 05. POST /v2/entities with JSON object in metadata::value
```
fgalan: Step should start at 01

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
fgalan: Actually this test seems to be a sub-case of the previous one... Maybe I'm missing something.

CHANGES_NEXT_RELEASE
```diff
@@ -15,3 +15,4 @@ Fix:  Checking for forbidden chars in many fields of JDON payload for API versio
 Fix:  Bug fix for receiving compound metadata values (which is not supported) (Issue #1110)
 Add:  GET /v2/entities/<entityId>/attrs/<attrName>/value - get a single attribute value (Issue #997)
 Add:  URI parameters id, idPattern and type for GET /v2/entities (Issues #952 #953 #969)
+Fix:  error handling for non-string entity::id/type (Issue #1108)
```
fgalan: Looking to the test, it seems to apply to additional fields.

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
kzangeli: In general, yes, of course.
However, the test is a part of another test (its name is non_string_fields_05 - meaning "part 05 of non_string_field).
I have used this for some time now ...

A test fails and I extract only the failing part, in a new functest.
Now that I have the new functest I can either remove it or keep it. I usually keep it

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
kzangeli: Yes, it is a sub-case (part 05) of non_string_fields_05.test

The name should change, of course. The name may no longer make sense.

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Cut 'n paste error ...

CHANGES_NEXT_RELEASE
kzangeli: True, it got a little bigger.
What's the best thing to do, add to the issue?  Create additional issues?

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Yeah, it's more like fields that should be strings but aren't

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Yes, I don't why I didn't add name for attribute/metadata ...
entity::id is just that ...   I don't remember.
Will add that check + functest covering it

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Sure, if you insist ... :-)
I am pretty sure we will notice the error in the test before we see the FIXME ;-)

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
fgalan: I understand the value of "partical .test" during the debug steps, but once the functionality is ready at PR stage, I think is better to remove them, in order to keep the test base clean and maintanable.

CHANGES_NEXT_RELEASE
fgalan: The simplest way would be to add the other fields covered by the checking in this same line.

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
fgalan: You are right, but once the error happens in the future it is better to have a FIXME in the .test explaining why is failing to help us to realize about it (now it seems very simple to realize, but who knows in a couple of months...).

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 6a46ebd

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Now I remember why there is no test for attribute::name and metadata::name.
In API v2, attributes and metadatas aren't like in v1:

    V2:
    "attr1" : { "type": "T1", "value": "" }

    V1:
    "attributes":
    {
        "name": "xxx", ...
    }

There is no need to check that attribute:name and metadata::name are strings. If they aren't, there will be a general JSON parser error (and also, no way for me to realize what they intended to do)

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Fixed in c787dda

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
kzangeli: Fixed in d7aafbf

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
kzangeli: ok

kzangeli: Fixed in 57de20f

test/functionalTest/cases/1108_non_string_fields/non_string_fields_05.test
kzangeli: Fixed in 57de20f

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
fgalan: NTC

test/functionalTest/cases/1108_non_string_fields/non_string_fields.test
fgalan: NTC

(At lest in the scope of this PR... attribute:name and metadata:name may need checking in the normalized mode https://github.com/telefonicaid/fiware-orion/issues/1007, when we implemented that)

fgalan:
LGTM

PR #1150, opened 2015-08-11
Task/no illegal char for FIWARE::StringQuery scope in payload

kzangeli:
Not checking for illegal characters in payload inside 'scope-values' if their scope-type is FIWARE::StringQuery


fgalan:
LGTM

PR #1152, opened 2015-08-12
FIX metadata not update bug

### LINKED ISSUE #1113: If attribute value is the same, metadatas are not updated
Part of a functional test to detect this bug:

    echo "01. POST /v2/entities with metadata 'm1' = 1"
    echo "============================================"
    payload='{
      "type":"T1",
      "id": "E1",
      "temp": {
        "value": "1",
        "m1": 1
      }
    }'
    orionCurl --url /v2/entities --payload "$payload" --json
    echo
    echo
    
    
    echo "02. POST /v2/entities with metadata 'm1' = 2"
    echo "============================================"
    payload='{
      "type":"T1",
      "id": "E1",
      "temp": {
        "value": "1",
        "m1": 2
      }
    }'
    orionCurl --url /v2/entities --payload "$payload" --json
    echo
    echo

Mongo dump after running this:

    {
      "_id" : {
        "id" : "E1",
        "type" : "T1",
        "servicePath" : "/" 
      },
      "attrNames" : [ "temp" ],
      "attrs" : {
        "temp" : { 
          "type" : "",
          "creDate" : 1438602795,
          "modDate" : 1438602795,
          "value" : "1",
          "md" : [
           {
              "name" : "m1",
              "value" : 1
            }
          ]
        }
      },
      "creDate" : 1438602795, "modDate" : 1438602795
    }

The value of metadata "m1" should be '2', not '1'.
If I change the value of the attribute "temp" in the second call, then the metadata is updated. Same value for the attribute, no update for the metadata ...

Seems like a bug to me ...

fgalan:
Fixes #1113

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -289,6 +289,46 @@ static bool matchMetadata(BSONObj& md1, BSONObj& md2)
     STR_FIELD(md1, ENT_ATTRS_MD_NAME) == STR_FIELD(md2, ENT_ATTRS_MD_NAME);
 }
 
+/* ****************************************************************************
+*
+* equalMetadataValues -
+*
+*/
+static bool equalMetadataValues(BSONObj& md1, BSONObj& md2)
+{
+
+  if (md1.getField(ENT_ATTRS_MD_VALUE).type() != md2.getField(ENT_ATTRS_MD_VALUE).type())
+  {
+    return false;
+  }
+
+  switch (md1.getField(ENT_ATTRS_MD_VALUE).type())
+  {
+    /* FIXME not yet
+    case Object:
+      ...
+      break;
+
+    case Array:
+      ...
+      break;
+    */
+
+    case NumberDouble:
+      return md1.getField(ENT_ATTRS_MD_VALUE).Number() == md2.getField(ENT_ATTRS_MD_VALUE).Number();
+
+    case Bool:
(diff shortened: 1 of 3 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -289,6 +289,46 @@ static bool matchMetadata(BSONObj& md1, BSONObj& md2)
...
+      return md1.getField(ENT_ATTRS_MD_VALUE).Number() == md2.getField(ENT_ATTRS_MD_VALUE).Number();
+
+    case Bool:
+      return md1.getBoolField(ENT_ATTRS_MD_VALUE) == md2.getBoolField(ENT_ATTRS_MD_VALUE);
+
+    case String:
+      return STR_FIELD(md1, ENT_ATTRS_MD_VALUE) == STR_FIELD(md2, ENT_ATTRS_MD_VALUE);
+
+    default:
+      LM_E(("Runtime Error (unknown metadata value type in DB: %d)", md1.getField(ENT_ATTRS_MD_VALUE).type()));
```
kzangeli: %d ...  Is it possible to convert this to a string?

fgalan: Yes, but I think it doesn't worth the effort. This error is supossed not to happen and, even in the rare case it happen, you can check the BSON reference to know to which type it refers.

NTC

kzangeli: ok

NTC

PR #1154, opened 2015-08-12
URI param 'q' to Scope for GET /v2/entities

kzangeli:
GET /v2/entities?q=XXX,   XXX moved to 'scope in payload for postQueryContext'


fgalan:
LGTM

Issue #999, opened 2015-06-12
API v2: Update single attribute value (by entity ID)


fgalan:
Implement the following operation in in v2:

```
PUT /v2/entities/<entityId>/attrs/<attrName>/value
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
Implemented in PR #1160

Issue #991, opened 2015-06-11
API v2: Update single attribute (by entity ID)


fgalan:
Implement the following operation in in v2:

```
PUT /v2/entities/<entityId>/attrs/<attrName>
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
Implemented in PR #1160

kzangeli:
By mistake, issue #999 was implemented instead of #991 in PR #1160

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/issues/1165

iariasleon:
LGTM 

See:
```
test/acceptance/behave/components/ngsiv2/attributes/update_attribute_data/update_attribute_data.feature
test/acceptance/behave/components/ngsiv2/attributes/update_attribute_data/update_attribute_data_part2.feature
test/acceptance/behave/components/ngsiv2/attributes/update_attribute_data/update_attribute_data_part3.feature
```

PR #1160, opened 2015-08-12
Feature/999 update single attr value by eid

kzangeli:
Implements #999

PUT /v2/entities/\<entityId\>/attrs/\<attrName\>/value


test/functionalTest/cases/0991_update_single_attr_by_eid/update_single_attr_by_eid.test
```diff
@@ -0,0 +1,337 @@
...
+--NAME--
+PUT Individual Entity Attribute
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
```
fgalan: I understand all these steps are related with changing attribute value. Some test to check updating attribute NGSI type and metadata should be also added (but in a different .test).

In addition, a test to check that, in a entity with 2 attributes, the operaiton only affects to the right one (leaving the other untouched) should be added.

kzangeli: Added a test of "leaving the other untouched" in 3cbf771 (also, I changed the behavior of the operation from REPLACE to UPDATE as this is not a normal PUT - it's a 'concentrated' PUT ...)

kzangeli: PR changed to resolve issue #999 and not #991.
I believe all necessary functests are present

Fixed in 3cbf771, cd9c877 and 361cbb6

test/functionalTest/cases/0999_update_single_attr_value_by_eid/update_single_attr_value_by_eid_error_handling.test
```diff
@@ -0,0 +1,100 @@
...
+Error handling For PUT Individual Entity Attribute
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. PUTT /v2/entities/E1/attrs/A1/value to see BAD METHOD, only GET, PUT
```
fgalan: Typo: PUTT -> PUT

test/functionalTest/cases/0999_update_single_attr_value_by_eid/update_single_attr_value_by_eid_error_handling.test
```diff
@@ -0,0 +1,100 @@
...
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. PUTT /v2/entities/E1/attrs/A1/value to see BAD METHOD, only GET, PUT 
+# 02. PUT /v2/entities/E1/attrs/A1/value without 'value' in payload
+# 03. PUT /v2/entities/E1/attrs/A1/value with 'value' and 'value2' in payload
```
fgalan: Testing with 'value' and 'type' and get an error would be also interesting.

test/functionalTest/cases/0999_update_single_attr_value_by_eid/update_single_attr_value_by_eid_error_handling.test
kzangeli: Actually no, not a typo.
The test here is to provoke "BAD METHOD"

test/functionalTest/cases/0999_update_single_attr_value_by_eid/update_single_attr_value_by_eid_error_handling.test
kzangeli: Sure, it's the same though - everything but 'value' gives an error.
However, 'type' is more interesting than 'asdasd' or 'value2'.

kzangeli: Fixed in 3b814c1

test/functionalTest/cases/0999_update_single_attr_value_by_eid/update_single_attr_value_by_eid_error_handling.test
fgalan: Maybe a more explicit token should be used, such as NOVERB or BADVERD.

kzangeli: BADVERB used in 99569c1

fgalan:
LGTM

PR #1165, opened 2015-08-13
Feature/991 real update single attribute

kzangeli:
Implements issue #991

PUT /v2/entities/\<entityId\>/attrs/\<attrName\>

### DIFF
--- a/test/functionalTest/harnessFunctions.sh
+++ b/test/functionalTest/harnessFunctions.sh
@@ -861,14 +861,22 @@ function orionCurl()
   dMsg Executing the curl-command
   _response=$(eval $command 2> /dev/null)
 
+  if [ ! -f /tmp/httpHeaders.out ]
+  then
+    echo "Broker seems to have died ..."
+  else
+    #
+    # Remove "Connection: Keep-Alive" and "Connection: close" headers
+    #
+    sed '/Connection: Keep-Alive/d' /tmp/httpHeaders.out  > /tmp/httpHeaders2.out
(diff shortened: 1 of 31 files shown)


test/functionalTest/harnessFunctions.sh
```diff
@@ -861,14 +861,22 @@ function orionCurl()
...
 
   #
-  # Remove "Connection: Keep-Alive" and "Connection: close" headers
+  # Unless we remove the HTTP header file, it will remain for the next execution
   #
-  sed '/Connection: Keep-Alive/d' /tmp/httpHeaders.out  > /tmp/httpHeaders2.out
-  sed '/Connection: close/d'      /tmp/httpHeaders2.out > /tmp/httpHeaders.out
-  sed '/Connection: Close/d'      /tmp/httpHeaders.out  
-  
+  \rm -f /tmp/httpHeaders2.out /tmp/httpHeaders.out
```
fgalan: `\rm` is correct?

kzangeli: Yes.
If you put a backslash before a unix shell command, then the shell doesn't look for any alias.
I myself always alias 'rm' to 'rm -i' - making rm ask me 'yes/no' for each file I remove.
Here, we don't want any alias to change the way 'rm' works.

fgalan: You are assuming ah hoc alias (yours) in a generic code (the one in github.com). I mean, following the same rule, why don't use `\sed`, as some of the potential user of this code could have an alias also for that?

However, it doesn't hurt, so NTC

fgalan:
LGTM

PR #1178, opened 2015-08-14
Feature/1169 uri param options map

kzangeli:
The broker now detects invalid items in URI parameter 'options' and returns **400 Bad Request** if necessary.
Also (implementation), all the items in URI parameter 'options' are saved in a map <string, bool> and can be looked up in ConnectionInfo::uriParamOptions:

    if (ciP->uriParamOptions["text"] == true)
    {
        ...
    }

### DIFF
--- a/test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
+++ b/test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
@@ -75,7 +75,7 @@ echo
 
 echo "06. Error: Test an arbitrary API v2 path"
 echo "========================================"
-orionCurl --url /v2/type?options=BAD --json
+orionCurl --url /v2/types?options=BAD --json
 echo
 echo
 
--- a/test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
+++ b/test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
@@ -30,148 +30,77 @@ brokerStart CB
 --SHELL--
 
 #
-# 01. OK: Send a /version request with one valid option (count)
-# 02. OK: Send a /version request with two valid options (count + normalized)
-# 03. Error: Send a /version request with one invalid option (canonical)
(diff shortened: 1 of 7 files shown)


test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
```diff
@@ -0,0 +1,188 @@
...
+URI Param OPTIONS
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. OK: Send a /version request with one valid option (count)
```
fgalan: Maybe /version behaviour would change in the future... I'd recommend to use other operation, e.g. GET /v2/entities.

kzangeli: We don't get down to that level, the options errors are detected before any payload is treated.
Are you thinking about the OK steps, that the REGEXPECT may change for /version?

kzangeli: Based functest on /v2/entities instead of /version in 056e9b3

test/functionalTest/cases/1169_uri_param_options_map/uri_param_options_map.test
```diff
@@ -0,0 +1,155 @@
...
+echo "01. OK: Send a request with one valid option (count)"
+echo "===================================================="
+orionCurl --url /v2/entities?options=count --json --payload '{ "id": "ID", "type": "type" }'
+echo
+echo
+
+
+echo "02. OK: Send a request with two valid options (count + normalized)"
+echo "=================================================================="
+orionCurl --url /v2/entities?options=count,normalized --json --payload '{ "id": "ID2", "type": "type" }'
```
fgalan: This will result in an error if at the end normalized mode is implemented for POST /v2/entities, as `{ "id": "ID2", "type": "type" }` is not normalized. However, I understand that will be easy to find&fix at that moment.

NTC

fgalan:
LGTM

PR #1180, opened 2015-08-14
Removed LM_Ms

kzangeli:
Removed all LM_Ms from the source code.
Also, a function stgringValue was removed as it was used only for the LM_Ms.
A few indentations was made as well


fgalan:
LGTM

PR #1184, opened 2015-08-23
Hardening/valgrind24

kzangeli:
Fixed the many many leaks for 0.24.0 ...
And quite a few new leaks, found after modifying the valgrind test suite ...
A total of perhaps 15 leaks have been fixed.

In the process, a lot of improvements to the valgrind suite have been made:
  * New option: '-fromIx ix'                 (test index from where to start the suite)
  * New option: '-ixList "ix1 ix2 ix3 ..."'  (list of indices of test cases to execute)
  * New option: '-dryLeaks                   (simulate leaks and functest errors) - only interesting for improvements to the valgrind suite
  * The time of each part is shown: seconds with two decimals (a little like the func test harness)
  * The time that the complete suite took is printed on stdout upon termination
  * The start time for the suite is printed on stdout when it starts
  * MOST IMPORTANT: In test cases involving more than **one** broker - valgrind is run only for the CB, not CP1 etc.
    Having valgrind run for *every broker* destroys the result of the main broker (the 'CB' broker).
    Thanks to this modification, 7-8 new (old :-)) leaks, until now undetected, have been found and fixed.
  * The older '*.vtest' cases are no longer executed by default. Use the parameter 'both' to include those 'pure' valgrind tests.
  * All functests, except two that I cannot fix, are now working under valgrind
  * All ftest* databases are removed before starting a new test case.
    This seems to not work 100% of the times ... May need some further debugging
    See special log-file /tmp/valgrindDbReset.log for this.
  * The "sleep 10" after starting valgrind and after terminating, awaiting valgrind to end have been replaced with
    loops using 'nc', just like in the func test suite - this modification has made the entire suite take perhaps 90 minutes less.
    The entire suite (excluding 'pure' cases) now take a little less than three hours.
  * Improved error reporting after the valgrind suite finishes

Now, only two func tests fail under the valgrind suite:
  * 0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_append.test - too many notifications ...
  * 0283_coap_support/coap_basic                                                   - Leandro supposed to fix this ... some day ...


src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -211,7 +211,9 @@ HttpStatusCode mongoSubscribeContext
     // 3. Create Subscription for the cache
     if (patternBased && onchange)
     {
-      subCache->insert(new orion::Subscription(tenant, servicePath, requestP, oid.toString(), expiration, notifyFormat));
+      std::string oidString = oid.toString();
+
+      subCache->insert(new orion::Subscription(tenant, servicePath, requestP, oidString, expiration, notifyFormat));
```
LeandroGuillen: Curious: is this a matter of style requirement?

kzangeli: This is a leak fix, believe it or not ...
When the scope ends, "std::string oidString" on the stack is destroyed.
If I do as before, the oid.toString() creates a std::string and it is lost.
It took me a while to fix this one ...

About style, I try to maintain an empty line between variable-declarations and the rest of the code

LeandroGuillen: That is crazy pointer semantics for you :|

NTC

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -368,10 +368,18 @@ std::string statisticsTreat
...
+  // The valgrind test suite uses REST GET /version to check that the broker is alive
+  // This fact makes the statistics change and some working functests fail under valgrindTestSuite
+  // due to the 'extra' version-request in the statistics.
+  // Removing version-requests from the statistics 'report' fixes this problem.
+  //
   if (noOfVersionRequests != -1)
   {
     out += TAG_ADD_COUNTER("versionRequests", noOfVersionRequests);
   }
+#endif
```
kzangeli: I left this "#if 0 / #endif" for a discussion on how to solve this.

Instead of doing this, I have solved the problem by adding a few "GET  /version  in XML" in functests and REGEX((1|2)) or similar in REGEXPECT part.

The 'problem' is due to the function stat starts the broker (actually, the function that awaits the broker start), sends a GET /version "in XML" to make sure the broker is running. And, of course, this alters the request statistics ...
One count in func-test, another count under valgrind.

test/functionalTest/cases/0000_https_support/https.test
```diff
@@ -27,7 +27,10 @@ HTTPS simple test
 dbInit CB
 ${SCRIPT_HOME}/httpsPrepare.sh --keyFileName /tmp/harnessTest.key --certFileName /tmp/harnessTest.pem
 extraParams="-https -key /tmp/harnessTest.key -cert /tmp/harnessTest.pem"
+
+export BROKER_AWAIT_SLEEP_TIME=6
 brokerStart CB 0-255 IPV4 "$extraParams"
+unset BROKER_AWAIT_SLEEP_TIME
```
kzangeli: Some functests fail "exit code 11" under valgrind if the broker is started with the new method of using nc and awaiting the broker that way.
I have fixed this problem by adding this alternate way - which is the old way, just a normal sleep (but with $BROKER_AWAIT_SLEEP_TIME instead of a hardcoded 10 seconds)

test/functionalTest/cases/0283_coap_support/coap_basic.test
```diff
@@ -31,9 +31,8 @@ proxyCoapStart
 
 --SHELL--
 
-valgrindSleep 2
+valgrindSleep 5
```
kzangeli: Just a vain intent to 'fix' coap_basic.test ...
This modification is worthless

test/functionalTest/cases/0725_multiservice_ignore/service_ignore_without_multiservice.test
```diff
@@ -101,10 +102,11 @@ Date: REGEX(.*)
 =====================================================
 1
 0
+ftest
 --TEARDOWN--
```
kzangeli: This may seem a little schizophrenic, but I have seen this test fali a few times, and always by finding TWO databases using the first command:

    echo "show dbs" | mongo | grep ${CB_DATABASE_NAME} | wc -l

This extra check will show us the exact name of the 'extra', unwanted database, and perhaps we can fix this sporadic failure thanks to this added check.

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -25,12 +25,13 @@ notifyFormat_ngsi10
 
 --SHELL-INIT--
 dbInit CB
-brokerStart CB 0 IPv4 -mutexTimeStat
+brokerStart CB 0 IPv4 -mutexTimeStat -subCacheIval 0
```
kzangeli: Disabling refresh of the subscription-cache

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_append.test
```diff
@@ -186,6 +186,7 @@ echo
 echo
 
 
+valgrindSleep 2
```
kzangeli: Vain intent to make this test-case work under valgrind

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1182,6 +1182,13 @@ void orionExit(int code, const std::string& reason)
 */
```
fgalan: Usually we don't include CNR entry for de-leaking PR, as the PR is supposed to fix the leaks introduced in the last version. However, due to:

> Thanks to this modification, 7-8 new (old :-)) leaks, until now undetected, have been found and fixed.

in this case I think we should include an entry.

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_append.test
fgalan: NTC

src/lib/serviceRoutines/statisticsTreat.cpp
fgalan: So, if In understand correctly we have two potencial solutions on the table a) using REGEX((1|2)), b) remove the code in the #if 0/#endif and we have to chose between one of them?

Maybe the less intrusive is the one based on REGEX((1|2)) in just one .test (i.e. the .test for statistics) explaining in a comment in that .test why the REGEX is needed.

test/functionalTest/cases/0000_https_support/https.test
fgalan: Does the BROKER_AWAIT_SLEEP_TIME overrides the nc mechanism? Or is a complementary mechanism acting as a timeout for the nc-based wait?

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval.test
```diff
@@ -439,7 +439,7 @@ Content-Type: application/xml
 </notifyContextRequest>
 =======================================
 NUMBER: 16
-TIMES: REGEX((0, 2, 4, 5, 9, 13, 14, 24, 34, 35, 37, 39, 53, 65, 67, 69|0, 2, 4, 5, 9, 13, 14, 24, 34, 35, 37, 39, 51, 63, 67, 69))
+TIMES: REGEX((0|1), (1|2|3), (3|4), (5|6), (9|10), (13|14), (14|15|16), (24|25|26), (34|35|36), (35|36|37|38), (37|38|39|40), (39|40|41|42), (51|52|53|54), (63|65|66), (67|72), (69|74))
```
fgalan: Not sure if it is a good idea using "blur" times this way (I have seen it also in other .test), as it makes the test harness weaker.

Maybe an way of specifying two alternative block could be better, something in this way:

```
 NUMBER: 16
#if valgrrind
TIMES: REGEX((0|1), (1|2|3), (3|4), (5|6), (9|10), (13|14), (14|15|16), (24|25|26), (34|35|36), (35|36|37|38), (37|38|39|40), (39|40|41|42), (51|52|53|54), (63|65|66), (67|72), (69|74))
#else
TIMES: REGEX((0, 2, 4, 5, 9, 13, 14, 24, 34, 35, 37, 39, 53, 65, 67, 69|0, 2, 4, 5, 9, 13, 14, 24, 34, 35, 37, 39, 51, 63, 67, 69))
#endif
...
```

However, probably that is very hard for this PR. Maybe we can create an issue to work on it in the future.

test/functionalTest/cases/0283_coap_support/coap_basic.test
fgalan: At the end, all the proxyCoap stuff (code, makefile targets, tests, etc.) will be moved to an "archive" directory, so it is not a problem if this 0283_coap_support tests have problems.

NTC

test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
```diff
@@ -52,24 +53,29 @@ echo "+++++ 2. Too many levels in path"
...
 orionCurl --url /v1/queryContext --payload "$payload" --servicePath 'X'
 echo
 
+
 echo "+++++ 5: four service paths"
 orionCurl --url /v1/queryContext --payload "$payload" --servicePath '/home/kz/1, /home/kz/2, /home/kz/3, /home/kz/4'
 echo
 echo "+++++ 6: service paths from log file"
-grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }'
+grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }' | grep -v '/#'
```
fgalan: Why the final `| grep -v '/#'` was needed?

test/functionalTest/cases/0725_multiservice_ignore/service_ignore_without_multiservice.test
fgalan: This is related with the "ftest-ftest-fest.... problem"? Which names have the TWO databases when they appear?

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: Why cache need to be disbled to make this test to work under valgrind?

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -268,7 +282,7 @@ Date: REGEX(.*)
 06. Statistics
 ==============
 HTTP/1.1 200 OK
-Content-Length: 488
+Content-Length: 51REGEX((3|5))
```
fgalan: Mmmm... this is dirty. I have to re-thought my previous comment on "REGEX((1|2)) is the best solution" :)

test/functionalTest/cases/0000_https_support/https.test
kzangeli: Yes, if BROKER_AWAIT_SLEEP_TIME is set, then nc is not used.
Just *sleep $BROKER_AWAIT_SLEEP_TIME*.

test/functionalTest/cases/0725_multiservice_ignore/service_ignore_without_multiservice.test
kzangeli: Yes, this is about the ftest-ftest... problem.
Will help us to detect exactly *where* the problem arises ...

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: Not sure how to respond on this ...
Timing issues, definitely, but ... how come we're not able to fix the problem in the atexit-function?
Frankly, I didn't have time enough to fix everything ...
This potential leak is only about when the broker dies. Not especially interesting.
An issue about removing this option is definitely in order though.  @fgalan ? :-)

src/app/contextBroker/contextBroker.cpp
kzangeli: Sure, I agree. After all, the behavior of the “black box“ (the broker) has somehow changed ...

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval.test
kzangeli: You are talking about running the REGEX part in cpp (C preprocessor) before real execution. Sure. Why not.  Personally  I find it a little exaggerated though. After all, this is a test that fails *a lot* without valgrind as well ...
This "fix" makes this test finally 'work' ...  It is all about timing, nothing else ...

src/app/contextBroker/contextBroker.cpp
fgalan: Maybe just a sentence like "Fix: several memory leaks has been fixed" is enough? Or are you able to be more specific about which old leaks has been fixed?

test/functionalTest/cases/0000_https_support/https.test
fgalan: NTC

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: This is 513 or 515.
Two bytes extra when timing gives that is takes TEN seconds instead of NINE.
Not dirty at all.
The tests must/should work under valgrind as well ...
A functest that fails under valgrind **fails** to get the leaks reported. *That* is important.
If it takes nine or ten seconds ... Not very important ...

However, if we run the REGEXPECT part under cpp before comparing, this 'shortage' can be avoided.   Still, I find this a little exaggerated ...  The functests aren't there to measure timing, but functionality.

test/functionalTest/cases/0725_multiservice_ignore/service_ignore_without_multiservice.test
fgalan: There was a bug related with this due to the subscriptoin cache that has been now fixed (https://github.com/telefonicaid/fiware-orion/issues/1186). Maybe this was related and it is no longer needed as the bug has been solved?

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: There was a bug related with this due to the subscriptoin cache that has been now fixed (https://github.com/telefonicaid/fiware-orion/issues/1186). Maybe this was related and it is no longer needed as the bug has been solved?

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval.test
fgalan: No, it is not about C, but about marking "blocks" for the test suite, so it uses one block in the case of "regular" functional test or the other in the case of test running under valgrind.

fgalan:
After merging develop (i.e. at commit 66dc824) the result of valgrind test suite is the following:

```
5 functional tests failed (not a leak, just a func-test failure):
   122: subscription_sequence_ontimeinterval_long_expiration.test (exit code 1)
   259: coap_basic.test (exit code 1)
   352: service_ignore_without_multiservice.test (exit code 1)
   430: creation_with_attribute_as_object.test (exit code 1)
   433: simple_creation.test (exit code 1)
---------------------------------------
Total test time: 5981.79 seconds (99 minutes)
Great, all valgrind tests ran without any memory leakage
```

which I interpret that after the last changes in develop the code stills being leak-free. The fails in function test could be due to timming or Python configuration in the environment where I have run the test (3.9999 vs. 4 problem).

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: I have checked that it is safe to remove `-subCacheIval 0`, running the test under valgrind three times:

```
[develenv@ci-fiware-01 fiware-orion]$ test/valgrind/valgrindTestSuite.sh -filter three_subscriptions.test
Thu Sep 10 13:51:05 CEST 2015
Test 001/1: 1048_subscription_cache/three_subscriptions ............................................................................................. OK (13.82 seconds)
Total test time: 13.94 seconds (0 minutes)
Great, all valgrind tests ran without any memory leakage
[develenv@ci-fiware-01 fiware-orion]$ test/valgrind/valgrindTestSuite.sh -filter three_subscriptions.test
Thu Sep 10 13:51:22 CEST 2015
Test 001/1: 1048_subscription_cache/three_subscriptions ............................................................................................. OK (13.34 seconds)
Total test time: 13.46 seconds (0 minutes)
Great, all valgrind tests ran without any memory leakage
[develenv@ci-fiware-01 fiware-orion]$ test/valgrind/valgrindTestSuite.sh -filter three_subscriptions.test
Thu Sep 10 13:51:37 CEST 2015
Test 001/1: 1048_subscription_cache/three_subscriptions ............................................................................................. OK (13.34 seconds)
Total test time: 13.47 seconds (0 minutes)
Great, all valgrind tests ran without any memory leakage
[develenv@ci-fiware-01 fiware-orion]$ test/valgrind/valgrindTestSuite.sh -filter three_subscriptions.test
Thu Sep 10 13:51:54 CEST 2015
Test 001/1: 1048_subscription_cache/three_subscriptions ............................................................................................. OK (13.38 seconds)
Total test time: 13.51 seconds (0 minutes)
Great, all valgrind tests ran without any memory leakage
```

fgalan: Fixed in 227371c

test/functionalTest/cases/0725_multiservice_ignore/service_ignore_without_multiservice.test
fgalan: Let's keep it, it doesn't hurt. However, I have provided a comment explaining where this come from.

Fixed in 53882e4

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval.test
fgalan: Issue created https://github.com/telefonicaid/fiware-orion/issues/1222

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -135,7 +144,7 @@ echo
 
 echo "04. Statistics"
 echo "=============="
-orionCurl --url /statistics --json
+orionCurl --url /statistics
```
fgalan: Why we have changed from JSON to XML? XML is now a deprecated format...

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: Maybe the correct way of dealing with any message with variable unbounded payload from a logical pooint of view, i.e., the ones using at least REGEX() `.*`, `.+`, `\d+`, `\d*`, etc. is to use `Content-Length: REGEX(\d+)`.

src/lib/serviceRoutines/statisticsTreat.cpp
kzangeli: No, both are needed.
Without REGEX((1|2)) it can never work, due to huge changes in timing with/without valgrind.

Removing the versionRequests from the statistics output is something else (that's why I say 'Apart from ...').
The problem is stat the statistics-output of ZERO version-requests get no line at all about version-requests, but if there has been one or more version-requests, then we have a line about it in the statistics-output.  I don't know how to express "Line or no line" in REGEX.
That is basically the reason for this '#if 0/#endif'.

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval.test
kzangeli: Yes of course, and the way you propose to do it is with the C preprocessor - CPP. The C preprocessor takes care of all '#-directives'.  #define, #include, #if¡, etc

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: At the end, let's use:

```
Content-Length: REGEX((513|515))
```

kzangeli:
About

    After merging develop (i.e. at commit 66dc824) the result of valgrind test suite is the following ...

Five errors seems a lot to me. To quote the comment on this PR:

    All functests, except two that I cannot fix, are now working under valgrind

I will look into this ...

src/lib/serviceRoutines/statisticsTreat.cpp
kzangeli: Fixed in 56c3fdf

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in 56c3fdf

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: Fixed in 56c3fdf

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: Fixed in ee67b3e

test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
kzangeli: It wasn't needed - probably some old stuff ...
Fixed in b0d7c18

fgalan:
LGTM

Issue #1158, opened 2015-08-12
POST /v2/entities must return error if the entity to create already exists


fgalan:
It would return the following error:

HTTP status code: 422 Unprocessable Entity
Payload:

```
{
   "error": "AlreadyExists",
   "description": "Entity already exists"
}
```

fortizc:
I send a pull request but it haven an error because my code broke the "update or append entity atribute". The main problem is know when is a update or an append. Any ideas?

fortizc:
I found a way to check when a operation is an "entity creation" or an "update or append entity atribute" if we check the uriParams map, when is a "update or append entity atribute" it have a "op" key, because in this operation some options can be added.

fortizc:
what if we move buildGeneralErrorResponse() method from MongoCommonUpdate.cpp to mongoUpdateContext.cpp and we catch all exceptions in mongoUpdateContext() ??

fgalan:
> what if we move buildGeneralErrorResponse() method from MongoCommonUpdate.cpp to mongoUpdateContext.cpp and we catch all exceptions in mongoUpdateContext() ?? 

Let's consider that, but please create a new issue to discuss about it (note that the present issue is about AlreadyExists error and it is better to keep independent topics in independent issues).

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1295. @iariasleon please have a look and close if if you find it ok.

iariasleon:
LGTM
tested with:
 - only id, 
 - id and type,
 - id, type and service_path
```
http code: 422
{
    "error":"InvalidModification",
    "description":"Entity already exists"
}
```

PR #1193, opened 2015-08-27
Replace previous pull request. This is a better fix to issue #1158

fortizc:


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2310,6 +2310,26 @@ void processContextElement
                     getEntitiesCollectionName(tenant).c_str(),
                     query.toString().c_str()));
 
+    std::map<std::string, std::string>::iterator it = uriParams.find("op");
```
fgalan: Have you tried `make unit_test` in this branch? I'm getting a segfault in mongoNotifyContextRequest.Ent1Attr1 (but it could be a false positive, which only happens in my system...).

In addition to `make unit_test`, `make functional_test` is also needed as "regression" in order to check that at PR is not breaking any existing functionality (this is standard Orion way-of-working, don't worry, it's easy to learn :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2310,6 +2310,26 @@ void processContextElement
...
+
+    // if is an append and uri has not possibility of options
+    if (strcasecmp(action.c_str(), "append") == 0 && it == uriParams.end())
+    {
+        EntityIdVector vec;
```
fgalan: Orion code style uses 2 whitespace for indentation. It seems (not fully sure, as github diff view is not always shows things correctly) that you are using 4 in this block.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2310,6 +2310,26 @@ void processContextElement
...
+        vec.push_back(enP);
+
+        AttributeList attList;
+        Restriction res;
+        ContextElementResponseVector rsp;
+        std::string err;
+        entitiesQuery(vec, attList, res, &rsp, &err, true, tenant, servicePathV);
```
fgalan: entityQuery() is a "heavy" operation at it involves an extra query to the DB each time an entity is going to be created. It would be great to detect the situation at DB insert time, i.e. if the the document to be inserted already exists then return the error at that point. Maybe Mongo operations parameter allow to do something like that... I'm not sure right now.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2310,6 +2310,26 @@ void processContextElement
...
+
+        if (rsp.size() && !rsp.get(0)->prune)
+        {
+            throw OrionCreateException("Entity already exists");
```
fgalan: I like the idea of using exception to progress error conditions on the code, it uses to lead to a more readable code. Maybe we should start using something like this at least at mongoBackend module.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2310,6 +2310,26 @@ void processContextElement
                     getEntitiesCollectionName(tenant).c_str(),
                     query.toString().c_str()));
 
+    std::map<std::string, std::string>::iterator it = uriParams.find("op");
+
+    // if is an append and uri has not possibility of options
+    if (strcasecmp(action.c_str(), "append") == 0 && it == uriParams.end())
```
fgalan: The "semantics" used by the mongoBackend module are APPEND and APPEND_ONLY (the op=append is only used in the serviceRoutine layer which invokese mongoBackend, i.e. if the cliente resquest uses op=append, then serviceRoutings translate it to APPEND_ONLY before invoking mongoBackend).

Thus, probably this statement could be simplified to:

```
if (strcasecmp(action.c_str(), "append_strict"))
```

and the `it` variable is not actually needed.

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -78,10 +79,20 @@ std::string postEntities
     ciP->httpHeaderValue.push_back(location);
     ciP->httpStatusCode = SccCreated;
   }
+  else
+  {
+    OrionError oe;
+    oe.code = ciP->httpStatusCode;
+    oe.reasonPhrase = "AlreadyExists";
+    oe.details = "An entity with the same id and type "
+                 "already exists. You can use POST /entities/{entityId} "
```
fgalan: Although this is the text that appear #1158, thinking a little bit on it, it could be missleanding as:

* The operation causing the problem could not include a type (i.e. only including id)"
* It is not fully cleear if the suggested "POST /entities/{entityId}" could include URI params and, in that case, if the user includes `op=append", the problem happens again.

Thus, maybe is better to use jsut:

```
oe.details = "Entity already exists"
```

(I will update #1158)

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -78,10 +79,20 @@ std::string postEntities
     ciP->httpHeaderValue.push_back(location);
     ciP->httpStatusCode = SccCreated;
   }
+  else
+  {
+    OrionError oe;
+    oe.code = ciP->httpStatusCode;
+    oe.reasonPhrase = "AlreadyExists";
```
fgalan: The "Error response" section in doc/apiary/v2/fiware-ngsiv2-reference.apib document should be edited to add this new error to the list.

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -78,10 +79,20 @@ std::string postEntities
     ciP->httpHeaderValue.push_back(location);
     ciP->httpStatusCode = SccCreated;
   }
+  else
```
fgalan: Maybe is safer to check the error condition instead of indiscrimined `else`. I mean:

```
else if ((ciP->httpStatusCode == SccUnprocessableEntity    )
```

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: After thinking on this a little bit, actually it should be an excesive overhead, as this check is only triggered when ?op=append / APPEND_STRICT is used, which I supposed will not be very often.

Thus, I think we can progress this PR with the current strategy. Improving that has been deferred to a tech debt issue (https://github.com/telefonicaid/fiware-orion/issues/1207).

NTC (=Nothing To Change, regarding this comment thread).

src/lib/mongoBackend/mongoUpdateContext.cpp
```diff
@@ -67,25 +67,33 @@ HttpStatusCode mongoUpdateContext
...
+              error gets "encapsulated" in the StatusCode of the corresponding ContextElementResponse and we
+              consider the overall mongoUpdateContext() as OK. */
+            responseP->errorCode.fill(SccOk);
         }
-
-        /* Note that although individual processContextElements() invocations return ConnectionError, this
-           error gets "encapsulated" in the StatusCode of the corresponding ContextElementResponse and we
-           consider the overall mongoUpdateContext() as OK. */
-        responseP->errorCode.fill(SccOk);
-    }
```
kzangeli: No changes in this block except indentation, right?
The indentation was ok before, not now ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Not that I am especially *against* exceptions, even though it is an extremely complicated method to simply return an error code, but in the source code of the broker we don't throw exceptions. We do catch them, as we have no other choice as e.g. the mongo client library throws exceptions.

The source code of orion uses return values and checks for the return value after each call - the simpler and faster 'C method' ...

You will soon understand that the broker is implemented in C, with very few features from C++.
In my personal opinion, the less C++, the better ...

I would prefer to uniformally keep using return values and avoid throwing exceptions.
Later we can discuss and perhaps change this choice of implementation method but until then ...

fgalan: @kzangeli is right, please ignore my first comment in this thread... Using exception would be a global change, so it is better to keep using return values.

fgalan:
Surpassed by PR https://github.com/telefonicaid/fiware-orion/pull/1295. Closing.

PR #1201, opened 2015-09-01
New dockerfile

LeandroGuillen:
This dockerfile generates an image about 270MB in size, about the same as the one created from an RPM installation -> Success!


fgalan:
LGTM

PR #1205, opened 2015-09-02
FIX docker compose scripts resersing mongo/orion startup order 

fgalan:
And adding the --smallfiles option (as comment)


LeandroGuillen:
I went through with the process and found a mistake carried from my previous change to the dockerfile. It was fixed and now everything works.

LGTM.

PR #1211, opened 2015-09-08
FIX CB crashes when DB stops due to cache refresh

### LINKED ISSUE #1210: [BUG] if mongo is restarted, Context Broker is stopped
### Options used

 `BROKER_EXTRA_OPS="-multiservice -t 0-255"` - with these options CB is stopped

 `BROKER_EXTRA_OPS="-multiservice -t 0-255 -subCacheIval 0"` - with these options CB is not stopped

fgalan:
Fixes #1210.

CNR entry not needed, as the fix is on functionality developed in the current version (subscriptions cache).

Manually tested.

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -270,29 +270,50 @@ extern void setDbPrefix(std::string _dbPrefix)
 *
 * getOrionDatabases -
 */
-extern void getOrionDatabases(std::vector<std::string>& dbs)
+extern bool getOrionDatabases(std::vector<std::string>& dbs)
 {
   BSONObj       result;
   DBClientBase* connection = getMongoConnection();
 
-  connection->runCommand("admin", BSON("listDatabases" << 1), result);
-  releaseMongoConnection(connection);
-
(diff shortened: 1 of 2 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -270,29 +270,50 @@ extern void setDbPrefix(std::string _dbPrefix)
 *
 * getOrionDatabases -
 */
-extern void getOrionDatabases(std::vector<std::string>& dbs)
+extern bool getOrionDatabases(std::vector<std::string>& dbs)
```
LeandroGuillen: Is this function never called anywhere? Because it may need checking whether the return value is true.

fgalan: The function is called from contextBrokerInit(), mongoInit() and subscriptionCache::fillFromDb(). None of that places check the return value, but anyway I think that from a logical point of view it is better is getOrionDatabase() signales sucess/failure on DB operation. Currently we are not using that information, but we may use in the future (e.g. to log error at a higher level).

LeandroGuillen:
LGTM

PR #1218, opened 2015-09-10
Remove manual parameter description to enable migration to FABRE Part I

jmcanterafonseca:


jmcanterafonseca:
WIP please do not review yet

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1,9 +1,33 @@
 FORMAT: 1A
-HOST: http://orion.lab.fiware.org/
+HOST: http://telefonicaid.github.io/fiware-orion/api/v2/
+TITLE: FIWARE-NGSI v2 Specification
+DATE: 15 September 2015
+VERSION: <placeholder>
```
jmcanterafonseca: nice to have would be to automatically add here the commit hash using a hook ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -372,13 +356,9 @@ temperature<=20
...
 </ul>
 
-</td></tr>
-<tr><td style="padding-right: 40px; width: 140px;">geometry</td><td>
-
-<p>Defines a geografical area so only the entities located in that area matches the query. It is composed of a tokens list
-separated by <code>;</code>. The first token is the shape of the geometry, the rest of the tokens (if any) depends
-on the shape. In particular, the following geometries can be defined so far:</p>
-
+<p><code>geometry</code> Defines a geografical area so only the entities located in that area matches the query.
```
jmcanterafonseca: in the near future all HTML tags will be removed from the document in preference of markdown conventions

jmcanterafonseca:
@fgalan this is ready to be reviewed. Please note that this is only the beginning and I plan to divide this in 4 or 5 parts

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: On the meanwhile, maybe we can use:

```
VERSION: 2.0-wip
```

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1,9 +1,27 @@
 FORMAT: 1A
-HOST: http://orion.lab.fiware.org/
+HOST: http://telefonicaid.github.io/fiware-orion/api/v2/
```
fgalan: I though that HOST was for the "reference host" holding the API implementation, not the URL hosting the apiary rendering of the spec. Could you clarify, please? Maybe a link to the piece of doc about HOST parameter in apiary documentaiton could help.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -31,9 +57,9 @@ The FIWARE NGSI (Next Generation Service Interface) API defines
 (whether to separate the two interfaces is currently under discussion).
 * a set of typical **roles** played by NGSI-compliant components
 
-### Context data modelling and exchange
+## Terminology
```
fgalan: Note the "Exchanging Context Availability Information", which is at the same section level than the (changed) "Context data modelling and exchange". Thus, probably the structure should be:

```
## Terminology
### Context data modelling and exchange
(subsections using ####)
### Exchanging Context Availability Information
(subsections using ####)
```

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -161,10 +189,12 @@ an opaque way. Nonetheless, the types described below are used to convey an spec
   greater than, lesser than, greater or equal, lesser or equal and range query operators. Eg:
 
 ```
+{
```
fgalan: The reading could (wrongly) think that the timestamp attribute appears always standalone. I would suggest using elipsis before and after the block in the following way:

```
{
  ...
  "timestamp": {
    "value": "2017-06-17T07:21:24.238Z",
    "type: "date"
  }
  ...
}
```

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -172,10 +202,12 @@ an opaque way. Nonetheless, the types described below are used to convey an spec
   check feasibility).
 
 ```
+{
```
fgalan: Same elipsis technique using `...` could be done here.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -241,61 +273,7 @@ Error list (HTTP response code in paranthesis):
...
-<tr><td style="padding-right: 40px; width: 140px;">q</td><td>
-
-<p>A query expression, composed of a list of statements separated by <code>;</code>, i.e.</p>
-
-<p><code>
-q=statement;statements;statement
-</code></p>
+## Simple Query Language
 
 <p>Each statement express a matching condition. The result of the query is the match on each one of the statements, e.g. a
```
fgalan: This has been cut-and-pasted from the List entities operation, but starting the section in this way is a bit "harsh". E.g. "each statement" assumes that previously you have explained that `q` is composed by a list of statement, separated by `;`.  In fact, in the previous version we have a text about it that could be also moved here:

> A query expression, composed of a list of statements separated by ;, i.e. q=statement;statements;statement 

Moreover, you are using the `Simple Query Language` to explain 3 different parameters (q, geometry, coords, etc). I'd suggest to use 3 subsections, one for each one of the parameters.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I understand you have started by removing the required/optional parameter tables, moving them to the "right panel" at apiary. However, doing so you are lossing information, as the required/optional nature of the attribute cannot be inferred from the "righ panel", as far as I checked.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -386,34 +360,52 @@ on the shape. In particular, the following geometries can be defined so far:</p>
...
+            "subscriptions_url": "/v2/subscriptions",
+            "registrations_url": "/v2/registrations"
+        }
+
+## Entities [/v2/entities{?limit,offset,options,type,id,idPattern,q,geometry,coords,attrs}]
+
+### List entities [GET /v2/entities{?limit,offset,options,type,id,idPattern,q,geometry,coords,attrs}]
+
+Retrieves a list of entities which match criteria defined by the following parameters: `id`, `type`, `idPattern`, `q`,
+`geometry` and `coords` attribute (see below for a detailed description of these parameters). A given entity
```
fgalan: I think the "see below for a detailed description of these parameters" is now obsolete... the "link" should point to somewhere (part in the Simple Query Language description).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -424,11 +416,34 @@ Response code:
...
 + Parameters
-    + limit: 20 (optional, number)
-    + offset: 20 (optional, number)
-    + type: Room (optional, string)
-    + id: Bcn_Welt (optional, string)
-    + idPattern: (optional, string)
+    + id: Boe_Idearium (optional, string) - A comma separated list of elements.
+    Retrieve entities which ID match one of the elements in the list.
+    
+    + type: Room (optional, string) -  comma separated list of elements.
```
fgalan: Related with my previous comment (https://github.com/telefonicaid/fiware-orion/pull/1218/files?diff=split#r39383034), it seems that apiary allows to specify mandataory/optional in an attribute... However, which effect has this in the "righ panel" rendering? Maybe is a subtle detail, but I cannot see how "optional" is marked in the web.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -492,18 +507,10 @@ Response:
 
 ## Entity by ID [/v2/entities/{entityId}{?attrs,options}]
 
-### Retrieve entity [GET /v2/entities/{entityId}{?attrs}]
++ Parameters
```
fgalan: It seems that the entityId is not shown in the "righ panel" operations under this ## point. Maybe parameter definition has to be done always at ### level?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: you are right, that change is spurious

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: but that information is going to be showed by Fabre anyway

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: too much levels that's why I took the liberty of simplifying

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: We have a problem with non-syntactically valid JSON and the syntax highlighting library

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I will check if that works

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: +1 to start the section with better wording

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I will check if an anchor is supported, it should be

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: that's something Fabre will show and I'm sure apiary will show in their standard rendering as well

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: that's one more time an Apiary rendering limitation and not a limitation of the format

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: However, there could be people that land into the apiary rendering without looking to FABRE rendering (in the same way there could be people that land into the Markdown documentation without looking to readthedocs rendering). Both renders (apiary and FABRE) should work fine, avoiding that making one happy ends with the other broken/damaged.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The problem is that the current approach is "flatering" sections that was original thought with an structure (one section for NGSI9 terms and another section for NGSI10 terms, each one wich subsections). A possible solution would be:

```
## Terminology for context data modelling and exchange
(subsections using ###)
## Terminology for context availability information
(subsectiosn using ###)
```
but I don't like it too much...

@tobiasjacobs feedback is valuable here, as he was the one proposing these sections.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: If we confirm that apiary rendering shows that, then it is ok. Otherwise, I'm afriad we are in a case in which improving one rendering damages the other and that should be avoided.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: However, if we move entityId definition under the ### sections, that makes FABRE and Apiary rendering to work?

doc/apiary/v2/fiware-ngsiv2-reference.apib
tobiasjacobs: From my point of view it is ok to have it flattened like suggested by @fgalan. 
I have not been following the formatting discussion too much, but in general I can support the approach to avoid having a too deep subsections tree.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: as we agreed today f2f for the current Apiary rendering convenience we are going to define parameters at verb level, even though they could be defined at resource level

fixed in https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed in https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I have kept the current section division .fixed in https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I have added a textual clarification. As I said we need a correct JSON in order to allow Fabre to highlight properly. fixed https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: yes, and fixed as per our conversation f2f today

https://github.com/jmcanterafonseca/fiware-orion/commit/01c2427080c7f44b53a89e26ce38de3faf33d990

jmcanterafonseca:
@fgalan version ready to be reviewed, thanks for the thorough review it improved a lot the final result

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The FABRE tool should adpat to the input it has to process and not the oposite, but I think we can live with this ;)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -240,63 +275,11 @@ Error list (HTTP response code in paranthesis):
 * LengthRequired (411). Zero/No Content-Length in PUT/POST/PATCH request
 * RequestEntityTooLarge (413). Payload too large
 * UnsupportedMediaType (415). Request content type is not supported.
-* InvalidModification (422). Some piece of information is missing in payload.
```
fgalan: Probably it doesn't matter as the Error List will be rewritten but... why the 422 has been removed in this PR?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -240,63 +275,11 @@ Error list (HTTP response code in paranthesis):
...
-<p>A query expression, composed of a list of statements separated by <code>;</code>, i.e.</p>
+## Simple Query Language
 
-<p><code>
-q=statement;statements;statement
-</code></p>
+The Simple Query Language provides a simplified syntax to retrieve entities which match a set of conditions. The
+supported syntax is:
 
 <p>Each statement express a matching condition. The result of the query is the match on each one of the statements, e.g. a
```
fgalan: Before expalining what an statement is, you have to explain that the syntax for the `q` parameters is `statement1;statement2;...;statementN` (that was already in the origininal .apib, but I guess it gets lost on the cut-and-paste operation).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: There is no separation between `geometry` and the `q` statement explanation. Thus, the reader may think (wrongly) that `geometry` is a kind of `q` statement. Let's remember that the list entities operations support three diferent parameters: `q` (make of statements), `geometry` and `coords`. I would suggest to describe `q` in a sub-section of the "Simple Query Language" section, then the geometry parameters (`geometry` and `coords`) in other sub-section.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I'm still seeing "see below" tokens that should be "see above" or "see section on Simple Query Language"...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -425,11 +422,34 @@ Response code:
...
+    Retrieve entities which type match one of the elements in the list.
+
+    + idPattern (optional, string) - A correctly formated regular expression.
+    Retrieve entities which ID matches the regular expression.
+    
+    + q (optional, string) - A query expression, composed of a list of statements separated by <code>;</code>,
+    i.e. q=statement;statements;statement. See [Simple Query Language specification](#simple_query_language). 
+    
+    + geometry (optional, string) - Defines a geografical area so only the entities located in
+    that area matches the query. See Simple Query Language specification.
```
fgalan: You should use `[Simple Query Language specification](#simple_query_language)` also here.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I'm realizing that `coords` is not mentioned in the current section... maybe it is intentional describing only `geometry`? (I tend to see `coords` highly tied to `geometry`, i.e. geometry doesn't make sense without coords or vice-versa).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -646,9 +646,6 @@ Response:
 * Successful operation uses 204 No Content
 * Errors use a non-200 and (optionally) an error payload. See subsection on Error Responses for more details.
 
-+ Parameters
```
fgalan: This has not been "recoved" from the resource level to the verb level (i.e. the APIary render doesn't show it at the right panel as required parameter).

fgalan:
LGTM

Once the branch gets squashed and rebased, I will merge it.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: oh that's unfortunate an spurious change

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed by including the Geographical Query Language

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed

jmcanterafonseca:
@fgalan thanks, I think we can land

PR #1235, opened 2015-09-15
All functests but coap_basic now work again under valgrind

kzangeli:
Made the last changes (there was no time to include this in the release 0.24.0) to fix the failing functests under the valgrind suite.

Now only coap_basic fails under valgrind.

### DIFF
--- a/test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
+++ b/test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
@@ -67,6 +67,13 @@ echo
 echo "+++++ 5: four service paths"
 orionCurl --url /v1/queryContext --payload "$payload" --servicePath '/home/kz/1, /home/kz/2, /home/kz/3, /home/kz/4'
 echo
+
+
+#
+# When started under valgrind, a "GET /version" is issued to make sure the broker is running, and this gives us ONE EXTRA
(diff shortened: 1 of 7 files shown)


test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
```diff
@@ -68,7 +68,7 @@ echo "+++++ 5: four service paths"
 orionCurl --url /v1/queryContext --payload "$payload" --servicePath '/home/kz/1, /home/kz/2, /home/kz/3, /home/kz/4'
 echo
 echo "+++++ 6: service paths from log file"
-grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }'
+grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }' | grep -v '/#'
```
fgalan: I would suggest to add an explanation about the grep, something like this (above the grep line):

````
# grep line is needed to make this test work under valgrind, as the /# is adding noise in the case of GET /version operations used to check CB in valgrind suite
```

or a better wording :)

kzangeli: Fixed in f465e69

fgalan:
LGTM

Issue #1198, opened 2015-09-01
[BUG] we can update multiples entities in update by ID


iariasleon:
We have two entities with the same entity_id: 
ex:
```
POST http://<host>:<port>/v2/entities
Fiware-Service: happy_path
Fiware-ServicePath: /test
{"type":"room_1", "id": "room_10", "my_attr": "34"}
```
```
POST http://<host>:<port>/v2/entities
Fiware-Service: happy_path
Fiware-ServicePath: /test
{"type":"room_2", "id": "room_10", "attr_test": "45"}
```
And try to update an entity by ID
```
POST http://<host>:<port>/v2/entities/room_10
Fiware-Service: happy_path
Fiware-ServicePath: /test
{"new_attr": "14"}
```
Both entities are updated:
ex in mongo:
```
{ "_id" : { "id" : "room_10", "type" : "room_1", "servicePath" : "/test" }, "attrNames" : [ "my_attr", "new_attr" ], "attrs" : { "my_attr" : { "type" : "", "creDate" : 1441095072, "modDate" : 1441095072, "value" : "34" }, "new_attr" : { "value" : "14", "type" : "", "creDate" : 1441095097, "modDate" : 1441095097 } }, "creDate" : 1441095072, "modDate" : 1441095097 }
```
```
{ "_id" : { "id" : "room_10", "type" : "room_2", "servicePath" : "/test" }, "attrNames" : [ "attr_test", "new_attr" ], "attrs" : { "attr_test" : { "type" : "", "creDate" : 1441095085, "modDate" : 1441095085, "value" : "45" }, "new_attr" : { "value" : "14", "type" : "", "creDate" : 1441095097, "modDate" : 1441095097 } }, "creDate" : 1441095085, "modDate" : 1441095097 }
```
In this case the response expected:
Status:  409 Conflict
```
{
error: "TooManyResults"
description: "There is more than one entity with that id. Refine your query."
}
```

fgalan:
Due to mongoBackend implications, @fgalan could be the best candidate for assigning this.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1240. @iariasleon please hava a look and close the issue if you find it solved.

iariasleon:
LGTM
```
{"error":"TooManyResults","description":"There is more than one entity that match the update. Please refine your query."}
```

Issue #1199, opened 2015-09-01
[BUG] some errors have leftover whitespaces


iariasleon:
```
| returned error                         | expected                                     |
|---------------------------------------------------------------------------------------|
| Request Entity Too Large               | RequestEntityTooLarge                        |
| Unsupported Media Type                 | UnsupportedMediaType                         |
| Length Required                        | LengthRequired                               |
```

rbl219:
This issue has been resolved in #PR 1239. @iariasleon, please have a look to it and close it if you find is solved

iariasleon:
LGTM

PR #1240, opened 2015-09-16
ADD checking for entities number in NGSIv2 updates

fgalan:
Fix for #1198 

1 new functional test added.
2 new unit tests added

100% pass in unit test and functional tests

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2208,7 +2208,7 @@ void processContextElement
   const std::vector<std::string>&      servicePathV,
   std::map<std::string, std::string>&  uriParams,   // FIXME P7: we need this to implement "restriction-based" filters
   const std::string&                   xauthToken,
-  const std::string&                   caller
+  const std::string&                   apiVersion
 )
 {
   DBClientBase* connection                  = NULL;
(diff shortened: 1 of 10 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2302,6 +2302,56 @@ void processContextElement
 
   auto_ptr<DBClientCursor> cursor;
 
+  // This block is to avoids that several entities with the same ID gets updated at the same time, which is
+  // not allowed in NGSIv2. The if is needed, as multiple update has been allowed in NGSIv1 (maybe without
+  // thinking too much about it, but NGSIv1 behaviour now will break backward compatibility)
+  if (apiVersion == "v2")
```
fgalan: I don't like too much using a second operations (`count()`) but I'm not sure the best way of doing. There is a currently open question about this in SOF (http://stackoverflow.com/questions/21832407/mongodb-logging-write-read-queries)

However, I think that the PR can be progressed without getting an answer to that. There are some other cases of count+query in the code and we could fix them all in the same shoot (in this case, I'd add this as technical debt).

src/lib/rest/HttpStatusCode.cpp
```diff
@@ -36,6 +36,7 @@ std::string httpStatusCodeString(HttpStatusCode code)
   case SccContextElementNotFound:            return "No context element found"; // Standard HTTP for 404: "Not Found"
   case SccBadVerb:                           return "Method Not Allowed";
   case SccNotAcceptable:                     return "Not Acceptable";
+  case SccConflict:                          return "TooManyResults";
```
fgalan: By the moment (see http://telefonicaid.github.io/fiware-orion/api/v2/), Conflict (HTTP code 409) corresponds only to "TooManyResults" error text. In the future, it may correspond with several error texts (as 400 may correspond to ParseError and BadRquest) and this part may break. However, lest's wait until that situation happens in order to fix.

src/lib/mongoBackend/MongoCommonUpdate.cpp

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2302,6 +2302,56 @@ void processContextElement
 
   auto_ptr<DBClientCursor> cursor;
 
+  // This block is to avoids that several entities with the same ID gets updated at the same time, which is
```
kzangeli: avoids -> avoid
gets -> get

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2302,6 +2302,56 @@ void processContextElement
...
+      LM_T(LmtMongo, ("count() in '%s' collection: '%s'",
+                      getEntitiesCollectionName(tenant).c_str(),
+                      query.toString().c_str()));
+
+      connection = getMongoConnection();
+      entitiesNumber = connection->count(getEntitiesCollectionName(tenant).c_str(), query);
+      releaseMongoConnection(connection);
+
+      LM_I(("Database Operation Successful (%s)", query.toString().c_str()));
+    }
```
kzangeli: I would propose to keep at a minimum what's inside 'try'.
- LM_T before, LM_I after (both catch do 'return').

Does getMongoConnection() throw exception?
If not, I would put it before, AND check 'if (connection == NULL) { ERROR + return } (or something similar).

src/lib/rest/HttpStatusCode.cpp
kzangeli: All older texts use spaces.
What is correct here?

rbl219: In issue #1199 I have solve this question including an if in OrionError.cpp, for example :

  else if (_reasonPhrase == "Length Required")
  {
    return "LengthRequired";
  }


And maintain White spaces in HttpStatusCode.cpp

  case SccLengthRequired:                    return "Length Required";

AS before has been mede with SccBadRequest

fgalan: Thanks @rbl219 ! I would use that function.

kzangeli: Ah, of course. I remember. 
We need one 'correct English phrase' and one 'enum-like string' ...

kzangeli: So, add spaces in the current place and add the conversion in OrionError::errorStringForV2()

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in bda9795.

Note I'm using a "Fatal Error" for that, given that if getMongoConnection() returns NULL something really bad is happening in the DB pool management logic (I have event though if that should cause the broker end, with LM_X).

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in bda9795.

src/lib/rest/HttpStatusCode.cpp
fgalan: Fixed in bda9795.

kzangeli:
LGTM

fgalan:
Rebased. Ready to merge, pls!

rbl219:
LGTM

PR #1242, opened 2015-09-17
ADD cleanFiles.sh script

fgalan:
A little script that may be useful to clean up the directory after failing testHarness.

It is deliverately simple. Please, don't look for `-v` or `-u` options ;)

### DIFF
--- a/test/functionalTest/cleanFiles.sh
+++ b/test/functionalTest/cleanFiles.sh
@@ -19,20 +19,20 @@
 # For those usages not covered by this license please contact with
 # iot_support at tid dot es
 
-# This script remove all the crap files that testHarness or valgrind suite 
-# use to produce in some cases (e.g. failing tests)
+# This script removes all the crap files that testHarness or valgrind suite 
+# produce in some cases (e.g. failing tests)
 
-find -name *.out -exec rm {} \;
-find -name *.diff -exec rm {} \;
-find -name *.regexpect -exec rm {} \;
-find -name *.shell -exec rm {} \;
-find -name *.shell.stdout -exec rm {} \;
-find -name *.shell.stderr -exec rm {} \;
-find -name *.shellInit -exec rm {} \;
-find -name *.shellInit.stderr -exec rm {} \;
-find -name *.shellInit.stdout -exec rm {} \;
-find -name *.teardown -exec rm {} \;
-find -name *.teardown.stderr -exec rm {} \;
-find -name *.teardown.stdout -exec rm {} \;
-find -name *.log -exec rm {} \;
-find . -type d -empty -exec rmdir {} \;
+find cases/ -name *.out -exec rm {} \;
+find cases/ -name *.diff -exec rm {} \;
+find cases/ -name *.regexpect -exec rm {} \;
+find cases/ -name *.shell -exec rm {} \;
+find cases/ -name *.shell.stdout -exec rm {} \;
+find cases/ -name *.shell.stderr -exec rm {} \;
+find cases/ -name *.shellInit -exec rm {} \;
+find cases/ -name *.shellInit.stderr -exec rm {} \;
+find cases/ -name *.shellInit.stdout -exec rm {} \;
+find cases/ -name *.teardown -exec rm {} \;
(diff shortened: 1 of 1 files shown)


LeandroGuillen:
Cool! Although so many _rm_ make me shiver :)

LGTM

test/functionalTest/cleanFiles.sh
```diff
@@ -0,0 +1,38 @@
...
+find -name *.shell.stdout -exec rm {} \;
+find -name *.shell.stderr -exec rm {} \;
+find -name *.shellInit -exec rm {} \;
+find -name *.shellInit.stderr -exec rm {} \;
+find -name *.shellInit.stdout -exec rm {} \;
+find -name *.teardown -exec rm {} \;
+find -name *.teardown.stderr -exec rm {} \;
+find -name *.teardown.stdout -exec rm {} \;
+find -name *.log -exec rm {} \;
+find . -type d -empty -exec rmdir {} \;
```
kzangeli: Perhaps you should do this under the 'test' directory only, and make sure you are under 'test' before continuing.
Especially the patterns *.out and *.log might hit files that we don't want to remove.

test/functionalTest/cleanFiles.sh
```diff
@@ -0,0 +1,38 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# This script remove all the crap files that testHarness or valgrind suite
```
LeandroGuillen: remove -> removes

test/functionalTest/cleanFiles.sh
```diff
@@ -0,0 +1,38 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# This script remove all the crap files that testHarness or valgrind suite 
+# use to produce in some cases (e.g. failing tests)
```
LeandroGuillen: use to produce -> produce

test/functionalTest/cleanFiles.sh
fgalan: What about checking that there is a subdirectory named cases? Not 100% save, but I'd say that 99% ;) 

```
find cases/ -name ...
```

test/functionalTest/cleanFiles.sh
fgalan: Fixed in 8b08bde

test/functionalTest/cleanFiles.sh
fgalan: Fixed in 8b08bde

test/functionalTest/cleanFiles.sh
fgalan: Fixed (using the `case/` solution) in 8b08bde

kzangeli: Sure, good enough.
But, you must be on toplevel and do

    find test/functionalTests/cases ...

fgalan: The typical usage case is:

```
cd test/functionaTests
./testHarness
(something bad happens, a lot of crap file appear under cases/ subdirectories)
./cleanFiles.sh
```

Using `find test/functionalTests/cases ...` would break that case, isn't it? Or maybe I'm not getting the point...

LeandroGuillen:
The Real LGTM

rbl219:
LGTM

test/functionalTest/cleanFiles.sh
kzangeli: I have test/functionalTests in my PATH, I always execute testHarness.sh from TOP level ...
Not important, I will not use this script. testHarness.sh itself does the cleaning before running a new test ... And the old files under test/functionalTests don't really bother me

Issue #1164, opened 2015-08-13
Clearer error on "PUT /v2/entities/<entityId>/attrs/<atrtrName>" without 'value' in payload


kzangeli:
A `PUT /v2/entities/<entityId>/attrs/<atrtrName>` without { "value": "xxx" } in the payload gives an error. The error message is not clear at all and should be improved.

The error is due to the broker demands an attribute to have a value. Doing a PUT, the entire state of the attribute is replaced by the new state and if no "value" is present ... then it is an error.

For now it seems the broker will keep working this way and we just need to return a clearer response. However, another approach would be to loosen the restriction on 'value' mandatory for attributes and let the request through.

**Effort: 1 man day**

fgalan:
The error will use "422 unprocessable entity" as HTTP status code and "InvalidModification" in the `error` text. The `description` feld would include an explanation about value is missing.

The .apib for v2 has to be edit as part of this issue to include the new error in the documentation.

rbl219:
With code modifications currently implemented to resolve this issue,  generated error response will be:

'''
HTTP/1.1 422 Unprocessable Entity
Content-Length: 96
Content-Type: application/json
Date: Wed, 16 Sep 2015 18:13:34 GMT

{"error":"Invalid Modification","description":"Empty attribute not allowed in APPEND or UPDATE"}
'''

You consider this response is correct, or some improvements should be made?

fgalan:
Just to know... before modifications, which error message was being generated?

fgalan:
Some recomendations on the new error message:

* Use `InvalidModifications` (without whitespace)
* I

rbl219:
Before modifications error was:



HTTP/1.1 472 Unknown
Content-Length: 0
Date: Wed, 16 Sep 2015 15:10:11 GMT

kzangeli:
Just one comment on this ...
We are now to allow requests with attributes without 'value'.
Perhaps it should be allowed here too ?

fgalan:
We are allowing request without 'value' in operations where it make sense (e.g. PATCH), but in this case the `PUT /v2/entities/<entityId>/attrs/<atrtrName>` semantics are "throw away value/type/metadata of the attribute and replace it by the value/type/metadata in the payload of the request". Given that 'value' is mandatory in NGSI, this particular operation has to have a 'value' field.

[...]

PR #1245, opened 2015-09-17
hardening/errorAttrsWithoutValue

rbl219:
Resolved issue   #1164

Clearer error on "PUT /v2/entities/<entityId>/attrs/<atrtrName>" without 'value' in payload


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -240,6 +240,7 @@ Error list (HTTP response code in paranthesis):
 * LengthRequired (411). Zero/No Content-Length in PUT/POST/PATCH request
 * RequestEntityTooLarge (413). Payload too large
 * UnsupportedMediaType (415). Request content type is not supported.
+* InvalidModification (422). Request payload is missing some piece of information.
```
fgalan: Maybe is better English to say?

"Request payload misses some piece of information"

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -240,6 +240,7 @@ Error list (HTTP response code in paranthesis):
 * LengthRequired (411). Zero/No Content-Length in PUT/POST/PATCH request
```
fgalan: CNR file entry corresponding to this PR is missing.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2239,7 +2240,7 @@ void processContextElement
       {
         ContextAttribute* ca = new ContextAttribute(aP);
 
-        buildGeneralErrorResponse(ceP, ca, responseP, SccInvalidParameter,
+        buildGeneralErrorResponse(ceP, ca, responseP, SccInvalidModification,
```
fgalan: Interesting point to think about it... :)

This change introduces a slight backward compability break, as we are returning 422 in places where 472 was returned previously. However, in this case we are moving from a general error (invalid parameter, in general) to a more specific error (invalid modification), so I think is actually an improvement.

We could avoid this and preserve previous NGSIv1 behaviour using the "if (apiVersion == v2)" technique already used in other PRs. However, in this particular case I think we should leave the modification as @rbl219 suggest.

Opinions?

test/functionalTest/cases/0000_ngsi10_update/update_sequence_empty_value_in_context_value.test
```diff
@@ -72,8 +72,8 @@ Date: REGEX(.*)
         </contextAttributeList>
       </contextElement>
       <statusCode>
-        <code>472</code>
-        <reasonPhrase>request parameter is invalid/not allowed</reasonPhrase>
+        <code>422</code>
+        <reasonPhrase>Invalid Modification</reasonPhrase>
         <details>action: APPEND - entity: [OfficeRoom, Room, false] - offending attribute: pressure - empty attribute not allowed in APPEND or UPDATE</details>
       </statusCode>
     </contextElementResponse>
```
fgalan: Could you include a test harness to test the fixing in the NGSIv2 API, pls? Contact me if you have doubt about it :)

doc/apiary/v2/fiware-ngsiv2-reference.apib
rbl219: Fix in commit ed18967

test/functionalTest/cases/0000_ngsi10_update/update_sequence_empty_value_in_context_value.test
rbl219: ok, I will include new test harness

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I would vote for not breaking backward compatibility ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: Better would be to turn it around:

Some piece of information (is) missing in payload

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: After a skype discussion, at the end we will keep the code as it is now, due to:

* Backward compabilitiy is really really small
* It involves an improvement
* Keep backward compability infolves extra work
* Kepp backward compability involves "dirty" code due to the "if (apiVersion == v2)"  exceptions.

However, let's make this explicit in the CNR, with an additiona entry like this:

"Fix: chaned error code from 472 (Invalid Parameter) changed to 422 (Invalid Moficiation) in NGSIv1 in the case of missing attribute value (#1164)"

test/functionalTest/cases/0000_ngsi10_update/update_sequence_empty_value_in_context_value.test
rbl219: Fix in commit 6d6b824

src/lib/mongoBackend/MongoCommonUpdate.cpp
rbl219: Fix in commit 14fa438

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed.

fgalan:
LGTM (but unable to merge yet due develop need to be merged...)

kzangeli:
LGTM, after pending merge ...

rbl219:
Merged in commit ff5521c

Issue #1188, opened 2015-08-25
Update an attribute to empty string


crbrox:
Changing an attribute to empty string ("") gives a 472 Unknown

Try https://github.com/telefonicaid/fiware-orion/blob/dbb0bfbad282132317fa28aba864f5e3905f397c/test/functionalTest/cases/1185_udpate_change_value_type/update_empty_string.test

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1247

Issue #1187, opened 2015-08-25
Creating an entity with attribute value empty string


crbrox:
POSTing an entity whit attr "" returns 200 OK, but the entity is not created (or not GETable)

Try https://github.com/telefonicaid/fiware-orion/blob/dbb0bfbad282132317fa28aba864f5e3905f397c/test/functionalTest/cases/1185_udpate_change_value_type/create_empty_string.test

fgalan:
This is maybe related with #1188. Probably it should be addressed after #1188 get solved.

fgalan:
Fixed along with #1188

PR #1250, opened 2015-09-18
Bug/1187 empty attribute in create

kzangeli:
The issue #1187 was fixed when #1188 was fixed.
This PR is just to add a functest for #1187 and to modify CNR.


fgalan:
LGTM

PR #1171, opened 2015-08-13
Feature/1027 get all entity types

kzangeli:
Implements #1027 

GET /v2/type

This request has two URI params:

* options=count
* options=values

*count* is implemented, but as **options** itself is not implemented yet (we need to save a vector for all the options in *options*, and a lookup function), the implementation isd based on a simple comparison:

    uriParamV["options"] == "count"

Good enough for now, until **options** itself is implemented.

**options=values** is not implemented.

Two issues created: #1169 and #1170

PR #1251, opened 2015-09-18
Use options=count when getting entities

### LINKED ISSUE #1041: options=count
(Originally included as comment on https://github.com/telefonicaid/fiware-orion/issues/947)

After latest discussion, the `count` option (within the `options` parameter) will be used instead of `count=true`. The code needs to be fixed (PR #958 was based on `count=true`).

crbrox:
Use options=count when getting entities, closes #1041

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -5,3 +5,4 @@ Fix:  Avoid updating several matching entities associated to the same ID, return
 Fix:  Permitting attribute values to be empty in API v2 (Issue #1187 and #1188)
 Fix:  Clearer error on NGSIv2 entity attribute update, without value in payload (Issue #1164)
 Fix:  Changed  error code from 472 (Invalid Parameter) to 422 (Invalid Moficiation) in NGSIv1 in the case of missing attribute value (Issue #1164)
+Add:  Using 'count' option when getting entities (#1041)
--- a/src/lib/serviceRoutinesV2/getEntities.cpp
+++ b/src/lib/serviceRoutinesV2/getEntities.cpp
@@ -70,11 +70,6 @@ std::string getEntities
   string    idPattern  = ciP->uriParam["idPattern"];
   string    q          = ciP->uriParam["q"];
 
-  if (ciP->uriParam["options"] == "count")
-  {
-    ciP->uriParam["count"] = "true";
-  }
-
   if ((idPattern != "") && (id != ""))
   {
     OrionError oe(SccBadRequest, "Incompatible parameters: id, IdPattern");
--- a/src/lib/serviceRoutinesV2/getEntities.cpp
+++ b/src/lib/serviceRoutinesV2/getEntities.cpp
@@ -70,6 +70,11 @@ std::string getEntities
   string    idPattern  = ciP->uriParam["idPattern"];
   string    q          = ciP->uriParam["q"];
 
+  if (ciP->uriParam["options"] == "count")
+  {
+    ciP->uriParam["count"] = "true";
+  }
+
   if ((idPattern != "") && (id != ""))
   {
     OrionError oe(SccBadRequest, "Incompatible parameters: id, IdPattern");
--- a/test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities_pagination.test
+++ b/test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities_pagination.test
@@ -35,7 +35,7 @@ brokerStart CB
 # 03. POST /v2/entities, created third  entity
 # 04. POST /v2/entities, created fourth entity
 # 05. POST /v2/entities, created fifth  entity
-# 06. GET  /v2/entities?limit=2&count=true
+# 06. GET  /v2/entities?limit=2&options=count,normalized
 # 07. GET  /v2/entities?offset=2
 #
 
@@ -99,9 +99,9 @@ echo
 echo
 
 
-echo "06. GET /v2/entities?limit=2&count=true"
-echo "======================================="
-orionCurl --url '/v2/entities?limit=2&count=true' --json
+echo "06. GET /v2/entities?limit=2&options=count,normalized"
+echo "====================================================="
+orionCurl --url '/v2/entities?limit=2&options=count,normalized' --json
 echo
 echo
 
@@ -160,8 +160,8 @@ Date: REGEX(.*)
 
 
 
-06. GET /v2/entities?limit=2&count=true
-=======================================
+06. GET /v2/entities?limit=2&options=count,normalized
+=====================================================
 HTTP/1.1 200 OK
 Content-Length: 49
 Content-Type: application/json
--- a/src/lib/serviceRoutines/postQueryContext.cpp
+++ b/src/lib/serviceRoutines/postQueryContext.cpp
@@ -311,7 +311,7 @@ std::string postQueryContext
   // In API version 2, this has changed completely. Here, the total count of local entities is returned
   // if the URI parameter 'count' is set to 'true', and it is returned in the HTTP header X-Total-Count.
   //
-  if ((ciP->apiVersion == "v2") && (ciP->uriParam["count"] == "true"))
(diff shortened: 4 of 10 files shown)


src/lib/serviceRoutinesV2/getEntities.cpp
```diff
@@ -70,6 +70,11 @@ std::string getEntities
   string    idPattern  = ciP->uriParam["idPattern"];
   string    q          = ciP->uriParam["q"];
 
+  if (ciP->uriParam["options"] == "count")
```
kzangeli: Not sure this is enough ....

What if option == 'xxx,count', or 'count,yyy' ?

I think there is a vector somewhere, to do something similar to:

    if (ciP->uriParamOptions['check'] == "true")
    {
       ...
    }

And, if there isn't, we'll have to implement it ...

kzangeli: Found it :-)

uriParamOptionsParse() in ConnectionInfo.cpp is called by uriArgumentGet() in rest/rest.cpp.
There, all options in the URI param 'options' are put in the vector ciP->uriParamOptions.

So, I would propose to use

    if (ciP->uriParamOptions["count"] == true)

wherever needed, and probably remove these four lines.
[ It is a little ugly also to add parameters to ciP->uriParam ... it's like lying ... ]

crbrox: I thought we were doing only one step after another one, I followed getEntityAllTypes
https://github.com/telefonicaid/fiware-orion/blob/develop/src/lib/serviceRoutinesV2/getEntityAllTypes.cpp#L59
which has an options=value possibility in api doc. As said in your PR #1171
"Good enough for now, until options itself is implemented.
options=values is not implemented."

kzangeli: In postQueryContext we have the following line:

    if ((ciP->apiVersion == "v2") && (ciP->uriParam["count"] == "true"))

Now, as there are **yet another** way to ask for count, we'd need to add the following code:

     if ((ciP->apiVersion == "v2") && (ciP->uriParamOptions["count"] == true))
    {
      countP = &count;
    }

With this, it's all taken care of in postQueryContext(), so no code needed in getEntities()

crbrox: OK, using uriParamOptions

CHANGES_NEXT_RELEASE
```diff
@@ -3,3 +3,5 @@ Fix:  Changed the description text for parse errors in v2 json requests (Issue #
 Fix:  Remove leftover whitespaces in error messages (Issue #1199) 
 Fix:  Avoid updating several matching entities associated to the same ID, returning TooManyResults error in that case (Issue #1198)
 Fix:  Permitting attribute values to be empty in API v2 (Issue #1188)
+Add:  Using options=count when getting entites (#1041)
```
fgalan: entites -> entities

src/lib/serviceRoutinesV2/getEntities.cpp
crbrox: Again, that was not the approach in your PR #1171, https://github.com/telefonicaid/fiware-orion/blob/develop/src/lib/serviceRoutinesV2/getEntityAllTypes.cpp#L59
I'll do it thé way you say

fgalan: `if (ciP->uriParam["options"] == "count")` seems to be something to be fixed (to use uriParamOptions[]), as I understand that it would fails if the client uses for examples `...?options=count,otherop`.

CHANGES_NEXT_RELEASE
fgalan: In addition, options=count may lead to think that options= has to be used literarly that way. Maybe is more precise:

Using 'count' option when getting entities.

src/lib/serviceRoutinesV2/getEntities.cpp
kzangeli: Yes, we need to fix these 'mistakes', we have them in earlier code also.
I would propose a new issue about it

fgalan: Actually, #1041 is about fixing this in all places, so in my opinion we don't need any extra issue.

crbrox: In that case, I think reference to #958  inside #1041 can be misleading.  I would change the text for this self PR also

fgalan: Yes, probably is misleanding, sorry...

What I tried to mean in #1041 is that in PR #958 "get all entities" operation was implemented using `count=true` and that that implementation needs to be fixed to use `count` option (not exclusively: I mean, it could happen that other operations appar from "get all entities" are currently using `count=true`, in that case, they also need to be fixed to use `count` option).

Bottom line: at the end only `count` option will be the valid way of getting the count in any NGSIv2 operation in which counting makes sense and this PR should address that (of course, only for currently implemented operations, e.g. list subscription is not).

kzangeli: About "Good enough for now" ...
That statement comes from before ciP->uriParamOptions was implemented.

But ... now it is implemented. That changes everything, right?

And of course, we need to fix those older 'good enough for then' to use uriParamOptions. It is not good enough any more.

I just added this to issue #1041

crbrox: Well, that comment was before I read your message about your "found" of the uriParamOptionsParse, which you don't remember if existed. I think both comments have crossed. *Obviously*, if the options param is parsed it should be used the parsed form. The strange thing  is not to use the parse function before implementing the version in getEntitiesAllTypes being all commits so close in time, both on Aug 14. (a3fa6aa, 38fb8ae) (not such older the 'raw' use of options)

The original issue #1041 talked about changing from ```count=true```to ```options=count```(in getEntities, could be understood) so the use in getEntitesAllTypes looked like a similar issue. It did not talk about allowing multiple values for options (which, I think, is not supported anywhere in the code and it's a orthogonal issue to changing count=true->options=count). So the quote of "good enough" wanted to express that this issue was not the best place to implement the parsing of options param, perhaps, in case it had to do from zero (as you thought for a3fa6aa)

I think changing the code for using the parsed form of options should be an issue by itself. As you wrote, it is not only for "count" but "anything" and the original issue was not related to things like "text" or "values" or any other options value. I would use the parsed form for 'options` in this issue, maybe for all cases of "count=true" in the code but I think it will break the "integrity" of  the PR if we changed a lot of different unrelated, in other way, files. Strictly speaking the change in getEntitesAllTypes should not be covered by this issue, but by another issue for refactoring of "options" param use

crbrox: Fix 5c9ef1a41b35af8e13c0f49166970700ee632d41

CHANGES_NEXT_RELEASE
crbrox: Fixed 5c9ef1a41b35af8e13c0f49166970700ee632d41

crbrox:
Ready for final review/approval
@fgalan 
@kzangeli

kzangeli:
LGTM

test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities_pagination.test
```diff
@@ -35,7 +35,7 @@ brokerStart CB
 # 03. POST /v2/entities, created third  entity
 # 04. POST /v2/entities, created fourth entity
 # 05. POST /v2/entities, created fifth  entity
-# 06. GET  /v2/entities?limit=2&count=true
+# 06. GET  /v2/entities?limit=2&options=count,normalized
```
fgalan: Probably 'normalized' will not be used at the end after the end, but it is ok to use it in this test as "random" option to check that the ability to mix count with other options works.

NTC

fgalan:
LGTM

Issue #1177, opened 2015-08-14
geometry= and coords= in GET /v2/entities


fgalan:
As described in http://telefonicaid.github.io/fiware-orion/api/v2/

Similarly to `q`, the gemoetry + coord specification (both come always together: an error should be returned in negative case), the idea is to build an Scope with them, then pass the query with that scope to mongoBackend.

The scope type to use is: "FIWARE::Location".

fgalan:
Implemented in PR #1177

PR #1254, opened 2015-09-20
Fixed issue 1177

kzangeli:
Added URI params **geometry** and **coords** for **GET /v2/entities"".
Implements issue #1177.


src/lib/parse/forbiddenChars.cpp
```diff
@@ -61,3 +61,36 @@ bool forbiddenChars(const char* s)
...
   return false;
 }
+
+
+
+/* ****************************************************************************
+*
+* forbiddenCharsInUriParam - 
+*/
+bool forbiddenCharsInUriParam(const char* s)
```
fgalan: The name forbiddenCharsInUriParam() is too generic, although the "straighforward" alternative is ugly: forbiddenCharsInGeomAndCoords().

Maybe forbiddenCharsInUriParamWithoutSemicolon() is better?

test/functionalTest/cases/0967_v2_query_op/illegal_chars_in_url_param.test
```diff
@@ -79,20 +77,6 @@ echo
 echo
```
fgalan: Maybe the deleted tests should be "recovered" once we have adjust the "semicolon relaxion" to only coords= and geometry= ?

kzangeli: Ah, yes, of course. I had that in mind 15 minutes beofre implementing the change. Then I forgot!

kzangeli: Fixed in 4610846

src/lib/parse/forbiddenChars.cpp
kzangeli: The equal sign '=' is removed also. forbiddenCharsInUriParamWithoutSemicolonAndEqualSign? :-D

I would simply defer this to the new issue #1255 ...
With that issue implemented this function will be removed.

fgalan: It could pass a long time until we implement #1255... So, it is better to have this fixed, to avoid "misleading" names in functions in the code (especially for developers that no have been involved in this PR and who know this discussion :).

If we don't have better alternative than forbiddenCharsInUriParamWithoutSemicolonAndEqualSign(), let's use that.

kzangeli: I have a better idea.

We add a parameter 'char* exceptions' to forbiddenChars, that defaults to the empty string (or NULL).

Where forbiddenCharsInUriParam is called, we call instead forbiddenChars("STRING", ";="), that way telling forbiddenChars to admit ';' and '='.  This easily implemented using **strchr**.

ok?

fgalan: +1

kzangeli: Fixed in 513e7b5

fgalan:
LGTM

Issue #1175, opened 2015-08-14
Redirect standard output and errors of the Orion service


dmoranj:
The command to start Orion in the RPM service currently looks like this:
```
su $BROKER_USER -c "${EXECUTABLE} ${BROKER_OPTS} &> /dev/null
```
It redirects the standard output and error to null. This should be changed, in order to redirect all this information to a secondary log in the `/var/log/contextBroker` folder.

**Effort: 1 man day**

fgalan:
Currently contextBroker.log contains the same that contextBroker bimnary prints in stdout (using a different line format), so we only need stderrr. We could use contextBroker.stderr as name.

This contextBroker.stderr is not a regular log file and it is expected to be small, so I don't think we need associated logrotate configuration.

My only doubt right now is if we should append or clean the file eacth time the contextBroker service is restarted...

kzangeli:
It should be enough with saving one old copy of the file, like this:

    mv contextBroker.stderr contextBroker.stderr.old
    contextBroker 2> contextBroker.stderr

fgalan:
Fixed in https://github.com/telefonicaid/fiware-orion/pull/1270

Issue #309, opened 2014-04-09
centos6:  cat: /var/log/contextBroker/contextBroker.pid: No such file or directory


anaselli:
I stopped context broker and dropped the database following the instructions here:
https://forge.fi-ware.org/plugins/mediawiki/wiki/fiware/index.php/Publish/Subscribe_Broker_-_Orion_Context_Broker_-_Installation_and_Administration_Guide#Delete_complete_database

when i restarted the service i got:
service contextBroker start
Starting contextBroker...                         cat: /var/log/contextBroker/contextBroker.pid: No such file or directory
Fail
but:
ps -edlaf | grep orion
1 S orion     3355     1  0  80   0 - 32515 hrtime 11:03 ?        00:00:00 /usr/bin/contextBroker -port 1026 -logDir /var/log/contextBroker -pidpath /var/log/contextBroker/contextBroker.pid -dbhost localhost -db orion
0 S root      3366  2384  0  80   0 - 25815 pipe_w 11:05 pts/0    00:00:00 grep orion

As an additional note, I'm not sure it's a problem or not, the orion db in mongo is created without any collections

**Effort: 3 man day**

fgalan:
The cat error message output is no longer printed after the fix in PR https://github.com/telefonicaid/fiware-orion/pull/1270

PR #1270, opened 2015-09-22
FIX log stderr in init script

fgalan:
Implements #1175 (and #309)

Not tested yet (I'm waiting to LGTM to test the final version).

Don't merge after LGTM (I'd check generating a RPM in a CentOS system to check that is working)

@kzangeli 
@dmoranj


etc/init.d/contextBroker.centos
```diff
@@ -91,7 +91,9 @@ broker_start()
 {
     printf "%-50s" "Starting ${NAME}..."
     if [[ -x ${EXECUTABLE} ]]; then
-        su $BROKER_USER -c "${EXECUTABLE} ${BROKER_OPTS} &> /dev/null "
+        # preserving existing error log (if it already exists)
+        su $BROKER_USER -C "mv ${BROKER_LOG_DIR}.${EXECUTABLE}.err ${BROKER_LOG_DIR}.${EXECUTABLE}.err.old 2> /dev/null || True"
```
kzangeli: -C => -c   (Windows might not be case-sensitive as it is not a real operating system ;-) but Unix/Linux is)
True => true  (same same. /bin/true is an executable that returns ... true - this is to avoid a e.g. make to stop executing if something goes wrong false || true evaluates to true).

Suggestion: ${BROKER_LOG_DIR}.${EXECUTABLE} is very long.
Adding a variable 

    logFileBase=${BROKER_LOG_DIR}.${EXECUTABLE}

would make the line easier to read ...

kzangeli: Also, I wonder if

    || true

should be outside the string to su, like this:

    su $BROKER_USER -c "mv ${BROKER_LOG_DIR}.${EXECUTABLE}.err ${BROKER_LOG_DIR}.${EXECUTABLE}.err.old 2> /dev/null" || true

Not an expert, but ... it doesn't make much sense to me otherwise.
Look it up!

dmoranj: I don't see the need of the " || True" statement, considering the result of the statement is not being tested, but I may be missing something.

dmoranj:
LGTM apart from some minor issues.

etc/init.d/contextBroker.centos
kzangeli: The only reason to use "|| true" is to avoid that the entire statement returns != true because if it dies, the execution ends there.   make works this way. contextBroker.centos is input to ... what executable?  Might work the way I describe. If so, the "|| true" makes sense, if not ... it doesn't.

fgalan: Regarding `|| True` I was thinking in the case something needs to capture the return value in the future... but you are right and it is an unneded overelaboration.

fgalan: Fixed in 85a032f

etc/init.d/contextBroker.centos
```diff
@@ -91,7 +91,10 @@ broker_start()
 {
     printf "%-50s" "Starting ${NAME}..."
     if [[ -x ${EXECUTABLE} ]]; then
-        su $BROKER_USER -c "${EXECUTABLE} ${BROKER_OPTS} &> /dev/null "
+        # preserving existing error log (if it already exists)
+        LOG_FILE_BASE=${BROKER_LOG_DIR}.${EXECUTABLE}
+        su $BROKER_USER -c "mv ${LOG_FILE_BASE}.err ${LOG_FILE_BASE}.err.old 2> /dev/null"
+        su $BROKER_USER -c "${EXECUTABLE} ${BROKER_OPTS} > /dev/null 2> ${LOG_FILE_BASE}.err"
         sleep 2 # wait some time to leave contextbroker start
```
kzangeli: Now that we're here ...

   wait some time to leave contextbroker ...   (Spanish: dejar el broker arrancar)

In English it would be "to **let** the broker start"

fgalan: Fixed in 85a032f

kzangeli:
LGTM

kzangeli:
LGTM, after merge con develop

Issue #1173, opened 2015-08-14
Fix the mess in names of types/files regarding /v2/type


kzangeli:
He have four different types to serve the requests /v2/type and /v2/type/<typeName>:

* TypeEntity
* TypeEntityVector
* EntityTypeAttributesResponse
* EntityTypesResponse

EntityTypeAttributesResponse is:

    TypeEntity
    StatusCode

EntityTypesResponse is:

    TypeEntityVector
    StatusCode

The names are **very** confusing.

I would change these type names to:

* TypeEntity => EntityType
* TypeEntityVector  => EntityTypeVector
* EntityTypeAttributesResponse => EntityTypeResponse
* EntityTypesResponse => EntityTypeVectorResponse

**Effort: 1 man day**

fgalan:
Fixed in PR #1243

Issue #1172, opened 2015-08-14
JSON encoder emits invalid strings


pjt33:
If I create an attribute with characters which need escaping in JSON strings:

	{
		"updateAction": "APPEND",
		"contextElements": [
			{
				"id": "Bug2",
				"isPattern": "false",
				"type": "Bug",
				"attributes": [
					{
						"name": "foo",
						"type": "string",
						"value": "line1\nline2"
					}
				]
			}
		]
	}

then when Orion encodes it in JSON for query responses or to notify subscribers it fails to escape:

	{
	  "contextElement" : {
		"type" : "Bug",
		"isPattern" : "false",
		"id" : "Bug2",
		"attributes" : [
		  {
			"name" : "foo",
			"type" : "string",
			"value" : "line1
	line2"
		  }
		]
	  },
	  "statusCode" : {
		"code" : "200",
		"reasonPhrase" : "OK"
	  }
	}

kzangeli:
Not sure that the broker should modify the values of the fields that is input to the broker.
What if you want to store e.g. a metadata as "newline": "\n" and then the broker goes and changes it to "\\n" ?
We will look into it, of course, but right now I don't really see how the broker could modify the data ...

pjt33:
I don't think it is modifying the data as such. I think it's correctly handling escaping on input but not on output.

fgalan:
Actually, from a JSON point fo view the followint two JSONs:

```{
    "updateAction": "APPEND",
    "contextElements": [
        {
            "id": "Bug2",
            "isPattern": "false",
            "type": "Bug",
            "attributes": [
                {
                    "name": "foo",
                    "type": "string",
                    "value": "line1\nline2"
                }
            ]
        }
    ]
}
```

and

```
{
    "updateAction": "APPEND",
    "contextElements": [
        {
            "id": "Bug2",
            "isPattern": "false",
            "type": "Bug",
            "attributes": [
                {
                    "name": "foo",
                    "type": "string",
                    "value": "line1
line2"
                }
            ]
        }
    ]
}
```

are the same. Thus, under my understanding, Orion is behaving correctly.

pjt33:
The [JSON spec](http://json.org/) specifically prohibits control characters (which includes literal newlines) from strings, requiring them to be escaped as `\n`. Orion is not escaping, and so it is not producing conforming JSON. My (third party) consumer does not have an option to enable processing of non-conforming input.

fgalan:
[...]

Issue #99, opened 2013-11-19
Use std namespace as default


fgalan:
Use a common/general.h file to include the using namespace declaration for std, so we can avoid that ugly "std::" prefix in strings and vector types.

This change is easily automatable, but given that it affects a lot of files, we need to do it in a moment in which there aren't open branches under work.

**Effort: 2 man day**

PR #1284, opened 2015-09-23
Add getAllSubscriptions w/o backend

crbrox:
BE not implemented yet


src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+using std::vector;
+
+string Subscription::toJson()
+{
+  JsonHelper jh;
+
+  jh.addString("id", this->id);
+  if (!this->duration.isEmpty())
+  {
+    jh.addRaw("duration: ",this->duration.render(JSON,"", false)); // render puede no escapar bien!!
```
fgalan: Spanish leftover at the end of line ;)

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+string Notification::toJson()
+{
+  JsonHelper jh;
+
+  jh.addString("callback", this->callback);
+  if (!this->throttling.isEmpty())
+  {
+    jh.addRaw("throttling", this->throttling.render(JSON,"",false)); // render puede no escapar bien
```
fgalan: Spanish leftover at the end

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+*
+* Author: Orion dev team
+*/
+#include "common/JsonHelper.h"
+
+std::string toJsonString(const std::string& input)
+{
+  std::ostringstream ss;
+  ss << '"';
+  for (std::string::const_iterator iter = input.begin(); iter != input.end(); ++iter) {
```
fgalan: Check style

src/lib/apiTypesV2/Subscription.h
```diff
@@ -0,0 +1,78 @@
...
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Orion dev team
+*/
+
+#include <string>
+#include <vector>
+
+#include "../ngsi/Duration.h"
```
fgalan: I think (not sure, it should be checked) that `ngsi/Duration.h` is enough. CMake makes the magic to have this header file available :)

src/lib/common/JsonHelper.h
```diff
@@ -0,0 +1,68 @@
...
+*/
+
+#include <sstream>
+#include <string>
+#include <vector>
+
+class JsonHelper
+{
+public:
+  JsonHelper();
```
fgalan: I like the JsonHelper idea as a way of making the render logic simpler. In addition, current implementation seem to solve the problem described in https://github.com/telefonicaid/fiware-orion/issues/1172. Maybe the usage of JsonHelper could be generalized to other NGSIv2 renders.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -376,6 +377,9 @@ PaArgument paArgs[] =
 #define ETT_COMPS_V2            2, { "v2", "types" }
 #define ETT_COMPS_WORD          ""
 
+#define GAS                     GetAllSubscriptionsRequest
```
kzangeli: *GET* should not be a part of the 'request name'.
What if we add a POST or PUT ... ?

I would propose **SubscriptionRequest**, unless it's already taken, of course ...

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "common/JsonHelper.h"
+
+using std::string;
+using std::vector;
+
+string Subscription::toJson()
+{
+  JsonHelper jh;
+
+  jh.addString("id", this->id);
```
kzangeli: this-> ...   any special reason to not just use 'id' ?

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <string>
+#include <sstream>
+#include <vector>
+
+#include "common/JsonHelper.h"
+
+using std::string;
+using std::vector;
+
+string Subscription::toJson()
```
kzangeli: A function should have a header, and three empty lines as separation.
Just open any source code file and copy the function header.

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "Subscription.h"
+
+#include <string>
+#include <sstream>
+#include <vector>
+
+#include "common/JsonHelper.h"
+
+using std::string;
+using std::vector;
```
kzangeli: Iustead of 'using std::xxx", I would propose to just maintain 'std::' where necessary.
ALL the code of the broker does that right now and I see no reason to change that behavior.
Especially not for a different behavior that contains less information

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -0,0 +1,104 @@
...
+  {
+    jh.addRaw("duration: ",this->duration.render(JSON,"", false)); // render puede no escapar bien!!
+  }
+  jh.addRaw("subject", this->subject.toJson());
+  jh.addRaw("notification", this->notification.toJson());
+
+  return jh.str();
+}
+
+string Notification::toJson()
```
kzangeli: 3 empty lines and function header

src/lib/apiTypesV2/Subscription.h
kzangeli: Yes, "ngsi/Duration.h" does the trick

src/lib/apiTypesV2/Subscription.h
```diff
@@ -0,0 +1,78 @@
...
+#include "../ngsi/Throttling.h"
+
+struct EntID
+{
+  std::string id;
+  std::string idPattern;
+  std::string type;
+  std::string toJson();
+};
```
kzangeli: We already have a class for Entity, called Entity ... (for v2, v1 uses another one, called EntityId.
I don't think we'll need two different classes in v2.

src/lib/apiTypesV2/Subscription.h
```diff
@@ -0,0 +1,78 @@
...
+
+struct Subscription
+{
+  std::string id;
+  Subject subject;
+  Duration duration;
+  Notification notification;
+  std::string toJson();
+};
```
kzangeli: Generally, each class should have its own header and source file.
However, I have a feeling these classes are solely used to render the JSON output for the Subscription class.
Perhaps they should all reside inside the Subscription class?

If they are to be 'global' classes, used by not only Subscription, they need to 'stand alone' ...
That means creating a lot of new files: Subject.h/Subject.cpp, etc etc

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+            case '\f': ss << "\\f"; break;
+            case '\n': ss << "\\n"; break;
+            case '\r': ss << "\\r"; break;
+            case '\t': ss << "\\t"; break;
+            default: ss << *iter; break;
+      }
+  }
+  ss << '"';
+  return ss.str();
+}
```
kzangeli: This function is called toJsonString, but what it really does is to escape some special characters and add double-quotes. Doesn't really have anything to do with JSON.

And by the way, this string-replace, copying char by char, using ostringstream ... Extremely inefficient way of implementing a string-replace ...
I would propose to reimplement that part in plain C. Probably at least 10 times faster in execution ...

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+template <>
+std::string vectorToJson(std::vector<std::string> &list)
+{
+  switch (list.size())
+    {
+    case 0:
+      return "[]";
```
kzangeli: Please add an empty line before every 'case XXX:'.
The code gets a lot easier to read that way.

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+      for(std::vector<std::string>::size_type i = 1; i != list.size(); ++i)
+      {
+          os << ", " << toJsonString(list[i]);
+      }
+      os << " ]";
+      return os.str();
+    }
+}
+
+JsonHelper::JsonHelper(): empty(true) {}
```
kzangeli: Not sure what good this does ...

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void JsonHelper::addString(const std::string& key, const std::string& value)
+{
+  if(!empty)
```
kzangeli: According to our unwritten style guide, between 'if' and '(' there should be a space.
Reason for that?
'if' is not a function, but a keyword for the compiler.
Let's make it show.
So. functions have no space before '(', all C keywords do (if, for, while, do, etc).

Really need to put all this down on paper ...

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+* along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+*
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Orion dev team
+*/
+#include "common/JsonHelper.h"
+
+std::string toJsonString(const std::string& input)
```
kzangeli: 3 empty lines and a function comment  header (now not only for the entire file, but for the entire PR)

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+  std::ostringstream ss;
```
kzangeli: An empty line every here and there makes a function easier to read (my opinion).
The rest of the broker code add an empty line after the variables ...

src/lib/apiTypesV2/Subscription.cpp
crbrox: It seems clearer to me. If Orion's style guide disapproves it, I`ll remove it

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+            case '/': ss << "\\/"; break;
+            case '\b': ss << "\\b"; break;
+            case '\f': ss << "\\f"; break;
+            case '\n': ss << "\\n"; break;
+            case '\r': ss << "\\r"; break;
+            case '\t': ss << "\\t"; break;
+            default: ss << *iter; break;
+      }
+  }
+  ss << '"';
```
kzangeli: ... and another empty line after a block

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+* Author: Orion dev team
+*/
+#include "common/JsonHelper.h"
+
+std::string toJsonString(const std::string& input)
+{
+  std::ostringstream ss;
+  ss << '"';
+  for (std::string::const_iterator iter = input.begin(); iter != input.end(); ++iter) {
+        switch (*iter) {
```
kzangeli: same here

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+  ss << '"';
+  return ss.str();
+}
+
+template <>
+std::string vectorToJson(std::vector<std::string> &list)
+{
+  switch (list.size())
+    {
```
kzangeli: indentation ('{' shoulld go right under the 's' in 'switch')

src/lib/apiTypesV2/Subscription.cpp
crbrox: I don`t see the point in that kind of header, without doc about params and return values, but if it is Orion's style, I'll change it

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,99 @@
...
+  empty = false;
+}
+
+void JsonHelper::addRaw(const std::string& key, const std::string& value)
+{
+  if(!empty)
+  {
+    ss << ", ";
+  }
+  ss << toJsonString(key) << ": " << value;
```
kzangeli: We want NO whitespace in the new JSON, please remove any whitespace from 'this new' json rendering mechanism

src/lib/common/JsonHelper.h
kzangeli: If were going to use it, I would propose to remove Helper from the name of the class.
Right now it's only a renderer, so JsonPrint ot JsonRender could be a good name.
However, the class could do more than just render some day.
I would call it **Json** simply.

src/lib/apiTypesV2/Subscription.h
crbrox: This is a substitute for EntityId, a far less complex version, for easing a future migration with a 'world" without v1, We can discuss it, but it is done intentionally. 'Duration' and 'Throttling' seemed simple enough ...

src/lib/common/JsonHelper.h
```diff
@@ -0,0 +1,68 @@
...
+  void addString(const std::string& key, const std::string& value);
+  void addRaw(const std::string& key, const std::string& value);
+  std::string str();
+private:
+ std::ostringstream ss;
+ bool empty;
+};
+
+std::string toJsonString(const std::string& input);
```
kzangeli: I know it's not mandatory (it should be), but please use the keyword 'extern' for external declarations of functions in header files. All the rest of the project does it. And ... dont forget the function header. Yes, in header files we use the function headers also

src/lib/common/JsonHelper.cpp
crbrox: Are you sure of that? How many allocations is doing ostringstream? Not all hand-crafted code is more performant than the STL. But it could be easy to measure ...

src/lib/common/string.h
```diff
@@ -159,4 +160,17 @@ extern void strReplace
 */
 extern std::string servicePathCheck(const char* servicePath);
 
+
+/* ****************************************************************************
+*
+* toString -
+*/
+template <typename T>
+std::string toString(T t) {
```
kzangeli: { on its own line

src/lib/common/JsonHelper.cpp
crbrox: I don't understand you ...

src/lib/common/tag.h
```diff
@@ -156,4 +159,6 @@ extern std::string startArray
 */
 extern std::string endArray(const std::string& indent, const std::string& tagName, Format format);
 
+
```
kzangeli: unnecessary empty lines  (the three empty lines go *before* the header, not after)

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -0,0 +1,64 @@
...
+* iot_support at tid dot es
+*
+* Author: Orion dev team
+*/
+
+#include "mongoBackend/mongoListSubscriptions.h"
+
+OrionError mongoListSubscriptions
+(
+  std::vector<Subscription>& subs,
```
kzangeli: If you indent 'subs' like the other three variables, it will be easier to read

src/lib/mongoBackend/mongoListSubscriptions.h
```diff
@@ -0,0 +1,48 @@
...
+#include "rest/OrionError.h"
+#include "apiTypesV2/Subscription.h"
+
+/* ****************************************************************************
+*
+* mongoListSubscriptions -
+*/
+OrionError mongoListSubscriptions
+(
+  std::vector<Subscription>& vec,
```
kzangeli: vec x space to the right (aligned with the other variables)

src/lib/apiTypesV2/Subscription.cpp
crbrox: It was used in another approved PR. I think the code is cleaner this way, using such basic types as 'string' and 'vector' directly, but I`ll do 'Roman' way

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
```diff
@@ -0,0 +1,73 @@
...
+std::string getAllSubscriptions
+(
+    ConnectionInfo*            ciP,
+    int                        components,
+    std::vector<std::string>&  compV,
+    ParseData*                 parseDataP
+)
+{
+  using std::string;
+  using std::vector;
```
kzangeli: As I said before, let's continue with the convention of writing out 'std' where necessary, and remove the 'using std::xxx' usage.

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
```diff
@@ -0,0 +1,73 @@
...
+    return oe.render(ciP,"");
+  }
+
+  if ((ciP->uriParamOptions["count"]))
+  {
+    ciP->httpHeader.push_back("X-Total-Count");
+    ciP->httpHeaderValue.push_back(toString(subs.size()));
+  }
+  return  vectorToJson(subs);
+
```
kzangeli: unnecessary empty line

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
```diff
@@ -0,0 +1,73 @@
...
+    std::vector<std::string>&  compV,
+    ParseData*                 parseDataP
+)
+{
+  using std::string;
+  using std::vector;
+
+  vector<Subscription> subs;
+
+  OrionError oe = mongoListSubscriptions(subs, ciP->uriParam, ciP->tenant, ciP->servicePathV);
```
kzangeli: There two stack-variable should be aligned and I see no reason for the empty line between them

src/lib/common/JsonHelper.cpp
crbrox: The function returns a JSON representation of a string ... I think it has everything to do with JSON

src/lib/serviceRoutinesV2/getAllSubscriptions.h
```diff
@@ -0,0 +1,49 @@
...
+extern std::string getAllSubscriptions
+(
+  ConnectionInfo*            ciP,
+  int                        components,
+  std::vector<std::string>&  compV,
+  ParseData*                 parseDataP
+);
+
+#endif  // SRC_LIB_SERVICEROUTINESv2_GETALLSUBCRIPTIONS_H_
+
```
kzangeli: empty line at EOF

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
```diff
@@ -0,0 +1,112 @@
...
+
+--SHELL--
+
+#
+# 01. GET /v2/subscriptions
+#
+
+echo "01. GET /v2/subscriptions"
+echo "========================="
+orionCurl --url /v2/subscriptions?options=count --in json --out json
```
kzangeli: orionCurl has another option '--json' that does '--in json --out json'

src/lib/apiTypesV2/Subscription.cpp
kzangeli: No, no probs. I was just curious :-)

NTC

src/lib/apiTypesV2/Subscription.cpp
kzangeli: Please.
What you say is true - most of the function headers only mark where the function begins.
ALL of them should have info on return values, parameters and the implementation.
We haven't had time for that I'm afraid. More complex functions do have a decent function
header, but most functions do not.
All functions have their header though.   It's a start ...

src/lib/apiTypesV2/Subscription.h
kzangeli: Yeah, EntityId is for APIv1.  Take a look at Entity (under apiTypesV2) and see if it better fits your needs.

src/lib/common/JsonHelper.cpp
kzangeli: 1. Yes, I am pretty sure I plain C function would be MUCH faster.
    To start with, probably 99% of all strings coming to this function will not contain any special chartacters, yet they are pushed to a new string char by char ... 

2. It only adds quotes and escapes a few special characters. A perfectly generic function

src/lib/common/JsonHelper.cpp
kzangeli: ok.
Ah, my fault. It's a constructor for JsonHelper.
It's clear, I was too fast.

Don't forget. {} on separate lines and function header ...

src/lib/apiTypesV2/Subscription.cpp
kzangeli: Thank you for that.
It's generally a good idea to follow the old conventions, already 'implemented'.
I am not against change, not at all, but to change a convention, it should be changed throughout the entire source code and there must be a pretty good reason to do so.

src/lib/apiTypesV2/Subscription.cpp
crbrox: Fixed 93306fe087d28bb8f2387c2ca05e76fb0062c2e8

src/lib/apiTypesV2/Subscription.cpp
crbrox: Fixed 93306fe087d28bb8f2387c2ca05e76fb0062c2e8

src/lib/common/JsonHelper.cpp
crbrox: Hopefully fixed in 910a029

src/lib/apiTypesV2/Subscription.h
crbrox: Fixed 93306fe087

src/lib/common/JsonHelper.cpp
kzangeli: Sorry ...
The '{' should go right under the start of the blocks keyword:

    switch (x)
    {
    case a:
      xxx();
      break;

    case ...
    }

'case' is a special case. It's not code, just a label, part of the 'switch infrastructure'.

for:

    for ()
    {
       ccc();
    }

src/lib/apiTypesV2/Subscription.cpp
fgalan: There is a very old issue slightly related with this at https://github.com/telefonicaid/fiware-orion/issues/99. If new overall changes regarding to namespaces usage needs to be done, maybe it could be used to described the common "style" we want at the end.

src/app/contextBroker/contextBroker.cpp
crbrox: Fixed 5567e1d

src/lib/common/JsonHelper.cpp
kzangeli: I also wanted to comment on the compact switch statement.

The unwritten style guide encourages to do these one-line cases, when the whole is very short, like in your case. When the 'entire case' doesn't fit on one line, each statement must be on a separate line and am empty line must be added before each new 'case'

src/lib/common/JsonHelper.cpp
fgalan: Let's avoit to get crazy about performance at this single point ;)

Performance optimization a general topic that needs to be analyzed globally, driven by a profiling process. For example, it could happen that although the C to C++ implementation performance difference is 99% faster in C, using it only have a 0.01% impact in the global CB performance due to other factors (e.g. bottlenecks at mongo driver level). Thus other criteria should be taking into account in this case (readibility, manteniability, etc.).

Having said that, it is ok for me to use a C implementation if you prefer that (I prefer don't give my personal opinion here in order not bias the discussion ;)

src/lib/apiTypesV2/Subscription.cpp
kzangeli: As long as all the code uses the same style, I will be satisfied. Uniformity is important.
If we later decide to use 'using' instead of explicitely showing std:: every time, then let's change the *entire* code at once. I will vote against such a change, but I have only *one* vote ...

src/lib/common/JsonHelper.cpp
kzangeli: At least, let's not implement any new code in a slower way on purpose.
It is of course true that we have **many** improvements pending to make the broker faster, but I see no meaning in making it slower on purpose ...
Right now, only GET Subscriptions use this function but if we step up and use JsonHelper instead of the already implemented rendering mechanism, them this would actually have an impact on the speed of the broker.

Writing a fast function to do the exact same thing is a question of 30 minutes. Tops. Actually, I believe we have one very similar already ...

src/lib/common/JsonHelper.cpp
crbrox: Is it said in somewhere ? (for improving the 'process' for the next time)

src/lib/common/JsonHelper.cpp
crbrox: (I'd love to see that x10 plain C function, if my vote counts)

src/lib/common/JsonHelper.cpp
kzangeli: Just a meeting we had with Juanjo in Madrid in April
No docs (unless Fermin sent minutes, as he normally does).

However, one important aspect of v2 is to minimize payload.
In v1, the render indents and adds newlines (beautifies). In v2 we plan to add a URI param for beautifying, but without that, we want no whitespace at all

src/lib/common/JsonHelper.cpp
kzangeli: You will, no problem

src/lib/common/JsonHelper.cpp
fgalan: As far as I remember (my memory is weak :) we were discusing about removing indentation withespace, which is a significative percentage of the total weight of the payload. Regarding the space between `:` and `"`is no so important, but if we want to be strict, probably it should be also removed.

What we are doing right now in other NGSIv2 renders? I would suggest to "copy" the same solution.

kzangeli: No whitespace at all, no spaces, no returns ... Zero

src/lib/apiTypesV2/Subscription.cpp
crbrox: "statistical" uniformity ... a few places in the code have 'using'
Fixed 847d1fd

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
crbrox: Fixed 847d1fd

src/lib/serviceRoutinesV2/getAllSubscriptions.h
crbrox: Fixed d544d3c

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
crbrox: Fixed 6502a07

src/lib/common/JsonHelper.cpp
kzangeli: Had some time over awaiting a 'make test' ...

    *
    * stringReplace -
    *
    * NOTE
    * The caller of this function supplies an allocated buffer for the
    * output and it is the callers responsibility to make sure the out-buffer
    * is large enough.
    */
    char* stringReplace(const char* in, char* out)
    {
      int  inIx   = 0;
      int  outIx  = 1;

      out[0] = '"';

      while (in[inIx] != 0)
      {
        switch (in[inIx])
        {
        case '\\':
        case '"':
        case '/':
        case '\b':
        case '\f':
        case '\n':
        case '\r':
        case '\t':
            out[outIx++] = '\\';

        default:
            out[outIx++] = in[inIx++];
        }
      }

      out[outIx++] = '"';
      out[outIx]   = 0;

      return out;
    }

fgalan: > The caller of this function supplies an allocated buffer for the output and it is the callers responsibility to make sure the out-buffer is large enough.

That means that the caller has to do a first pass on the `in` string in order to know `out` buffer size (I guess that counting the number of `\`), then allocate memory (and releasing after use) or relay in a buffer large enough to hold whatever output may be (which is segfault prone as sooner or later the initial assumption on size is not honored).

This is the kind of things that make "plain C" implementation harder (in the mantenability aspect I refer in a previous comment) compared with using  `ostringstream` 

(Don't interpreted this comment as a support for the C++ implementation, as I tool before I don't want to bias the decission ;) I just only want to illustrate that performance-mantenability-readibility is a multivariable trade off which solution is not necesarily maximizing performance as the cost of minimizing the others).

kzangeli: No, it does not mean that at all.
Normally a user would know what is a sufficient amount of data, and if not, just allocate a char buffer of twice the size of the buffer to be replaced. Mas size of the output buffer is 2 * strlen(inBuffer).

It is common praxis for a function of a library to not decide for its user. If the user wants to use a 'slow' heap variable (slow due to the call to malloc), let him do that, if he prefers to use a faster allocation on the stack, he can do that too.
I would strongly recommend to use the stack if possible. What I would pretty much never do in a function of a library is to not let this decision be up to the user.

fgalan: Not sure if we can use static allocation in this case... we don't know the size of the `in` buffer (e.g. attribute values hasn't a maximum length) and `out` (as it is a function of `in`) buffers.

src/lib/apiTypesV2/Subscription.h
```diff
@@ -0,0 +1,78 @@
...
+
+struct Subject
+{
+  std::vector<EntID> entities;
+  Condition condition;
+  std::string toJson();
+};
+
+struct Subscription
```
fgalan: This class name may colide with the the Subscription class defined at src/lib/cache/Subscription.h (although the later belongs to the "orion" namespace).

fgalan: Maybe define a "ngsiv2::" namespace for new NGSIv2 types?

src/lib/common/JsonHelper.cpp
kzangeli: Yeah, a fixed sized stack-variable, without any checks, we cannot use.

But, we can use a stack variable of size 1024 (for example) and before allocating we check to see whether the in-variable is shorter than 512. If so (probably 99% of the cases), we use the stack-variable and we avoid the malloc+free.

And that is the reason why a function like this should never make the decision to do an alloc, its users should always decide that.

If you look at common functions like 'strcpy' and alike, they don't allocate.
It's more, they don't even check whether the buffer to copy to is valid.
strcpy("123", NULL) gives a SIGSEGV.
Why? 
Well, it is not strcpy:s decision to do that check. We're talking about 3-4 assembler instructions here. That is 3-4 too many for some uses. Let the user decide, not the library. Important rules for library writing under C. I know these detail may seem almost silly for a Java or Python developer but in C it's different, it's the real thing. It is not uncommon to count assembler instructions or clock cycles while implementing ...

src/lib/apiTypesV2/Subscription.h
kzangeli: Not a bad idea ...
What about the older name, 'apiv2', should we change that to ngsiv2 ... ?

src/lib/apiTypesV2/Subscription.h
crbrox: We don't need de the vector of attributes associated. As in v1 there was a class for entittyid, I think the concept is better represented by a separated class

src/lib/apiTypesV2/Subscription.h
crbrox: I think it is clearer without nested classes and ease building the object that has a few levels of composition. . One file for each of these small and ultra-simple class looks too heavy

src/lib/common/JsonHelper.h
crbrox: I orefer JsonHelper., I think it express what is properly enough

src/lib/apiTypesV2/Subscription.h
kzangeli: associations are to be removed completely.
However, just because a class has a field or two that you don't need ... no reason to create a new one, right?
To have two different classes about the same concept ... very confusing and must be avoided.
It's a completely different thing if you maintain these classes inside another class, no confusion in that case.

src/lib/common/JsonHelper.cpp
crbrox: I think this is not a library function, and comparing with strcpy is not the right approach. It's a helper function for a little class for building a string of JSON. We cannot do every function we code as a library function. Sometimes we need to fit the specific need, not save the world

As all our cases for this use dynamic string coming from the user, basically you are trading space for speed, but "at least x10 faster"? why not x2, x1.1 or x1.01? Without measures it is sci-fi. And also the overall effect over the performance of Orion without measurement it is unknown (Sometimes the results in perform field can be counter-intuitive, these micro-benchmarks are widely discouraged)

What it is sure is your approach makes the code more unreadable, error-prone, and uncomfortable to write

The basic algorithm is the same, and the original implementation is very normal, nothing weird, so you has changed the design of the function and the callers doing another thing not functionally equivalent. Not improving the function, but another function no interchangeable at all. Please, don`t call "extremely inefficient" such code just because you think C++ is "extremely inefficient". I would prefer you use it if the complexity is O(n^2) instead of a O(n·logn) or O(n). Talking about uniformity, I don't think all the code (nor the most part) follow the design of pure C library function ... Happy to see my PRs are source of inspiration and creativity

By the way, your approach could be implemented in C++ with iterators and char sequences. It has nothing to do with plain C but with the strategy of allocating all the double of memory at first  or trying waste less memory with a smart and highly optimized implementation of the implemetarors of the STL. It's about speed vs space. But we should check the overall effect ...

src/lib/common/JsonHelper.cpp
crbrox: The communication should be improved for next decisions. Let's take note

src/lib/apiTypesV2/Subscription.h
crbrox: Strange for a man worried about cycles and assembler instructions, to have a unecessary vector going back and forth in a lll assignments, copies, ... Just for efficiency, the calls should have only the necessary member. If we have a vector of 10000 entititesID we should carry with 10000 vectors of attributes?
I see two concepts, an entity and an entityID. Sometimes a concept is representable in a basic type (perfect as a member) and sometimes it requieres a more elaborated representation -> a class. In this case , it has id, idPattern and type
Maybe the Entity class in v2 should be refactored to use EntID.
Using entity ids without the need  of the whole entity seems reasonable and convenient

crbrox: how it is used in v1 ...

src/lib/apiTypesV2/Subscription.h
kzangeli: I see two options here.
1. If these classes are only to be used for json rendering of the subscription list it can all stay the way it is. I would change the name of the classes though, to reflect what they actually do.
It would be clearer if the classes were a part of the Subscription class, though.

2. If the classes are to be used for something more than only rendering of JSON, they must be includable separately. If somebody needs the class Subject (just an example) but doesn't want the class Notification, that somebody should be able to get only Subject. Perhaps that include already has a class called Notification ...
This implies a separate file for each class.

src/lib/apiTypesV2/Subscription.h
kzangeli: I agree to your comment about carrying a std::vector that may be empty.
We started to use std::string and std::vector because they make the code so much easier to read and even implement. My first choice was to use a pointer, of course, and allocate the vector when needed but that made the code harder to read (much harder), so vectors are used.
Son't expect to see the code 100% perfect, all over. It is not perfect. But let's not use that fact as an excuse to not trying to make the new code as good as possible.

Perhaps we can try to make a new Entity class that it good enough for all three uses, or we maintain three different classed but with some prefix to clearly show what is the use of each one.

Right now we have   ... EntityId, Entity and EntID ... just too confusing.

crbrox: Sorry,, the code i have made is as good as possible, I do not see the excuse you are referring to. Maybe adding namespaces (which I did not see widely used in the "uniform" orion code) can make things clearer
Entity, EntID -> v2
EntityId -> v1, to be deprecated and removed. Not re-used because unwanted complexity

crbrox: Using a pointer instead a vector does not look a great improvement. A vector is a little more than a pointer ... It a issue of clarity and not carrying innecesary data. Copying and assinig to a pointer also waste cycles and words in memory, but must important it does not belong to the concept of identity

src/lib/apiTypesV2/Subscription.h
fgalan: If "apiv2" namespace is already used in some place, let's use that in order not make up a new namespace. If "apiv2" is not yet used in any place, we have yet some freedom to choose. Both, "apiv2" and "ngsiv2" are ok with me (I tend to slightly prefer the second one, as it provides more clarity on the "nature" of the API, but the first is also fine).

src/lib/apiTypesV2/Subscription.h
crbrox: This classes are for more than representing JSON. In fact, I think Fermin will use them in BE 
As there is nothing in API referring to an isolated condition, subject o notification, I think is very unlikely none of them will be used apart. The day we need only a condition, we will be able to extract a .h/.cpp files without much trouble, so I'd not change a thing right now (except namespaces, what I think it`s good)

fgalan: In NGSIv2 we have the following operations related with subscriptions that in principle will be using this class:

```
GET	/v2/subscriptions
GET	/v2/subscriptions/<subId>
POST /v2/subscriptions
PATCH /v2/subscriptions/<subId>
```

The first and the second correspond to know operations at mongoBackend, but how to do in the third and fourth case is not clear yet:

* POST /v2/subscriptions will probably reuse the existing mongoSubscribeContext operation at mongoBackend, which is based in a different subscription class (so some v2 to v1 conversion would be needed).
* I don't think that PATCH /v2/subscriptions could reuse directly existing mongoUpdateSubscription operation in mongoBackend, as the PATCH semantics are not exactly the same than the NGSIv1 subscription update.

However, in the avove scenarios don't see the need of using for example Subject outside the Subscription. Thus, probably we could stay as it is now (only one file) and split if we find we need in the future. Makes sense?

src/lib/apiTypesV2/Subscription.h
kzangeli: I agree to use ngsiv2 instead of apiv2, it is clearer.
No, apiv2 is not a namespace (as far as I can remember), but it is the name of a library, well 'apiTypesV2'.
Would it be better to use the name 'ngsiTypesV2' fir the library (and inside use namespace 'ngsiv2') ?

src/lib/common/JsonHelper.h
fgalan: Json is too broad in my opinion. JsonHelper and JsonRender are also good names. Maybe we can keep the current name and adjust in the future if we find the class adquires new semantics in the future and the name is no longer correct (renaming class is pretty easy, see https://github.com/telefonicaid/fiware-orion/issues/1173).

src/lib/common/JsonHelper.cpp
fgalan: Let's also use PR as a vehicle for comunication (as I think we are succesfully doing).

src/lib/apiTypesV2/Subscription.h
fgalan: I think it makes sense that the name of the library directoy is aligned with the name of the namespace used by tle libray. However, probably it could done in a separate PR.

src/lib/apiTypesV2/Subscription.h
kzangeli: With a namespace, the risk of colliding is avoided. That is the important part.
The rest is 'cosmetica' if you like, so sure, while no real need present, we can keep it the way it is.

src/lib/apiTypesV2/Subscription.h
```diff
@@ -0,0 +1,78 @@
...
+{
+  std::string id;
+  Subject subject;
+  Duration duration;
+  Notification notification;
+  std::string toJson();
```
kzangeli: If you align the field names horizontally, it is easier to read (part of the still unwritten style guide)

src/lib/apiTypesV2/Subscription.h
kzangeli: Definitely.
I create an issue about it, saying "rename apiTypesV2 to ngsiV2Types".
Feel free to edit the issue, both of you

src/lib/mongoBackend/mongoListSubscriptions.cpp
crbrox: Don't worry too much about this file its a "mock" that Fermin will replace completely in its PR implemting BE

src/lib/mongoBackend/mongoListSubscriptions.h
crbrox: idem , to bre replaced by Fermin

src/lib/common/string.h
crbrox: hopefully fixed 82e1760467b3a1ff072140

src/lib/common/tag.h
crbrox: hopefully fixed 82e1760

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
crbrox: hopefully fixed 82e1760

src/lib/serviceRoutinesV2/getAllSubscriptions.cpp
crbrox: hopefully fixed 82e1760

src/lib/apiTypesV2/Subscription.h
crbrox: hopefully fixed 82e1760

src/lib/mongoBackend/mongoListSubscriptions.cpp
kzangeli: Sure, let's Fermin promise to do it! :-)

NTC

src/lib/mongoBackend/mongoListSubscriptions.h
kzangeli: Yup

NTC

src/lib/common/JsonHelper.cpp
crbrox: I think it's a little too late ... I`d like having known it when implementing JSOn rendering

Fixed fd1bc7d3468f33df1a46f6f9b558af3ae820b24

src/lib/common/JsonHelper.h
crbrox: hopefully fixed 82e1760

src/lib/common/tag.h
kzangeli: tag.h doesn't seem to be part of the commit

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/apiTypesV2/Subscription.cpp
crbrox: hopefully fixed 82e1760

src/lib/apiTypesV2/Subscription.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

src/lib/common/JsonHelper.cpp
crbrox: hopefully fixed 82e1760

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
```diff
@@ -0,0 +1,56 @@
...
+
+--SHELL--
+
+#
+# 01. GET /v2/subscriptions
+#
+
+echo "01. GET /v2/subscriptions"
+echo "========================="
+orionCurl --url /v2/subscriptions?options=count --in json --out any
```
kzangeli: If '--out any' is used, then orionCurl doesn't know that the JSON beautifier should be used, hence the result on one single line, no whitespace and no JSON check.
This is not very good, of course.
If it is *really* necessary, ok, but if not, let's use the beautifier - much easier to read.

src/lib/common/tag.h
crbrox: glups, that file should not be even referenced, slipped from a former approach using JSON_VALUE. I'll restore it

crbrox: fixed 983a37c0039ec87ca472ac3f0f18a6ec806b8e34

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
crbrox: I know (I think I added 'text' and 'any' options). but I can't see the way of checking that our JSON does not contain spaces, new lines, ... if the beautifier changes the output. Maybe all test should be changed this way or some day we can be adding white spaces in the JSON we generate and passing the test ...

crbrox: Thinking about it, we could trust in content-length. I'll change it

fgalan: >  I can't see the way of checking that our JSON does not contain spaces, new lines, ... if the beautifier changes the output

Good point :)

You are rigth that the current way test are written could hide fails in the whitespace functionality. However, it would be very hard to mantain test without beautifier.

Maybe a good solution is to have one test case (composed of one or several .test) to check that whitespaces removal is ok without using beautifier, but let the existing (and future) .test as they are now (with beautifiers).

Not for this PR (if you like the idea, a tech debt issue will be added for the task).

kzangeli: It would be more or less easy to run a short function from inside orionCurl to check for unwanted whitespace. Only problem: not check inside strings. I'll create an issue

crbrox: Using the beautifier again. Content-length should be a good caged canary ...

ff25db4

src/lib/common/JsonHelper.cpp
kzangeli: ok, not even going to answer that.
And no meaning to add more comments either, that is clear.

As long as this function is only used for the list of subscriptions, lets keep it as is.
If expanded to more requests, we'll fix it.

NTC

kzangeli:
FYI, I did a quick performance test. 

    kz@dellie:~> jsonStringReplacer
    stringReplace:        0.000001
    stringReplace+malloc: 0.000050
    toJsonString:         0.000106

stringReplace without malloc (which will not be necessary 99% of the times) is not 10 but 100 times faster than toJsonString. Not very surprising, really.
With malloc+free, stringReplace is twice as fast.

The string I tested was this:

    std::string a = toJsonString("1234567890 \"qwerty\" uiopasddsgfdfhgfghdfhdfh");
       vs
    stringReplace("1234567890 \"qwerty\" uiopasddsgfdfhgfghdfhdfh", out);

Trying with the string "E1", the result is pretty much the same.
To measure the time the loop takes in stringReplace it is not enough with microseconds, I would need to go no nanoseconds (it gives 0 almost every time).

So, my proposition for this project (just the way it has always been):
1. Let's use C++ when it gives us something that C cannot give, like the templates introduced in this PR.
2. When a task can be done in C, we will always use C

Can we agree to that?

src/lib/common/JsonHelper.h
kzangeli: NTC

src/lib/apiTypesV2/Subscription.h
kzangeli: Sure, it may make sense to have a different class for different uses of Entity, and all other concepts/classes also.
It adds to the number of lines of the project, but it makes it faster and probably easier to read too.
Some parts of the code will be duplicated but that's not really a problem.

NTC

src/lib/common/JsonHelper.cpp
```diff
@@ -0,0 +1,153 @@
...
+    case '\f': ss << "\\f"; break;
+    case '\n': ss << "\\n"; break;
+    case '\r': ss << "\\r"; break;
+    case '\t': ss << "\\t"; break;
+    default:
+      if (ch >= 0 && ch <= 0x1F)
+      {
+        static const char intToHex[16] =  { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' } ;
+
+        ss << "\\u00" << intToHex[(ch & 0xF0) >> 4] << intToHex[ch & 0x0F];
```
kzangeli: Took me awhile to understand this ...
It's a 'hex2acsii', like a printf("%x", ch), covering all special chars before SPACE (32)
I would add a comment - something like "Converting the rest of special chars 0-31 to \u00xx".
Perhaps there is even a name for \u00xx?

What about chars 0x80 - 0xFF - should they also be covered by this conversion?

src/lib/apiTypesV2/Subscription.h
fgalan: NTC

src/lib/apiTypesV2/Subscription.h
fgalan: I guess the issue number is https://github.com/telefonicaid/fiware-orion/issues/1290

src/lib/common/JsonHelper.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/42b75e839d602badea756d2ddfe4d597cccdd14b

fgalan:
LGTM

kzangeli:
LGTM

PR #1285, opened 2015-09-23
Task/null value in parse step

kzangeli:
'bool typeGiven' in Entity, ContextAttribute and Metadata.
If typeGiven == true, then the 'type' was present during the JSON parse of the payload.

### DIFF
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -38,6 +38,7 @@
 */
 Entity::Entity()
 {
+  typeGiven = false;
 }
 
 
--- a/src/lib/apiTypesV2/Entity.h
+++ b/src/lib/apiTypesV2/Entity.h
@@ -47,6 +47,7 @@ public:
   ErrorCode               errorCode;        // Optional - mandatory if not 200-OK
 
   std::string             servicePath;      // Not part of payload, just an internal field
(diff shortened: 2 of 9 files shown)


src/lib/apiTypesV2/Entity.cpp
```diff
@@ -38,6 +38,7 @@
 */
 Entity::Entity()
 {
+  typeGiven = false;
```
fgalan: Actually, the title of the PR should be "null type in parse step", right? I mean, the ability from the parser layer to distinguish between the situation in which the input payload includes `type: ""` (typeGiven = false) and absences of `type` (typeGiven = true).

Just to clarify this PR is not about the usage of `null` as value for an attribute, as boolean or numbers are.

fgalan: NTC (it seems pretty clear reading the code and its comments)

fgalan:
LGTM

src/lib/apiTypesV2/Entity.cpp
kzangeli: Yes, of course, the title is poorly chosen.

However, a clarification to your comment:

 type: "" (typeGiven = **true**)
absences of type (typeGiven = **false**)

fgalan: You are right, It was a typo... I tried meaning just the opossite :$

Issue #1259, opened 2015-09-21
Fixing JSON representation format for entities and attributes in APIv2


kzangeli:
This issue implements a set of changes of the entities/attributes representation format discussed during September 2015 by 
* @jmcanterafonseca,
* @fgalan,
* @mrutid,
* @kzangeli and
* @antoniojoseln 

See http://telefonicaid.github.io/fiware-orion/api/v2/ for details (yet to be added).

Parts to implement:
1. ~~Creation/Update of Entities~~
2. ~~Rendering of Entity~~
3. ~~Creation/Update of Attribute~~
4. ~~Rendering of Attribute~~
5. ~~Update of Attribute Value~~
6. ~~Rendering of Attribute Value~~
7. ~~Selector to make the broker accept different input payloads for /v2/entities~~
8. ~~Make attribute::value **non-mandatory** - if not present, value is NULL~~
9. ~~Modify input format of strings in *5. Update of Attribute Value*, not to need citation-marks~~ (not finally needed)
10. ~~(Introduced by FGM at 26/10/2015) Replace `op=append` by `options=[...,]append[,...]` usage (I have looked to feature/1259_new_json_for_v2 branch and it seems that the old way is still in use)~~

    1. Creation/Update of Entities

    1A) POST /v2/entities
    1B) POST/PATCH/PUT /v2/entities/<id>
    
    {
        "id": "ID",            # Mandatory for 1A, ERROR for 1B
        "type": "TYPE",        # Optional for 1A, ERROR for 1B
        "attr1": <value>,      # string, number, bool, null, json object (if value present), compound (vector or object)
        "attr2": {             # Not a compound as it contains "value"
            "value": <value>,
            "type": "type",
            "metadata": {
                "meta1": <value>,    # string, number, bool, null (compounds are not allowed)
                "meta2": "STRING" || 12 || 3.14 || true || false || null,
                "meta3": {           # Not ERROR as it contains "value" or "type"
                    "type": "TYPE",
                    "value": <value>,  # string, number, bool, null
                }
            },
            "anything_else_for_attribute": ERROR
        }
    }

    2. Rendering of Entity

    Response from
    * GET /v2/entities
    * GET /v2/entities/<EID>
    
    2A. Compact    (?options=keyvalue)
    2B. Normalized (default)
    2C. Value      (?options=value)
    
    
    2A. Compact
    {
      "id": "ID",
      "type": "TYPE" || null,
[...]

PR #1288, opened 2015-09-24
Task/part2 rendering of entities

kzangeli:
Part II of issue #1259, rendering the output of /v2/entities[/< eid >]

Three renering modes implemented:
* normalized (default),
* keyValues, and
* values (tentative - format is not even fully decided yet - I think)

The mechanism right now to pick a rendering mode is via URL parameters.
This is not really decided yet - might change to URL PATH instead.
But, that is a minor change in the implementation.

### DIFF
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -55,9 +55,28 @@ Entity::~Entity()
 /* ****************************************************************************
 *
 * Entity::render - 
+*
+* The rendering of JSON in APIv2 depends on the URI param 'options'
+* Rendering methods:
+*   o 'normalized' (default)
+*   o 'keyValues'  (less verbose, only name and values shown for attributes - no type, no metadatas)
+*   o 'values'     (only the values of the attributes are printed, in a vector)
 */
 std::string Entity::render(ConnectionInfo* ciP, RequestType requestType, bool comma)
 {
+  std::string renderMode = "normalized";
+
+  if (ciP->uriParamOptions["keyValues"] == true)
+  {
+    renderMode = "keyValues";
+  }
+  else if (ciP->uriParamOptions["values"]== true)
+  {
+    renderMode = "values";
+  }
+
+  LM_M(("KZ: renderMode == %s", renderMode.c_str()));
+
   if ((errorCode.description == "") && ((errorCode.error == "OK") || (errorCode.error == "")))
   {
     std::string out = "{";
@@ -69,7 +88,8 @@ std::string Entity::render(ConnectionInfo* ciP, RequestType requestType, bool co
     if (attributeVector.size() != 0)
     {
       out += ",";
-      out += attributeVector.toJson(true, false);
+
+      out += attributeVector.toJson(true, false, renderMode);
     }
 
     out += "}";
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -97,7 +97,6 @@ std::string Entity::render(ConnectionInfo* ciP, RequestType requestType, bool co
       out += ",";
     }
 
-    LM_M(("KZ: Done: '%s'", out.c_str()));
     return out;
   }
 
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -75,8 +75,6 @@ std::string Entity::render(ConnectionInfo* ciP, RequestType requestType, bool co
     renderMode = "values";
   }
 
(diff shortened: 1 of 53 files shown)


src/lib/apiTypesV2/Entity.cpp
```diff
@@ -55,25 +55,39 @@ Entity::~Entity()
...
+*   o 'keyValues'  (less verbose, only name and values shown for attributes - no type, no metadatas)
+*   o 'values'     (only the values of the attributes are printed, in a vector)
 */
 std::string Entity::render(ConnectionInfo* ciP, RequestType requestType, bool comma)
 {
+  std::string renderMode = "normalized";
+
+  if (ciP->uriParamOptions["keyValues"] == true)
+  {
+    renderMode = "keyValues";
```
fgalan: Unfortunatelly, "keyValues" seems that will not be used at the end. However, the issue is still under discussion, so probably the best strategy is to progress this PR without changes, then fix it in a next PR once it gets clear. What do you think?

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -527,44 +527,50 @@ std::string ContextAttribute::toJson(bool isLastElement, bool types)
...
+      out += JSON_STR(stringValue);
     }
     else if (valueType == orion::ValueTypeBoolean)
     {
-      out = JSON_VALUE_BOOL(name, boolValue);
+      out += (boolValue == true)? "true" : "false";
     }
   }
-  else
+  else  // Render mode: normalized
```
fgalan: Code comnets should be aligned with spec. In this particular case, I'm not sure if "normalized" will be a term used in the spec.

However, again, until things get clear I'm afraid cannot fix this.

NTC

test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities.test
```diff
@@ -112,14 +112,12 @@ echo
...
 payload='{
-      "id":   "E2",
-      "type": "T1",
-      "A0" : 
-        {
-          "value" : { "i1": "1", "s1": "s" }
-        }
-      
-
+  "id":   "E2",
```
fgalan: Actual change or just indentation?

src/lib/apiTypesV2/Entity.cpp
kzangeli: Yes, I really believe that would be easier.

test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities.test
kzangeli: I found a few functests that had really bad indentations, among the failing ones.
Now that I was there, I fixed that also ...

The only 'real' changes in all these functests are to add 'metadata: { *** }'  and 'type: null'

test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities.test
```diff
@@ -156,83 +154,77 @@ Date: REGEX(.*)
...
+        "id": "E1",
         "type": "T1"
     }
 ]
 
 
-04. GET /v2/entities?options=count,normalized
-=============================================
+04. GET /v2/entities?options=count,keyValues
+============================================
```
kzangeli: OK, this is also a 'big change' in a functest.
As the previous GET (03) was without options (normalized), I decided to exercise 'keyValues' in this one.

NTC

test/functionalTest/cases/0947_v2_list_all_entities/GET_v2_entities.test
fgalan: I know that "Now that I was there" feeling, in fact I sometimes also do it :)

However, please keep this to the minimum in future PRs. If you feel that something related with indentation needs to be changed, the use a PR for that, different for the one with functional changes. Otherwise, it is hard to do the review.

NTC

src/lib/apiTypesV2/Entity.cpp
fgalan: At the end seems that "keyValues" will be used. Keep tunned :)

NTC

fgalan:
LGTM (although I have to admit that the "identation changes noisy" make me not be too confident about my review...)

kzangeli:
Sorry about that.
It's just ... so extremely difficult to fix a failing functest (or source code for that matter) that is so badly indented that you have problems reading it.
I understand your point and I agree. At the same time ... It's difficult. I promise to try ...

Any case, if you ever have doubts on functests, execute them in your own machine.
It's not 100% proof but it means a lot that they actually work.

PR #1291, opened 2015-09-25
FIX typo in test harness comment

fgalan:
It seems we leave some comment without fix in a past PR (I have spoted by coincidence)

### DIFF
--- a/test/functionalTest/cases/0953_list_all_ents_id/GET_v2_entities_pattern.test
+++ b/test/functionalTest/cases/0953_list_all_ents_id/GET_v2_entities_pattern.test
@@ -33,7 +33,7 @@ brokerStart CB
 # 01. GET /v2/entities?idPattern=E.*
 # 02. POST /v2/entities, creating E1
 # 03. GET /v2/entities?idPattern=E.*
-# 04. GET /v2/entities?idPattern=E1&count=true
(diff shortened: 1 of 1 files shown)


test/functionalTest/cases/0953_list_all_ents_id/GET_v2_entities_pattern.test
```diff
@@ -33,7 +33,7 @@ brokerStart CB
 # 01. GET /v2/entities?idPattern=E.*
 # 02. POST /v2/entities, creating E1
 # 03. GET /v2/entities?idPattern=E.*
-# 04. GET /v2/entities?idPattern=E1&count=true
+# 04. GET /v2/entities?idPattern=E1&options=count,normalized
```
kzangeli: Did this not make the test fail?
Perhaps we're still supporting count=true but we should signal an error instead?

fgalan: No, it isn't. If you look the .test carefully, we will see that the typo is only in the preamble comment (even the echo statement is using the right text).

NTC

kzangeli: Aaah, ok :-D

Yes, NTC

kzangeli:
LGTM

PR #1295, opened 2015-09-25
Fix Bug #1158

fortizc:
Implements #1158

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2240,7 +2240,8 @@ void processContextElement
   const std::vector<std::string>&      servicePathV,
   std::map<std::string, std::string>&  uriParams,   // FIXME P7: we need this to implement "restriction-based" filters
   const std::string&                   xauthToken,
-  const std::string&                   apiVersion
+  const std::string&                   apiVersion,
+  bool                                 checkEntityExistance
 )
 {
   DBClientBase* connection                  = NULL;
@@ -2382,7 +2383,12 @@ void processContextElement
       return;  // Error already in responseP
     }
 
-    if (entitiesNumber > 1)
+    if (entitiesNumber > 0 && checkEntityExistance && action == "APPEND_STRICT")
+    {
+        buildGeneralErrorResponse(ceP, NULL, responseP, SccInvalidModification, "Already Exists");
+        return;
+    }
+    else if (entitiesNumber > 1)
     {
       buildGeneralErrorResponse(ceP, NULL, responseP, SccConflict, "There is more than one entity that match the update. Please refine your query.");
       return;
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2450,9 +2450,8 @@ void processContextElement
   //
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
-  int docs = 0;
-  // to check updates
-  bool isUpdate = false;
+  int docs      = 0;
+  bool isUpdate = false; // to check updates
 
   while (cursor->more())
   {
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2451,6 +2451,8 @@ void processContextElement
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
   int docs = 0;
+  // to check updates
+  bool isUpdate = false;
 
   while (cursor->more())
   {
@@ -2686,6 +2688,8 @@ void processContextElement
       connection->update(getEntitiesCollectionName(tenant).c_str(), query, updatedEntityObj);
       releaseMongoConnection(connection);
 
+      // update is successful
+      isUpdate = true;
       LM_I(("Database Operation Successful (update %s)", query.toString().c_str()));
     }
     catch (const DBException &e)
@@ -2745,7 +2749,6 @@ void processContextElement
     responseP->contextElementResponseVector.push_back(cerP);
   }
   LM_T(LmtServicePath, ("Docs found: %d", docs));
-
 
   /*
    * If the entity doesn't already exist, we create it. Note that alternatively, we could do a count()
@@ -2834,6 +2837,11 @@ void processContextElement
 
       responseP->contextElementResponseVector.push_back(cerP);
     }
+  }
+  else if (apiVersion == "v2" && action == "APPEND" && !isUpdate)
+  {
+    responseP->contextElementResponseVector[0]->statusCode.fill(SccInvalidModification, "Entity already exists");
+    return;
   }
 
   if (attributeAlreadyExistsError == true)
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2456,7 +2456,7 @@ void processContextElement
   //
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
-  int docs      = 0;
+  int docs = 0;
 
   while (cursor->more())
   {
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2451,7 +2451,6 @@ void processContextElement
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
   int docs      = 0;
-  bool isUpdate = false; // to check updates
 
   while (cursor->more())
   {
@@ -2687,8 +2686,6 @@ void processContextElement
       connection->update(getEntitiesCollectionName(tenant).c_str(), query, updatedEntityObj);
       releaseMongoConnection(connection);
 
-      // update is successful
-      isUpdate = true;
       LM_I(("Database Operation Successful (update %s)", query.toString().c_str()));
     }
     catch (const DBException &e)
@@ -2836,11 +2833,6 @@ void processContextElement
 
       responseP->contextElementResponseVector.push_back(cerP);
     }
-  }
-  else if (apiVersion == "v2" && action == "APPEND" && !isUpdate)
-  {
-    responseP->contextElementResponseVector[0]->statusCode.fill(SccInvalidModification, "Entity already exists");
-    return;
   }
 
   if (attributeAlreadyExistsError == true)
--- a/src/lib/serviceRoutinesV2/postEntities.cpp
+++ b/src/lib/serviceRoutinesV2/postEntities.cpp
@@ -29,6 +29,7 @@
 #include "ngsi/ParseData.h"
 #include "apiTypesV2/Entities.h"
 #include "rest/EntityTypeInfo.h"
+#include "rest/OrionError.h"
 #include "serviceRoutinesV2/postEntities.h"
 #include "serviceRoutines/postUpdateContext.h"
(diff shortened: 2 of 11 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2451,6 +2451,8 @@ void processContextElement
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
   int docs = 0;
+  // to check updates
+  bool isUpdate = false;
```
kzangeli: If you put the comment on the same line (EOL), you gain a line.
These short comments we try to put on the same line.
Longer comments go on a separate line, of course.

Also, align the variable names and the eq-sign, like this:

    int  docs     = 0;
    bool isUpdate = false;  // to check updates

It's easier to read this way ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2835,6 +2838,11 @@ void processContextElement
       responseP->contextElementResponseVector.push_back(cerP);
     }
   }
+  else if (apiVersion == "v2" && action == "APPEND" && !isUpdate)
+  {
+    responseP->contextElementResponseVector[0]->statusCode.fill(SccInvalidModification, "Entity already exists");
+    return;
+  }
```
kzangeli: What about that new mode called *APPEND_STRICT* or something similar ... ?
I may be wrong, but, this looks like a typical use for that 'mode'.

At least, the old APPEND (v1) becomes an UPDATE if the entity exists.
While new mode (I understood) would flag an error on 'already existing'.

Might be I have misunderstood the while thing, but ... enlighten me, please :-)

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -68,9 +69,13 @@ std::string postEntities
   // 02. Call standard op postUpdateContext
   postUpdateContext(ciP, components, compV, parseDataP);
 
+  HttpStatusCode rcode = parseDataP->upcrs.res.contextElementResponseVector[0]->statusCode.code;
+
+  // Operation answer
+  std::string answer;
```
kzangeli: Same as before, comment on same line and variable alignment.
Actually, is this comment even necessary?
At least I don't get any wiser with the comment ...

Normally (not always - some are really complex), if a variable needs an explanation, then the name of the variable is not good and should be changed.

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -78,10 +83,19 @@ std::string postEntities
     ciP->httpHeaderValue.push_back(location);
     ciP->httpStatusCode = SccCreated;
   }
+  else if (rcode == SccInvalidModification)
+  {
+    OrionError oe;
+    oe.code = SccContextElementNotFound;
+    oe.reasonPhrase ="AlredyExists";
+    oe.details = "Entity alredy exists";
```
kzangeli: This seems like API v1 (OrionError with reasonPhrase and details).
There is another class/struct (ErrorCode) that is used for API v2.

Also, both these classes have a contructor accepting parameters, something like:

    ErrorCode errorCode("InvalidModification", "Entity already exists");

[ Not sure about the text to use for InvalidModification ... ]

It is also a bit strange to see SccContextElementNotFound in oe.code and SccInvalidModification in
ciP->httpStatusCode ...

Looking closer at OrionError (which unfortunately resides in the rest library and not in orionTypes) I see that OrionError is prepared for v2 as well (ciP carries the info on apiVersion), so OrionError can be used as well.
[ Not very good to have two different classes for this ... mea culpa ]
However, if you use OrionError, never set reasonPhrase, the constructor sets it for you, depending on the code.
Use this constructor:

    OrionError oe(SccInvalidModification, "Entity alredy exists");

    // Here is the implementation of that constructor:
    OrionError::OrionError(HttpStatusCode _code, std::string& _details)
    {
      code          = _code;
      reasonPhrase  = httpStatusCodeString(code);
      details       = _details;
    }

fortizc: I use OrionError because is used in src/lib/serviceRoutinesV2/deleteEntity.cpp but if its ErrorCode is a better class for this task, I can chenge it.
The SccContextElementNotFound is my mistake, sorry.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: I think you are right, the if should use APPEND_STRICT instead of APPEND.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2835,6 +2838,11 @@ void processContextElement
       responseP->contextElementResponseVector.push_back(cerP);
     }
   }
+  else if (apiVersion == "v2" && action == "APPEND" && !isUpdate)
```
fgalan: Looking to this else in context:

```
 if (docs == 0)
 {
 ...
 }
 else if (apiVersion == "v2" && action == "APPEND" && !isUpdate)
 {
 ...
 }
```

Note that the `if(doc == 0)` block is processed after the `while (cursor->more())` block in which the result of the query that get the entity matching with the update is processed. Thus, as far as I understand, you capture the error *after* the existing entity has been modified in DB, which wouldn't be a right solution as modifications in the existing entity have to be avoided.

If my theory is correct, we should see the fail when executing the the .test case you are now implementing to cover this fix. Thus I'd recomment first you develop the .test and once we check with that .test that current way doesn't work, look for a better solution (probably similar to the one you proposed in a previous attemp https://github.com/telefonicaid/fiware-orion/pull/1193#discussion_r38410801, based in detecting duplicated entities before the `while (cursor->more())` block.

fgalan: In a second thought, we probably need some other modifications to make it work. I would try to explain the rationale behind the needed modifications (but if my explanation is confusing, don't hesitate to ask for help in the Orion Dev skype chat :)

First, a modification is needed in the `POST /v2/entities` serviceRoutine to change APPEND to APPEND_STRICT in the mongoBackend invokation.

After that, we still having a problem, due to mongoUpdateContext() is not able to distinguish between the following situations:

* `POST /v2/entities`, which requires to check entity existence to solve #1158 
* `POST /v2/entities/<id>?op=append`, which doesn't required to check for entity existence as the operation my be used to add attributes to an existing entity.

Both operation are using APPEND_STRICT and the payload they use follow the same structure, so they cannot be distinguishes.

In order to solve this, we need to:

1. Add a new parameter to mongoUpdateContext() function signature, a parameter `bool checkEntityExistence` which default is `false`.
2. Set `checkEntityExistence` to `true` in `POST /v2/entities`
3. The #1158 error detection logic should be only done when APPEND_STRICT *and* checkEntityExistence are used

Anyway, first step is finishing the .test to cover the #1158 fix, as it is very valuable as guide to tune the implementation in the code.

fortizc: I see in src/lib/serviceRoutinesV2/ two differents files for this operation. First postEntities.cpp and with some debug I saw that this file is used, the other file is postEntity.cpp but when this file is used? if you see the name seems to be for same operation but one for many entities and other for a single entity. BTW postEntity.cpp use APPEND_STRICT

src/lib/mongoBackend/MongoCommonUpdate.cpp
fortizc: Fixed

src/lib/serviceRoutinesV2/postEntities.cpp
fortizc: Fixed

fgalan: Some PR prodecure" :) tip:

It is better to use: "Fixed in X", where X is a commit hash. That way, it is very easy for the reviewer to have a look to the commit where the fix has been addressed, without reviewing all the changes in the PR again (this is specially important in large PR).

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: The broker supports two similar requests:

    POST /v2/entities              (service routine postEntities)
    POST /v2/entities/{EntityId}   (service routine postEntity)

The first one is meant for creating an entity and if it already exists, it should an error (the bug addressed in #1158).
In the first request, the field "id" must be present.

The second one is more about creating attributes, not entities.
No field "id" is allowed. If present, error.
Also, if no entity with that ID exists, you should get a 404 Not Found.

It (the second request) has a URI parameter, **op**, that can take one single value (or empty): "append".
If "op==append", then APPEND_STRICT is used, meaning that trying to append an attribute that already exists gives an error.

test/functionalTest/cases/1158_v2_entity_already_exists/create_already_exists.test
```diff
@@ -0,0 +1,87 @@
...
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0
+
+--SHELL--
+
+#
+# 01. POST /v2/entities (Room1-Room)
+# 02. POST /v2/entities (Room1-Room)
+#
```
fgalan: Please add an additional step, to get the entity:

```
#03. GET /v2/entitites/Room1
```

test/functionalTest/cases/1158_v2_entity_already_exists/create_already_exists.test
```diff
@@ -0,0 +1,87 @@
...
+echo
+echo
+
+
+echo "02. POST /v2/entities (Room1-Room)"
+echo "=================================="
+payload='{
+  "id": "Room1",
+  "type": "Room",
+  "temp": "29"
```
fgalan: Change 29 to 30 and see what's happend in the (newley added) step 3:

* If the behaviour is the expected, then we will see 29, as the sedond POST shouldn't have been processed.
* If the behaviour is not the expected, then we will see 30, meaning that the second POST was wrongly processed (I bet for this :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2450,7 +2456,7 @@ void processContextElement
   //
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
-  int docs = 0;
+  int docs      = 0;
```
fgalan: Extra whitespace to remove?

test/functionalTest/cases/1158_v2_entity_already_exists/create_already_exists.test
fgalan: > Change 29 to 30...

I mean: step 1 uses 29, step 2 uses 30.

test/functionalTest/cases/1158_v2_entity_already_exists/create_already_exists.test
fortizc: Fixed in 5417881a656ad2ec976dfbad2ab1db24d57da0a6

test/functionalTest/cases/1158_v2_entity_already_exists/create_already_exists.test
fortizc: Fixed in 5417881a656ad2ec976dfbad2ab1db24d57da0a6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fortizc: Fixed in b55f25e2d29a16d5a50b00b8ad056a780e4f5934

fgalan:
LGTM (passing the ball to @kzangeli for final LGTM + merge + delete)

kzangeli:
LGTM

PR #1297, opened 2015-09-28
Remove deprecated ConfMan functionality

fgalan:
Removing around ~6,300 lines of dead code + tests related with ConfMan deprecated functionality.

Important to check: 

* Apart from the removed .test (related with ConfMan functionality), no other .test has been touched (thus other functionality has not been touched).
* The changes in the valueTag() functions to remove the associations stuff are correct


src/lib/ngsi/AttributeList.cpp
```diff
@@ -65,7 +65,10 @@ std::string AttributeList::render(Format format, const std::string& indent, bool
   out += startTag(indent, xmlTag, jsonTag, format, true, true, false);
 
   for (unsigned int ix = 0; ix < attributeV.size(); ++ix)
-    out += valueTag(indent + "  ", "attribute", attributeV[ix], format, ix != attributeV.size() - 1, false, true);
+  {
+    //out += valueTag(indent + "  ", "attribute", attributeV[ix], format, ix != attributeV.size() - 1, false, true);
```
fgalan: Ups... a leftover comment still here :(

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -479,6 +479,7 @@ std::string ContextAttribute::render
       }
 
       out += valueTag(indent + "  ", ((ciP->outFormat == XML)? "contextValue" : "value"),
+                        "value",
```
fgalan: Not sure if the previous invocation was using this funtion incorrectly or if the addition of the "value" is due to the assocations removal. Could you  have a look?

Definetively, mixing polymorphism (specially when function signature is very close among them) with default values makes thinks very hard to follow :) (default values are ok, but in simple cases).

test/unittests/common/commonTag_test.cpp
```diff
@@ -120,7 +120,8 @@ TEST(commonTag, valueTag)
    out = valueTag(indent, tag, value, JSON);
    EXPECT_EQ(jsonNoComma, out);   
 
-   out = valueTag(indent, tag, value, JSON, true, true);
+   // FIXME: after removing association code, maybe the test for 'jsonCommaAndAssociation' is no longer needed
```
fgalan: What do you think? Should this be removed?

src/lib/ngsi/ContextAttribute.cpp
kzangeli: Yes, this set of functions 'valueTag' are a nightmare.
Overloading functions used to be my favorite C++ feature, among with default values for parameters.
Right now I feel both these features are really bad ideas.

I will take a look at this particular change.
However, if the functests aren't complaining ...  This is a VERY centric piece of code - rendering of an attribute ...

test/unittests/common/commonTag_test.cpp
kzangeli: From the text 'jsonCommaAndAssociation' it **really** seems it should be removed.

src/lib/common/tag.h
```diff
@@ -130,7 +128,6 @@ extern std::string valueTag
   const std::string&  value,
   Format              format,
   bool                showComma           = false,
-  bool                isAssociation       = false,
   bool                valueIsNumberOrBool = false
 );
```
kzangeli: The first and the third valueTag are identical, as far as I can see.
Does this complie?

src/lib/ngsi/EntityId.cpp
```diff
@@ -121,7 +119,7 @@ std::string EntityId::render
        indent2 += "  ";
     }
 
-    out += (isInVector? indent + (isAssoc? "\"" + assocTag + "\" : ": "") + "{\n": "");
+    out += (isInVector? indent + "" + "{\n": "");
```
kzangeli: indent + "" ? Probably not necessary ... :-)
Also, please use a space before and after ':' in these 'special ifs'.

src/lib/parse/CompoundValueNode.cpp
```diff
@@ -581,7 +581,8 @@ std::string CompoundValueNode::render(ConnectionInfo* ciP, Format format, const
   if (valueType == orion::ValueTypeString)
   {
     LM_T(LmtCompoundValueRender, ("I am a String (%s)", name.c_str()));
-    out = valueTag(indent, tagName, stringValue, format, jsonComma, false, container->valueType == orion::ValueTypeVector);
+    //out = valueTag(indent, tagName, stringValue, format, jsonComma, false, container->valueType == orion::ValueTypeVector);
```
kzangeli: One more comment left in the code ...

test/unittests/xmlParse/xmlRegisterContextRequest_test.cpp
```diff
@@ -1,142 +0,0 @@
...
-  EXPECT_EQ("OK", testDataFromFile(testBuf, sizeof(testBuf), inFile)) << "Error getting test data from '" << inFile << "'";
-  EXPECT_EQ("OK", testDataFromFile(expectedBuf, sizeof(expectedBuf), outFile)) << "Error getting test data from '" << outFile << "'";
-
-  ci.inFormat  = XML;
-  ci.outFormat = XML;
-  out  = xmlTreat(testBuf, &ci, &parseData, RegisterContext, "registerContextRequest", NULL);
-  EXPECT_STREQ(expectedBuf, out.c_str());
-
-  utExit();
-}
```
kzangeli: So much work, to later after almost never using it, throw it away ...
Very good to cleanup our source code, of course!

NTC

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1567,7 +1549,7 @@ int main(int argC, char* argV[])
   SemRequestType policy = policyGet(reqMutexPolicy);
   orionInit(orionExit, ORION_VERSION, policy, mutexTimeStat);
   mongoInit(dbHost, rplSet, dbName, user, pwd, dbTimeout, writeConcern, dbPoolSize, mutexTimeStat);
-  contextBrokerInit(ngsi9Only, dbName, mtenant);
+  contextBrokerInit(dbName, mtenant);
   curl_global_init(CURL_GLOBAL_NOTHING);
```
kzangeli: Isn't there a vector of services also, in this file?
The one used when running as CM ...
Or, have we cleaned that up already?

src/lib/common/tag.h
fgalan: These are the 3 valueTag() functions defined in tag.h:

```
extern std::string valueTag
(
  const std::string&  indent,
  const std::string&  tagName,
  const std::string&  value,
  Format              format,
  bool                showComma           = false,
  bool                isVectorElement     = false
);

extern std::string valueTag
(
  const std::string&  indent,
  const std::string&  tagName,
  int                 value,
  Format              format,
  bool                showComma     = false
);

extern std::string valueTag
(
  const std::string&  indent,
  const std::string&  xmlTag,
  const std::string&  jsonTag,
  const std::string&  value,
  Format              format,
  bool                showComma           = false,
  bool                valueIsNumberOrBool = false
);
```

As far as I understand, where would be the conflict? It seems that all the signatures are differents (including the "variability" that the usage or not usage of optional arguments may introduce).

src/app/contextBroker/contextBroker.cpp
fgalan: Good catch! I completly forgot about that vector... Removal on the way.

src/lib/common/tag.h
kzangeli: ok, I didn't see the entire picture in the diff here in guthub.
Sorry for not taking the time to open the file before I commented.
It is clear that there can be no compiler-confusion.

NTC
NTC

src/lib/ngsi/ContextAttribute.cpp
fgalan: Functest (and unit tests) work fine.

So, NTC?

src/lib/ngsi/AttributeList.cpp
fgalan: Fixed in 4601741f14d843e91c611d871a34cefc8757051f

test/unittests/common/commonTag_test.cpp
fgalan: Fixed in 4601741f14d843e91c611d871a34cefc8757051f

src/lib/ngsi/EntityId.cpp
fgalan: Fixed in 4601741f14d843e91c611d871a34cefc8757051f

src/lib/parse/CompoundValueNode.cpp
fgalan: Fixed in 4601741f14d843e91c611d871a34cefc8757051f

src/app/contextBroker/contextBroker.cpp
fgalan: Fixed in 4601741f14d843e91c611d871a34cefc8757051f

src/lib/ngsi/ContextAttribute.cpp
kzangeli: Yes!

NTC

kzangeli:
LGTM

PR #1299, opened 2015-09-29
Task/part5 attribute value parse

kzangeli:
Part V of issue #1259


src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
```diff
@@ -325,3 +325,144 @@ std::string parseContextAttributeCompoundValue
...
+      cvnP->container  = parent;
+      cvnP->rootP      = parent->rootP;
+      cvnP->level      = parent->level + 1;
+      cvnP->siblingNo  = counter;
+      cvnP->path       = parent->path + "[" + itemNo + "]";
+
+
+      if (nodeType == "String")
+      {
+        cvnP->stringValue       = iter->GetString();
```
fgalan: Too much whitespace?

src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
```diff
@@ -325,3 +325,144 @@ std::string parseContextAttributeCompoundValue
...
+      cvnP->valueType  = stringToCompoundType(nodeType);
+      cvnP->container  = parent;
+      cvnP->rootP      = parent->rootP;
+      cvnP->level      = parent->level + 1;
+      cvnP->siblingNo  = counter;
+      cvnP->path       = parent->path + cvnP->name;
+
+      if (nodeType == "String")
+      {
+        cvnP->stringValue       = iter->value.GetString();
```
fgalan: Too much whitespace?

src/lib/orionTypes/OrionValueType.cpp
```diff
@@ -43,6 +43,7 @@ const char* valueTypeName(const orion::ValueType _type)
   case ValueTypeBoolean:      return "Boolean";
   case ValueTypeObject:       return "Object";
   case ValueTypeVector:       return "Vector";
+  case ValueTypeNull:         return "Null";
```
fgalan: Good! This will be useful in the future at mongoBackend to rightly persist this kind of values.

NTC

src/lib/rest/RestService.cpp
```diff
@@ -52,9 +52,111 @@
...
+
+  //
+  // 4. Valid Double?
+  // FIXME P4: this is much more complex than just (atof(string) != 0 || string == "0")
+  //           0.000 also is a valid float and it given 0 - 0.0000000000, and 0e0 also ...
+  //           And, even worse, 123K gives 123.0 back, we would need to analyze the string to try to
+  //           find garbage bytes after it if we want to detect this error.
+  //           However, all of this is not so extremely important and for now, (atof(string) != 0 || string== "0") is good enough
+  //
+  else if (((d = strtod(ciP->payload, NULL)) != 0) || ((ciP->payload[0] == '0') || (ciP->payload[1] == 0)))
```
fgalan: Not sure of understanding the logic behind this if clause... specially the last token about `payload[1]`.

(In addition, it seems that the text in the comment is not aligned with the if clause, e.g. comment mentions atof while the if uses strtod).

src/lib/rest/RestService.cpp
```diff
@@ -52,9 +52,111 @@
...
+  return "OK";
+}
+
+
+
+/* ****************************************************************************
+*
+* textRequestTreat - to be moved to ... own library?
+*
+* FIXME P10: Where do we put this function?
```
fgalan: If it related with request processing, rest/ library seems to be its place. If it related with parsing, the library about parsing (not remember the name) should be the place. Maybe not so important to be a P10 ;)

src/lib/rest/rest.cpp
```diff
@@ -370,12 +370,12 @@ static Format wantedOutputSupported(const std::string& apiVersion, const std::st
...
      std::string format = vec[ix].c_str();
      if (format == "*/*")              { xml  = true; json = true; text=true;}
-     if (format == "*/xml")            xml  = true;
+     if (format == "*/xml")            { xml  = true; }
      if (format == "application/*")    { xml  = true; json = true; }
-     if (format == "application/xml")  xml  = true;
-     if (format == "application/json") json = true;
-     if (format == "*/json")           json = true;
-     if (format == "text/plain")       text = true;
+     if (format == "application/xml")  { xml  = true; }
```
fgalan: I guess that we can break a little bit the rule that mandates to have `{` and `}` is separated lines ;)

src/lib/rest/rest.cpp
```diff
@@ -666,9 +667,9 @@ static int contentTypeCheck(ConnectionInfo* ciP)
   // Five cases:
   //   1. If there is no payload, the Content-Type is not interesting
   //   2. Payload present but no Content-Type 
-  //   3. text/xml used and acceptTextXml is setto true (iotAgent only)
+  //   3. text/xml used and acceptTextXml is set to true (iotAgent only)
```
fgalan: Off-topic: the existence of "iotAgent" only make me think that we have deprecated code related with old iotAgent (which as far as I know it no longer under developed). That code should be removed, in the same way ConfMan code was recently removed.

Could you add a tech debt issue about it, pls? I think that "tech debt should" would be fine.

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
```diff
@@ -30,9 +30,11 @@ brokerStart CB
 --SHELL--
 
 #
-## 01. POST /v2/entities to create E1/A1=0
-## 02. PUT /v2/entities/E1/attrs/A1/value (empty)
-
+# 01. POST /v2/entities to create E1/A1=0
+# 02. GET /v2/entities/E1/attrs/A1
+# 03. PUT /v2/entities/E1/attrs/A1/value (empty)
```
fgalan: Step 3 description doesn match with step 3 implementation, which uses `PUT /v2/entities/E1/attrs/A1` (ie. without `/value).

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
```diff
@@ -41,14 +43,27 @@ orionCurl --url /v2/entities --payload "$payload" --json
...
+echo "================================"
+orionCurl --url /v2/entities/E1/attrs/A1 --json
+echo
+echo
+
+
+echo "03. PUT /v2/entities/E1/attrs/A1 (empty)"
+echo "========================================"
 payload='{ }'
 orionCurl --url /v2/entities/E1/attrs/A1 --payload "$payload" --json -X PUT
```
fgalan: Maybe is due to the mismatch described in a comment above, but this operaiton maybe should return an error. I doesn't make sense to do a PUT on /v2/entities/E1/attrs/A1 without including at least one of the following properties: `value`, `type` or `metadata`.

If the operation where ` PUT /v2/entities/E1/attrs/A1/value`, then it would make sense to have `{ }` as payload. In this case, I would suggest to use "(empty JSON object)" instead of just "(empty)" in the test description, to make it more clear.

fgalan: Moreover, from @mrutid email (sorry for the Spanish, it is verbatim):

> Si no aparece ningún campo: type, value, metadata sería un ERROR.

test/functionalTest/cases/1259_new_json_for_v2/put_attribute_value.test
```diff
@@ -0,0 +1,582 @@
...
+echo '22. Check that A1 == 123'
+echo '========================'
+orionCurl --url /v2/entities/E1/attrs/A1 --json
+echo
+echo
+
+
+echo '23. PUT /v2/entities/E1/attrs/A1/value with TEXT payload 3.14'
+echo '============================================================='
+payload='3.14'
```
fgalan: Maybe including a text with negative number (pe. -3.14) and a test with cientific notation (pe. 3.2e2) ?

kzangeli: Good idea, that I haven't even tested! :-)

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
kzangeli: Perhaps we should create an issue (about PUT on /v2/entities/E1/attrs/A1) ... 
This is not really part of the PR, just 'other' functests that needed correction ...

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
kzangeli: True, I didn't look very closely at the functests, just made them work again ...
But, I did remove the extra #s, and I will fix this also, of course ...

src/lib/rest/rest.cpp
kzangeli: It looks better with the {} for all of these lines.
Doesn't hurt ...

src/lib/rest/RestService.cpp
kzangeli: I put P10 to make sure it is fixed in this PR, not because it was especially important per se.
We have three different libraries already for parsing stuff, xmlParse and two for jsonParse.
To follow our own style, we should make a new library for 'textParse'.

However, it is a little bit overkill now that we have one single function ...

There is a library containing common functions for the different parse libraries (the library is called **parse**), but that is also not the place for this test-parsing function ...

If we want to do it "well", we add a new library, called textParse, for example.
If not ... then the best place for this function is right where it resides right now (rest/RestService.cpp), IMO.

src/lib/rest/rest.cpp
fgalan: I agree. NTC

src/lib/rest/RestService.cpp
fgalan: > There is a library containing common functions for the different parse libraries (the library is called parse), but that is also not the place for this test-parsing function 

parse/ can be seen as a place for common functionality that doesn't fit in other places. Under that interpretation, textParseAttributeValue() could be there. A new textParse/ library would be also ok.

src/lib/rest/RestService.cpp
kzangeli: Yeah, atof and strtod is pretty much the same. The comment was just for a broader understanding, valid for atof, atod, atoi, strtod, strtoul, etc.  I will change it, to strtod, of course. Unnecessary confusion is always bad.

About

    ((ciP->payload[0] == '0') || (ciP->payload[1] == 0)) ...

... this is the fastest way to do

    strcmp(ciP->payload, "0") == 0

The first char in the char* is '0' (ASCII 48) and the second char is 0  (string terminator), thus "0".

The problem with this part is the difficulty to decide that it is a valid float when strtod(XXX) returns 0.

Actually, now that I think of it ... even if it returns != 0, the problem is still there ...
 
strtod("3x") returns 3.0, as far as I know ...

*I do know that atoi works this way, even did a minor demonstration not long ago. Makes sense to assume strtod works the very same way*

So, examples of 'OK when 0':

    0
    0.00000
    0e2  // Not sure about this syntaxis - I have nerve used it !!!

Examples of 'NOT OK when 0', and an error should be returned:

    Hi
    0x

So, my very simple solution is to assume that, if strtod gives back zero and the string is NOT EXACTLY "0", then it was not a float. Hence the FIXME.
I know that @fgalan was struggling with this very same problem not so long ago ...
Help, please ... :-)

src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
kzangeli: Oh yes ...
All 'cvnP->xxxValue =' were supposed to be horizontally aligned. Don't know what happened to this one ...

src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
kzangeli: Oh yes ...
All 'cvnP->xxxValue =' were supposed to be horizontally aligned. Don't know what happened to this one ...

But, now I know what happened to the 'previous one' (nextcoming - I'm reading the PR comment starting from the end) - cut 'n paste ...

src/lib/rest/RestService.cpp
kzangeli: ok, let's add it to the 'parse' library for now.
If we ever implement another text parsing function, we'll create a library for them.

src/lib/rest/RestService.cpp
fgalan: I'm afraid that I don't have the solution... on the contrary (if I'm remembering correctly) I have the same "0 gap" problem in other parts of the code :( 

My only advice here is to study how other C/C++ projects have solved the same problem. It no solution comes in a reasonable time, them create an issue about this (not only for textParseAttributeValue(), but also to analyze in which other points of the code we have the same problem) and live with that FIXME while we find a solution.

kzangeli: OK.

Just so that I don't forget:

    ((ciP->payload[0] == '0') || (ciP->payload[1] == 0))

Is incorrect, the OR must be an AND, of course!!!
Perhaps that's the source of the confusion earlier.

kzangeli: And, strtod and its second parameter solves this problem.
Two functions to add to the 'common' library:

    bool isFloat(char*);
    bool string2double(char*, double*);

src/lib/rest/rest.cpp
kzangeli: Issue created: #1302

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
kzangeli: Issue #1303 created

src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
kzangeli: Fixed in 447cec7

src/lib/jsonParseV2/parseContextAttributeCompoundValue.cpp
kzangeli: Fixed in 447cec7

src/lib/rest/RestService.cpp
kzangeli: Fixed in 447cec7

Except isFloat and string2double that will be added (and used here and anywhere else needed) in a separate PR

src/lib/rest/RestService.cpp
kzangeli: Fixed in 447cec7

test/functionalTest/cases/1164_v2_update_entities_attr_empty_value/update_entities_attr_empty_value.test
kzangeli: Fixed in 447cec7

test/functionalTest/cases/1259_new_json_for_v2/put_attribute_value.test
kzangeli: It worked!
Fixed in 447cec7

src/lib/rest/RestService.cpp
kzangeli: Issue added: #1306
I will start with this fix tomorrow morning, or perhaps even tonight ...

kzangeli: Actually, during the implementation of isFloat and string2double (yes, tonight :-)), I found a function called isFloat (I will use isDouble instead). And it made me remember ...
We did find a solution to this problem some time ago (on SOF):

    bool isFloat(const std::string& s)
    {
      std::istringstream iss(s);
      float              f;
    
      // Check the entire string was consumed and if either failbit or badbit is set
      iss >> std::noskipws >> f; // noskipws considers leading whitespace invalid
      return iss.eof() && !iss.fail();
    }

However, this is one of those 'nightmare' C++ solutions that we (read: **I**) hate so much.
A constructor, an operator, and two methods are needed for this solution (4 function calls).
In the strtod solution it is enough with ONE function call.
Some day I will measure the time these both functions need to decide whether a string is a valid float/double or not.
My guess?  the strtod solution will be at least 4 times faster, probably 10 times ...

fgalan: Blame buttom (https://github.com/telefonicaid/fiware-orion/blame/develop/src/lib/common/Format.cpp) shows that function was introduced in one of my PRs (it seems related with FIWARE::QueryString fucntionality).

Feel free to replace that fuction if you have a more efficient implementation with a similar readibility-mantenibility level (I will see the new one in the PR).

kzangeli: Yes, you will see it soon.
Hopefully I can present the timing information by then as well.

fgalan: I guess that in this case timing information is not necessary (specially if it costly to calculate that) if you have an argument that support the efficient improvement (e.g. 4 functional calls vs. 1).

kzangeli: On 01/10/15 00:22, Fermín Galán Márquez wrote:

In src/lib/rest/RestService.cpp<https://github.com/telefonicaid/fiware-orion/pull/1299#discussion_r40860178>:

> +
> +  // 3. Null ?
> +  else if ((strlen(ciP->payload) == 4) && ((strcmp(ciP->payload, "null") == 0) || (strcmp(ciP->payload, "Null") == 0) || (strcmp(ciP->payload, "NULL") == 0)))
> +  {
> +    caP->valueType   = orion::ValueTypeNull;
> +  }
> +
> +  //
> +  // 4. Valid Double?
> +  // FIXME P4: this is much more complex than just (atof(string) != 0 || string == "0")
> +  //           0.000 also is a valid float and it given 0 - 0.0000000000, and 0e0 also ...
> +  //           And, even worse, 123K gives 123.0 back, we would need to analyze the string to try to
> +  //           find garbage bytes after it if we want to detect this error.
> +  //           However, all of this is not so extremely important and for now, (atof(string) != 0 || string== "0") is good enough
> +  //
> +  else if (((d = strtod(ciP->payload, NULL)) != 0) || ((ciP->payload[0] == '0') || (ciP->payload[1] == 0)))


I guess that in this case timing information is not necessary (specially if it costly to calculate that) if you have an argument that support the efficient improvement (e.g. 4 functional calls vs. 1).

—
Reply to this email directly or view it on GitHub<https://github.com/telefonicaid/fiware-orion/pull/1299/files#r40860178>.

Its easy enough to measure (I have it implemented for other cases already),
It's really just a few calls to clock_gettime and a function I implemented for the semaphore waiting times - clock_difftime.
Takes ... 10 min tops.
It is interesting to know the difference ...

________________________________

Este mensaje y sus adjuntos se dirigen exclusivamente a su destinatario, puede contener información privilegiada o confidencial y es para uso exclusivo de la persona o entidad de destino. Si no es usted. el destinatario indicado, queda notificado de que la lectura, utilización, divulgación y/o copia sin autorización puede estar prohibida en virtud de la legislación vigente. Si ha recibido este mensaje por error, le rogamos que nos lo comunique inmediatamente por esta misma vía y proceda a su destrucción.

The information contained in this transmission is privileged and confidential information intended only for the use of the individual or entity named above. If the reader of this message is not the intended recipient, you are hereby notified that any dissemination, distribution or copying of this communication is strictly prohibited. If you have received this transmission in error, do not read it. Please immediately reply to the sender that you have received this communication in error and then delete it.

Esta mensagem e seus anexos se dirigem exclusivamente ao seu destinatário, pode conter informação privilegiada ou confidencial e é para uso exclusivo da pessoa ou entidade de destino. Se não é vossa senhoria o destinatário indicado, fica notificado de que a leitura, utilização, divulgação e/ou cópia sem autorização pode estar proibida em virtude da legislação vigente. Se recebeu esta mensagem por erro, rogamos-lhe que nos o comunique imediatamente por esta mesma via e proceda a sua destruição

kzangeli: String tested: "223.123456"

    kz@dellie:fiware-orion> isFloat 
    C:   0.000017770
    C++: 0.000173541

My initial guess was pretty good ...
The C-implementation is 9-10 times faster than the one in C++

fgalan: > Fixed in 447cec7

Maybe I'm not looking to the right place, but the code in 447cec7 doesn't seem to change anything related with this comment thread (so I understand that this rather a NTC, deferred to be fixed -globally- in https://github.com/telefonicaid/fiware-orion/issues/1306).

fgalan:
LGTM

src/lib/rest/RestService.cpp
kzangeli: The whole isFloat thing will be fixed later, already ready to integrate with the work here.
With 'fixed' I referred to fixing this:

    ((ciP->payload[0] == '0') || (ciP->payload[1] == 0))

Which I mentioned earlier has to be like this:

     ((ciP->payload[0] == '0') && (ciP->payload[1] == 0))

I.e. 

    (std::string(payload) == "0")

just a hundred times faster :-)


The isFloat fix, the REAL fix will be done in issue #1306

Issue #1290, opened 2015-09-25
apiTypesV2 library name


kzangeli:
Rename apiTypesV2 to ngsiV2Types.
The namespace used in apiTypesV2 as been decided to be **ngsiv2**, so, it makes sense to change the name of the library to **ngsiV2Types**.

**Effort: 1 man day**

fgalan:
In addition, classes inside ngsiV2Types (today: apiTypesV2) have to be modified to use ngsiv2:: as part of this issue.

Issue #136, opened 2013-12-03
Exahustive exception capture in mongoBackend


fgalan:
*(Original name of this issue was: "Disable assertions in production?", but is has evolved during the discussion to the current name)*

It seems that mongo driver code includes assertions. Sometimes, we have got messages such as the following one:

```
Thu Oct 17 16:54:08 Assertion: 13111:wrong type for field (value) 16 != 4
terminate called after throwing an instance of 'mongo::MsgAssertionException'
  what():  wrong type for field (value) 16 != 4
```

Is it a good idea that code installed in production may abort due to an assertion fail? Should we compile using NDEBUG to avoid this behavior in production?

**Effort: 4 man day**

fgalan:
Partially addressed in PR #1394. What is pending after that PR:

* Check that all the code is *not* using directly BSONObj get methos *after* landing the "new cache"
* Stop using STR_FIELD and C_STR_FIELD (as now the syntax is completely aligned with the function to get the field data).

fgalan:
Implementation finished in PR https://github.com/telefonicaid/fiware-orion/pull/1416

PR #1312, opened 2015-10-02
Feature/1126 get subscription list at mongo backend

fgalan:
Get all subscription functionality in mongoBackend, except some things deferred to a next PR:

* Pagination
* duration/expiration/throlling, as they are not clear yet in NGSIv2

### DIFF
--- a/src/app/contextBroker/contextBroker.cpp
+++ b/src/app/contextBroker/contextBroker.cpp
@@ -191,7 +191,7 @@
 #include "cache/subCache.h"
 #include "cache/SubscriptionCache.h"
 
-
+using namespace orion;
 
 /* ****************************************************************************
 *
--- a/src/lib/apiTypesV2/Subscription.cpp
+++ b/src/lib/apiTypesV2/Subscription.cpp
@@ -31,6 +31,9 @@
 
 #include "common/JsonHelper.h"
 
+namespace ngsiv2
+{
+
 /* ****************************************************************************
 *
 * Subscription::toJson -
@@ -41,8 +44,8 @@ std::string Subscription::toJson()
 
   jh.addString("id", this->id);
   if (!this->duration.isEmpty())
-  {
-    jh.addRaw("duration: ",this->duration.render(JSON,"", false));
+  {    
+    jh.addRaw("duration: ", "\"" + this->duration.get() + "\"");
   }
   jh.addRaw("subject", this->subject.toJson());
   jh.addRaw("notification", this->notification.toJson());
@@ -62,8 +65,8 @@ std::string Notification::toJson()
 
   jh.addString("callback", this->callback);
   if (!this->throttling.isEmpty())
-  {
-    jh.addRaw("throttling", this->throttling.render(JSON,"",false));
+  {    
+    jh.addRaw("throttling", "\"" + this->throttling.get() + "\"");
   }
   jh.addRaw("attributes", vectorToJson(this->attributes));
 
@@ -126,4 +129,6 @@ std::string EntID::toJson()
   jh.addString("type", this->type);
 
   return jh.str();
+}
+
 }
--- a/src/lib/mongoBackend/collectionOperations.cpp
+++ b/src/lib/mongoBackend/collectionOperations.cpp
@@ -1,0 +1,88 @@
+/*
+*
+* Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+*
+* This file is part of Orion Context Broker.
+*
+* Orion Context Broker is free software: you can redistribute it and/or
+* modify it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* Orion Context Broker is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+* General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+*
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Fermín Galán
+*/
+
+#include "mongoBackend/MongoGlobal.h"
+#include "logMsg/traceLevels.h"
+
+#include "mongo/client/dbclient.h"
+
+using namespace mongo;
+
+/* ****************************************************************************
+*
+* query -
+*
+*/
+bool query
+(
+    const std::string&             col,
+    const BSONObj&                 q,
+    std::auto_ptr<DBClientCursor>* cursor,
+    std::string*                   err
+)
+{
+  DBClientBase* connection = getMongoConnection();
+
+  if (connection == NULL)
+  {
+     LM_E(("Fatal Error (null DB connection)"));
+     *err = "null DB connection";
+     return false;
+  }
+
+  LM_T(LmtMongo, ("query() in '%s' collection: '%s'", col.c_str(), q.toString().c_str()));
+
+  try
+  {
+    *cursor = connection->query(col.c_str(), q);
+
+    // We have observed that in some cases of DB errors (e.g. the database daemon is down) instead of
+    // raising an exception, the query() method sets the cursor to NULL. In this case, we raise the
+    // exception ourselves
+    //
+    if (cursor->get() == NULL)
+    {
+      throw DBException("Null cursor from mongo (details on this is found in the source code)", 0);
+    }
+    releaseMongoConnection(connection);
+    LM_I(("Database Operation Successful (%s)", q.toString().c_str()));
+  }
+  catch (const DBException &e)
+  {
+    releaseMongoConnection(connection);
+    LM_E(("Database Error (%s)", e.what()));
+    *err = "Database Error: " + std::string(e.what());
+    return false;
+  }
+  catch (...)
+  {
+    releaseMongoConnection(connection);
+    LM_E(("Database Error (generic exception)"));
+    *err = "Database Error: generic exception";
+    return false;
+  }
+
+  return true;
+}
--- a/src/lib/mongoBackend/collectionOperations.h
+++ b/src/lib/mongoBackend/collectionOperations.h
@@ -1,3 +1,6 @@
+#ifndef SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_
+#define SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_
+
 /*
 *
 * Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
@@ -20,50 +23,26 @@
 * For those usages not covered by this license please contact with
 * iot_support at tid dot es
 *
-* Author: Orion dev team
+* Author: Fermín Galán
 */
 
-#include "serviceRoutinesV2/getAllSubscriptions.h"
+#include "mongo/client/dbclient.h"
 
-#include <string>
-#include <vector>
-
-#include "apiTypesV2/Subscription.h"
-#include "common/JsonHelper.h"
-#include "common/string.h"
-#include "mongoBackend/mongoListSubscriptions.h"
-#include "ngsi/ParseData.h"
-#include "rest/ConnectionInfo.h"
-#include "rest/OrionError.h"
+using namespace mongo;
 
 /* ****************************************************************************
 *
-* getAllSubscriptions -
-*
-* GET /v2/subscriptions
+* query -
 *
 */
-std::string getAllSubscriptions
+extern bool query
 (
-    ConnectionInfo*            ciP,
-    int                        components,
-    std::vector<std::string>&  compV,
-    ParseData*                 parseDataP
-)
-{
+    const std::string&             col,
+    const BSONObj&                 q,
+    std::auto_ptr<DBClientCursor>* cursor,
+    std::string*                   err
+);
 
-  std::vector<ngsiv2::Subscription> subs;
-  OrionError                oe = mongoListSubscriptions(subs, ciP->uriParam, ciP->tenant);
 
-  if (oe.code != SccOk)
-  {
-    return oe.render(ciP,"");
-  }
 
-  if ((ciP->uriParamOptions["count"]))
-  {
-    ciP->httpHeader.push_back("X-Total-Count");
-    ciP->httpHeaderValue.push_back(toString(subs.size()));
-  }
-  return  vectorToJson(subs);
-}
+#endif // SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_
--- a/src/lib/mongoBackend/mongoListSubscriptions.cpp
+++ b/src/lib/mongoBackend/mongoListSubscriptions.cpp
@@ -23,42 +23,169 @@
 * Author: Orion dev team
 */
 
-#include "mongoBackend/mongoListSubscriptions.h"
+#include "common/sem.h"
 
+#include "logMsg/logMsg.h"
+#include "logMsg/traceLevels.h"
+
+#include "mongoBackend/mongoListSubscriptions.h"
+#include "mongoBackend/MongoGlobal.h"
+#include "mongoBackend/collectionOperations.h"
+
+#include "mongo/client/dbclient.h"
+
+using namespace ngsiv2;
+
+/* ****************************************************************************
+*
+* setSubscriptionId -
+*/
+static void setSubscriptionId(Subscription* s, const BSONObj& r)
+{
+  s->id = r.getField("_id").OID().toString();
+}
+
+
+/* ****************************************************************************
+*
+* setSubject -
+*/
+static void setSubject(Subscription* s, const BSONObj& r)
+{
+  // Entities
+  std::vector<BSONElement> ents = r.getField(CSUB_ENTITIES).Array();
+  for (unsigned int ix = 0; ix < ents.size(); ++ix)
+  {
+    BSONObj ent           = ents[ix].embeddedObject();
+    std::string id        = ent.getStringField(CSUB_ENTITY_ID);
+    std::string type      = ent.getStringField(CSUB_ENTITY_TYPE);
+    std::string isPattern = ent.getStringField(CSUB_ENTITY_ISPATTERN);
+
+    EntID en;
+    if (isFalse(isPattern))
+    {
+      en.id = id;
+    }
+    else
+    {
+      en.idPattern = id;
+    }
+    en.type = type;
+
+    s->subject.entities.push_back(en);
+  }
+
+  // Condition
+  std::vector<BSONElement> conds = r.getField(CSUB_CONDITIONS).Array();
+  for (unsigned int ix = 0; ix < conds.size(); ++ix)
+  {
+    BSONObj cond = conds[ix].embeddedObject();
+    // The ONCHANGE check is needed, as a subscription could mix different conditions types in DB
+    if (std::string(cond.getStringField(CSUB_CONDITIONS_TYPE)) == "ONCHANGE")
+    {
+      std::vector<BSONElement> condValues = cond.getField(CSUB_CONDITIONS_VALUE).Array();
+      for (unsigned int jx = 0; jx < condValues.size(); ++jx)
+      {
+        std::string attr = condValues[jx].String();
+        s->subject.condition.attributes.push_back(attr);
+      }
+    }
+  }
+
+  // Note that current DB model is based on NGSIv1 and doesn't consider expressions. Thus
+  // subject.condition.expression cannot be filled. The implemetion will be enhanced once
+  // the DB model gets defined
+  // TBD
+
+}
+
+
+/* ****************************************************************************
+*
+* setNotification -
+*/
+static void setNotification(Subscription* s, const BSONObj& r)
+{
+  // Attributes
+  std::vector<BSONElement> attrs = r.getField(CSUB_ATTRS).Array();
+  for (unsigned int ix = 0; ix < attrs.size(); ++ix)
+  {
+    std::string attr = attrs[ix].String();
+    s->notification.attributes.push_back(attr);
+  }
+
+  // Callback
+  s->notification.callback = r.getStringField(CSUB_REFERENCE);
+
+  // Throttling
+  long long throttling = r.hasField(CSUB_THROTTLING)? r.getField(CSUB_THROTTLING).numberLong() : -1;
+  if (throttling > -1)
+  {
+    // FIXME P10: int -> str conversion needed
+    s->notification.throttling.set("FIXME");
+  }
+}
+
+
+/* ****************************************************************************
+*
+* setDuration -
+*/
+static void setDuration(Subscription* s, const BSONObj& r)
+{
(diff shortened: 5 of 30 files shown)


src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -42,7 +45,8 @@ std::string Subscription::toJson()
   jh.addString("id", this->id);
   if (!this->duration.isEmpty())
   {
-    jh.addRaw("duration: ",this->duration.render(JSON,"", false));
+    //jh.addRaw("duration: ",this->duration.render(JSON,"", false));
```
fgalan: I needed to change this due to the current code was producing a wrong JSON token:
```
"duration: ":
```

Maybe it is a way of using the Duration render() that fix it, without needing to composed using the duration.get().

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -63,7 +67,8 @@ std::string Notification::toJson()
   jh.addString("callback", this->callback);
   if (!this->throttling.isEmpty())
   {
-    jh.addRaw("throttling", this->throttling.render(JSON,"",false));
+    //jh.addRaw("throttling", this->throttling.render(JSON,"",false));
```
fgalan: I needed to change this due to the current code was producing a wrong JSON token:
```
"throttling":"throttling" : "FIXME"
```

Maybe it is a way of using the Throttling render() that fix it, without needing to composed using the throttling.get().

src/lib/mongoBackend/collectionOperations.cpp
```diff
@@ -0,0 +1,88 @@
...
+*/
+
+#include "mongoBackend/MongoGlobal.h"
+#include "logMsg/traceLevels.h"
+
+#include "mongo/client/dbclient.h"
+
+using namespace mongo;
+
+/* ****************************************************************************
```
fgalan: We have a lot of places in mongoBackend with the same repeated pattern: getConnection / try / catch Exception / catch (...), which make some part of mongoBackend difficult to read. Thus, the idea is to have "wrappers" for query/update/insert/remove/etc and other MongoDB operations in this module, making its usage simpler from the caller side.

I have started to use the pattern in mongoListSubscriptions but at the end all the mongoBackend will be refactored to work in this way. My only doubt if it the name (collectionOperations, as they are operations that can be applied on collections) could be improved. Feedback welcome :)

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,224 @@
...
+
+/* ****************************************************************************
+*
+* setDuration -
+*/
+extern void setDuration(Subscription* s, const BSONObj& r)
+{
+  //long long expiration = r.hasField(CSUB_EXPIRATION)? r.getField(CSUB_EXPIRATION).numberLong() : 0;
+  // FIXME P10: int -> str conversion needed
+  s->duration.set("FIXME");
```
fgalan: @jmcanterafonseca this is the part that is pending on a NGSIv2 spec discussion.

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,224 @@
...
+
+  // Callback
+  s->notification.callback = r.getStringField(CSUB_REFERENCE);
+
+  // Throttling
+  long long throttling = r.hasField(CSUB_THROTTLING)? r.getField(CSUB_THROTTLING).numberLong() : -1;
+  if (throttling > -1)
+  {
+    // FIXME P10: int -> str conversion needed
+    s->notification.throttling.set("FIXME");
```
fgalan: @jmcanterafonseca this is the part that is pending on a NGSIv2 spec discussion.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -191,7 +191,7 @@
 #include "cache/subCache.h"
 #include "cache/SubscriptionCache.h"
 
-
+using namespace orion;
```
kzangeli: I would like to open a discussion about using

    using namespace XXX

instead of writing the namespace out everywhere it is used.

My point of view: If we have more than one type with the same name (we already have Subscription, under different namespaces), it makes it very complicated to locate the use of these types.

A 'grep' of "Subscription" gives us both xxx::Subscription and yyy::Subscription.
This is not very good ...

I vote for keeping our namespaces very short (they all are), and writing them out everywhere. Like that there is no room for confusion.

src/lib/ngsi/Scope.h
```diff
@@ -31,7 +31,7 @@
 #include "common/Format.h"
 #include "orionTypes/areas.h"
 
-using namespace orion;
+//using namespace orion;
```
fgalan: It seems that having this defined in Scope.h was polutting (due to chained includes, I guess) the usage of namespaces in mongo list subscription at mongoBackend, thus ambiguety existed between orion::Subscription and ngsiv2::Subscription.

Changing this has caused a cascade of little changes in other files, but I think they are an improvements. It seems it is not a good idea to use "use namespace" in a .h file, after this experience...

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
```diff
@@ -30,40 +30,140 @@ brokerStart CB
...
+  ]
+}'
+orionCurl --url /v1/subscribeContext --payload "${payload}" --json
+echo
+echo
+
+echo "02. GET /v2/subscriptions"
 echo "========================="
-orionCurl --url /v2/subscriptions?options=count --json
+orionCurl --url /v2/subscriptions --json
```
fgalan: Pagination is not yet implemented in this PR and will be addressed (which its associated .test) in a different PR.

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -31,6 +31,9 @@
 
 #include "common/JsonHelper.h"
 
+namespace ngsiv2
```
fgalan: A first step in the issue https://github.com/telefonicaid/fiware-orion/issues/1290 solution...

src/app/contextBroker/contextBroker.cpp
fgalan: It depends... in some places you only need a punctual usage of the calss inside the namespace, so xxx::theClass is ok. In other places, you are continuously using the namespaced clase, and using xxx:: al the time make end in a noisy and less readable code.

I think it's a matter of common sense when to use one way or the other, so I wouldn't suggest not being too "normative" with this.

kzangeli: OK.
I am of another opinion, but, never mind ...

NTC

src/lib/mongoBackend/collectionOperations.cpp
kzangeli: Sounds like a great idea!

The file will only contain one function, and its name is query, right?
Why not call the files simply query.cpp/h ?

src/lib/mongoBackend/collectionOperations.cpp
```diff
@@ -0,0 +1,88 @@
...
+  catch (...)
+  {
+    releaseMongoConnection(connection);
+    LM_E(("Database Error (generic exception)"));
+    *err = "Database Error: generic exception";
+    return false;
+  }
+
+  return true;
+
```
kzangeli: extra empty line

src/lib/mongoBackend/collectionOperations.h
```diff
@@ -0,0 +1,47 @@
+#ifndef COLLECTIONOPERATIONS_H
```
kzangeli: COLLECTIONOPERATIONS_H is not according to the (still unwritten) style guide.
Actually, this is not even mentioned yet in the 'style guide work in progress' ...

This part comes from the Google style guide (the Python script) and it states that the 
definition has to be SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_

Also, it must be present at the #endif, like this (note TWO spaces before '//'):

    #endif  // SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_

Not my idea - googles. But it IS a good idea to include the directories and not only the file name in the name of the definition.

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+
+using namespace ngsiv2;
+
+/* ****************************************************************************
+*
+* setSubscriptionId -
+*/
+extern void setSubscriptionId(Subscription* s, const BSONObj& r)
+{
+  s->id = r.getField("_id").OID().toString();
```
kzangeli: Doubt. "_id" must always be there, right?
I mean, if "_id" is not found, what would happen?  SIGSEGV?

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+}
+
+
+/* ****************************************************************************
+*
+* setSubject -
+*/
+extern void setSubject(Subscription* s, const BSONObj& r)
```
kzangeli: *extern* is for header files.
The fact is is not *static* tells us it is 'extern'.

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+  // Entities
+  std::vector<BSONElement> ents = r.getField(CSUB_ENTITIES).Array();
+  for (unsigned int ix = 0; ix < ents.size(); ++ix)
+  {
+    BSONObj ent           = ents[ix].embeddedObject();
+    std::string id        = ent.getStringField(CSUB_ENTITY_ID);
+    std::string type      = ent.getStringField(CSUB_ENTITY_TYPE);
+    std::string isPattern = ent.getStringField(CSUB_ENTITY_ISPATTERN);
+
+    EntID en;
```
kzangeli: Why not move "EntID en" up one line and align it with the rest of the local vars?

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+        std::string attr = condValues[jx].String();
+        s->subject.condition.attributes.push_back(attr);
+      }
+    }
+  }
+
+  // Note that current DB model is based in NGSIv1 and doesn't consider expressions. Thus
+  // subject.condition.expression cannot be filled. The implemetion will be enhanced once
+  // the DB model gets defined
+  // TBD
```
kzangeli: type: based on

Now, would this comment not be more visible if in the function header?

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+
+  LM_T(LmtMongo, ("Mongo List Subscriptions"));
+
+  /* ONTIMEINTERVAL subscription are not part of NGSIv2, so they are excluded.
+   * Note that expiration is not taken into account (in the future, a q= query
+   * could be added to the operation in order to filter results) */
+  std::auto_ptr<DBClientCursor>  cursor;
+  std::string                    err;
+  std::string                    conds = std::string(CSUB_CONDITIONS) + "." + CSUB_CONDITIONS_TYPE;
+  BSONObj                        q = BSON(conds << "ONCHANGE");
```
kzangeli: align '='

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+    BSONObj r = cursor->next();
+    LM_T(LmtMongo, ("retrieved document: '%s'", r.toString().c_str()));
+
+    Subscription s;
+    setSubscriptionId(&s, r);
+    setSubject(&s, r);
+    setNotification(&s, r);
+    setDuration(&s, r);
+    subs.push_back(s);
+
```
kzangeli: unwanted empty line

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+  }
+
+  reqSemGive(__FUNCTION__, "Mongo List Subscriptions", reqSemTaken);
+  return OrionError(SccOk);
+
```
kzangeli: unwanted empty line

src/lib/mongoBackend/mongoListSubscriptions.cpp
```diff
@@ -23,42 +23,171 @@
...
+  while (cursor->more())
+  {
+    BSONObj r = cursor->next();
+    LM_T(LmtMongo, ("retrieved document: '%s'", r.toString().c_str()));
+
+    Subscription s;
+    setSubscriptionId(&s, r);
+    setSubject(&s, r);
+    setNotification(&s, r);
+    setDuration(&s, r);
```
kzangeli: These four functions could be methods of the Subscription class instead, right?

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -456,6 +458,8 @@ std::string ContextAttribute::render
       std::string effectiveValue      = "";
       bool        valueIsNumberOrBool = false;
 
+      using namespace orion;
+
```
kzangeli: using namespace orion *twice*.

Perhaps this 'directive' should be in the beginning of the file, right after includes.
That way the risk to repeat it becomes very small.

src/lib/ngsi/Scope.h
kzangeli: +1

Yeah, if in a header file it gets spread to *every* source file that somehow includes the header file.
Seems like a very bad idea.
Remove the comment (I mean, the entire line :-)) ...

src/lib/mongoBackend/collectionOperations.cpp
fgalan: I think there is value in grouping all the operations in the same module, as they are closely related in their purpose (and in its structure, e.g. all them start getting the connection, have same try/catch with 2 catch clauses). Even more, it could happen (we will see as new function are being added) that some common functionality between functions may be factorized.

Probably query() is not a good name for the function, as is very close to mongo driver query() method. Maybe doQuery() could be better...

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
```diff
@@ -30,40 +30,140 @@ brokerStart CB
...
-Content-Length: 513
+Content-Length: 733
 Content-Type: application/json
-X-Total-Count: 2
 Date: REGEX(.*)
 
 [
     {
-        "id": "s1",
+        "duration: ": "FIXME",
```
kzangeli: This must be fixed before next release, right? :-)

src/lib/mongoBackend/collectionOperations.cpp
kzangeli: "do" doesn't add anything to the name ...

dbQuery?
mongoQuery?

Also, all requests aren't queries, really ...

dbRequest?

src/lib/mongoBackend/collectionOperations.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "mongo/client/dbclient.h"
+
+using namespace mongo;
+
+/* ****************************************************************************
+*
+* query -
+*
+*/
+bool query(
```
kzangeli: '(' on its own line - style guide police ...

src/lib/mongoBackend/collectionOperations.cpp
fgalan: In future versions of this module, functions for insert, update, remove, etc. will be added, so dbRequest doesn't seem to be suitable.

Thus, let's keep query() by the moment, and if some naming problem arises in the future, change it.

src/lib/mongoBackend/collectionOperations.h
fgalan: The IDE chose COLLECTIONOPERATIONS_H unnoticelly. I will change to SRC_LIB_MONGOBACKEND_COLLECTIONOPERATIONS_H_

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: `_id` is a mandatory field in every document in every collection in MongoDB. Only if the driver fails or some other very critical and werid situation occur we will get here a SGIEGV.

Somehow related and more dangerous (as MongoDB doesn't have any "protection" in this case) is not finding a field in the returned BSON or having the field but with a different type (e.g. it is a string but mongoBackend code expect and integer). The fix for this is included in https://github.com/telefonicaid/fiware-orion/issues/136

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Mmmm... not sure if that would be a good idea. Right now the comment is in the place were the new code to fill the expression field would be in the future.

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: It would be an option. But they are too "tied" to the mongoBackend operation (I mean it doesn't seem to be a functionality that could be exploited by another), so by the moment maybe is better to have them here.

src/lib/ngsi/ContextAttribute.cpp
fgalan: Yes, you are right... Not sure why I have ended with a duplicated declaration. I'll check.

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
fgalan: Yes. 0.25.0 will not released with this ugly FIXME here.

src/lib/mongoBackend/collectionOperations.cpp
fgalan: No alternative name has been suggested, so let's go with collectionOperations.cpp/c

NTC

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: NTC (it will not be fixed in this PR, but in a next one; discussion is yet ongoing)

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: NTC (it will not be fixed in this PR, but in a next one; discussion is yet ongoing)

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
fgalan: NTC (just informative)

src/lib/apiTypesV2/Subscription.cpp
fgalan: NTC (just informative)

src/lib/apiTypesV2/Subscription.cpp
kzangeli: Throttling::render seems to give back the entire json string already. Feeding it to addRaw also, gives you the ** "throttling": ** twice ...
Or, so it seems

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: NTC

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: I think this was an stupid copy-paste error... I'll change them to *static*.

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Becasue in the current position it is more "tied" with the if/else that fills the variable.

fgalan: NTC

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: NTC

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: At the end, they are quite "tied" to the mongoBacked model, so they will stay as they are now (however, if we changer our mind in the future, it is very easy to move them to the ngsiv2::Subscription class).

NTC

src/lib/apiTypesV2/Subscription.cpp
fgalan: Fixed in 2d1b4b8 (just the removal of the comment line)

src/lib/apiTypesV2/Subscription.cpp
fgalan: Fixed in 2d1b4b8 (just the removal of the comment line)

src/lib/ngsi/Scope.h
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/collectionOperations.cpp
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/collectionOperations.h
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/mongoListSubscriptions.cpp
fgalan: Fixed in 2d1b4b8

src/lib/ngsi/ContextAttribute.cpp
fgalan: Fixed in 2d1b4b8

test/functionalTest/cases/1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
fgalan: NTC

src/lib/mongoBackend/collectionOperations.cpp
fgalan: Fixed in 2d1b4b8

src/lib/mongoBackend/mongoListSubscriptions.cpp
kzangeli: What about the typo?:

    based in  => based on

kzangeli:
LGTM

PR #1313, opened 2015-10-02
Hardening/1306 is double string2double

kzangeli:
New function for isFloat - right now only used by mongoBackend, but to be integrated also in code for the new JSON format in v2.

### DIFF
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -718,3 +718,80 @@ std::string servicePathCheck(const char* servicePath)
 
   return "OK";
 }
+
+
+
+/* ****************************************************************************
+*
+* isDouble - is there a double value in the string 's' ?
+*
+* From the manual page of strtod:
+*   double strtod(const char *nptr, char **endptr);
+*
+*   If endptr is not NULL, a pointer to the character after the last character
+*   used in the conversion is stored in the location referenced by endptr.
+*
+*   If no conversion is performed, zero is returned and the value of nptr is stored 
+*   in the location referenced by endptr.
+*
+* Now, this is just perfect. It is just what we need to assure that a string is
+* in fact 'expressing' a double.
+* All we need to do is to check that endptr only contains whitespace (or nothing at all)
+* after the call to 'strtod'. And also, that a covervsion has actually been performed.
+* If so, the string sent to isDouble is indeed a valid 'double'.
+*
+* More from the man page:
+*   If the correct value would cause overflow, plus or minus HUGE_VAL (HUGE_VALF, HUGE_VALL)
+*   is returned (according to the sign of the value), and ERANGE is stored in errno.
+*   If the correct value would cause underflow, zero is returned and ERANGE is stored in errno.
+*
+* So, we will need to check errno for ERANGE also.
+*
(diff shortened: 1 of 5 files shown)


src/lib/common/string.cpp
```diff
@@ -718,3 +718,66 @@ std::string servicePathCheck(const char* servicePath)
...
+*   is returned (according to the sign of the value), and ERANGE is stored in errno.
+*   If the correct value would cause underflow, zero is returned and ERANGE is stored in errno.
+*
+* So, we will need to check errno for ERANGE also.
+*
+* This in turn will make a 'HUGE_VAL double' 12e999999 (is that enough? :-)) be treated as an error.
+* It IS a double, but this function will say it is not, as it actually is not a valid double for
+* this computer as the computer cannot represent it as the C builtin type 'double' ... OK!
+*/
+bool isFloat(char* s, double* dP)
```
fgalan: Thus, this functions sums the functionality of a function to check it a string represent a float and to convert the float, right?

I understand that it doesn't make sense to have two separate functions (or maybe I'm wrong?) but, what about of renaming to `str2double`? The rationale is that, from the two functionalities, I think the most "complete" one is the string parsing. Format checking (as the `isFloat` name suggest) is a "side-effect" of the conversion.

kzangeli: Yes, I removed the other one. No sense in having them both.
And no, I am not very convinced that 'isFloat' is the best possible name. Far from it.
str2double is just so ugly ... but, yes, it describes better what the function actually does.

I was thinking of atod also (atof and atoi already exists bue atod doesn't)

fgalan: str2double is ugly but at the same time describes clearly the functionality :)

Another suggestion: toDouble()

fgalan: ...although `toDouble()` may lead to think that the return value of the function is a double (and it is the bool that means sucess/failure). Forget it :)

src/lib/common/string.h
```diff
@@ -161,16 +161,26 @@ extern void strReplace
...
+
+
+
 /* ****************************************************************************
 *
 * toString -
 */
-template <typename T>
-std::string toString(T t) {
+template <typename T> std::string toString(T t)
```
fgalan: In which part of the code is this function used?

src/lib/common/string.cpp
fgalan: > Yes, I removed the other one. No sense in having them both.

You mean in this PR? There is no trace in the diff of that removal... or at least I'm not able to find it.

src/lib/common/string.h
kzangeli: this template is part of the new json render proposal

src/lib/common/string.cpp
kzangeli: Yes, I had two functions initially, but I decided to remove one ...
New name: str2double

src/lib/common/string.h
fgalan: NTC

src/lib/common/string.cpp
kzangeli: Name changed to str2double in e215979
Removing one of the two functions might have been done before I even sent the PR, that could be why you can't see it ...

fgalan:
LGTM

PR #1325, opened 2015-10-06
Task/part6 attribute value render

kzangeli:
Part 6 of #1259 - rendering of attribute value

### DIFF
--- a/src/lib/ngsi/ContextAttribute.h
+++ b/src/lib/ngsi/ContextAttribute.h
@@ -85,6 +85,7 @@ typedef struct ContextAttribute
   std::string  renderAsJsonObject(ConnectionInfo* ciP, RequestType request, const std::string& indent, bool comma, bool omitValue = false);
   std::string  renderAsNameString(ConnectionInfo* ciP, RequestType request, const std::string& indent, bool comma = false);
   std::string  toJson(bool isLastElement, bool types, const std::string& renderMode);
+  std::string  toJsonAsValue(ConnectionInfo* ciP);
   void         present(const std::string& indent, int ix);
   void         release(void);
   std::string  toString(void);
--- a/test/functionalTest/cases/0000_content_related_headers/in_out_formats.test
+++ b/test/functionalTest/cases/0000_content_related_headers/in_out_formats.test
@@ -111,7 +111,7 @@ echo '================================================================='
 echo
 
 echo "2.4: Content-Type OK, Accepts text/plain: 406, Response in XML"
-orionCurl --url "$url" --payload "${payload}" --in "application/xml" --out "text/plain"
+orionCurl --url "$url" --payload "${payload}" --in "application/xml" --out "text/plain" --payloadCheck xml
 echo '================================================================='
 echo
 
@@ -222,7 +222,12 @@ Content-Length: 160
 Content-Type: application/xml
 Date: REGEX(.*)
 
-<orionError> <code>415</code> <reasonPhrase>Unsupported Media Type</reasonPhrase> <details>not supported content type: text/plain</details> </orionError>
+<?xml version="1.0"?>
+<orionError>
+  <code>415</code>
+  <reasonPhrase>Unsupported Media Type</reasonPhrase>
+  <details>not supported content type: text/plain</details>
+</orionError>
 =================================================================
 
 1.5: Unrecognized Content-Type, Accepts NOTHING: 415, response in XML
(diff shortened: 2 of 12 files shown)


src/lib/ngsi/ContextAttribute.h
```diff
@@ -85,6 +85,7 @@ typedef struct ContextAttribute
   std::string  renderAsJsonObject(ConnectionInfo* ciP, RequestType request, const std::string& indent, bool comma, bool omitValue = false);
   std::string  renderAsNameString(ConnectionInfo* ciP, RequestType request, const std::string& indent, bool comma = false);
   std::string  toJson(bool isLastElement, bool types, const std::string& renderMode);
+  std::string  toJsonAsValue(ConnectionInfo* ciP);
```
fgalan: What's the different between toJson() and toJsonAsValue() ?

test/functionalTest/cases/0000_content_related_headers/in_out_formats.test
```diff
@@ -86,12 +86,12 @@ echo '================================================================='
 echo
 
 echo "1.4: Unrecognized Content-Type, Accepts text/plain: 415, response in XML"
-orionCurl --url "$url" --payload "${payload}" --in "text/plain" --out "text/plain"
+orionCurl --url "$url" --payload "${payload}" --in "text/plain" --out "text/plain" --payloadCheck xml
```
fgalan: Why this change was needed in such old .test?

test/functionalTest/cases/0997_GET_v2_entity_single_attribute_value/GET_v2_entity_single_attribute_value.test
```diff
@@ -180,65 +180,41 @@ Date: REGEX(.*)
...
 Date: REGEX(.*)
 
-{
-    "value": {
-        "metadata": {},
-        "type": null,
-        "value": "Mercedes Benz"
-    }
-}
+Mercedes Benz
```
fgalan: Should this be

```
Mercedes Benz
```

or

```
"Merces Benz"
```

?

In order to kep "symmetry" with PUT operation, it should have "". In order to keep simpler, it should be without "".

I don't have an answer for this :) Please, cast the question in the NGSIv2 skype chat spec.

test/functionalTest/cases/1179_options_test/GET_v2_ent_attr_val_options_text.test
```diff
@@ -42,49 +42,56 @@ echo "====================="
 payload='{
   "id": "E1",
   "type": "T1",
-  "a_string":"some value",
```
fgalan: Only indentation changes in this part, right?

src/lib/ngsi/ContextAttribute.h
kzangeli: Normally ContextAttribute is rendered as 

    "name": {
       "type": "x",
       "value": "123",
      ...

This new method I added for the case  "GET /v2/entities/EEE/attr/attrrName/value", where only the value is rendered (and as standalone).
I could have added it to **toJson()**, with a new parameter, but it would make that method innecessarily complex.

test/functionalTest/cases/0000_content_related_headers/in_out_formats.test
kzangeli: Fixes in orionCurl.
If out format is json or xml, orionCurl does the payloadCheck (beautifier).
In this case the out-format is text/plain, but orionCurl STILL did the payloadCheck.
The new outputs for text/plain, that are just a string or a Number fail, of course.
I had to modify the *payload check mechanism* in orionCurl.
Here, the output is actually XML, and I want to force the payload check mechanism.

test/functionalTest/cases/0997_GET_v2_entity_single_attribute_value/GET_v2_entity_single_attribute_value.test
kzangeli: Yeah, I was also unsure about this.

Would it not be nice to have:
  Content-Type: text/number
or something similar?

"Mercedes Benz" is not very good (apart from being a bit ugly), as the users will have to strip the citation marks ...

test/functionalTest/cases/1179_options_test/GET_v2_ent_attr_val_options_text.test
kzangeli: Now that I had to modifiy this file, I fixed the indentations also. It was ugly ...

The real change is that **?options=text** is removed and that the response is in JSON for compound values

kzangeli:
LGTM

src/lib/ngsi/ContextAttribute.h
fgalan: NTC

test/functionalTest/cases/0000_content_related_headers/in_out_formats.test
fgalan: NTC

test/functionalTest/cases/0997_GET_v2_entity_single_attribute_value/GET_v2_entity_single_attribute_value.test
fgalan: Already discussed. NTC in this PR

test/functionalTest/cases/1179_options_test/GET_v2_ent_attr_val_options_text.test
fgalan: NTC

fgalan:
LGTM

Issue #1126, opened 2015-08-06
APIv2: Get subscription list


fgalan:
Implement the following operation in in v2:

```
GET /v2/subscriptions
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
mongoBackend part partically addressed in PR #1312. What is pending after that is:

* Pagination
* duration/expiration/throlling, as they are not clear yet in NGSIv2

fgalan:
mongoBackend part partial implementation in PR https://github.com/telefonicaid/fiware-orion/pull/1330/files. What is pending after that is:

* Pagination

fgalan:
Finished in PR https://github.com/telefonicaid/fiware-orion/pull/1391

Issue #1328, opened 2015-10-06
Pause subscriptions


fgalan:
Use the `status` field in a subscription in order to "pause" (so status changes to "paused") or "resume" (so status changes to "active" or "expired").

jmcanterafonseca:
We need to take into account that the initial status of a subscription could not be active, but suspended or paused.

fgalan:
.apib changes in PR #1926

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1998. Moving to QATestPending milestone so @iariasleon can test and finally close it.

iariasleon:
LGTM
implemented in https://github.com/telefonicaid/fiware-orion/blob/develop/test/acceptance/behave/components/ngsiv2/subscriptions/create_a_new_subscription/create_a_new_subscription_status.feature

PR #1330, opened 2015-10-07
ADD expires, count and lastNotification in NGSIv2 subscription rendering

fgalan:
Part of #1126 implementation.


doc/manuals/admin/database_model.md
```diff
@@ -86,55 +86,55 @@ Fields:
 Example document:
 
 ```
- {
```
fgalan: My IDE has ruin the whole formatting (only one line was touched). I will fix it.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -855,7 +855,10 @@ A subscription is represented by a JSON object with the following fields:
 + `id`: Subscription unique identifier. Automatically created at creation time.
 + `subject`: It is an object that describes the subject of the subscription.
 + `notification`: It is an object that describes the notification received by the subscriber.
-+ `duration`: Duration of the subscription in ISO8601 format. Infinite if not specified.
++ `expires`: Subscription expiration date in ISO8601 format. Permanent subscriptions must omit
+   this field.
++ `status` (not editable, only in GET operations): Either `active` (for permanent subscriptions or if
+   the subscription hasn't expired yet) or `expired` (for expired subscriptions).
```
jmcanterafonseca: nit: has not

jmcanterafonseca: maybe in the future we could have a 'suspended' status as well.

src/lib/common/JsonHelper.cpp
```diff
@@ -149,6 +154,46 @@ void JsonHelper::addRaw(const std::string& key, const std::string& value)
...
+* JsonHelper::addDate -
+*/
+void JsonHelper::addDate(const std::string& key, long long timestamp)
+{
+  if (!empty)
+  {
+    ss << ',';
+  }
+
+  // 80 bytes are large enough to store any ISO8601 string safely
```
fgalan: "1970-04-26T17:46:40.00Z" is 24 char long. Event assuming some more space for the timezone in the future, I think that 80 bytes will suffice al cases.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -855,7 +855,10 @@ A subscription is represented by a JSON object with the following fields:
 + `id`: Subscription unique identifier. Automatically created at creation time.
 + `subject`: It is an object that describes the subject of the subscription.
 + `notification`: It is an object that describes the notification received by the subscriber.
-+ `duration`: Duration of the subscription in ISO8601 format. Infinite if not specified.
++ `expires`: Subscription expiration date in ISO8601 format. Permanent subscriptions must omit
+   this field.
++ `status` (not editable, only in GET operations): Either `active` (for permanent subscriptions or if
```
jmcanterafonseca: for the moment ok for only in GET operations but I can imagine creating subscriptions which are not initially active ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -855,7 +855,10 @@ A subscription is represented by a JSON object with the following fields:
 + `id`: Subscription unique identifier. Automatically created at creation time.
 + `subject`: It is an object that describes the subject of the subscription.
 + `notification`: It is an object that describes the notification received by the subscriber.
-+ `duration`: Duration of the subscription in ISO8601 format. Infinite if not specified.
++ `expires`: Subscription expiration date in ISO8601 format. Permanent subscriptions must omit
+   this field.
```
kzangeli: ISO8601 is a very wide format.
We will only accept exact date, including hour, minute and seconds, and ONLY UTC is to be used. Right?
This is very important information and must be documented. 

Emphasize on that it is a **date**, not a duration.

For example, "P1M" is ISO8601 but will not be accepted.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -873,8 +876,10 @@ A `notification` object contains the following subfields:
   in the notification
 + `callback` : URL pointing to the service which will be invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas (e.g. schemas for web sockets) could also be supported.
-+ `throttling`: Minimal period of time (in ISO8601 format) which must elapse between two consecutive notifications. It
++ `throttling`: Minimal period of time in seconds which must elapse between two consecutive notifications. It
   is optional.
++ `count` (not editable, only in GET operations): Number of notifications sent due to this subscription.
++ `lastNotification`: Last notification date in ISO6601 format.
```
jmcanterafonseca: lastNotification seems to be a bit redundant with the field 'notification' . I think we can find another name for it ... but I do not have a good one right now

kzangeli: 8601?  :-)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -873,8 +876,10 @@ A `notification` object contains the following subfields:
   in the notification
 + `callback` : URL pointing to the service which will be invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas (e.g. schemas for web sockets) could also be supported.
-+ `throttling`: Minimal period of time (in ISO8601 format) which must elapse between two consecutive notifications. It
++ `throttling`: Minimal period of time in seconds which must elapse between two consecutive notifications. It
   is optional.
++ `count` (not editable, only in GET operations): Number of notifications sent due to this subscription.
```
jmcanterafonseca: Could it be ```timesSent``` ?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -896,6 +901,13 @@ Response:
 * Successful operation uses 200 OK
 * Errors use a non-200 and (optionally) an error payload. See subsection on Error Responses for more details.
 
++ Parameters
+    + limit: 10 (optional, number) - Limit the number of types to be retrieved
+    + offset: 20 (optional, number) - Skip a number of records
+    + options (optional, string) - Options dictionary
+      + Members
+          + count - when used, the total types number is returned as a HTTP in the response named `X-Total-Count`
```
jmcanterafonseca: cut & paste problem. It should say total subscription number

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: Also, would not "expirationDate" be a **much** better name than "expires"?

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -63,10 +64,18 @@ std::string Notification::toJson()
...
+    jh.addLongLong("throttling", this->throttling);
+  }
+  if (this->count > 0)
+  {
+    jh.addLongLong("count", this->count);
+  }
+  if (this->lastNotification > 0)
+  {
+    jh.addDate("lastNotification", this->lastNotification);
   }
```
kzangeli: An empty line between every if () {}  makes the function easier to read.
Not yet in style-guide, but it will come (as SHOULD, not MUST)

src/lib/common/JsonHelper.cpp
```diff
@@ -149,6 +154,46 @@ void JsonHelper::addRaw(const std::string& key, const std::string& value)
...
+  // We use gmtime() to get UTC strings, otherwise we would use localtime()
+  // Date pattern: 1970-04-26T17:46:40.00Z
+  char buffer[80];
+  time_t rawtime = (time_t) timestamp;
```
kzangeli: style-guide: align vars horizontally

src/lib/common/JsonHelper.cpp
```diff
@@ -149,6 +154,46 @@ void JsonHelper::addRaw(const std::string& key, const std::string& value)
...
+  strftime (buffer, sizeof(buffer),"%Y-%m-%dT%H:%M:%S.00Z", gmtime(&rawtime));
```
kzangeli: style-guide: no space between function name and '('

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: lastNotificationTime

src/lib/common/JsonHelper.cpp
```diff
@@ -149,6 +154,46 @@ void JsonHelper::addRaw(const std::string& key, const std::string& value)
 }
 
 
+/* ****************************************************************************
+*
+* JsonHelper::addDouble -
```
crbrox: addLongLong ?

fgalan: Let's use addNumber at the end.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: As far as I understand, the exact encoding is part of the ISO8601 string, so it can be univocally identified (or generated) by the client. I mean, if you use:

2016-04-05T14:00:00.00Z

you are meaning a date in UTC format while if you use

1994-11-05T08:15:30-05:00

you are referring to a time in GMT-5 timezone.

So, no ambiguety is possible within the ISO8601 standard.

Moreover, currently Orion generates dates in UTC but 1) this may change in the future, 2) we haven't defined yet what to do with dates provided by clients not using UTC (and maybe it is easy to process them if the library we use from parsing convert them to the right timestamp time_t for free). Moreover, both can be considered subtle detailt related with our particular NGSIv2 implementation.

Bottom line: I think is ok have the current wording in the spec. However, as I understand the whole specification will be reviewed once NGSIv2 get fully implemented, we have an opportunity in the future to change it if it doesn't match the final implementation in Orion.

Regarding emphasizing "date", I also think that it is pretty clear with the current working. Adding more stuff could make understanding harder (*excusatio non petita accusatio manifesta* ;).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: It would depend on how https://github.com/telefonicaid/fiware-orion/issues/1328 gets implemented at the end. Could you add a comment about this in that issue, pls?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Regarding "expirationDate", I like "expires" as it is shorter, but let's discuss offline about it...

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: My idea was adjusting the .apib to add the new status when we address https://github.com/telefonicaid/fiware-orion/issues/1328. Some uncertainty exist at the present moment.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Yes, probably it is better... "count" is the same work we are using for one of the options, so it could misselead readers.

src/lib/common/JsonHelper.cpp
kzangeli: Sounds good.
Might have to add a

    addNumber(const std::string& key, double value)

in the future?

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: So. I might have misunderstood this.
If this part refers to a **response** from the broker, then I agree.
I got the impression (don't know from where) that it was a **request** being described.
Your comments make me believe it's a response ...

src/lib/common/JsonHelper.cpp
fgalan: That's the idea... to add new addNumber() functions for other kind of number types if we need in the future.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Actually, it refers to both (JSON definition is common... maybe looking the whole document could help to see how it fits: http://telefonicaid.github.io/fiware-orion/api/v2/).

kzangeli: In that case, I would insist on clearly explaining that "expires" needs to be an exact date, that "P1M" is ISO8601, but the broker will not accept it.

Also, as discussed somewhere, the date must be complete with hour, minute and second (that *might* be part of ISO8601, but still, it is never a bad idea to explain it).

About "expires" vs "expirationDate" - "expires" gives me the feeling of a boolean. "expirationDate" on the other hand is very clear. It also removes the doubt about using "P1M", which is not a date, but a period of time.

src/lib/apiTypesV2/Subscription.cpp
fgalan: I think in this case (wich one line statement, very regular among them, all using add*() method), having in this way is better. Thus, I'm applying the SHOULD ;)

NTC

kzangeli: Actually, this is a pretty good candidate for the one-liner ifs (if you want to save lines and improve readibility). Look:

    if (this->throttling > 0)           { jh.addLongLong("throttling", this->throttling);         }
    if (this->count > 0)                { jh.addLongLong("count", this->count);                   }
    if (this->lastNotification > 0)     { jh.addDate("lastNotification", this->lastNotification); }

Just a suggestion.

src/lib/common/JsonHelper.cpp
fgalan: NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: When the referred standard describes the behavior I think the best approach is not including the details in the document pointing to the standard for two main reasons:

* Avoid potential contractions (noboby will provide a better description than the standard itself)
* Avoid making the document unnecesarry long or complex

In this particular case, I think is pretty clear: use ISO8601 dates. On the one hand, within the ISO8601 specification I think is pretty clear what a date is and what a interval (duration) is. On the other hand, think we are very "biassed" to the "P1M" notation because we have suffered a long time :) but most of poeple doesn't think in duration when the think ISO8601. Thus, I don't see the need to stress this.

However, it is not so important in this PR to have a perfect wording for this, as I said in a previous comment. We don't know yet all the details on how the ISO8601 stuff will be implemented at the end.

kzangeli: OK, I give up.

NTC

doc/manuals/admin/database_model.md
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed (the nit) in 18aa61203ef827009f1f4d4b64019cb6025110b8

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed (the ISO number in) 18aa61203ef827009f1f4d4b64019cb6025110b8

Regarding field name, lastNotification was decided at the end (see skype chat discussion).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

src/lib/common/JsonHelper.cpp
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

src/lib/common/JsonHelper.cpp
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

src/lib/common/JsonHelper.cpp
fgalan: Fixed in 18aa61203ef827009f1f4d4b64019cb6025110b8

fgalan:
I have needed to put the test related with "get a particular subscription" in .DISABLED in this PR, but it will be recovered very soon (when the "get a particular subscription" mongoBackend part starts to be implemented).

crbrox:
LGTM

kzangeli:
LGTM

Issue #1289, opened 2015-09-25
[BUG] when is exceeded the index key limit in mongo


iariasleon:
the index key limit in mongo is 1024, see issue https://github.com/telefonicaid/fiware-orion/issues/1283

if `type` or `id` exceed index key limit ,CB responds 201 and nothing is stored in mongo.

#### Request:
```
 POST http://qa-orion-fe-01:1026/v2/entities
  Content-Type: application/json
  Fiware-Service: test_entities_type
  Fiware-ServicePath: /test
[...]

PR #1344, opened 2015-10-08
REFACTOR MongoCommonUpdate operations using connectionOperations functions

fgalan:
This PR concludes the refactor to use all connection operations from connectionsOperations.cpp/h module. It is being progressed now as it may conflict with the fix for https://github.com/telefonicaid/fiware-orion/issues/1289

In addition, I have take the opportunity to add one CNR entry missing from a previous PR.


kzangeli:
LGTM

PR #1349, opened 2015-10-10
FIX hotfix to recover $where

fgalan:
Some .test are failing, but they seem to be things under control (take into account this is not a "regular" PR, but a hotfix).


kzangeli:
LGTM

PR #1350, opened 2015-10-11
valgrind stuff

kzangeli:
PR just to open a dialogue ...


fgalan:
Probably this PR is no longer meaninful, but let's keep as long as the "cache crisis" continues, just in case.

kzangeli:
I have taken care of this 'fix' in the 'new sub cache' code

PR #1352, opened 2015-10-13
Removed old subcache, prepared for new

kzangeli:
Old subcache removed via #if SUB_CACHE. new function in mongoSubCache cpp logs all subs that will be in the cache in the new implementation


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1212,10 +1214,11 @@ static bool addTriggeredSubscriptions_withCache
   //
   // Now, take the 'patterned subscriptions' from the Subscription Cache and add more TriggeredSubscription to subs
   //
+#if SUB_CACHE_ON
   std::vector<Subscription*> subVec;
 
   subCache->semTake();
-  //subCache->lookup(tenant, servicePath, entityId, entityType, attr, &subVec);
+  subCache->lookup(tenant, servicePath, entityId, entityType, attr, &subVec);
```
rbl219: OK, here this line is uncommentes 
NTC

rbl219:
LGTM

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, and the whole 'block' is removed by the

    #if SUB_CACHE_ON
    #endif

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1417,8 +1421,8 @@ static bool addTriggeredSubscriptions_noCache
           lastNotification,
           sub.hasField(CSUB_FORMAT) ? stringToFormat(STR_FIELD(sub, CSUB_FORMAT)) : XML,
           STR_FIELD(sub, CSUB_REFERENCE),
-          //subToAttributeList(sub)); siganture changed in TriggeredSubscription() constructor in 0.24.0
-          subToAttributeList(sub), NULL);
+          //subToAttributeList(sub)); signature changed in TriggeredSubscription() constructor in 0.24.0
+          subToAttributeList(sub), "");
```
kzangeli: This here was a failure, a remain from the TriggeredSubscription accepted a pointer to the subscription in the cache.
Made broker crash if started with -noCache

NTC

fgalan:
LGTM

PR #1355, opened 2015-10-13
subcache filled and emptied but not used

kzangeli:
cache created, emptied and recreated.
Not used, still a lot missing ...

### DIFF
--- a/src/lib/mongoBackend/mongoSubCache.cpp
+++ b/src/lib/mongoBackend/mongoSubCache.cpp
@@ -1,3 +1,30 @@
+/*
+*
+* Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
+*
+* This file is part of Orion Context Broker.
+*
+* Orion Context Broker is free software: you can redistribute it and/or
+* modify it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* Orion Context Broker is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+* General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+*
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Ken Zangelin
+*/
+#include <regex.h>
+#include <string>
+
 #include "mongo/client/dbclient.h"
 
 #include "logMsg/logMsg.h"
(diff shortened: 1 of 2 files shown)


src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -13,10 +40,237 @@ using std::auto_ptr;
...
+
+
+
+/* ****************************************************************************
+*
+* EntityInfo::present - 
+*/
+void EntityInfo::present(const std::string& prefix)
+{
+  LM_F(("%sid:    regex describing EntityId::id (the original string is not saved)", prefix.c_str()));
```
fgalan: Maybe LM_M would me this more easy to "catch" to be removed at the end? (I think that LM_F are used for present() methods, which is ok, but probably I don't remembering well...).

fgalan: Argghh... ignore my comment... I have just realize that the LM_F are inside within a present() method :$

NTC

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -26,28 +280,155 @@ static void subCacheItemUpdate(const char* tenant, BSONObj* subP)
...
-  int64_t                   expiration        = subP->hasField(CSUB_EXPIRATION)? subP->getField(CSUB_EXPIRATION).Long() : 0;
-  int64_t                   throttling        = subP->hasField(CSUB_THROTTLING)? subP->getField(CSUB_THROTTLING).Long() : -1;
+  //
+  // 03. Create CachedSubscription
+  //
+  CachedSubscription* cSubP = (CachedSubscription*) calloc(1, sizeof(CachedSubscription));
+
+  if (cSubP == NULL)
+  {
+    LM_E(("Runtime Error (cannot allocate memory for a cached subscription: %s)", strerror(errno)));
```
fgalan: Probably this should be Fatal Error and invoke a orionExit(). The broker shouldn't continue running if memory has been exahusted.

However, don't need to fix this right now. Just use a "FIXME" above this line to indicate that behaviour for a next round.

fgalan:
Mostly LGTM (only a minor comment). However, before merging, just to check... is the code in this branch passin the "jmeter test" (DEBUG & RELEASE, several times each one) without breaking?

(For next PRs, probably this should be mentioned in the PR body :)

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Sure, no real meaning to continue when we have no more RAM ...
Fixing it in my current branch, as to save time.

kzangeli:
I will *not* send a PR without the branch running OK on my local machine, of course not :-)

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Ok. NTC

fgalan:
LGTM

PR #1369, opened 2015-10-15
Task/trial1 create update delete subs

kzangeli:
Rest of new sub-cache implemented (but not tested):

The sub-cache is now modified when a subscription is:
- created
- updated
- deleted

### DIFF
--- a/src/lib/mongoBackend/mongoSubCache.cpp
+++ b/src/lib/mongoBackend/mongoSubCache.cpp
@@ -441,6 +441,32 @@ void mongoSubCacheDestroy(void)
 
 /* ****************************************************************************
 *
+* tenantMatch - 
+*/
+static bool tenantMatch(const char* tenant1, const char* tenant2)
+{
+  if ((tenant1 == NULL) && (tenant2 == NULL))
+  {
+    return true;
+  }
+
+  if ((tenant1 == NULL) || (tenant2 == NULL))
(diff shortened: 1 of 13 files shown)


src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -481,19 +441,73 @@ void mongoSubCacheDestroy(void)
...
   //
-  BSONElement  idField = subP->getField("_id");
+  BSONElement  idField = sub.getField("_id");
 
   if (idField.eoo() == true)
   {
-    LM_E(("Database Error (error retrieving _id field in doc: '%s')", subP->toString().c_str()));
-    return;
+    LM_E(("Database Error (error retrieving _id field in doc: '%s')", sub.toString().c_str()));
+    return -1;
```
fgalan: Maybe a dictionary of codes that the fucntion may return should be added to the function comment preamble.

kzangeli: Sure, that is always good

rbl219: LGTM

kzangeli: Fixed in 8ac8f45 :-)

fgalan:
LGTM

Issue #1358, opened 2015-10-14
APIv1 empty values support.


fgalan:
Currently, trying to create/update an empty value (i.e. value set to "") or not include attribute at all leads to this error: `"action: APPEND - entity: [Room2, Room, false] - offending attribute: temperature - empty attribute not allowed in APPEND or UPDATE"`

Thus, in some way, we are breaking backward compatibility as something that results in an error in a previous version, now it is not  an error anymore. However, it shouldn't be a problem (it would be a problem in the opossite case).

This issue is somehow related with https://github.com/telefonicaid/fiware-orion/issues/1077 (already closed) for APIv2.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1373

iariasleon:
LGTM
```
  <version>0.25.0-next</version>
  <git_hash>4b8e7621ec3e742a89bb7dcee0a06a0609163173</git_hash>
```

PR #1373, opened 2015-10-16
ADD empty attribute support in NGSIv1

fgalan:
Implementing #1358

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -385,23 +385,26 @@ static bool equalMetadataVectors(BSONObj& mdV1, BSONObj& mdV2)
 *
 * attributeValueAbsent -
 *
-* Check that the attribute doesn't have any value, taking into account its multi-valuated nature.
-*
-* NOTE
-*   1. ValueTypeNumber and ValueTypeBoolean are always non-empty.
-*   2. In API version 2, the value of a ContextAttribute CANNOT be left out
+* Check that the attribute doesn't have any value
 *
 */
-bool attributeValueAbsent(ContextAttribute* caP, const std::string& apiVersion)
+bool attributeValueAbsent(ContextAttribute* caP)
 {
-  if (apiVersion == "v1")
-  {
-    return (caP->valueType == ValueTypeString) && (caP->stringValue == "") && (caP->compoundValueP == NULL);
-  }
-  else // NGSIv2
-  {
-    return caP->valueType == ValueTypeNone;
-  }
+  return caP->valueType == ValueTypeNone;
+}
+
+/* ****************************************************************************
+*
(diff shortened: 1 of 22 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2262,7 +2264,6 @@ bool processOnChangeCondition
       AttributeList                 emptyList;
 
       // FIXME P10: we are using dummy scope by the moment, until subscription scopes get implemented
-      // FIXME P10: we are using an empty service path vector until serive paths get implemented for subscriptions
```
fgalan: I think this is an old leftover... as we are using servicePath since time ago.

test/functionalTest/runFailing.sh
```diff
@@ -0,0 +1,40 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: New helping script, to re-run only the failing tests.

rbl219: LGTM

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yes, seems like old garbage. 
Just remove it

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -385,23 +385,26 @@ static bool equalMetadataVectors(BSONObj& mdV1, BSONObj& mdV2)
...
-* Check that the attribute doesn't have any value, taking into account its multi-valuated nature.
+* Check that the attribute doesn't have any value
 *
-* NOTE
-*   1. ValueTypeNumber and ValueTypeBoolean are always non-empty.
-*   2. In API version 2, the value of a ContextAttribute CANNOT be left out
+*/
+bool attributeValueAbsent(ContextAttribute* caP)
+{
+  return caP->valueType == ValueTypeNone;
```
kzangeli: I would propose using parenthesis for returning comparisons as bool:

    return (caP->valueType == ValueTypeNone);

This is only my opinion. We need to decide what to do with this and put it is the style-guide.

src/lib/mongoBackend/MongoGlobal.cpp

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: If we don't have yet a decission on this, I'd suggest to progress the PR as it is, then align this globally in all the code base. If you agree, mark this as NTC :)

kzangeli: Yes, that's ok.

NTC

kzangeli:
LGTM - but, merge with develop is needed it seems

PR #1378, opened 2015-10-16
Task/trial1 subcache more functests

kzangeli:
- Functests about throttling and expiration of subs in subcache.
- Removed some traces and put back the INFO messages.
- Made broker fail at atartup on -reqMutexPolicy != "all" AND subCache ON


fgalan:
LGTM

PR #1379, opened 2015-10-16
FIX uncontrolled traces

fgalan:


CHANGES_NEXT_RELEASE
```diff
@@ -24,3 +24,4 @@ Fix:  Fix escaping dots in attrNames (Issue #1323)
 Fix:  POST /v2/entities returns error if the entity to create already exists (Issue #1158)
 Fix:  Add charset=utf-8 for notifications (Issue #1340)
 Add:  Functional test to PUT/POST/PATCH with 'empty' JSON object as payload (Issue #1305)
+Fix:  performance problem in NGSIv2 API due to uncontrolled log trace printing
```
fgalan: I have checekd (using blame) that one of the present was added in a commit around Jun 12th this year. Thus, the problem is present in previous versions and the CNR entry is justified.

NTC

kzangeli:
LGTM

PR #1394, opened 2015-10-20
FIX hardening mongoBackend checking for field existence and proper type to avoid broker crashes due to DB corruption

fgalan:
Partial implementation of #136 

Instead of using directly BSON class method, I have defined "wrapper" functions that do all the checking and prevent cases in the safeBsonGet.h module.

At the end, no code should use directly BSON methods (except unit test code). However, note that some leftover remains in cache/ (I have preferred not to touch there, as the cache is being re-done and the work will eventually land in develop branch).

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.h
+++ b/src/lib/mongoBackend/MongoGlobal.h
@@ -133,9 +133,15 @@ using namespace mongo;
 /*****************************************************************************
 *
 * Macros to ease extracting fields from BSON objects
+*
+* FIXME P10: probably the STR_FIELD macro (and maybe C_STR_FIELD) are not needed at
+* the end
 */
-#define STR_FIELD(i, sf) std::string(i.getStringField(sf))
-#define C_STR_FIELD(i, sf) i.getStringField(sf)
+//#define STR_FIELD(i, sf) std::string(i.getStringField(sf))
+//#define C_STR_FIELD(i, sf) i.getStringField(sf)
+#include "mongoBackend/safeBsonGet.h"
+#define STR_FIELD(i, sf) getStringField(i, sf)
+#define C_STR_FIELD(i, sf) getStringField(i, sf).c_str()
 
 
 
--- a/src/lib/mongoBackend/mongoQueryTypes.cpp
+++ b/src/lib/mongoBackend/mongoQueryTypes.cpp
@@ -72,7 +72,7 @@ static std::string attributeType
     /* It could happen that different entities within the same entity type may have attributes with the same name
      * but different types. In that case, one type (at random) is returned. A list could be returned but the
      * NGSIv2 operations only allow to set one type */
-    return r.getField(ENT_ATTRS).embeddedObject().getField(attrName).embeddedObject().getStringField(ENT_ATTRS_TYPE);
+    return getStringField(getField(getField(r, ENT_ATTRS).embeddedObject(), attrName).embeddedObject(), ENT_ATTRS_TYPE);
   }
 
   return "";
@@ -199,7 +199,7 @@ HttpStatusCode mongoEntityTypes
   // Processing result to build response
   LM_T(LmtMongo, ("aggregation result: %s", result.toString().c_str()));
 
-  std::vector<BSONElement> resultsArray = result.getField("result").Array();
+  std::vector<BSONElement> resultsArray = getField(result, "result").Array();
 
   if (resultsArray.size() == 0)
   {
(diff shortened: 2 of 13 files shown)


src/lib/mongoBackend/MongoGlobal.h
```diff
@@ -133,9 +133,15 @@ using namespace mongo;
 /*****************************************************************************
 *
 * Macros to ease extracting fields from BSON objects
+*
+* FIXME P10: probably the STR_FIELD macro (and maybe C_STR_FIELD) are not needed at
```
fgalan: Fixing this will be part of the next PR about #136

src/lib/mongoBackend/MongoGlobal.h
```diff
@@ -133,9 +133,15 @@ using namespace mongo;
...
+* the end
 */
-#define STR_FIELD(i, sf) std::string(i.getStringField(sf))
-#define C_STR_FIELD(i, sf) i.getStringField(sf)
+//#define STR_FIELD(i, sf) std::string(i.getStringField(sf))
+//#define C_STR_FIELD(i, sf) i.getStringField(sf)
+#include "mongoBackend/safeBsonGet.h"
+#define STR_FIELD(i, sf) getStringField(i, sf)
+#define C_STR_FIELD(i, sf) getStringField(i, sf).c_str()
```
kzangeli: Yeah, I agree.
A definition to just change the name of the 'function' doesn't help much. Just makes it more confusing.
I would definitely vote for *not* using STR_FIELD and C_STR_FIELD and instead use the functions themselves

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -72,7 +72,7 @@ static std::string attributeType
     /* It could happen that different entities within the same entity type may have attributes with the same name
      * but different types. In that case, one type (at random) is returned. A list could be returned but the
      * NGSIv2 operations only allow to set one type */
-    return r.getField(ENT_ATTRS).embeddedObject().getField(attrName).embeddedObject().getStringField(ENT_ATTRS_TYPE);
+    return getStringField(getField(getField(r, ENT_ATTRS).embeddedObject(), attrName).embeddedObject(), ENT_ATTRS_TYPE);
```
kzangeli: Pretty complex line. [ and yes, I know it is not new to this PR, it was just as bad before :-) ]
I would propose to split it up in 3 lines, using help variables (don't know what name to use for these).

Something like:

    XXX xObj = getField(r, ENT_ATTRS).embeddedObject());
    XXX yObj = getField(xObj, attrName).embeddedObject());

    return getStringField(yObj, ENT_ATTRS_TYPE);

src/lib/mongoBackend/safeBsonGet.cpp
```diff
@@ -0,0 +1,130 @@
...
+    LM_E(("Runtime Error (object field '%s' is missing in BSONObj <%s>)", field.c_str(), b.toString().c_str()));
+    return BSONObj();
+  }
+  if (b.getField(field).type() != Object)
+  {
+
+    LM_E(("Runtime Error (field '%s' was supposed to be an object but type=%d in BSONObj <%s>)",
+          field.c_str(), b.getField(field).type(), b.toString().c_str()));
+    return BSONObj();
+  }
```
kzangeli: Probably a stupid question ... are no exceptions thrown if you do **b.getObjectField(field);** directly and *field* doesn't exist?

The reason I ask this is the impact all these safety checks will have on the efficiency of the broker.

One thing that could be done to improve efficiency is (to not empty the CPU internal pipeline) to first ask for the "normal" case, like this (I understand this *is* the normal case):

    if (b.hasField(field) && (b.getField(field).type() == Object))
    {
       return b.getObjectField(field);
    }

    // Now detect the error
    if (!b.hasField(field)) ...

Same goes for all functions in here, of course

fgalan: Make sense... I first do some test with that approach and, if works, I will use it.

src/lib/mongoBackend/MongoGlobal.h
fgalan: NTC (just informative)

src/lib/mongoBackend/MongoGlobal.h
fgalan: To be fixed in the next PR.

NTC

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Fixed in 3755765

src/lib/mongoBackend/safeBsonGet.cpp
fgalan: Fixed in 3755765

kzangeli:
LGTM

PR #1395, opened 2015-10-21
FIX Remove comment in smallfiles and build.

dmoranj:
Remove comment in --smallfiles flag and change from building the image to downloading from the official repositories.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,18 +1,15 @@
 # Note that mongo container is started before Orion, as Orion needs it
 # as dependency and reserve the order has been found problematic in some
 # low resource hosts
-
-# Uncomment the --smallfiles line if you host doesn't have too much free space
 mongo:
   image: mongo:2.6
-  #command: --smallfiles
+  command: --smallfiles --nojournal
 
 orion:
-  build: .
+  image: fiware/orion
   links:
     - mongo
   ports:
     - "1026:1026"
   command: -dbhost mongo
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -1,18 +1,15 @@
 # Note that mongo container is started before Orion, as Orion needs it
 # as dependency and reserve the order has been found problematic in some
 # low resource hosts
-
-# Uncomment the --smallfiles line if you host doesn't have too much free space
 mongo:
```
fgalan: This should be aligned with https://github.com/telefonicaid/fiware-orion/blob/develop/docker/README.md#run-a-container-pulling-an-image-from-the-cloud-recommended. In particular, apply the same "uncomment" to smallfiles and add the `--nojournal`.

fgalan: In addition, maybe the following explanation about the mongo command would be useful in the .md setion:

"Regarding --smallfiles, you can ommit it if your hosting machine have enough disk space. Regarding --nojournal it is not recommened for production, but it speeds up mongo container start up and avoid some "race conditions" problems if Orion containers is faster and doesn't find the DB up and ready"

fgalan: Taking into account the new PR by @LeandroGuillen which address the .md, I think we can work .md changes in that PR. Thus, this becomes a NTC.

fgalan:
LGTM

PR #1396, opened 2015-10-21
Task/trial1 functests

kzangeli:
A few more bugs fixed and a leak.
Also, all tests (except coap_basic, that is deprecated) now work under valgrind


scripts/hdiff.sh
```diff
@@ -0,0 +1,8 @@
+base=$(basename $1 .test)
```
fgalan: License Header file missing.

scripts/hdiff.sh
```diff
@@ -0,0 +1,8 @@
...
+
+out=$(find . -name ${base}.out)
+exp=$(find . -name ${base}.regexpect)
+
+ORION_DIFF=${ORION_DIFF:-tkdiff}
+
+$ORION_DIFF $out $exp
```
fgalan: Other places use CB_DIFF_TOOL (e.g. viewDiff.sh). I would suggest to use the same name for the env variable.

fgalan: I mean, either CB_DIFF_TOOL or ORION_DIFF in both places.

src/lib/common/sem.h
```diff
@@ -65,6 +65,23 @@ extern int semInit
...
+* reqSemTryToTake - try to take take semaphore
+*
+* This function is only used by the exit-function, in order to clean the subscription cache.
+* The exit-function runs when the broker is shutting down and instead of just wating for the
+* semaphore, it is taken if it is free, if not, the subscription cache is simply cleaned.
+* A little dangerous, no doubt, but, the broke is shutting down anyway, if it dies because
+* of a SIGSEGV in the subscription cache code, it is not such a big deal ... perhaps ...
+*
+* FIXME P4
+* This implementation will stay like this until we find a better way attack the problem.
```
fgalan: My concern is that SIGSEGV errors appear in production (e.g. when the service is restarted) controled by a team that doesn't know this deep detail.

What about not cleaning the cache in production at exitFunction and use this only in DEBUG (for valgrind) using conditional compilation?

scripts/hdiff.sh
kzangeli: Yeah, I was thinking about that and the reason NOT to do it is that, I NEVER set CB_DIFF_TOOL in any startup script, if I want graphical diffs in my functest, I give it to testHarness.sh like this:

    CB_DIFF_TOOL=tkdiff testHarness.sh

hdiff.sh is different, it needs an env var to be set, and it would be set in the startup files, of course.

That's why I chose another env var

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -151,6 +151,5 @@ std::string parseEntity(ConnectionInfo* ciP, Entity* eP, bool eidInURL)
     }
   }
 
-  eP->present("");
```
fgalan: Given that you are fixing the pressent() issue, I would suggest to do it completely. In that sense, have a look to https://github.com/telefonicaid/fiware-orion/pull/1379 and remove also the other present() line and include the same CNR comment about it.

src/lib/common/sem.h
kzangeli: Sure, not a bad idea

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: To REALLY fix it, I would need to replace all LM_F for LM_Ts ...
Do we want this here?

test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
```diff
@@ -67,8 +67,11 @@ echo
 echo "+++++ 5: four service paths"
 orionCurl --url /v1/queryContext --payload "$payload" --servicePath '/home/kz/1, /home/kz/2, /home/kz/3, /home/kz/4'
 echo
+
+
+# The grep -v '/#' is to remove the "GET /version" that is used by the valgrind test suite
 echo "+++++ 6: service paths from log file"
-grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }'
+grep servicePathSplit /tmp/contextBroker.log | awk -F\| '{ print $6 }' | awk -F: '{ printf "%s:%s\n", $2, $3 }' | grep -v '/#'
```
fgalan: I think we have a discussion about the need of `| grep -v '/#'` in summer when valgrind suite was improved... but I don't remember the conclusion :$. Why this is needed at the end?

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
```diff
@@ -0,0 +1,581 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 101
+accumulatorStart
+
+--SHELL--
+
+#
+# 01. Create subscription S1 - without pattern
```
fgalan: I have looked the test "script" and seem to be ok (I haven't reviewed the details of the test itself).

NTC (just informative)

test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
kzangeli: Yes, we discussed it. For once, my memory is better than yours!!! :-D

The valgrind test suite uses a "GET /version"  to make sure the broker is alive.
This request is logged with servicePath /#.
I need to remove it in order for the functest to work under valgrind

scripts/hdiff.sh
kzangeli: Removing hdiff.sh and modifying viewDiff.sh.

The problem with CB_DIFF_TOOL is solved with this alias (just for me):

    alias hdiff='CB_DIFF_TOOL=tkdiff viewDiffs.sh'

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: The other call to 'present' (in jsonRequestTreat.cpp) removed as well.

LM_F for LM_Ts ... for an issue, right?

kzangeli: Fixed in f106612

src/lib/common/sem.h
kzangeli: Fixed in f106612

scripts/hdiff.sh
kzangeli: Fixed in f106612

scripts/hdiff.sh
kzangeli: Fixed in f106612  (hdiff.sh removed)

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Issue: #1398

test/functionalTest/cases/0392_service_path_update_and_query/service_path_http_header.test
fgalan: NTC

src/lib/common/sem.h
```diff
@@ -65,6 +65,23 @@ extern int semInit
 
 /* ****************************************************************************
 *
+* reqSemTryToTake - try to take take semaphore
```
fgalan: Now that this only runs in DEBUG, I think the FIXME could be removed. In fact, we can replace that FIXME with:

"This method only runs in DEBUG compilations, so SIGSGV are not so critical"

Or a better wording explaining this thing

src/lib/jsonParseV2/jsonRequestTreat.cpp
```diff
@@ -79,7 +79,6 @@ std::string jsonRequestTreat(ConnectionInfo* ciP, ParseData* parseDataP, Request
   case EntityAttributeRequest:
     releaseP->attribute = &parseDataP->attr.attribute;
     answer = parseContextAttribute(ciP, &parseDataP->attr.attribute);
-    parseDataP->attr.attribute.present("Parsed Attribute: ", 0);
```
fgalan: I'd suggest to add the folloing to CNR, related with this:

"Fix:  performance problem in NGSIv2 API due to uncontrolled log trace printing"

as I think it is important to note in the 0.24.0 branch CNR this fix.

Of course, this doesn't preclude a new CNR entry about  #1398 once we address it.

kzangeli: Fixed in 45083b2

src/lib/common/sem.h
kzangeli: Fixed in 45083b2

kzangeli: DBUG ... :-(
New commit coming. Look at this one for now ...

kzangeli: Fixed again in f652b10

fgalan:
LGTM (waiting until @kzangeli 's valgrind test result in order to merge)

PR #1399, opened 2015-10-22
New docker readme

LeandroGuillen:
The docker readme file has undergone substantial changes in structure. Hopefully is more clear how to use the different methods now.

Comments welcome, especially for section names :)

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,32 +1,31 @@
 
-# How to use this Dockerfile
+# How to use Orion Context Broker with Docker
 
-You can build a docker image based on this Dockerfile. This image will contain only an Orion Context Broker instance, exposing port `1026`. This requires that you have [docker](https://docs.docker.com/installation/) installed on your machine.
+You can run Orion Context Broker very easily using docker. There are several ways to accomplish this. These are (in order of complexity):
 
-If you just want to have an Orion Context Broker running as quickly as possible jump to section *The Fastest Way*.
+- _"Have everything automatically done for me"_. See Section **1. The Fastest Way** (recommended).
+- _"Only run a Context Broker, I'll take care of where I put my database"_. See Section **2. Run one container**.
+- _"Let me see how this docker thing works from the inside"_ or _"I want to customize my Orion Docker file"_ : See Section **3. Build a docker image**.
 
-If you want to know what is behind the scenes of our container you can go ahead and read the build and run sections.
+These are alternative ways to do the same thing, you do not need to do all three of them.
 
-## The Fastest Way
+You do need to have docker in your machine. See the [documentation](https://docs.docker.com/installation/) on how to do this.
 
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -34,58 +33,112 @@ If you do not have or want to download the Orion repository, you can create a fi
 	  ports:
 	    - "1026:1026"
 	  command: -dbhost mongo
+```
+> Uncomment the `--smallfiles` line if you host doesn't have too much free space.
```
fgalan: This should be aligned with the .yml file we provide after changes in PR https://github.com/telefonicaid/fiware-orion/pull/1395, which uses (without comment):

`command: --smallfiles --nojournal`

The rationale is that it is better to use the "safest" configuration to run in every system (e.g. the ones which small disk space of slow). It doesn't hurt to newbie users and advanced users would know that this is about and can remove that commands if they don't want to use it.

This text could be used to explain a little bit what they are (maybe with a better wording):

"Regarding --smallfiles, you can ommit it if your hosting machine have enough disk space. Regarding --nojournal it is not recommened for production, but it speeds up mongo container start up and avoid some "race conditions" problems if Orion containers is faster and doesn't find the DB up and ready".

docker/README.md
```diff
@@ -34,58 +33,112 @@ If you do not have or want to download the Orion repository, you can create a fi
...
+3. Using the command-line and within the directory you created type: `sudo docker-compose up`.
```
fgalan: Check rendering... it seems that the bullet number 3 is put alongside the "grey part".

docker/README.md
LeandroGuillen: Fixed in 4fc7c55

docker/README.md
LeandroGuillen: It's funny how vanilla markdown will render something fine while github will not.

Fixed in 4fc7c55

LeandroGuillen:
@dmoranj Take a look also!

fgalan:
LGTM (passing the ball to @dmoranj for final LGTM + merge + delete)

docker/README.md
```diff
@@ -1,91 +1,145 @@
...
-whose name is `mongodb`. In case you need one type `sudo docker run --name mongodb -d mongo:2.6`.
-* In `-p 1026:1026` the first value represents the port to listen on localhost. If you want to run a second context broker
+If you want to know more about images and the building process you can find it in [Docker's documentation](https://docs.docker.com/userguide/dockerimages/).
+
+## 4. Other info
+
+Things to keep in mind while working with docker containers and Orion Context Broker.
+
+### 4.1 Data persistence
+Everything you do with Orion Context Broker when dockerized is non-persistent. *You will lose all your data* if you turn off the MongoDB container. This happens with either this method or the manual one.
```
dmoranj: It's not clear what method you are referring to in the following sentence "This happens with either this method or the manual one".

LeandroGuillen: Good catch. It was actually a left over from the previous version.

Fixed in 830e21d

dmoranj:
LGTM

Issue #1401, opened 2015-10-22
Semaphore for subCache when mutex policy is none


kzangeli:
Right now, the subscription cache uses the mongo request semaphore, but this has an *important* limitation - the broker cannot be started with -mutexPolicy != all, unless also -noCache is present.

To fix this, the subscription cache needs a semahpore of its own, in case the mongo request semaphore is not in use.

And, this must be removed from contextBroker.cpp:

    if ((strcmp(reqMutexPolicy, "all") != 0) && (noCache == false))
    {
      LM_E(("Bad Input (reqMutexPolicy != 'all'  AND  subscription-cache CANNOT be used together in 0.24.0 (cache can be disabled using -noCache))"));
      exit(1);
    }

**Effort: 2 man day**

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1435

Issue #1359, opened 2015-10-14
GCC stack protector


kzangeli:
When adding the options 

    -fstack-check -fstack-protector

to the compiler, the broker doesn't compile.
These extra warnings might be very useful for the broker, we should use this (only as debugging tools)

**Effort: 3 man day**

fgalan:
Current CMakeLists.txt is ready for this. It is only a matter of uncomenting the line:

```
add_definitions(-fstack-check -fstack-protector)
```

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1418

PR #1402, opened 2015-10-22
Hardening/sub cache in 24 to develop

kzangeli:
Merge of the sub cache fix in 0.24.0.
That and a few fixes and tweaks ...


CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ execute_process(COMMAND            sh ../scripts/rapidxmlVersion.sh
 
 add_definitions(-DLM_ON)
 add_definitions(-fPIC)
+# add_definitions(-fstack-check -fstack-protector)
```
fgalan: Preparing for https://github.com/telefonicaid/fiware-orion/issues/1359 ? Good!

src/app/contextBroker/contextBroker.cpp
```diff
@@ -204,6 +205,14 @@ using namespace orion;
 
 /* ****************************************************************************
 *
+* Global vars
+*/
+static bool isFather = false;
```
fgalan: isFather is a bit generic... maybe isProcessFather (or something like that) would be more clear.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1472,6 +1489,17 @@ int main(int argC, char* argV[])
   paParse(paArgs, argC, (char**) argV, 1, false);
   lmTimeFormat(0, (char*) "%Y-%m-%dT%H:%M:%S");
 
+  //
+  // FIXME P8: for release 0.24, -reqMutexPolicy != "all" ("all" is default) cannot
```
fgalan: I think that we have an issue number for this. Maybe it could be references here. However, I understand that that issue is the next thing after merging this PR, so probably it doesn't worth the effort...

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1472,6 +1489,17 @@ int main(int argC, char* argV[])
...
+  //           be used together with mongo subscription cache.
+  //           This limitation will be fixed in the next release (0.25.0)
+  //
+  if ((strcmp(reqMutexPolicy, "all") != 0) && (noCache == false))
+  {
+    LM_E(("Bad Input (reqMutexPolicy != 'all'  AND  subscription-cache CANNOT be used together in 0.24.0 (cache can be disabled using -noCache))"));
```
fgalan: In fact, the error message is not correct, as this code (in develop) is no longer 0.24.0. However, it doesn't matter if this is going to be resolved before 0.25.0 releasing.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1485,8 +1513,6 @@ int main(int argC, char* argV[])
     LM_X(1, ("dbName too long (max %d characters)", DB_NAME_MAX_LEN));
   }
 
-  LM_I(("Orion Context Broker is running"));
```
fgalan: This removal is part of the cache merge from 0.24.0 sub-branch? Or a modification on the code after merging?

src/lib/cache/SubscriptionCache.cpp
```diff
@@ -219,6 +219,30 @@ Subscription* SubscriptionCache::lookupById
 
 /* ****************************************************************************
 *
+* SubscriptionCache::lookupById -
```
fgalan: The line `ADD_SUBDIRECTORY(src/lib/cache)` has been removed from CMakeFile.txt. However, it seems that not only we still having files under src/lib/cache but also this PR includes modification in these files. I'm missing something :)

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -79,17 +79,18 @@ std::string parseEntity(ConnectionInfo* ciP, Entity* eP, bool eidInURL)
     {
       LM_W(("Bad Input (No entity id specified"));
       eP->errorCode.fill("ParseError", "no entity id specified");
-      ciP->httpStatusCode = SccBadRequest;;
+      ciP->httpStatusCode = SccBadRequest;
+
       return eP->render(ciP, EntitiesRequest);
     }
   }
-  else if (document.Empty())
```
fgalan: What is the different between `.Empty()` and `.ObjectEmpty()` ?

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -79,17 +79,18 @@ std::string parseEntity(ConnectionInfo* ciP, Entity* eP, bool eidInURL)
...
     }
   }
-  else if (document.Empty()) 
+  else if (document.ObjectEmpty()) 
   {
     LM_W(("Bad Input (Empty payload)"));
     eP->errorCode.fill("ParseError", "empty payload");
-    ciP->httpStatusCode = SccBadRequest;;
-    return eP->render(ciP, EntitiesRequest);
+    ciP->httpStatusCode = SccBadRequest;
```
fgalan: I understand that from now on all the changes in this file are style things (probably related with traces insertion/removal). I see several empty lines... not sure if that is intentional.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1220,49 +1223,275 @@ static bool addTriggeredSubscriptions
...
   return true;
 }
 
 
+
+/* ****************************************************************************
+*
+* addTriggeredSubscriptions_noCache
+*
+* Recovered almost verbatim from release 0.23.0
```
fgalan: A fixme (and issue reference) should be created about the unification between _noCache and _cache functions. Copying verbatim code was a temporal solution for the 0.24.0 hotfix, but it is not something that we want to mantain in the long run.

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: I looked for that info. But without any luck. The documentation on rapidjson is not very good.
It might be Empty works on vectors and ObjectEmpty on objects?
Sorry, I don't know.

What I **do** know is that there are plenty of complaints about "very small json docs and crashes" on the internet and especially that the broker crashes using Empty, but not when using ObjectEmpty ...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -268,6 +268,17 @@ extern void setDbPrefix(std::string _dbPrefix)
 
 /*****************************************************************************
 *
+* dbPrefixGet -
+*/
+const char* dbPrefixGet(void)
```
fgalan: Takin into account that the "dual" metho is setDbPrefix() this should be getDbPrefix().

In addition, for homogenity, given that setDbPrefix() uses a std::string as argument, getDbPrefix() should return a std::string also.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -310,10 +321,12 @@ std::string tenantFromDb(std::string& database)
 {
   std::string r;
   std::string prefix  = dbPrefix + "-";
+
   if (strncmp(prefix.c_str(), database.c_str(), strlen(prefix.c_str())) == 0)
   {
     char tenant[MAX_SERVICE_NAME_LEN];
```
fgalan: Now that I see this MAX_SERVICE_NAME_LEN is remember that it was too long (1024 bytes). Probably we should make it shorter, to reduce "stress" in the stack space.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1163,25 +1169,6 @@ static void qStringFilters(std::string& in, std::vector<BSONObj> &filters)
       }
     }
 
-#if 0
```
fgalan: Not related with the cache stuff, but fine for me removing this at the end ;)

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -310,12 +310,12 @@ HttpStatusCode mongoQueryContext
                        limit,
                        &limitReached,
                        countP);
-    reqSemGive(__FUNCTION__, "ngsi10 query request", reqSemTaken);
```
fgalan: I didn't remember this part... we move reqSemGive in mongoQueryContext to cover a wide scope (in fact, the scope that actually correspond to the operation, not sure why was so "reduced" before this change). Nice.

NTC

src/lib/jsonParseV2/parseEntity.cpp
fgalan: Maybe we should add a comment here like "// We have experienced crash problems with Empty(), so ObjectEmpty() is used instead" to avoid that in the future we forget this and somebody changes back to Empty()....

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Sure, about the name change, even though much better would be to change setDbPrefix to dbPrefixSet - object-action, a more 'object-oriented' naming convention that I use since ... since ever.
We'll talk more about this when discussing the style-guide.

Now, NO function should ever return a std::string. This is something we do a lot, but we really need to stop sending **objects** over the stack. Once we have lint or cppcheck we will get a lot of complaints about this.
The user of dbPrefixGet don't need/want std::string and it would just be a complete waste of time (and especially run-time) to make it return a std::string.
If we want them to use the same type (I agree that homogenity is a good thing), I would propose to change setDbPrefix from std::string to char*.

Either that or we try to return a "const std::string&", which I don't fully understand. However, if we look for speed, char* is the way to go. [ not that this function will even slightly modify the total speed of the broker ]

CMakeLists.txt
```diff
@@ -170,7 +171,6 @@ if (error EQUAL 0)
   ADD_SUBDIRECTORY(src/lib/xmlParse)
   ADD_SUBDIRECTORY(src/lib/rest)
   ADD_SUBDIRECTORY(src/lib/mongoBackend)
-  ADD_SUBDIRECTORY(src/lib/cache)
```
kzangeli: Note that the library 'cache' is removed from compilation and linking, but the files are still in git.
Once the sub-cache is fully accepted, we'll remove the files under lib/cache as well.

src/app/contextBroker/contextBroker.cpp
kzangeli: ok.
Very clear to me. But I am not 'everybody' ...

src/app/contextBroker/contextBroker.cpp
kzangeli: Issue number added the the comment (#1401)

src/app/contextBroker/contextBroker.cpp
kzangeli: This might be a mistake. There should be ONE

    LM_I(("Orion Context Broker is running"));

    in the code (I copied this line and added a whole bunch for troubleshooting. At cleanup I must have removed them all, including the original one).  Good catch!

I will make sure there is only one.

src/app/contextBroker/contextBroker.cpp
kzangeli: New error text:

    LM_E(("This must be fixed for the 0.25.0 release!!!"));

Soon to be fixed

src/lib/cache/SubscriptionCache.cpp
kzangeli: Yeah, I commented on that above.
I haven't removed the source code of the old cache yet, so there will be a few nmodifications of this "dead code" in this PR.
Just ignore it, the library will soon be completely removed.

src/lib/mongoBackend/mongoSubCache.h
```diff
@@ -0,0 +1,226 @@
...
+/* ****************************************************************************
+*
+* CachedSubscription - 
+*/
+typedef struct CachedSubscription
+{
+  std::vector<EntityInfo*>    entityIdInfos;
+  std::vector<std::string>    attributes;
+  NotifyConditionVector       notifyConditionVector;
+  char*                       tenant;
```
fgalan: Could we use `std::string` instead of `char*`?

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Sure, good idea

CMakeLists.txt
fgalan: Do you mean to remove in the PR after this one?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Let's assume that our functions should return always `char*` (not an opinion yet about that... I have to think about it). Would be the following work?

```
std::string s = getDbPrefix()
```

or I need to write something like:

```
std::string s = std::string(getDbPrefix());
```

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  LM_T(LmtMongoSubCache, ("Refreshing subscription cache for DB '%s'", database.c_str()));
+
+  BSONObj                   query      = BSON("conditions.type" << "ONCHANGE" << CSUB_ENTITIES "." CSUB_ENTITY_ISPATTERN << "true");
+  DBClientBase*             connection = getMongoConnection();
+  auto_ptr<DBClientCursor>  cursor;
+
+  std::string tenant = tenantFromDb(database);
+  try
+  {
+    cursor = connection->query(getSubscribeContextCollectionName(tenant).c_str(), query);
```
fgalan: This is no longer the way of accesing DB... Please have a look to the methods in the connectionOperations.cpp/h

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  return false;
+}
+
+
+
+/* ****************************************************************************
+*
+* mongoSubCacheItemLookup - 
+*
+* FIXME P7: lookups would be A LOT faster if the subCache used a hash-table instead of
```
fgalan: Probably std::map could help. But not for this PR, of course.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1220,49 +1223,275 @@ static bool addTriggeredSubscriptions
...
+  LM_T(LmtMongo, ("query() in '%s' collection: '%s'",
+                  getSubscribeContextCollectionName(tenant).c_str(),
+                  query.toString().c_str()));
+
+  /* Do the query */
+  auto_ptr<DBClientCursor> cursor;
+  try
+  {
+    connection      = getMongoConnection();
+    cursor = connection->query(getSubscribeContextCollectionName(tenant).c_str(), query);
```
fgalan: This is no longer the way of accesing DB... Please have a look to the methods in the connectionOperations.cpp/h

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  reqSemGive(__FUNCTION__, "subscription cache", reqSemTaken);
+}
+
+
+
+/* ****************************************************************************
+*
+* mongoSubCacheRefresherThread - 
+*/
+static void* mongoSubCacheRefresherThread(void* vP)
```
fgalan: So now refresh is done in an idenpendent thread... Where this thread is created (to have a look)?

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -82,8 +81,11 @@ HttpStatusCode mongoSubscribeContext
     sub.append(CSUB_REFERENCE, requestP->reference.get());
 
     /* Throttling */
-    if (!requestP->throttling.isEmpty()) {
-      sub.append(CSUB_THROTTLING, (long long) requestP->throttling.parse());
+    long long throttling = 0;
```
fgalan: Not sure of understanding this change... could you elaborate?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Either char* or "const std::string&" (that should work right?), or even std::string* (a little trickier).
Never an object on the stack. We need to stop doing that. Extremely stupid to copy a whole object instead of just passing the pointer to it. 

Now,

    std::string s = getDbPrefix();

works just fine.
Just like

    std::string s = "27";

What does NOT work is this:

    std::string s = getDbPrefix()  + "abc";

You can't add a char* to a char*.
This only works if you have a std::string to the left.
Then, how much time do we lose on calling a method (operator+) on the std::string every time, instead of a strcat ... that would be nice to know.
Here, the C++ way is SO MUCH nicer to read and that is why we're using it. We lose time though. Probably a lot.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I think I did that ... Perhaps in one of the many branches during the 'cache-crisis' and it is not here ...
Yes, important to change that 1024 to ... 50 was it?
Let's include this change in this PR.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+    if (eIdP->isPattern == "false")
+    {
+      continue;
+    }
+
+    EntityInfo* eP = new EntityInfo(eIdP->id, eIdP->type);
+    
+    cSubP->entityIdInfos.push_back(eP);
+  }
+
```
fgalan: What about if the entityIdVector doesn't contain any pattern subscription? Maybe the entityIdInfo vector size should be checked (if 0, then exist without effective insertion).

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: This is not me, the merge with develop must have done this ...
I think.
Take a good look at it and decide whether we need the semaphore all the way to the end.
I am not so sure ...  It's not good to take the semaphore any more time than really needed ...

src/lib/ngsiNotify/senderThread.h
```diff
@@ -2,47 +2,63 @@
 #define SENDERTHREAD_H
```
fgalan: Is there any real change in this file?

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: I think is ok as it is. NTC.

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -183,10 +183,12 @@ HttpStatusCode mongoSubscribeContext
     {
       std::string oidString = oid.toString();
 
-      orion::subCache->insert(new orion::Subscription(tenant, servicePath, requestP, oidString, expiration, notifyFormat));
+      LM_T(LmtMongoSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
+      mongoSubCacheItemInsert(tenant.c_str(), servicePath.c_str(), requestP, oidString.c_str(), expiration, throttling, notifyFormat);
     }
 
 
+    reqSemGive(__FUNCTION__, "ngsi10 subscribe request", reqSemTaken);
```
fgalan: Another reqSemGive move... "This is not me, the merge with develop must have done this ..." again? Just to know if there is some intentional change here.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Not sure... maybe you could have a look to the documentation and calculate the maximum size taking into account all the rules we have (number of levels, max lengh in each level, etc.).

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: A hash table, that is the way to go here.
If you want to we can compare the speed of a std::map solution against a hash-table solution.
My guess? hash-table in C will be 10 times faster.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Ah, of course, I saw it during the the merge.
Yes, I will change that to the new way. Much better.
This code was implemented "before" ...

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: In my opinion this is clearly a case in which readibility is most important that performance. In addition, std::string has been in CB from the very beginning. Thus, I would suggest to continue with the same style.

src/lib/mongoBackend/mongoSubCache.h
kzangeli: Of course we could, but there is no reason.
The code would just be much slower.
When there is no need for adding strings, or comparing them, there is no reason whatsoever to use std::string,

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: We already use maps for URL params, thus I don't see a prioriy problems of using also maps for this. Again, I think readibility is more important than performance at this point (specially when we expect many people contributing to this code).

CHANGES_NEXT_RELEASE
```diff
@@ -27,3 +27,5 @@ Add:  Functional test to PUT/POST/PATCH with 'empty' JSON object as payload (Iss
 Fix:  performance problem in NGSIv2 API due to uncontrolled log trace printing
 Add:  empty attribute values supported in NGSIv1 API operations (Issue #1358)
 Add:  Add CLI argument for notification mode (Issue #1370)
+Fix:  Recovering $where from 0.23.0 code bases to search for subscriptions (-noCache flag)
```
fgalan: Probably we should mention this `-noCache` flag in documentation.

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: This one is probably me - look at the call to mongoSubCache.
Those where mongoSubCache is found, those moves of the semaphore are mine.
If no mongoSubCache is there ... then it's not mine

src/lib/ngsiNotify/senderThread.h
kzangeli: Only

    #define NOTIFICATION_WAIT_MODE false

as it seems.

Remember this is a merge. Not all modifications are mine ...

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Look at mongoSubCacheStart

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Yes, I need throttling later in this funcion, for the subCache.
That's why I "extract" the value as its own variable.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: This condition is taken care of before calling the function (this function is only called in one place - in mongoSubscribeContext;

    //
    // Add the subscription to the subscription cache.
    // But only if any of the entities in entityIdVector is pattern based -
    // AND it is a subscription of type ONCHANGE
    //
    bool patternBased = false;
    bool onchange     = false;

    // 1. Pattern-based?
    for (unsigned int ix = 0; ix < requestP->entityIdVector.size(); ++ix)
    {
      if (requestP->entityIdVector[ix]->isPattern == "true")
      {
       	patternBased = true;
	break;
      }
    }

    // 2. ONCHANGE?
    for (unsigned int ix = 0; ix < requestP->notifyConditionVector.size(); ++ix)
    {
      if (strcasecmp(requestP->notifyConditionVector[ix]->type.c_str(), ON_CHANGE_CONDITION) == 0)
      {
        onchange = true;
        break;
      }
    }

    // 3. Create Subscription for the cache
    if (patternBased && onchange)
    {
      std::string oidString = oid.toString();

      mongoSubCacheItemInsert(tenant.c_str(), servicePath.c_str(), requestP, oidString.c_str(), expiration, throttling, notifyFormat);
    }

Other insert-functions implement that check, whereas in this function is is not necessary (OK, this could be argued with "what if somebody goes and uses this function from some other place?" And it would be true, of course.  At the same time, having the same check in each layer doesn't make any sense it is just a waste of time. I'll add a mention of this as a prerequisite in the function-header.)

src/lib/mongoBackend/mongoSubCache.h
kzangeli: Let's postpone the 'std::string vs char*' discussion, or 'speed vs readability' for some other time.
We need this discussion, it's part of the style-guide, it just doesn't have to be right during this PR.

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -183,10 +183,12 @@ HttpStatusCode mongoSubscribeContext
     {
       std::string oidString = oid.toString();
 
-      orion::subCache->insert(new orion::Subscription(tenant, servicePath, requestP, oidString, expiration, notifyFormat));
+      LM_T(LmtMongoSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
+      mongoSubCacheItemInsert(tenant.c_str(), servicePath.c_str(), requestP, oidString.c_str(), expiration, throttling, notifyFormat);
```
kzangeli: Here *throttling* is used, in the call to mongoSubCacheItemInsert.

src/lib/ngsiNotify/senderThread.h
kzangeli: Ah wait, I have touched this file also - I added Format to the SenderThreadParams struct !

CMakeLists.txt
kzangeli: I would leave the old code until we're sure the new code is fully working

src/app/contextBroker/contextBroker.cpp
kzangeli: Changed to isFatherProcess

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Just style changes, yes.
For readability, it is good to have an empty line after the variable declarations and before the return statement.
This is my opinion, of course. But also, most of the source code of the broker is written this way, so it could be called to fix the style as well.

I mention this in the style guide.
Empty lines DO help readability, I don't think anyone can argue that.
They just need to be in the right place.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: FIXME inserted. I can't find the issue though. Did you perhaps mean to create a new issue about it?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, will definitely be modified

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: This is a tenant, not a service-path ... I think we looked this up not so long ago and it was 50, right?

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: This is another discussion I would propose to delay to some other time.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: MAX_SERVICE_NAME_LEN is defined in MongoGlobal.h.
I just changed it from 1024 to 50.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I changed the return type from _const char*_ to _const std::string&_.  The compiler accepts it, so I have no reason to believe it will give any problems.

My only concern about this right now, having the following code:

    const std::string& stringFix()
    {
        return  "A long string";
    }

    std::string sss = stringFix();

is that I don't know how the compiler translates this ...
Supposedly *stringFix* returns a pointer to a std::string.
My guess: std::string has an operator '=' that accepts a pointer to a std::string.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Now, what should we do if the tenant name is too long?
Right now it is just silently shortened to 50 chars. Probably not the best option ...

CMakeLists.txt
fgalan: A hint about this has been added to #1359 

NTC

CMakeLists.txt
fgalan: Ok, let's hold for a while. Anyway, even if it is removed it can be checked afterwards (git is a "time machine" :))

NTC

src/lib/cache/SubscriptionCache.cpp
fgalan: NTC

src/lib/jsonParseV2/parseEntity.cpp
fgalan: NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: You are right, the issue doesn't exist. Could you create it? Or prefer I do so?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Issue #1403 created

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1220,49 +1223,275 @@ static bool addTriggeredSubscriptions
...
+      long long throttling       = sub.hasField(CSUB_THROTTLING) ? sub.getField(CSUB_THROTTLING).numberLong() : -1;
+      long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION) ? sub.getIntField(CSUB_LASTNOTIFICATION) : -1;
+
+      TriggeredSubscription* trigs = new TriggeredSubscription
+        (
+          throttling,
+          lastNotification,
+          sub.hasField(CSUB_FORMAT) ? stringToFormat(STR_FIELD(sub, CSUB_FORMAT)) : XML,
+          STR_FIELD(sub, CSUB_REFERENCE),
+          //subToAttributeList(sub)); signature changed in TriggeredSubscription() constructor in 0.24.0
```
fgalan: Leftover comment?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Maybe we can use max_expectec_limit x 2 (i.e. 100) ?

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: I agree in posposing it. However, in order to have all the options on the table, I'd suggest to mention the std::map alternative in the FIXME comment text.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, but not mine.
I saw it and I thought it was yours, so I left it :-)
Removing it.

src/lib/mongoBackend/mongoSubCache.h
fgalan: Yes, let's pospone.

In fact, there is a very powerful argument to keep as it is now: it works :). I understand that changing `char*` to `std::string` will be a big change in the current cache implementation which is rather "delicate".

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: mmm. You are right. Now I remember.

Don't remove it... it could help when the _noCache and _cache copies of this function will be merged.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
+/*
```
fgalan: We have the "open question" about keep the cache management classes in cache/ module or in mongoBackend/ module.

Maybe it is better to have them in mongoBackend, as they are very "tied" to the DB technologies (i.e. they use connectionOperationsfunctions and, soon, they will use safeBson functions).

Just a thought...

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+    return;
+  }
+
+  // Call the treat function for each subscription
+  int subNo = 0;
+  while (cursor->more())
+  {
+    BSONObj sub = cursor->next();
+
+    mongoSubCacheItemInsert(tenant.c_str(), sub);
```
fgalan: As far as I understand by the answer of another comment, the usage of pattern entities in subscription is checked before mongoSubCacheItemInsert() is invoked. However, I don't find the checking in the mongoSubCacheRefresh() method. Maybe I'm missing something?

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: Your or not, what we should check in this PR is if this (and others) semGive movements make sense. What do you think in this particular case? For me, it could make sense.

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -208,34 +207,43 @@ HttpStatusCode mongoUpdateContextSubscription
   }
   responseP->subscribeResponse.subscriptionId = requestP->subscriptionId;
 
-  reqSemGive(__FUNCTION__, "ngsi10 update subscription request", reqSemTaken);
```
fgalan: Another reqSemGive movement here. Makes sense to me. What do you think?

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -208,34 +207,43 @@ HttpStatusCode mongoUpdateContextSubscription
...
   // 
   // All we need to do now for the cache is to:
-  //   1. Remove 'sub' from sub-cache (if present)
-  //   2. Create 'newSub' in sub-cache (if applicable)
+  //   1. Remove 'cSubP' from sub-cache (if present)
+  //   2. Create 'newSubObject' in sub-cache (if applicable)
+  //
+  // The subscription is already updated in mongo.
+  //
+  // FIXME P7: mongoSubCache stuff should be avoided if subscription is not patterned
```
fgalan: This means that currently isPattern is not checked in updateContextSubscription logic?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  return items;
+}
+
+
+
+/* ****************************************************************************
+*
+* mongoSubCacheItemInsert - 
+*/
+static void mongoSubCacheItemInsert(CachedSubscription* cSubP)
```
fgalan: Maybe the implementation of this method should be closer to the other instances of mongoSubCacheItemInsert() methods.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Of course.
However, it would be more visible if we created an issue about it.
We are very good looking at issues, but the FIXMEs ... not so good ...

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: I agree in not having the same check in two layers (although doing so add extra protection in the code... humans are error prone and tend to forget things ;). However, not sure if the "caller level" is better than the "called level" in this case. I think is better in the "called level" due to 1) implement the check in only one place, 2) it is less error-prone (if you forget to include the check in the caller, then you introduce a bug).

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, I had the same thought. That's why I put "v2" of the cache in mongoBackend.
At least, it has its own file ...

fgalan: After looking deeper to the code I have to "ammend" my previous thought :)

The functions that interact with MongoDB (i.e. the ones that are going to use connectionOperations and safeBson functions) should remain here. But there is logic currently in this module that has nothing to do with DB interaction (e.g. the function to create the refresh thread).

I don't suggest to move anything in this PR but probably mark with a FIXME the candidate functions to be moved is a good idea.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  std::string               formatString  = sub.hasField(CSUB_FORMAT)? sub.getField(CSUB_FORMAT).String() : "XML";
+  std::vector<BSONElement>  eVec          = sub.getField(CSUB_ENTITIES).Array();
+  std::vector<BSONElement>  attrVec       = sub.getField(CSUB_ATTRS).Array();
+  std::vector<BSONElement>  condVec       = sub.getField(CSUB_CONDITIONS).Array();
+
+
+  cSubP->tenant = (tenant[0] == 0)? NULL : strdup(tenant);
+
+  //
+  // FIXME: Check all strings about empty strings
```
fgalan: Should this FIXME resolved in the context of this PR?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
```
fgalan: Should this FIXME resolved in the context of this PR?

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Look up a few lines, line 1077:

    query      = BSON("conditions.type" << "ONCHANGE" << CSUB_ENTITIES "." CSUB_ENTITY_ISPATTERN << "true");

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  //        What happens if you call strdup on an empty string?
+  //
+  cSubP->subscriptionId        = strdup(subscriptionId);
+  cSubP->servicePath           = strdup(servicePath);
+  cSubP->reference             = strdup(sub.hasField(CSUB_REFERENCE)?    sub.getField(CSUB_REFERENCE).String().c_str() : "NO REF");  // Mandatory
+  cSubP->notifyFormat          = stringToFormat(formatString);
+  cSubP->throttling            = sub.hasField(CSUB_THROTTLING)? sub.getField(CSUB_THROTTLING).Long() : -1;
+  cSubP->expirationTime        = sub.hasField(CSUB_EXPIRATION)? sub.getField(CSUB_EXPIRATION).Long() : 0;
+
+  // FIXME P6: Doubt: should lastNotificationTime be reset when a subscription is updated?
```
fgalan: Not sure if I understand this correctly, but if this code is for updating cache with DB contents (or contents that are also to be synced in DB) then I think lastNotifcationTime should be updated also.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  *refreshes = mongoSubCache.noOfRefreshes;
+  *inserts   = mongoSubCache.noOfInserts;
+  *removes   = mongoSubCache.noOfRemoves;
+  *updates   = mongoSubCache.noOfUpdates;
+  *items     = mongoSubCacheItems();
+
+  CachedSubscription* cSubP = mongoSubCache.head;
+
+  //
+  // FIXME?
```
fgalan: Should this FIXME resolved in the context of this PR?

src/lib/mongoBackend/mongoUnsubscribeContext.cpp
```diff
@@ -111,13 +111,23 @@ HttpStatusCode mongoUnsubscribeContext(UnsubscribeContextRequest* requestP, Unsu
     /* Destroy any previous ONTIMEINTERVAL thread */
     getNotifier()->destroyOntimeIntervalThreads(requestP->subscriptionId.get());
 
-    responseP->statusCode.fill(SccOk);
-    reqSemGive(__FUNCTION__, "ngsi10 unsubscribe request", reqSemTaken);
+
+    // FIXME P7: mongoSubCache stuff should be avoided if subscription is not patterned
```
fgalan: Does this mean that the isPattern check is currently missing in the unsubscirbeContext logic?

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Yes, this one definitely makes sense. The sem must protect the subCache also. This is very clear. I moved it on purpose.
That other case, where no subCache calls are ... that one I am not so sure about. The "narrower" the sem is the less other threads have to wait.

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Yes, this one is also to 'cover subCache' as well

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Yeah, the check is done later, unfortunately. 
In:

    mongoSubCacheItemInsert(tenant.c_str(), newSubObject, subscriptionId, servicePath);

It would be faster to do the check before, and not even call mongoSubCacheItemInsert if the subscription is not  patterned.

But also, doing the check in mongoSubCacheItemInsert, saves us from bad future uses of an 'unprotected' mongoSubCacheItemInsert.

This is the exact opposite of another insert function we discussed yesterday.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, of course. Moving it down

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Issue #1405 created

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -112,18 +113,26 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
...
+                        tenant,
+                        spathList,
+                        xauthToken,
+                        uriPath,
+                        content_type,
+                        payload,
+                        true,
+                        NOTIFICATION_WAIT_MODE);
+
+    if ((r != "") && (r != "error"))
```
fgalan: `(r != "") && (r != "error")` means that the notification was send and the other end responsed ok. Is that correct?

src/lib/ngsiNotify/senderThread.h
fgalan: Could you elaborate on NOTIFICATION_WAIT_MODE, please? I think we introduced that in some point in the past time ago, but I don't see the "connection" with the cache stuff...

src/lib/mongoBackend/mongoUnsubscribeContext.cpp
kzangeli: It's the same thing as before, the call to mongoSubCacheItemLookup could be avoided, if we know that the sub is not patterned.

src/lib/rest/rest.cpp
```diff
@@ -1259,16 +1258,6 @@ void restInit
     strncpy(bindIp, _bindAddress, MAX_LEN_IP - 1);
   }
 
-  if ((_ipVersion == IPV4) || (_ipVersion == IPDUAL))
```
fgalan: This is related with the valgrind warn about "coyping into yourself", right?

src/lib/ngsiNotify/Notifier.cpp
kzangeli: It means that everything worked, yes.
I made an issue about this.
httpRequestSend should NOT return a string. Some status-code to check OK/NOT OK would be much better

src/lib/ngsiNotify/senderThread.h
kzangeli: The NOTIFICATION_WAIT_MODE stuff is by Carlos, I think.  Comes from the merge.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -309,11 +321,11 @@ PaArgument paArgs[] =
...
   { "-writeConcern",  &writeConcern, "WRITE_CONCERN",  PaInt,    PaOpt, 1,          0,      1,     WRITE_CONCERN_DESC },
 
-  { "-corsOrigin",       allowedOrigin,     "ALLOWED_ORIGIN",    PaString, PaOpt, _i "",          PaNL, PaNL,     ALLOWED_ORIGIN_DESC   },
-  { "-cprForwardLimit",  &cprForwardLimit,  "CPR_FORWARD_LIMIT", PaUInt,   PaOpt, 1000,           0,    UINT_MAX, CPR_FORWARD_LIMIT_DESC},
-  { "-subCacheIval",     &subCacheInterval, "SUBCACHE_IVAL",     PaInt,    PaOpt, 10,             0,    3600,     SUB_CACHE_IVAL_DESC   },
-  { "-notificationMode", &notificationMode, "NOTIF_MODE",        PaString, PaOpt, _i "transient", PaNL, PaNL,     NOTIFICATION_MODE_DESC},
-
+  { "-corsOrigin",       allowedOrigin,     "ALLOWED_ORIGIN",    PaString, PaOpt, _i "",          PaNL,  PaNL,     ALLOWED_ORIGIN_DESC    },
+  { "-cprForwardLimit",  &cprForwardLimit,  "CPR_FORWARD_LIMIT", PaUInt,   PaOpt, 1000,           0,     UINT_MAX, CPR_FORWARD_LIMIT_DESC },
+  { "-subCacheIval",     &subCacheInterval, "SUBCACHE_IVAL",     PaInt,    PaOpt, 10,             1,     3600,     SUB_CACHE_IVAL_DESC    },
```
fgalan: Refresh is only needed in some other (typically, other CB in an Active-Active configuration) can modify the DB. Thus, mono-CB deployment doesn't need refresh. Should we "recover" `-subCacheIval 0` as a way of using the cache *but* not refreshing it?

src/lib/rest/rest.cpp
kzangeli: Yes, I just removed those line as they didn't make any sense to me.

src/app/contextBroker/contextBroker.cpp
kzangeli: True, normally the refresh is NOT used, but the cache is always good to have.
I think it is a VERY VERY good idea.

Perhaps we should even have 0 (no refresh) as default value for -subCacheIval ... ?

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Ok to issue creating, moving the current comment text to the issue (+ std::map alternative) and replacing current comment text to a reference to the issue.

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Appart from issue (which is ok), what do you think about adding *in this PR* hints about which functions to migrate? Have you a clear list of which are the candidate ones?

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Right. However, probably not so obvious (note that it has passed unnotices to me... and I supposed to know how the query is done in MongoDB). Probably a comment about "MongoDB query ensures that we only process isPattern subscriptions" just above mongoSubCacheItemInsert() callc ould help.

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: Ok. Thus NTC (I guess)

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
fgalan: NTC

src/lib/ngsiNotify/Notifier.cpp

src/lib/ngsiNotify/senderThread.h
fgalan: Looking to the diff again, it seems that NOTIFICATION_WAIT_MODE has not been touched by this merge... What has been added has been the comment preamble to the define. So, asking again "Is there any real change in this file?" is seems that there isn't any change (appart from comments, of course).

src/lib/rest/rest.cpp
fgalan: NTC

src/app/contextBroker/contextBroker.cpp
fgalan: +1 to `-subCacheIval 0` and using is as default. In fact, that way we will recover backward compability with 0.24.0, because I think that it was the way it worked in that version.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: I haven't taken a look at it, but it is extremely easy. Only those functions that actually *do* something with mongo should stay in mongoBackend.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I've put it back again :-)

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Proposition on exactly *what* to move away from mongoBackend added to issue #1405

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Sure, it would help. No doubt

test/functionalTest/cases/1308_subscription_cache/cache_reset_from_db.test
```diff
@@ -0,0 +1,667 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: I understand that this new test in this merge was already reviewed in a previos PR in the 0.24.0 sub-branches. Is that correct? Just to save some reviewing time... ;)

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
```diff
@@ -0,0 +1,578 @@
+# Copyright 2015 Telefonica InvesTigacion y Desarrollo, S.A.U
```
fgalan: I understand that this new test in this merge was already reviewed in a previos PR in the 0.24.0 sub-branches. Is that correct? Just to save some reviewing time... ;)

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpiration.test
```diff
@@ -0,0 +1,292 @@
+# Copyright 2015 Telefonica InvesTigacion y Desarrollo, S.A.U
```
fgalan: I understand that this new test in this merge was already reviewed in a previos PR in the 0.24.0 sub-branches. Is that correct? Just to save some reviewing time... ;)

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
```diff
@@ -0,0 +1,328 @@
+# Copyright 2015 Telefonica InvesTigacion y Desarrollo, S.A.U
```
fgalan: I understand that this new test in this merge was already reviewed in a previos PR in the 0.24.0 sub-branches. Is that correct? Just to save some reviewing time... ;)

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I think at the end we have decided to change MAX_SERVICE_NAME_LEN to 50, use MAX_SERVICE_NAME_LEN + 1 to size the array and find a better fix for this in the context of #1403.

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: NTC

src/lib/ngsiNotify/senderThread.h
fgalan: Which is the relation between adding Format to  SenderThreadParams struct and the cache work? Just to understand...

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Sorry for this stupid remove-add cycle... Thanks!

NTC

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -0,0 +1,1204 @@
...
+    if (entity.hasField(CSUB_ENTITY_TYPE))
+    {
+      type = entity.getStringField(CSUB_ENTITY_TYPE);
+    }
+
+    EntityInfo* eiP = new EntityInfo(id, type);
+    cSubP->entityIdInfos.push_back(eiP);
+  }
+
+  if (cSubP->entityIdInfos.size() == 0)
```
fgalan: This should be moved upwards, as the first check in this function. In addition, this whould be an error (i.e. not use LM_E).

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -208,34 +207,43 @@ HttpStatusCode mongoUpdateContextSubscription
...
   // 0. Lookup matching subscription in subscription-cache
-  std::string servicePath;
+  CachedSubscription* cSubP = mongoSubCacheItemLookup(tenant.c_str(), requestP->subscriptionId.get().c_str());
 
-  servicePath = (servicePathV.size() == 0)? "" : servicePathV[0];
-  orion::Subscription* subP = orion::subCache->lookupById(tenant, servicePath, requestP->subscriptionId.get());
-
-  // 1. Remove 'sub' from sub-cache (if present)
-  if (subP)
+  if (cSubP == NULL)
```
fgalan: As discussed as skype, this need a fix to cover cases in which the original subscription hasn't pattern (so cSupP is NULL) but the new subscriptions has patterns.

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: We have discussed at skype about this... Not sure if at the end the code need to be changed or the comment to be remove (as the code is correct), but a modification in the file is needed in either case :)

src/lib/mongoBackend/mongoUnsubscribeContext.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, there where the notification is sent due to the sub-update, the lastNotifcationTime will be set.
Still need to be 100% sure that notifications **are** send un subUpdates

CHANGES_NEXT_RELEASE
fgalan: Taking care in PR https://github.com/telefonicaid/fiware-orion/pull/1410

NTC (in this PR).

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in a1a4c55

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in a1a4c55

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in a1a4c55

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in a1a4c55

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Issue #1408

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in a1a4c55

src/lib/ngsiNotify/senderThread.h
kzangeli: Format is part of the subscription, needed for the cache

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, notifications are sent in subUpdates

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Not really a FIXME, just a mark so that you would see it.
Please read and give me your opinion.
I've changed the "FIXME?" for a NOTE.

This information is **extremely** useful for debugging - as there are no other way to see the contents of the subscription cache right now.

src/app/contextBroker/contextBroker.cpp
kzangeli: Coming in next commit

test/functionalTest/cases/1308_subscription_cache/cache_reset_from_db.test
kzangeli: All tests are "old"

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
kzangeli: All tests are "old"

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpiration.test
kzangeli: All tests are "old"

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
kzangeli: All tests are "old"

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in a1a4c55

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Fixed in a1a4c55

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in e7feea2

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in e7feea2

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in c474073

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: It seems that the changes wasn't implemetned at the end... at least gibthub still showing the code using connection->query...

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Thus the FIXME text has been removed? Have been needed any modification in the code? A "Fixme in" message is missing to conclude this comment thread...

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Fixed in e7feea2c5d887fa1c59ee638922be6eec3e7bf10 (I guess).

test/functionalTest/cases/1308_subscription_cache/cache_reset_from_db.test
fgalan: NTC

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
fgalan: NTC

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpiration.test
fgalan: NTC

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Here you are!!! :-D

The doubt was clarified at seeing that subscriptions, when updated, the last notification time is updated if the change provokes a notification. If not, the old notification time is kept.

All good, FIXME/DOUBT removed in e7feea2

src/lib/ngsiNotify/senderThread.h
kzangeli: Actually, in this particular function **format** was needed for statistics, the call to *statisticsUpdate*

fgalan: NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Sorry about that, I got lost in the jungle of PR comments ...
Finally fixed in 6481808

fgalan:
LGTM

PR #1412, opened 2015-10-26
FIX empty attribute values in PUT/POST /v1/contextEntities/<id>/attributes/<attr> operation

### LINKED ISSUE #1392: POST/PUT convenience operation not working with attribute empty values
Reported by @anabelengp

This is a bug that seems to be related with the functionality recently implemented in 0.25.0 milestone to support empty attribute values.

The following request should work (instead of returning an error):

```
curl 'localhost:1026/v1/contextEntities/type/thing/id/thing:dev_ana1/attributes/ping' -X PUT -H 'Fiware-Service: smartvalencia' -H 'Fiware-ServicePath: /testAna' -H 'Content-Type: application/json;charset=UTF-8' -H 'Accept: application/json;charset=UTF-8' -d'{"value":"","type":"command"}'

{
  "statusCode" : {
    "code" : "400",
    "reasonPhrase" : "Bad Request",
    "details" : "empty context value"
  }
}

curl 'localhost:1026/v1/contextEntities/type/thing/id/thing:dev_ana1/attributes/ping' -X POST -H 'Fiware-Service: smartvalencia' -H 'Fiware-ServicePath: /testAna' -H 'Content-Type: application/json;charset=UTF-8' -H 'Accept: application/json;charset=UTF-8' -d'{"value":"","type":"command"}'

{
  "statusCode" : {
    "code" : "400",
    "reasonPhrase" : "Bad Request",
    "details" : "empty context value"
  }
}
```

fgalan:
Fixes #1392 

CNR entry is not included (as this is a fix in a issue adressed in 0.25.0, not present in previous versions).

### DIFF
--- a/src/lib/convenience/UpdateContextAttributeRequest.h
+++ b/src/lib/convenience/UpdateContextAttributeRequest.h
@@ -43,10 +43,10 @@ struct ConnectionInfo;
 typedef struct UpdateContextAttributeRequest
 {
   std::string                type;                // Optional
-  std::string                contextValue;        // Mandatory
-  orion::ValueType           valueType;           // Type of value: either string or none
+  std::string                contextValue;        // Mandatory  
   MetadataVector             metadataVector;      // Optional
 
+  orion::ValueType           valueType;           // Type of value: either string or none
   std::string                typeFromXmlAttribute;
   orion::CompoundValueNode*  compoundValueP;
 
--- a/src/lib/convenience/UpdateContextAttributeRequest.h
+++ b/src/lib/convenience/UpdateContextAttributeRequest.h
@@ -44,6 +44,7 @@ typedef struct UpdateContextAttributeRequest
 {
   std::string                type;                // Optional
   std::string                contextValue;        // Mandatory
+  orion::ValueType           valueType;           // Type of value: either string or none
   MetadataVector             metadataVector;      // Optional
 
   std::string                typeFromXmlAttribute;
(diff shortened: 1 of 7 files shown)


src/lib/convenience/UpdateContextAttributeRequest.h
```diff
@@ -44,6 +44,7 @@ typedef struct UpdateContextAttributeRequest
 {
   std::string                type;                // Optional
   std::string                contextValue;        // Mandatory
+  orion::ValueType           valueType;           // Type of value: either string or none
```
kzangeli: Please put this new field down a few lines.
The three first fields in this struct is "the payload", after that comes the help variables, like typeFromXmlAttribute et al.

src/lib/ngsi10/UpdateContextRequest.cpp
```diff
@@ -288,6 +288,10 @@ void UpdateContextRequest::fill
   else
   {
     caP = new ContextAttribute(attributeName, ucarP->type, ucarP->contextValue);
+    if (ucarP->valueType == orion::ValueTypeNone)
+    {
+      caP->valueType = orion::ValueTypeNone;
+    }
   }
```
kzangeli: Why set caP->valueType *only* when it is ValueTypeNone?
Why not **always** ?

fgalan: From a logical point of view, it makes sense... as this fragment is covering what the construct is not able to do, so it should be an unconditional copy.

I'd modify and re-run test. If works, I would use your suggested approach.

src/lib/convenience/UpdateContextAttributeRequest.h
fgalan: Fixed in 93655eb

src/lib/ngsi10/UpdateContextRequest.cpp
fgalan: Fixed in 93655eb

kzangeli:
LGTM

PR #1416, opened 2015-10-26
Hardening/0136 safe bson operations 2

fgalan:
Finished #136 implementation

Appart from that, unit test broken in a previous PR today has been fixed.

### DIFF
--- a/src/lib/serviceRoutines/postQueryContext.cpp
+++ b/src/lib/serviceRoutines/postQueryContext.cpp
@@ -155,7 +155,7 @@ static void queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, Format 
   if ((out == "error") || (out == ""))
   {
     qcrsP->errorCode.fill(SccContextElementNotFound, "");
-    LM_E(("Runtime Error (error forwarding 'Query' to providing application)"));
+    LM_W(("Runtime Error (error forwarding 'Query' to providing application)"));
     return;
   }
 
--- a/test/unittests/testInit.h
+++ b/test/unittests/testInit.h
@@ -29,6 +29,8 @@
 #include "ngsi9/NotifyContextAvailabilityRequest.h"
 #include "ngsi10/NotifyContextRequest.h"
 
+#include "mongoBackend/safeBsonGet.h"
+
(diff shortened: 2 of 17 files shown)


test/unittests/testInit.h
```diff
@@ -43,6 +45,8 @@
 #define RES_CER_STATUS(i)       res.contextElementResponseVector.get(i)->statusCode
 #define RES_CER_ATTR(i, j)      res.contextElementResponseVector.get(i)->contextElement.contextAttributeVector.get(j)
 
+#define C_STR_FIELD(b, f)       getStringField(b, f).c_str();
```
fgalan: Although removed from the non-test code base, the macro still have sense in the unit test code.

NTC

test/unittests/testInit.h
```diff
@@ -43,6 +45,8 @@
 #define RES_CER_STATUS(i)       res.contextElementResponseVector.get(i)->statusCode
 #define RES_CER_ATTR(i, j)      res.contextElementResponseVector.get(i)->contextElement.contextAttributeVector.get(j)
 
+#define C_STR_FIELD(b, f)       getStringField(b, f).c_str()
```
fgalan: Although removed from the non-test code base, the macro still have sense in the unit test code.

NTC

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -155,7 +155,7 @@ static void queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, Format
   if ((out == "error") || (out == ""))
   {
     qcrsP->errorCode.fill(SccContextElementNotFound, "");
-    LM_E(("Runtime Error (error forwarding 'Query' to providing application)"));
+    LM_W(("Runtime Error (error forwarding 'Query' to providing application)"));
```
fgalan: This is not actually part of the topic of the PR, but I have seem while fixing other things.

This condition shouldn't be loged as ERROR, as it is not a problem in the CB itself, but in a external system. Thus, I think that the righ log level is WARNING.

kzangeli:
LGTM

PR #1418, opened 2015-10-27
Compiling debug with stack protection flags (-fstack-check -fstack-protector)

### LINKED ISSUE #1359: GCC stack protector

fgalan:
Fixes #1359 

Some steps has been done enabling -fstack-check -fstack-protector in debug mode but qStringFilters() still needs to be refactored (currently, this PR branch doesn't compile in debug mode).


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -948,7 +949,7 @@ static void qStringFilters(const std::string& in, std::vector<BSONObj> &filters)
     }
     else
     {
-      printf("\nNo OP found\n");
+      //printf("\nNo OP found\n");
```
fgalan: Remove leftover

CMakeLists.txt
```diff
@@ -80,6 +79,7 @@ if (${CMAKE_BUILD_TYPE} STREQUAL DEBUG)
 
     MESSAGE("cmake: DEBUG compilation")
     add_definitions(-DDEBUG -DDEBUG_$ENV{USER} -DDEBUG_${USER})
+    add_definitions(-fstack-check -fstack-protector)
```
fgalan: Suggestion for the comment:

"Note that stack protection is activated only in debug compilation. This is due that this directive makes the compiler introduce code at the begin/end of the functions, so has impact in performance (see <links where that is explained in more detail>)"

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 9165059

CMakeLists.txt
kzangeli: Fixed in d8d8681

fgalan:
LGTM

PR #1420, opened 2015-10-27
FIX accumulator-server.py to work with IPv6 in Debian 8.2 (maybe it also helps others, e.g. Ubuntu)

fgalan:
This PR is supposed to fix the ipv4_ipv6.test problem in some distributions. In order to test, you need to re-install the accumulator-server.py, typically running:

```
make install_scripts
```
(Use the proper INSTALL_DIR if you need so)

### DIFF
--- a/scripts/accumulator-server.py
+++ b/scripts/accumulator-server.py
@@ -152,7 +152,14 @@ def record():
         #times.append(t)
 
     # Store verb and URL
-    s += request.method + ' ' + request.url + '\n'
+    #
+    # We have found that request.url can be problematic in some distributions (e.g. Debian 8.2)
+    # when used with IPv6, so we use request.scheme, request.host and request.path to compose
+    # the URL "manually"
+    #
+    #  request.url = request.scheme + '://' + request.host + request.path
+    #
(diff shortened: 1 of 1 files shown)


kzangeli:
LGTM, but perhaps this should be tested in each environment ...

fgalan:
Jenkins will test in in CentOS. I agree we need a second success case before merging, probably @LeandroGuillen or @raconte could help.

LeandroGuillen:
It does not work on Linux Mint (which is basically Ubuntu). It hangs forever on the functional test *0000_ipv6_support/ipv4_ipv6_both.test*. I'm investigating to see if the cause may be elsewhere.

raconte:
I don´t know if this is still been active . Due to the fact that I have reinstalled the virtual machine with Ubuntu I have proved all and  it works perfect . In my case the problem was that the .test was not able to find the route of the directory where the accumulator script was stored .

fgalan:
It's still active :)

@LeandroGuillen has you done any progress?

LeandroGuillen:
The problem seems to have been solved from my part.

LGTM

PR #1421, opened 2015-10-27
FIX xmlExtractor script to replace REGEX() in providingApplication lines

fgalan:
This PR is supposed to fix the problems we are having at jenkins with the payload check job.


crbrox:
LGTM

PR #1425, opened 2015-10-27
Feature/1384 cli memory connection

crbrox:
(Something weird in the rebase ... sorry)


src/app/contextBroker/contextBroker.cpp
```diff
@@ -243,6 +243,7 @@ unsigned        cprForwardLimit;
 int             subCacheInterval;
```
fgalan: CNR entry missing.

In addition, cli.md could include the following entry regarding the new `-connectionMemory`:

"""
-connectionMemory, set the size of the connection memory buffer (in Kb) [is "connection memory buffer" correct? otherwise use the right term] used internally by the HTTP server library. Default value is 64 Kb.
"""

crbrox: Fixed be329cd7ccfd38319c3fcd67bbc835fa14e4b242
Fixed 6f7d2eb878b86458a048aa707684cdf30500ada8

fgalan:
LGTM

PR #1431, opened 2015-10-28
Fix #1340 GET subscription list count not working

### LINKED ISSUE #1340: Adding charset UTF-8 to notifications
It seems a matter of using:
 
```  
Content-Type: application/json; charset=utf-8
```

instead of only

```
```

In notifications

### LINKED ISSUE #1430: GET subscription list (NGSIv2) count not working
Reported by @anabelengp 

Get subscriptions list operation in NGSIv2 is not honouring `count` option in order to get the total number of subscriptions in `X-Total-Count` header. The bug has been identified and it shouldn't be difficult to solve.

crbrox:
Fixes #1430


fgalan:
LGTM

Issue #1377, opened 2015-10-16
CLI '-reqMutexPolicy' silently accepts any value


kzangeli:
Usage says:

    [option '-reqMutexPolicy' <mutex policy (none/read/write/all)>]

Now, if -reqMutexPolicy is fed "nada", then "all" (which also is the default value) is silently picked.

Perhaps the broker should die with an error message instead?

**Effort: 1 man day**

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1435

Issue #1308, opened 2015-10-01
Possible bug detected about patterned subscriptions [cache not working]


kzangeli:
The following subscription seems to have stopped working in 0.24.0. It worked in 0.23.0, prior to the Subscription Cache implementation:

    {
      "entities": [
        {
          "type": "device",
          "isPattern": "true",
          "id": ".*”
        }
      ],
      "reference": "{{endpoint}}",
      "duration": "P1D",
      "notifyConditions": [
        {
          "type": "ONCHANGE",
          "condValues": [ "h" ]
        }
      ]
    }

The expected notifications no longer arrive.
If the Subscription is modified to use an exact **entity::id** (e.g. "id": "E1", and isPattern == false) instead of **.***, then everything works as before.
Reported by @antoniojoseln, Sep 9, 2015

fgalan:
> If the Subscription is modified to use the exact entity::id instead of .*, then everything works as before.

You mean that it works with <id="entityId", isPattern="true"> or with <id="entityId", isPattern="false">? A subtle but important detail that may help to debug the issue...

kzangeli:
Yes, it was not 100% clear, perhaps.
I edited the initial description to include "isPattern": "false".

kzangeli:
However, the simple functest I created to try to reproduce the bug **did not** reproduce the bug.
It worked just the way it should ...
We need to compare my test case to the real test case, the one that has the problem of not being notified.

kzangeli:
Found the problem, I believe.
tenantFromDb() in MongoGlobal.cpp **always** returns "" (empty string).

[ This function is only used by the subscription cache ]

fgalan:
After working on this issue in deep, it seems that the problem is more complex. After fixing tenantFromDB(), the whole list of tenant is processes. This "open" a new bug, found in some environment (Integration) with around 40 tenant, some of them with more that 800 subscriptions.

In this situation, it seems that the cache logic (maybe related with concurrent write access to the cache table) corrupt memory and eventually crash the broker.

dmabtrg:
I'm testing with the latest docker image from dockerhub (https://hub.docker.com/r/fiware/orion/builds/brtokibglv829ensgaxbsfg/) and I see no notifications sent from orion.  This is the payload we're using to generate the subscription:
```
{
    "entities": [
        {
            "type": "thing",
            "isPattern": "true",
            "id": "SENSOR_TEMP_*"
        }
    ],
    "attributes": [
[...]

PR #1435, opened 2015-10-28
Feature/1401 own semaphore for sub cache

kzangeli:
Implements #1401 and https://github.com/telefonicaid/fiware-orion/issues/1377

The mongo sub cache is now protected by its own semaphore and now "req mutex policy" con coexist with the subscription cache


CHANGES_NEXT_RELEASE
```diff
@@ -33,4 +33,5 @@ Add:  Hardening: mongoBackend checking for field existence and proper type to av
 Add:  githash in --version (Issue #1363)
 Add:  Added CLI argument for notification mode. Default mode is 'transient' (previous version used 'permanent' implicitly) (Issue #1370)
 Add:  Added CLI argument for connection memory limit.  Default value is 64Kb (Issue #1384)
+Fix:  Separate semahpore for subscription cache, so that sub cache is independent of mutex policy (Issue #1401)
```
fgalan: Probably this is an "aspect" of the #1308 fix, but it doesn't hurt having it as an idepdennt entry, if you prefer so. Up to you (just mark this as NTC if you think is ok as it is now).

kzangeli: As it is a "separate issue", I thought it convenient to mark it here.
What we perhaps should do is to make sure the issue 1401 is clearly marked as "related" to 1308

fgalan: What about "merging" both line, something like:

```
Fix: Fixed broken subscription cache (Issue #1308), including the semaphore for subscription cache (#1401)
```

kzangeli: Sure, no probs

kzangeli: Fixed in 72bff47

fgalan:
LGTM

Issue #1375, opened 2015-10-16
Possible problem in functests due to database-name/tenant 


kzangeli:
dbInit and dbDrop accept a database-name as input, when they should instead have a tenant-name as input, prepending the 'basename' of the DB.
The basename should be set in testEnv.sh and used by dbInit/dbDrop

[ CB_DATABASE_NAME **is** the basename, but a different variable should be used ]

The potential problem is that scripts/testEnv.sh might be modified, changing the base-name (currently 'ftest'), here:

    export CB_DATABASE_NAME=${CB_DATABASE_NAME:-ftest}
    export CM_DATABASE_NAME=${CM_DATABASE_NAME:-ftest_cm}
    export CP1_DATABASE_NAME=${CP1_DATABASE_NAME:-ftest_cp1}
    export CP2_DATABASE_NAME=${CP2_DATABASE_NAME:-ftest_cp2}
    export CP3_DATABASE_NAME=${CP3_DATABASE_NAME:-ftest_cp3}
    export CP4_DATABASE_NAME=${CP4_DATABASE_NAME:-ftest_cp4}
    export CP5_DATABASE_NAME=${CP5_DATABASE_NAME:-ftest_cp5}

This should be something like:

    export DATABASE_NAME_BASE=ftest
    export CB_DATABASE_NAME=${CB_DATABASE_NAME:-$DATABASE_NAME_BASE}
    etc. (not even sure that last line even is correct ...)

[ This problem all originates in the fact that the broker will not recognize **any** database. The databases **must** start with the prefix 'DATABASE_NAME_BASE', which is 'orion' for a normal run of the broker and 'ftest' when the broker is running in a functional test ]

The functests (one line per dangerous db) with this possible problem are:

    0713_propagate_tenant/propagate_tenant_for_update_and_query_standard_ops.test:dbInit ftest-t1
    0713_propagate_tenant/propagate_tenant_for_update_and_query_standard_ops.test:dbInit ftest-t2
    0713_propagate_tenant/propagate_tenant_for_update_and_query_standard_ops.test:dbInit ftest_cp1-t1
    0713_propagate_tenant/propagate_tenant_for_update_and_query_standard_ops.test:dbInit ftest_cp1-t2
    1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test:dbInit ftest-t1
    1308_subscription_cache/cache_reset_from_db.test:dbInit ftest-t1
[...]

PR #1436, opened 2015-10-28
Hardening/1375 problem functest database name/tenant

raconte:
Implements #1375

Change the names of the database
-Changed to 
export DATABASE_NAME_BASE=CB
export CB_DATABASE_NAME=${CB_DATABASE_NAME:-$DATABASE_NAME_BASE}
-And created a function : dbInitTenant and dbDropTenant


scripts/testEnv.sh
```diff
@@ -129,9 +129,24 @@ export ORION_DATABASE_PASSWORD=${ORION_DATABASE_PASSWORD:-orion}
...
 #
-export CB_DATABASE_NAME=${CB_DATABASE_NAME:-ftest}
-export CP1_DATABASE_NAME=${CP1_DATABASE_NAME:-ftest_cp1}
-export CP2_DATABASE_NAME=${CP2_DATABASE_NAME:-ftest_cp2}
-export CP3_DATABASE_NAME=${CP3_DATABASE_NAME:-ftest_cp3}
-export CP4_DATABASE_NAME=${CP4_DATABASE_NAME:-ftest_cp4}
-export CP5_DATABASE_NAME=${CP5_DATABASE_NAME:-ftest_cp5}
+
+
+export DATABASE_NAME_BASE=CB
```
fgalan: The name of the default CB database should be `ftest`, not `CB`.

scripts/testEnv.sh
```diff
@@ -129,9 +129,24 @@ export ORION_DATABASE_PASSWORD=${ORION_DATABASE_PASSWORD:-orion}
...
+export DATABASE_NAME_BASE_T2=t2
+export CB_DATABASE_NAME=${CB_DATABASE_NAME:-$DATABASE_NAME_BASE}
+export CP1_DATABASE_NAME=${CP1_DATABASE_NAME:-$DATABASE_NAME_BASE_CP1}
+export CP2_DATABASE_NAME=${CP2_DATABASE_NAME:-$DATABASE_NAME_BASE_CP2}
+export CP3_DATABASE_NAME=${CP3_DATABASE_NAME:-$DATABASE_NAME_BASE_CP3}
+export CP4_DATABASE_NAME=${CP4_DATABASE_NAME:-$DATABASE_NAME_BASE_CP4}
+export CP5_DATABASE_NAME=${CP5_DATABASE_NAME:-$DATABASE_NAME_BASE_CP5}
+export CBT1_DATABASE_NAME=${CBT1_DATABASE_NAME:-$DATABASE_NAME_BASE:-$DATABASE_NAME_BASE_T1}
+export CBT2_DATABASE_NAME=${CBT2_DATABASE_NAME:-$DATABASE_NAME_BASE:-$DATABASE_NAME_BASE_T2}
+export CP1T1_DATABASE_NAME=${CP1T1_DATABASE_NAME:-$DATABASE_NAME_BASE_CP1:-$DATABASE_NAME_BASE_T1}
```
kzangeli: I think this is exagerrating a little.   "t1" is simply the name of a tenant.
We can't really create new env vars for each tenant name, we probably have 20 already in the functests and new tenant names can come in every day ...

The tenant name is the second parameter to the 'new dbInit/Drop', right?
Let's keep it as a simple text, freely selectable by the author of the functest.

*My Opinion, of course - more opinions are more than welcome*

scripts/testEnv.sh
kzangeli: +1

test/functionalTest/harnessFunctions.sh
```diff
@@ -87,6 +87,50 @@ function dMsg()
 }
 
 
+# ------------------------------------------------------------------------------
+#
+# dbInitTenant - 
+#
+function dbInitTenant()
```
kzangeli: Could not the old **dbInit** be reused for this?
Accepting either **one** or **two** params and if the second is present, it is simply concatenated to the name of the DB ... ?

test/functionalTest/harnessFunctions.sh
```diff
@@ -87,6 +87,50 @@ function dMsg()
...
+  if [ "$port" == "" ]
+  then
+    port="27017"
+  fi
+  
+  db=$host:$port/$1 
+
+  dMsg initializing database;
+
+  if [ "$role" == "CB t1" ]
```
kzangeli: This looks funny ...
If the function is called with two params (CB t1, for example),  then  $1 will be "CB" and $2 will be "t1".
As role is set to $1 and after that not modified, the only way to get here is to call the function like this:

    dbInitTenant "CB t1"     OR
    dbInitTenant CB\ t1

Have a look at this, please :-)

test/functionalTest/harnessFunctions.sh
```diff
@@ -153,7 +198,22 @@ function dbDrop()
   fi
 }
 
+# ------------------------------------------------------------------------------
+#
+# dbDropTenant - 
+#
+function dbDropTenant()
+{
+  db=$1
```
kzangeli: Here we have a problem also, the second param is not taken into consideration.

    if [ $# == 2 ]
    then 
      db=${db}$2
    fi

Something like that would fix it.
And ... this can be done in the 'old' dbDrop, right?
We don't need a new function for an optional second variable, I would say

scripts/testEnv.sh
fgalan: Let's try to use a shorter name, using DB instead of DATABASE in env var names.

scripts/testEnv.sh
fgalan: +1

scripts/testEnv.sh
raconte: Fixed in 3ef4ea3 and bd93c95, the name of the variable has been changed from CB_DATABASE_NAME to CB_DB_NAME .If it is better to change to FTEST_DB_NAME tell me and I will change it.It has been changed for all the test that uses that variable.

scripts/testEnv.sh
raconte: Fixed in 3ef4ea3 and bd93c95

test/functionalTest/harnessFunctions.sh
raconte: Fixed in 3ef4ea3 and bd93c95, dbInitTenant and dbDropTenant has been removed

test/functionalTest/harnessFunctions.sh
raconte: Fixed in 3ef4ea3 and bd93c95

test/functionalTest/harnessFunctions.sh
raconte: Fixed in 3ef4ea3 and bd93c95, dbInitTenant and dbDrop has been removed and the code included in the function.

test/functionalTest/harnessFunctions.sh
```diff
@@ -1239,7 +1244,6 @@ export -f brokerStart
 export -f localBrokerStop
 export -f localBrokerStart
 export -f brokerStop
-export -f localBrokerStop
```
fgalan: Not sure if this is a good idea... @kzangeli what do you think?

test/functionalTest/harnessFunctions.sh
```diff
@@ -109,33 +109,39 @@ function dbInit()
   fi
   
   db=$host:$port/$1
+  
+  if [ $# == 2 ]
+  then 
+    db=$db$2
```
fgalan: Is this really work? I mean, if I use the function with: `dbInit CB t1`, then db=CB and $2=t1. Thus, the final DB  name will be `ftestt1` (i.e. missing the `-`).

Maybe it should be?

```
db=$db-$2
```

test/functionalTest/harnessFunctions.sh
```diff
@@ -109,33 +109,39 @@ function dbInit()
   fi
   
   db=$host:$port/$1
+  
+  if [ $# == 2 ]
```
fgalan: The same logic (once fixed) should be use din dbDrop

test/functionalTest/cases/0322_multitenancy/http_tenant_subscribe_and_update.test
```diff
@@ -318,5 +318,5 @@ brokerStop CB
 accumulatorStop $LISTENER_PORT
 accumulatorStop $LISTENER2_PORT
 dbDrop CB
-dbDrop ${CB_DATABASE_NAME}-t_01
-dbDrop ${CB_DATABASE_NAME}-t_02
+dbDrop ${CB_DB_NAME}-t_01
```
fgalan: Affer adjusting dbDrop, all this should be like:

```
dbDrop ${CB_DB_NAME} t_01
```

test/functionalTest/harnessFunctions.sh
fgalan: It makes sense... it was a duplicated line.

However, given that we are modifying this, probalby it is a good idea to take the opportunity to reorder them in this way:

brokerStart
brokerStrop
localBrokerStart
localBrokerStop

kzangeli: We call brokerStart and brokerStop from the functests, so those two must be exported (otherwise we cannot see them). Now, brokerStart calls localBrokerStart. It shouldn't be necessary to export localBrokerStart, IMHO.

I would do a test, exporting only brokerStart and brokerStop and if it works, then NOT export the other two.

test/functionalTest/harnessFunctions.sh
raconte: I think I solved this sorry for the delay .

Fixed in c170369

test/functionalTest/harnessFunctions.sh
raconte: I think I solved this sorry for the delay .

Fixed in c170369

fgalan:
LGTM (although I probably do some small modification to fix some little things at github edit interface before merging)

test/functionalTest/harnessFunctions.sh
```diff
@@ -1,4 +1,4 @@
-# Copyright 2013 Telefonica Investigacion y Desarrollo, S.A.U
+# # Copyright 2013 Telefonica Investigacion y Desarrollo, S.A.U
```
kzangeli: This might give an error in 'make check_files'

fgalan: Fixed in d164ca69457dade5cf1a01d5f9775482543557d5

PR #1437, opened 2015-10-28
Compile support for Linux Mint

LeandroGuillen:
I added support for compiling on Linux Mint.

It is similar to what's needed on the Ubuntu case so I combined them. Also, there were a few Ubuntu versions with the same packages linking and I refactored it a bit.


fgalan:
LGTM

fgalan:
LGTM

PR #1450, opened 2015-10-30
Hardening/1429 request sizes

### LINKED ISSUE #1429: Fix disabled "large request" test
Upon merging https://github.com/telefonicaid/fiware-orion/pull/1425/files, the "large payload" .test are failing (they are now as .DISABLED to avoid a fail at jenkins). This issue is about enabling them back and fix CB in order to make them work.

kzangeli:
Fixes #1429 

Better implementation of connectionTreat, part II.
Now large requests work again.


src/lib/rest/rest.cpp
```diff
@@ -1072,6 +1101,18 @@ static int connectionTreat
     LM_T(LmtUriParams, ("'default' value for notifyFormat (ciP->outFormat == %d)): '%s'", ciP->outFormat, ciP->uriParam[URI_PARAM_NOTIFY_FORMAT].c_str()));
   }
 
+  //
+  // Here, if the incoming request was too big, return error abouyt it
+  //
+  if (ciP->httpHeaders.contentLength > PAYLOAD_MAX_SIZE)
```
fgalan: What about if the client is not honoring the Content-Length header? For example, he uses `Content-Length: 50` but the paylod is 1.5 MB long. Maybe an extra check in the "copy chunk" logic to detect when the payload has grown too much (in fact, I'm not sure if with the current "copy chunk" logic the ciP->payload could be "overflow" in that situation...)?

(That case would be a rare situation, we don't need to cover this is this PR if you find it is hard. We can open an issue about it).

CHANGES_NEXT_RELEASE
```diff
@@ -33,6 +33,7 @@ Add:  githash in --version (Issue #1363)
 Add:  Added CLI argument for notification mode. Default mode is 'transient' (previous version used 'permanent' implicitly) (Issue #1370)
 Add:  Added CLI argument for connection memory limit.  Default value is 64Kb (Issue #1384)
 Fix:  Fixed broken subscription cache (Issue #1308), including the semaphore for subscription cache (#1401)
-Fix:  compound attribute values support in CPr update forwardking (Issue #1440)
+Fix:  compound attribute values support in CPr update forwarding (Issue #1440)
 Fix:  Return error response when the (mongo) index key limit is exceeded (Issue #1289)
 Fix:  Incorrect 'details' field rendering (leading to illegal JSON) for error responses in some cases (Issue #843) 
+Fix:
```
fgalan: Actually, this PR is part of the connection memory limit work (already included in the CNR), so I don't think we need new entries.

src/lib/rest/rest.cpp
fgalan: Or maybe the MHD detect the Content-Length vs. actual payload disalignment and send the error itself... We need to test what hapepend with a crafted HTTP request (maybe curl can do that).

CHANGES_NEXT_RELEASE
kzangeli: That came in there by mistake ...
I was looking for the issue number and finally decided nothing more was to be added.
I only committed CNR as I fixed a spelling error. This line ... by mistake

src/lib/rest/rest.cpp
kzangeli: Let's create an issue. No time for this today, and the rest of the fix is necessary.
Not 100% easy to test this, but it might be a good idea to protect ourselves against
bad usage.

Issue #1451

src/lib/rest/rest.cpp
```diff
@@ -974,46 +993,56 @@ static int connectionTreat
   //
   if (dataLen != 0)
   {
-    if (dataLen == ciP->httpHeaders.contentLength)
+    //
+    // If the HTTP header says the request is bigger than our PAYLOAD_MAX_SIZE,
+    // just silently "eat" the entire message
+    //
+    if (ciP->httpHeaders.contentLength > PAYLOAD_MAX_SIZE)
     {
```
crbrox: Question: Is it necessary to read the whole body?  Could we send an error earlier? closing connection maybe ... (only asking)

src/lib/rest/rest.cpp
```diff
@@ -974,46 +993,56 @@ static int connectionTreat
   //
   if (dataLen != 0)
   {
-    if (dataLen == ciP->httpHeaders.contentLength)
+    //
+    // If the HTTP header says the request is bigger than our PAYLOAD_MAX_SIZE,
+    // just silently "eat" the entire message
+    //
+    if (ciP->httpHeaders.contentLength > PAYLOAD_MAX_SIZE)
```
crbrox: Question: Is it necessary to read the whole body?  Could we send an error earlier? closing connection maybe ... (only asking)

kzangeli: And it is not a bad question.
Unfortunately MHD doesn't let you write on the socket until the entire message is read.
This is pretty strange, as the socket is full duplex, but tests have been made and an error is returned if you try to write on the socket before it 'is ready'.

We could always close the socket (which is not a bad option), but I don't think we'd be able to return any error response.

Perhaps the best option here is just to **close** the socket ... ?

crbrox: addde #1451
NTC

CHANGES_NEXT_RELEASE
fgalan: Fixed in 1fa2cdd730c99ad4dd918286d8b107e896ee8450

fgalan:
LGTM

crbrox:
LGTM

PR #1452, opened 2015-10-30
Bug/int or long in last notification

kzangeli:
Fixed a bug about reading mongo int as longs and vice versa.
Some log messages are in this PR as well, as this is what I was doing when I discovered the bug ...


src/lib/mongoBackend/safeBsonGet.cpp
```diff
@@ -105,8 +102,31 @@ int getIntField(const BSONObj& b, const std::string& field)
 
 /* ****************************************************************************
 *
-* getBoolField -
+* getLongField -
+*/
+long getLongField(const BSONObj& b, const std::string& field)
```
fgalan: Great! So probably now we can replace all the usage of `getField(XXX).Long()` and `getField(XXX).numberLong()` usages in the code for this new function.

Probably not in this PR, but in a separted one. What do you think?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1056,9 +1069,17 @@ void mongoSubCacheStatisticsGet(int* refreshes, int* inserts, int* removes, int*
   {
     while (cSubP != NULL)
     {
+      char msg[256];
+
       unsigned int bytesLeft = listSize - strlen(list);
 
-      if (strlen(cSubP->subscriptionId) + 2 > bytesLeft)
+#if UpdateBugTests
```
fgalan: Leftover?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1056,9 +1069,17 @@ void mongoSubCacheStatisticsGet(int* refreshes, int* inserts, int* removes, int*
...
+      snprintf(msg, sizeof(msg), "%s|N:%lu|E:%lu|T:%lu|P:%d", cSubP->subscriptionId, cSubP->lastNotificationTime, cSubP->expirationTime, cSubP->throttling, cSubP->pendingNotifications);
+#else
+      snprintf(msg, sizeof(msg), "%s", cSubP->subscriptionId);
+#endif
+
+      if (strlen(msg) + 2 > bytesLeft)
```
fgalan: This +2 probably deserves a comment explaining this code :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1213,7 +1214,19 @@ static bool addTriggeredSubscriptions_withCache
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
       long long throttling       = sub.hasField(CSUB_THROTTLING) ? getField(sub, CSUB_THROTTLING).numberLong() : -1;
-      long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION) ? getIntField(sub, CSUB_LASTNOTIFICATION) : -1;
+      long long lastNotification = -1;
+
+      if (sub.hasField(CSUB_LASTNOTIFICATION))
```
fgalan: Pontentially, we could have the same problem for NGSI9 subscription (look for CASUB_LASTNOTIFICATION field). I wonder it is is worth the effort also to implement a similar fix for them (probably simpler, as casubs doesn't use yet a cache). The point to take into account is that NGSI9 subscription are not very used nowadays.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -55,6 +55,7 @@ using std::string;
 using std::map;
```
fgalan: Maybe we should include this CNR entry?

"Fix: Hardening: safer input/ouput logic for csub documents in MongoDB and cache"

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yeah, not really a leftover, see I need this trace (even if it is temporal) to catch the "possible" bug about updating subcsriptions in the cache

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Sure, why not. It's to make sure ", " has room in the char-vector - but that you knew already

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 069b4dd

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: So the plan is to remove this after debugging, finding and fixing the bug? In that case, I'd suggest to add a FIXME comment about it just before the `#if`.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in 069b4dd

src/lib/mongoBackend/safeBsonGet.cpp
kzangeli: I guess we could.
It will get a little slower in run-time, as more checks are made.
I think I would prefer to *always* use **Long**, and never use **Int** instead ...

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: I removed it in d0c923d (and I have already put it back in my personal copy :-))

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Actually, any integer field could have this problem.

CASUB_LASTNOTIFICATION does not, it is read/written only in two places:

    newSub.append(CASUB_LASTNOTIFICATION, getIntField(sub, CASUB_LASTNOTIFICATION));
    BSONObj     update = BSON("$set" << BSON(CASUB_LASTNOTIFICATION << getCurrentTime()) << "$inc" << BSON(CASUB_COUNT << 1));

Now, as getCurrentTime() returns **int** we are good.

I would like to stress a little on the danger of using the << operator. It's not the first time I see this.
If getCurrentTime() changes to reutn a long, the **<< getCurrentTime()** would still work, no warning would be issued but a Long field would be created in mongo and later when reading that fiels (as Int), the broker crashes ...

src/lib/mongoBackend/safeBsonGet.cpp
fgalan: Being a littler slower (just internal function calls and ifs) is a very small price to pay for getting safer code. Remember the parachute metaphore ;)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: The alternative to BSON() macro (the one which use that `<<` style) is to use BSONObjBuilder and append() method, then use the obj() method to flush to a BSONObj. I guess that BSON() macro (which composes the BSONObj in a shoot) is more efficient, so my recommendation it not stopping using it (in the places it make sense, i.e. for dynamic composition you will still needing BSONObjBuilder) but using it with care.

NTC

src/lib/mongoBackend/safeBsonGet.cpp
kzangeli: Sure, but if we use Long always, the risk disappears, right?
The price to pay is to get the databases a little bigger.
To me, this is the best option, risk free, fast an a little bigger.

We should create an issue about this. The decision can be made later

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: What I'm looking for is to use a function that will actually warn about a change in type.
If append() methods of BSONObjBuilder (I imagine there is more than one) accept Int **or** Long, then we have the same problem again - we will not be warned.

So C++ ... elegant perhaps, but **so** dangerous ...

fgalan: Having a quick lock to http://api.mongodb.org/cplusplus/current/classmongo_1_1_b_s_o_n_obj_builder.html it seems that both "idioms" are possible for append().

src/lib/mongoBackend/safeBsonGet.cpp
fgalan: Pending bits deferreed to a new PR.

fgalan:
LGTM

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: If so, we win nothing ...

PR #1465, opened 2015-11-02
(Codebase 0.24.0) Hardening/new subcache trial1

fgalan:
This is the first of a two steps landing process to update release/0.24.0 branch with the most updated fix for csubs cache.

This lands hardening/new_subcache_trial1 into hardening/new_subcache. A next PR will land hardening/new_subcache into release/0.24.0.


kzangeli:
LGTM

PR #1466, opened 2015-11-02
FIX coverage for functional tests

fgalan:
Functional test were not generating coverage in report since DB pool implementation. This PR is supossed to fix that.

The PR is long, but most of the changes in *_test.cpp files are all the same. The interesting parts to review are in: MongoGlobal.cpp/h, mongoConnectionPool.cpp and testInit.cpp.

### DIFF
--- a/test/unittests/mongoBackend/mongoOntimeintervalOperations_test.cpp
+++ b/test/unittests/mongoBackend/mongoOntimeintervalOperations_test.cpp
@@ -280,6 +280,7 @@ TEST(mongoOntimeintervalOperations, mongoGetContextSubscriptionInfo_dbfail)
 
     /* Set MongoDB connection (prepare database first with the "actual" connection object) */
     prepareDatabase();
+    DBClientBase* connectionDb = getMongoConnection();
     setMongoConnectionForUnitTest(connectionMock);
 
     /* Do operation */
@@ -291,8 +292,10 @@ TEST(mongoOntimeintervalOperations, mongoGetContextSubscriptionInfo_dbfail)
               "- findOne(): { _id: ObjectId('51307b66f481db11bf860001') } "
               "- exception: boom!!)", err);
 
-    /* Release mock */
-    setMongoConnectionForUnitTest(NULL);
+    /* Restore real DB connection */
+    setMongoConnectionForUnitTest(connectionDb);
+
+    /* Release mocks */
     delete connectionMock;
 }
 
@@ -510,6 +513,7 @@ TEST(mongoOntimeintervalOperations, mongoGetContextElementResponses_dbfail)
 
     /* Set MongoDB connection (prepare database first with the "actual" connection object) */
     prepareDatabase();
+    DBClientBase* connectionDb = getMongoConnection();
     setMongoConnectionForUnitTest(connectionMock);
 
     /* Do operation */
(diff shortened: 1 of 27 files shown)


test/unittests/mongoBackend/mongoOntimeintervalOperations_test.cpp
```diff
@@ -522,8 +526,10 @@ TEST(mongoOntimeintervalOperations, mongoGetContextElementResponses_dbfail)
               "attrNames: { $in: [ \"A1\", \"A2\", \"A3\", \"A4\" ] } }, orderby: { creDate: 1 } } - "
               "exception: boom!!)", err);
 
-    /* Release mock */
-    setMongoConnectionForUnitTest(NULL);
+    /* Restore real DB connection */
+    setMongoConnectionForUnitTest(connectionDb);
```
kzangeli: All these *setMongoConnectionForUnitTest(connectionDb);* are because utInit/utExit aren't used in these functions, right?

fgalan: They are used to "recover" the actual connection to DB (previously overriden by `setMongoConnectionForUnitTest(connectionMock)`) so next next find it correctly (otherwise, the next test will bet `connectionMock` and that crashed most of the times).

I think is not related with utInit/utExit (these functions create and release mocks, but they don't do anything related connection objects, as far as I remember...).

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptions_test.cpp
```diff
@@ -3087,13 +2948,11 @@ TEST(mongoUpdateContext_withOnchangeSubscriptions, DISABLED_MongoDbQueryFail)
     BSONObj sub = connection->findOne(SUBSCRIBECONTEXT_COLL, BSON("_id" << OID("51307b66f481db11bf860001")));
     EXPECT_EQ(20000000, sub.getIntField("lastNotification"));
 
-    /* Release connection */
-    setMongoConnectionForUnitTest(NULL);
-
+    /* Restore real DB connection */
+    setMongoConnectionForUnitTest(connectionDb);
```
kzangeli: I was expecting to see a *setMongoConnectionForUnitTest(connectionDb);* in utExit ...
Not sure I understand why "Restore real DB connection" is done only sometimes and not in all test cases ...

fgalan: The "default" status for the internal connection is the actual connection to DB. Some test change this to a connectionMock. Thus, you only need to set again the connection to the actual connection to DB in these tests.

kzangeli: ok.

I would like to put all these things in utInit/Exit (and use the functions in all tests).
utInit would have parameters (bool  setMongoConnection) etc to 'change' its behavior, to fit all needs, but implemented only in *one* place.

NTC

test/unittests/mongoBackend/mongoOntimeintervalOperations_test.cpp
kzangeli: NTC

kzangeli:
LGTM

Issue #1455, opened 2015-10-30
Enforce DB types


fgalan:
First, identify for each field described in the DB model (described at http://fiware-orion.readthedocs.org/en/develop/admin/database_model/index.html) the corresponding (BSON) type. Modify the documentation to make that information explicit.

Next, create a DB migration script to "colapse" ambiguos case (typically, the ones that can be either an Int or a Long) into the decided one.

Finally, change the code to align it with the decided types.

**Efort: 2 man day**

Issue #1406, opened 2015-10-23
[BUG] missing notifications (duration/throttling error?) for an updated subscription


iariasleon:
if we create a subscription with a duration of 50s and we wait to the subscription expire, we update the same subscription with duration of 1000s, **only one notification** is sent. 

#### create subscription request
```
POST http://<host>:<port>/NGSI10/subscribeContext
Content-Type: application/json
Accept: application/json
Fiware-Service: happy_path
Fiware-ServicePath: /test
{  
     "entities": [ {                                         
            "type": "house",                                   
             "isPattern": "true",                        
             "id": "room_*"                               
      }],                                              
     "attributes": [ "temperature" ],                                
     "reference": "http://localhost:5678/notify",                                 
     "duration": "PT50S",                                 
     "notifyConditions": [ {                                           
             "type": "ONCHANGE",                          
             "condValues": ["temperature" ]                                          
       }]                                                      
       "throttling": "PT40S"
}
```

#### update subscription request
```
POST http://<host>:<port>/v1/updateContextSubscription
Content-Type: application/json
Accept: application/json
Fiware-Service: happy_path
Fiware-ServicePath: /test
{       
      "subscriptionId": "5629e7be256202e888cd1378", 
      "duration": "PT1000S"    
'}
```

fgalan:
(From Ken's report)

New info on this issue, testing in the QA machine qa-orion-fe-01:

The test fails **only** if we use sub-cache-refresh *and* subscription-update.

sub-cache-refresh is only interesting for **active-active** and subscription-update is not really interesting at all.

NOTE:
After sub-update, the **first** notification (sometimes 2 notifications) arrives, so, it seems it is the cache-refresh that "destroys" the subscription.
A test with sub-cache-refresh-interval of 1000 secs just verified this hypothesis.

iariasleon:
It is still failing just after 0.26.1, tested with:
```
 <version>0.26.1_20151210101150</version>
  <git_hash>1736125cdb6b4b8bffc1dd253d69fd60f0019ce2</git_hash>
```
The problem is when we try to update a subcription but it has expired previously.

fgalan:
We have checked that expireation values is updated at DB after subcription update. However, we don't know if that update happens also in the cache: failing in doing such update could be causing the problem.

We have to check if a trace could help to check this (either an existing one or one explicitely added for the test).

fgalan:
[...]

PR #1468, opened 2015-11-02
Bug/int or long in last notification

kzangeli:
lastNotificationTime as Long everywhere


test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
```diff
@@ -96,12 +96,12 @@ static void prepareDatabaseV1Subs(void) {
...
                                                        "value" << BSON_ARRAY("AX2" << "AY2")
                                                        )) <<
-                        "throttling" << 5.0
+                        "throttling" << (long long) 5
                         );
 
     BSONObj sub3 = BSON("_id" << OID(SUB_OID3) <<
-                        "expiration" << 20000000 <<
-                        "lastNotification" << 25000000 <<
+                        "expiration" << (long long) 20000000 <<
```
kzangeli: These changes here of expiration, count and throttling I don't like too much.
I tried to remove the **(long long)**  for these fields but everything went south, so finally I rollbacked that change.

We will take care of all this in issue #1455

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1457,7 +1467,7 @@ static bool addTriggeredSubscriptions_noCache
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
       long long throttling       = sub.hasField(CSUB_THROTTLING) ? getField(sub, CSUB_THROTTLING).numberLong() : -1;
-      long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION) ? getIntField(sub, CSUB_LASTNOTIFICATION) : -1;
+      long long lastNotification = getLongField(sub, CSUB_LASTNOTIFICATION);
```
fgalan: This should be:

```
long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION) ? geLongField(sub, CSUB_LASTNOTIFICATION) : -1;
 ``
From a caller point of view, you have to be sure that the field exists when invoking get*Field() fucntion in safeBson module. Don't rely in the function returning a -1, as it also generate a Runtime Error.

File: src/lib/mongoBackend/mongoGetSubscriptions.cpp [hunk 3]
```diff
@@ -123,10 +123,10 @@ static void setNotification(Subscription* s, const BSONObj& r)
   s->notification.throttling = r.hasField(CSUB_THROTTLING)? getField(r, CSUB_THROTTLING).numberLong() : -1;
 
   // Last Notification
-  s->notification.lastNotification = r.hasField(CSUB_LASTNOTIFICATION) ? getField(r, CSUB_LASTNOTIFICATION).numberLong() : -1;
+  s->notification.lastNotification = getLongField(r, CSUB_LASTNOTIFICATION);
```
fgalan: This should be:

```
s->notification.lastNotification = r.hasField(CSUB_LASTNOTIFICATION) ? getLongField(r, CSUB_LASTNOTIFICATION) : -1;
```

From a caller point of view, you have to be sure that the field exists when invoking get*Field() fucntion in safeBson module. Don't rely in the function returning a -1, as it also generate a Runtime Error.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1072,8 +1075,16 @@ void mongoSubCacheStatisticsGet(int* refreshes, int* inserts, int* removes, int*
       char          msg[256];
       unsigned int  bytesLeft = listSize - strlen(list);
 
-      snprintf(msg, sizeof(msg), "%s", cSubP->subscriptionId);
-
+#if 0
```
fgalan: To be removed after checking the branch in this PR is ok with QA?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1213,9 +1213,12 @@ static bool addTriggeredSubscriptions_withCache
     {
```
fgalan: Not sure...

Is this PR related with the fix to https://github.com/telefonicaid/fiware-orion/issues/1406?

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
```diff
@@ -85,17 +85,19 @@ HttpStatusCode mongoGetContextSubscriptionInfo
...
 
-    csiP->expiration = getField(sub, CSUB_EXPIRATION).numberLong();
+    csiP->expiration = sub.hasField(CSUB_EXPIRATION)? getField(sub, CSUB_EXPIRATION).numberLong() : -1;
 
     csiP->url = getStringField(sub, CSUB_REFERENCE);
-    if (sub.hasElement(CSUB_LASTNOTIFICATION)) {
-        csiP->lastNotification = getIntField(sub, CSUB_LASTNOTIFICATION);
+    if (sub.hasElement(CSUB_LASTNOTIFICATION))
+    {
+      csiP->lastNotification = getLongField(sub, CSUB_LASTNOTIFICATION);
```
fgalan: Not sure but... maybe this is other case in which "CSUB_LASTNOTIFICATION might be either Int or Long ..." and we need the "dual read" you have implemented in other places to be safer?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -839,9 +842,9 @@ int mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub, const char*
   std::vector<BSONElement>  attrVec       = sub.getField(CSUB_ATTRS).Array();
   std::vector<BSONElement>  condVec       = sub.getField(CSUB_CONDITIONS).Array();
 
-  if ((lastNotificationTime == -1) && (sub.hasField(CSUB_LASTNOTIFICATION)))
+  if (lastNotificationTime == -1)
   {
-    lastNotificationTime = sub.getField(CSUB_THROTTLING).Long();
+    lastNotificationTime = getLongField(sub, CSUB_LASTNOTIFICATION);
```
fgalan: Not sure but... maybe this is other case in which "CSUB_LASTNOTIFICATION might be either Int or Long ..." and we need the "dual read" you have implemented in other places to be safer?

(In fact, the "dual read" function seems is used in a lot of places, I'd suggest to move it to a separate and common function)

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -187,7 +188,7 @@ HttpStatusCode mongoUpdateContextSubscription
     /* The hasField checks are needed as lastNotification/count might not be present in the original doc */
     if (sub.hasField(CSUB_LASTNOTIFICATION))
     {
-      newSub.append(CSUB_LASTNOTIFICATION, getIntField(sub, CSUB_LASTNOTIFICATION));
+      newSub.append(CSUB_LASTNOTIFICATION, getLongField(sub, CSUB_LASTNOTIFICATION));
```
fgalan: Not sure but... maybe this is other case in which "CSUB_LASTNOTIFICATION might be either Int or Long ..." and we need the "dual read" you have implemented in other places to be safer?

(In fact, the "dual read" function seems is used in a lot of places, I'd suggest to move it to a separate and common function)

src/lib/mongoBackend/safeBsonGet.cpp
```diff
@@ -104,11 +104,18 @@ int getIntField(const BSONObj& b, const std::string& field)
 *
 * getLongField -
 */
-long getLongField(const BSONObj& b, const std::string& field)
+long long getLongField(const BSONObj& b, const std::string& field)
```
fgalan: Not sure if I like this...

This function shouldn't be so clever. If I'm trying to read a long field that actually is an int, I don't want this function "hides" that: I want to get a Runtime Error that point me where the problem is and fix it. That's the whole idea behind safeBsonGet functions.

It we need "intelligence" to read fields that at the same time may be int or long, then that intelligence has to be put in the caller, not in the get*Field() function.

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdate.test
```diff
@@ -0,0 +1,329 @@
...
+
+--SHELL--
+
+#
+# 00. Create subscription for s0.* entities, attribute A, w/o tenant, expiration time 10 secs
+# 01. Update the subscription
+# 02. Create entity s01 w/ attr A w/o tenant
+# 03. Modify entity s01 w/ attr A w/o tenant - notification lost by throttling
+# 04. Modify entity s01 w/ attr A w/o tenant - notification lost by throttling
+# 05. Wait 5 secs so that the throttling of 4 secs is OK
```
fgalan: However, the payload shows that thottling is P1M (1 month), very small compared with 4 seconds. On the contrary, subscription duration is PT1S (1 second). Maybe both fields are incorrectly swapped...

If that's the case, I wonder how this .test is working... the difference between both situation is big.

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdateWhenSubExpired.test
```diff
@@ -0,0 +1,330 @@
...
+echo "========================================================================================="
+payload='{
+  "entities": [
+    {
+      "isPattern": "true",
+      "id": "s0.*"
+    }
+  ],
+  "reference": "http://localhost:'${LISTENER_PORT}'/notify",
+  "duration": "PT1S",
```
fgalan: Same problem seems to have with duration/throttling here...

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
```diff
@@ -96,12 +96,12 @@ static void prepareDatabaseV1Subs(void) {
...
+                        "lastNotification" << (long long) 25000000 <<
```
fgalan: We applied the approach "read both long/int, write always long" to lastNotifcation. Thus, I don't understand why 25000000 (as int) doesn't work... I think it should be researched, this could be the tip of an iceberg of a more important bug.

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
fgalan: The problem is that we can have in existing databases expirations in the int form (i.e., without `(long long)`), so the original test is a "safety measure" to be sure that we will work for these. Maybe we should apply the "read both long/int, write always long" also to expiration?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: getLongField implements hasField, getField and -1

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: same same

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
kzangeli: getLongField implements that dual read

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, getLongField is that function

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yup, hopefully today :-)

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: getLongField implements this

src/lib/mongoBackend/safeBsonGet.cpp
kzangeli: This is the exact function you've been asking for.
I am not going to implement this part in the 10 times where it is called. That is a really bad idea.
Perhaps you just don't like the name of the function?

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdate.test
kzangeli: The comment is incorrect as you have probably seen.
The subscription is created with 'impossible' values and the sub-update makes throttling and duration OK.
It is done like this on purpose, so that without the sub-update nothing would work. The comment should be removed though.

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdateWhenSubExpired.test
kzangeli: No problem here either, in step '01. Update the subscription' the sub is put in order

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
kzangeli: Yeah, this 'long long' is probably not necessary

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
kzangeli: I don't know if that is the way to go.
Very unfortunate this whole int/long thing ...
All these checks make us lose time.
If we could avoid them, it would be great.

However, we have an issue for this, the #1455

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Not really, but the branch worked for those tests. If it still works after all the recent changes, then we may add the connection to 1406.

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
fgalan: > I don't know if that is the way to go.

I'm my opinion:

* Long term: issue #1455 
* Short term: keep the prepareDatabase() as they are now (i.e. without `(long long)`) as this is probably the case we found in real databases now and rely in our "read long/int, write long" temporal fix.

fgalan: >  and rely in our "read long/int, write long" temporal fix.

Actually, I'm not sure about this... this is "expiration" (not "lastNotiication") and probably the current logic may work.

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdateWhenSubExpired.test
kzangeli: Fixed in 1dc46df

test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdate.test
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/safeBsonGet.cpp
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: Fixed in 1dc46df

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 1dc46df

test/unittests/mongoBackend/mongoGetSubscriptions_test.cpp
kzangeli: Fixed in 4bba343

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
```diff
@@ -80,22 +80,17 @@ HttpStatusCode mongoGetContextSubscriptionInfo
...
+    for (unsigned int ix = 0; ix < attrs.size(); ++ix)
+    {
+      csiP->attributeList.push_back(attrs[ix].String());
     }
 
-    csiP->throttling = sub.hasField(CSUB_THROTTLING) ? getField(sub, CSUB_THROTTLING).numberLong() : -1;
+    csiP->url              = getStringField(sub, CSUB_REFERENCE);
+    csiP->expiration       = sub.hasField(CSUB_EXPIRATION)?       getIntOrLongFieldAsLong(sub, CSUB_EXPIRATION)       : -1;
+    csiP->lastNotification = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLong(sub, CSUB_LASTNOTIFICATION) : -1;
+    csiP->throttling       = sub.hasField(CSUB_THROTTLING)?       getField(sub, CSUB_THROTTLING).numberLong()         : -1;
```
fgalan: Maybe this should use getIntOrLongFieldAsLong() also?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -620,24 +620,12 @@ int mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub)
   cSubP->servicePath           = strdup(sub.hasField(CSUB_SERVICE_PATH)? sub.getField(CSUB_SERVICE_PATH).String().c_str() : "/");
   cSubP->reference             = strdup(sub.hasField(CSUB_REFERENCE)?    sub.getField(CSUB_REFERENCE).String().c_str() : "NO REF");  // Mandatory
   cSubP->notifyFormat          = stringToFormat(formatString);
-  cSubP->throttling            = sub.hasField(CSUB_THROTTLING)? sub.getField(CSUB_THROTTLING).Long() : -1;
-  cSubP->expirationTime        = sub.hasField(CSUB_EXPIRATION)? sub.getField(CSUB_EXPIRATION).Long() : 0;
-  cSubP->lastNotificationTime  = 0;
+  cSubP->throttling            = sub.hasField(CSUB_THROTTLING)?       getLongField(sub, CSUB_THROTTLING) : -1;
```
fgalan: Maybe this should use getIntOrLongFieldAsLong() also?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -839,9 +827,9 @@ int mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub, const char*
   std::vector<BSONElement>  attrVec       = sub.getField(CSUB_ATTRS).Array();
   std::vector<BSONElement>  condVec       = sub.getField(CSUB_CONDITIONS).Array();
 
-  if ((lastNotificationTime == -1) && (sub.hasField(CSUB_LASTNOTIFICATION)))
+  if (lastNotificationTime == -1)
```
fgalan: Could you elaborate on this part, please? Maybe we have discussed at skype, but I cannot remember about this no...

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -849,7 +837,7 @@ int mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub, const char*
   cSubP->servicePath           = strdup(servicePath);
   cSubP->notifyFormat          = stringToFormat(formatString);
   cSubP->reference             = strdup(sub.hasField(CSUB_REFERENCE)? sub.getField(CSUB_REFERENCE).String().c_str() : "NO REF");  // Mandatory
-  cSubP->throttling            = sub.hasField(CSUB_THROTTLING)?       sub.getField(CSUB_THROTTLING).Long() : -1;
+  cSubP->throttling            = sub.hasField(CSUB_THROTTLING)?       getLongField(sub, CSUB_THROTTLING) : -1;
```
fgalan: getIntOrLongAsLong() ?

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -63,22 +63,29 @@ HttpStatusCode mongoSubscribeContext
...
-    OID oid;
+    BSONObjBuilder  sub;
+    OID             oid;
+
     oid.init();
+
     sub.append("_id", oid);
     sub.append(CSUB_EXPIRATION, expiration);
-    sub.append(CSUB_REFERENCE, requestP->reference.get());
+    sub.append(CSUB_REFERENCE,  requestP->reference.get());
```
fgalan: Extra whitespace?

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
```diff
@@ -127,8 +127,9 @@ HttpStatusCode mongoUpdateContextAvailabilitySubscription
   newSub.append(CASUB_ATTRS, attrs.arr());
 
   /* Duration (optional) */
-  if (requestP->duration.isEmpty()) {
-      newSub.append(CASUB_EXPIRATION, getField(sub, CASUB_EXPIRATION).numberLong());
+  if (requestP->duration.isEmpty())
+  {
+    newSub.append(CASUB_EXPIRATION, getField(sub, CASUB_EXPIRATION).numberLong());
```
fgalan: getIntOrLongAsLong() ?

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -234,7 +234,7 @@ static void prepareDatabase(void) {
 
     BSONObj sub1 = BSON("_id" << OID("51307b66f481db11bf860001") <<
                         "expiration" << 10000000 <<
-                        "lastNotification" << 15000000 <<
+                        "lastNotification" << (long long) 15000000 <<
```
fgalan: Let's check if the final "read long or int, write long" logic works" by removing `(long long)` in all the prepareDatabase() places.

(Note that other modifications to the unit test, in particular `sub.getField("lastNotification").Long()` would be ok, in the case of fields re-writen during the test, as our write policy is to use long).

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1213,20 +1213,8 @@ static bool addTriggeredSubscriptions_withCache
...
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
-      long long throttling       = sub.hasField(CSUB_THROTTLING) ? getField(sub, CSUB_THROTTLING).numberLong() : -1;
-      long long lastNotification = -1;
```
fgalan: Probably this PR deserver a line like this in CNR:

"Hardening: safer read of lastNotification, expiration and throttling fields in csub collection, assuming that the field can be either int or long (always written as long)"

or something similar.

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
kzangeli: I only implemented getIntOrLongFieldAsLong for expirationTime (lastNotification was already done), For throttling it was not needed. All tests passed.

But sure, we can do it for throttling as well, if you think it is worth the effort

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: sub.hasField was removed at is was done inside tghe old version of getLongField.
Now this has changed. hasField must come back here too.

Good catch.  However, let's take a look at the entire function ...

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Nope, just an extra space to align "requestP->reference.get()" and "expiration" (on the line above)

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
kzangeli: This is ngsi9, I decided not to touch that part ...

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
fgalan: I think it could worth the effort... just in case we avoid some unexpected problem in the short future and it should be to hard to implemented.

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, I believe the "hasField" must be brought back

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
fgalan: Ok, I missed the A in CASUB_EXPIRATION...

NTC

src/lib/mongoBackend/mongoOntimeintervalOperations.cpp
kzangeli: Fixed in f5612c9

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in f5612c9

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in f5612c9

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in f5612c9

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 4c5126a

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: It seems this PR doesn't solve the https://github.com/telefonicaid/fiware-orion/issues/1406, so this will remain.

NTC

fgalan:
LGTM

Issue #881, opened 2015-04-15
[Reducing DB interactions at mongoBackend] Fill ONCHANGE notifications information from thread memory instead of entityQuery()


fgalan:
As in version 0.20.0, the "filling" of ONCHANGE notification is done using entitiesQuery() operation, which involves a new query to DB (i.e. suboptimal, given that that information is already in the thread memory processing the updateContext).

This issue is about changing the way it works, so avoiiding the additional query to DB.

fgalan:
Not sure at the end if all information is already known, e.g. in the case of subscriptions with geo-location scopes associated. We need entitiesQuery() to solve that notifications.

fgalan:
After analyzing this a little bit, there is a couple of "spin off" issues also related with reducing the number of DB interactions in mongoBackend:

* https://github.com/telefonicaid/fiware-orion/issues/1475
* https://github.com/telefonicaid/fiware-orion/issues/1476

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1483

PR #1483, opened 2015-11-10
Avoid double query to compose ONCHANGE notifications triggered by update context

fgalan:
Implements #881 

100% unit test and 100% ft are passing. Some part has been heavily refactored, in order to make the fixes functionality fit in a appropaite way.

Valgrind suite has not been (yet) executed. First, I need some feedback on the existing code.


src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -85,7 +85,7 @@ ContextAttribute::ContextAttribute(ContextAttribute* caP)
   numberValue           = caP->numberValue;
   boolValue             = caP->boolValue;
   compoundValueP        = caP->compoundValueP;
-  caP->compoundValueP   = NULL;
+  caP->compoundValueP   = NULL;   // FIXME P10: really?? shouldn't it be a clone?
```
fgalan: I need a second opinion here... 

This seems to be a bug, as the constructor shouldn't modify the "source" being copied. In addition, a clone() for portencial compounds is missing, i.e.:

```
compoundValueP = caP->compoundValueP == NULL ? NULL : caP->compoundValueP->clone();
```

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2866,7 +3092,18 @@ void processContextElement
...
+        ContextElementResponse* notifyCerP = new ContextElementResponse(ceP);
+        notifyCerP->contextElement.entityId.servicePath = servicePathV.size() > 0? servicePathV[0] : "";
+
+        processSubscriptions(subsToNotify, notifyCerP, &errReason, tenant, xauthToken);
+
+        // FIXME P10: I would say that the memory allocated previously for notifyCerP should
+        // be released, but uncomentaing this line produces a double-free crash. Mabye somebody
+        // has already released before?
+        //notifyCerP->release();
+        //delete notifyCerP;
```
fgalan: I understand that `delete notifyCerP` should be used here. However, I'm having some problems doing so... maybe something that I don't understand (or is wrong) in the chained releasing of the internal objects.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2771,6 +2991,12 @@ void processContextElement
     //
     releaseTriggeredSubscriptions(subsToNotify);
 
+    // FIXME P10: I would say that the memory allocated previously for notifyCerP should
+    // be released, but uncomentaing this line produces a double-free crash. Mabye somebody
+    // has already released before?
+    //notifyCerP->release();
+    //delete notifyCerP;
```
fgalan: I understand that `delete notifyCerP` should be used here. However, I'm having some problems doing so... maybe something that I don't understand (or is wrong) in the chained releasing of the internal objects.

src/lib/ngsi/ContextAttribute.cpp
kzangeli: This is a "special" construction, letting the new attribute 'inherit' the compound of the 'older' attribute:

    compoundValueP        = caP->compoundValueP;   // Give compond to new attr
    caP->compoundValueP   = NULL;                  // caP no longer has the compound

This has been tested to work (doesn't mean it will keep working for all future - it clearly depends on its usage).
The advantage with this implementation is that it is **very** fast, the pointer -assignment is immediate. To clone the entire compound and, later on, free not one, but *two* compounds might be the **right** way to do this, but I opted for this 'inheriting-mechanism' instead.

Uptil now it has been enough. If this stops to be true, then we'd have to implement the cloning of the compound and the 'extra free', to avoid leaks.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: processOnChangeConditionForUpdateContext:

    caP = notifyCerP->contextElement.contextAttributeVector.get(ix);

It seems you point to an attribute of notifyCerP.
If you copy the attribute instead, you might be able to free  notifyCerP  where you want

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1500,6 +1501,83 @@ static bool addTriggeredSubscriptions
...
+)
+{
+  NotifyContextRequest   ncr;
+  ContextElementResponse cer;
+  cer.contextElement.entityId.fill(&notifyCerP->contextElement.entityId);
+
+  /* Fill NotifyContextRequest with cerP, filtering by attrL */
+  for (unsigned int ix = 0; ix < notifyCerP->contextElement.contextAttributeVector.size(); ix++)
+  {
+    ContextAttribute* caP = notifyCerP->contextElement.contextAttributeVector.get(ix);
```
kzangeli: Right here!

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1500,6 +1501,83 @@ static bool addTriggeredSubscriptions
...
+
+  ncr.contextElementResponseVector.push_back(&cer);
+
+  /* Complete the fields in NotifyContextRequest */
+  ncr.subscriptionId.set(subId);
+  // FIXME: we use a proper origin name
+  ncr.originator.set("localhost");
+
+  getNotifier()->sendNotifyContextRequest(&ncr, notifyUrl, tenant, xauthToken, format);
+  //ncr.contextElementResponseVector.release();  safe??
```
kzangeli: Probably, but only if you don't point to the same attribute from more than one place

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1709,6 +1779,103 @@ static unsigned int howManyAttrs(BSONObj& attrs, std::string& attrName)
...
+* deleteAttrInNotifyCer -
+*
+*/
+static void deleteAttrInNotifyCer
+(
+  ContextElementResponse* notifyCerP,
+  ContextAttribute*       targetAttr
+)
+{
+  /* The algorithm is based in copying the old notify CER object excluding the delted attribute */
```
kzangeli: This comment in better in the function header, that's what the header is for ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1709,6 +1779,103 @@ static unsigned int howManyAttrs(BSONObj& attrs, std::string& attrName)
...
+  for (unsigned int ix = 0; ix < notifyCerP->contextElement.contextAttributeVector.size(); ix++)
+  {
+    ContextAttribute* caP = notifyCerP->contextElement.contextAttributeVector.get(ix);
+    if (caP->name == targetAttr->name)
+    {
+      //newNotifyCerP->contextElement.contextAttributeVector.push_back(caP);
+      caP->skip = true;
+    }
+  }
+  //delete notifyCerP;  unexplicable crash...
```
kzangeli: Was it a "trying to free an unallocated object" type of crash?
Might be the same problem as before

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1872,7 +1591,7 @@ static void processEntity(ContextRegistrationResponse* crr, EntityIdVector enV,
 *
 * processAttribute -
 */
-static void processAttribute(ContextRegistrationResponse* crr, AttributeList attrL, BSONObj attribute)
+static void processAttribute(ContextRegistrationResponse* crr, const AttributeList& attrL, BSONObj attribute)
```
kzangeli: Good! :-)
But, BSONObj is also an object, right?

src/lib/ngsi/ContextElementResponse.cpp
```diff
@@ -75,6 +82,178 @@ ContextElementResponse::ContextElementResponse(ContextElementResponse* cerP)
...
+/* ****************************************************************************
+*
+* ContextElementResponse::ContextElementResponse -
+*
+* This constructor builds the CER object based in a BSON object taken from the
+* entities collection at DB
+*/
+ContextElementResponse::ContextElementResponse(const mongo::BSONObj& entityDoc, const AttributeList& attrL, bool includeEmpty)
+{
+  /* Note that statusCode is not touched by this constructor */
```
kzangeli: Comment better in func-header

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Rigth... I focused on `attrL` when chanching to `const &` but of course it can also be used in the others arguments in the signature (in this case, the `attribute`). I'll do it.

src/lib/ngsi/ContextAttribute.cpp
fgalan: I think is ok, as long as the behaviour is clear and and the code (taking into account the changes in this PR) works. I'll add a comment in the constructor preamble to explain it.

(Btw, note that compound are not cloned, but metadata aren't... This different approach in two different "children" ContextAttribute is also a bit confusing. I'll hope to clarify also this in the comment).

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Actually, that comment is not correct (a previous attempt of implementing this function was based on creating a new CER object; however final version is simpler and relies in the `skip` CA field).

I will add a comment about the algorithm finally implemented in the function preamble.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Actually, that comment is a leftover, as the algorithm at the end is not based in create new CER.

NTC

src/lib/ngsi/ContextElementResponse.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in f747df6

src/lib/ngsi/ContextAttribute.cpp
fgalan: Fixed in f747df6

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Actually, NTC (it seems I don't need to touch this to make memory stuff to work...)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in ed0205145c6ef266149002419009afede139b384 (just removed the comment)

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yeah ....   There will be a leak here that will demand this to be done.
No worries, I have it in mind and I'll take a look ...

fgalan:
LGTM (to Ken's changes... thanks for the de-leaking ;)

kzangeli:
LGTM

Issue #1485, opened 2015-11-11
Notification queue statistics


fgalan:
It would be useful to have the following statisticis about que queue used for notifications:

* Size of the queue (current value)
* Number of items that have entered in the queue (accumulation)
* Number of items that have exited out the queue (accumulation)
* Number of items that were droped due to the queue was full (accumulation)
* Time that itmems have waited in the queue (accumulation). The best way of doing so is probably expand SenderThreadParams to include a timestamp (to be set in the moment the item is put in the queue), then get current time when the SenderThreadParam exits the queue and sum the different to the accumulator for this metric.

fgalan:
Partial implementation in PR https://github.com/telefonicaid/fiware-orion/pull/1489

fgalan:
Finished in PR https://github.com/telefonicaid/fiware-orion/pull/1501

PR #1489, opened 2015-11-11
Add statistics for notif queue (part A, w/o time)

crbrox:
Partially implements #1485 

(In a following PR for part B, with time measures, we can refactor and choose better names for variables etc)

### DIFF
--- a/src/lib/ngsiNotify/QueueWorkers.cpp
+++ b/src/lib/ngsiNotify/QueueWorkers.cpp
@@ -106,13 +106,15 @@ static void *workerFunc(void* pSyncQ)
     {
       statisticsUpdate(NotifyContextSent, params->format);
     }
+
+    // FIXME: These counters should be incremented in the other notification modes
     if (r != "error" && r != "")
     {
-      ++QueueStatistics::noOfNotificationsQueueOK;
+      ++QueueStatistics::noOfNotificationsQueueSentOK;
     }
     else
     {
-      ++QueueStatistics::noOfNotificationsQueueError;
+      ++QueueStatistics::noOfNotificationsQueueSentError;
     }
 
     // Free params memory
--- a/src/lib/ngsiNotify/QueueWorkers.cpp
+++ b/src/lib/ngsiNotify/QueueWorkers.cpp
@@ -32,6 +32,7 @@
 #include "logMsg/traceLevels.h"
 #include "ngsi10/NotifyContextRequest.h"
 #include "rest/httpRequestSend.h"
+#include "ngsiNotify/QueueStatistics.h"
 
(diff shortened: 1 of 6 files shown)


src/lib/ngsiNotify/QueueWorkers.cpp
```diff
@@ -103,6 +106,14 @@ static void *workerFunc(void* pSyncQ)
     {
       statisticsUpdate(NotifyContextSent, params->format);
     }
+    if (r != "error" && r != "")
+    {
+      ++QueueStatistics::noOfNotificationsQueueOK;
```
fgalan: I'd suggest noOfNotificationsQueueSentOK. 'Sent' is important to mean that an actual HTTP interaction take place (i.e. it was not only movements in the internal queue).

In addition, probably this SendOk and SendError counters should be general for all notification modes. Please, add a FIXME about it around this code.

src/lib/ngsiNotify/QueueWorkers.cpp
```diff
@@ -103,6 +106,14 @@ static void *workerFunc(void* pSyncQ)
...
+    }
+    else
+    {
+      ++QueueStatistics::noOfNotificationsQueueError;
```
fgalan: In a similar line, it could be noOfNotificationsQueueSentError.

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -457,7 +457,7 @@ std::string statisticsTreat
 
   if (noOfDroppedNotifications != -1)
   {
-    out += TAG_ADD_COUNTER("droppedNotifications", noOfDroppedNotifications);
+    out += TAG_ADD_COUNTER("droppedNotifications", noOfDroppedNotifications + 1);
```
fgalan: I'd suggest to add a comment about why the +1 is neded. Just something like, "Given that the noOfDroppedNotifications  starts at -1, a +1 adjustement is needed".

src/lib/ngsiNotify/QueueWorkers.cpp
crbrox: Fixed 706b1e9

src/lib/ngsiNotify/QueueWorkers.cpp
crbrox: Fixed 706b1e9

src/lib/serviceRoutines/statisticsTreat.cpp
crbrox: Fixed 706b1e9

fgalan:
LGTM

kzangeli:
LGTM

Issue #1482, opened 2015-11-10
"contextBroker -db 1234567890" lefover after ft run


fgalan:
Sometimes a leftover contextBroker process appears after `make ft`:

```
fermin@neodeb:~/src/fiware-orion/test$ ps ax | grep context
104594 pts/0    S+     0:00 grep context
122755 ?        Ssl    0:00 contextBroker -db 1234567890
```

That's weird, as CB run by the test harness suite should use 9999 port...

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1496

PR #1496, opened 2015-11-12
Small fixes in unittest and functional tests

### LINKED ISSUE #1481: "uniittest" leftover DB afer unit_test run
It seems that a leftover `unittest` DB remains after unit test run. Probably, even using `unittest` is wrong (as we changed unit test DB name time ago to `utest`).

At the begining:

```
> show dbs
local  0.000GB
```

After running `make unit_test`

```
> show dbs
local     0.000GB
unittest  0.000GB
```

fgalan:
Fixes #1481 and #1482


crbrox:
LGTM

Issue #1488, opened 2015-11-11
Doing twice unsubscribeContext with empty ID hangs broker


fgalan:
Just after starting CB it seems that first unsubscribeContext goes ok, send on hangs. It seems related with a sem that is not correctly released.

Curl trace:

```
fermin@neodeb:~/src/fiware-orion$ sh test.sh
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 1026 (#0)
> POST /v1/unsubscribeContext HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:1026
> Fiware-service: unknownT
> Fiware-servicepath: /
> Content-Type: application/json
> Accept: application/json
> Content-Length: 21
> 
* upload completely sent off: 21 out of 21 bytes
< HTTP/1.1 200 OK
< Connection: Keep-Alive
< Content-Length: 144
< Content-Type: application/json
< Date: Wed, 11 Nov 2015 13:04:19 GMT
< 
{
  "subscriptionId" : "000000000000000000000000",
  "statusCode" : {
    "code" : "404",
    "reasonPhrase" : "No context element found"
  }
}
* Connection #0 to host localhost left intact
fermin@neodeb:~/src/fiware-orion$ sh test.sh
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 1026 (#0)
> POST /v1/unsubscribeContext HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:1026
> Fiware-service: unknownT
> Fiware-servicepath: /
> Content-Type: application/json
> Accept: application/json
> Content-Length: 21
> 
* upload completely sent off: 21 out of 21 bytes
[hangs forever]
```

test.sh script:

```
(curl localhost:1026/v1/unsubscribeContext -s -S  -v --header 'Fiware-service: unknownT' --header 'Fiware-servicepath: /' --header 'Content-Type: application/json' --header 'Accept: application/json' -d @- ) <<EOF
{
"subscriptionId":""
}
EOF
```

Log:

```
DEBUG@13:04:19  sem.cpp[266]: transactionIdSet taking the 'trans' semaphore for 'changing the transaction id'
DEBUG@13:04:19  sem.cpp[287]: transactionIdSet has the 'trans' semaphore
DEBUG@13:04:19  sem.cpp[363]: transactionIdSet gives the 'trans' semaphore for 'changing the transaction id'
[...]

PR #1501, opened 2015-11-13
Add time and size stats for notif queue

crbrox:


src/lib/common/clockFunctions.cpp
```diff
@@ -32,7 +32,7 @@
 *
 * clock_difftime - 
 */
-void clock_difftime(struct timespec* endTime, struct timespec* startTime, struct timespec* diffTime)
+void clock_difftime(const struct timespec* endTime, const struct timespec* startTime, struct timespec* diffTime)
```
fgalan: I understand that endTime and startTime are input parametesr and diffTime output parameter, so the best signature would be:

```
clock_difftime(const struct timespec& endTime, const struct timespec& startTime, struct timespec* diffTime)
```

Although I'm not sure if is possible, as startTime and endTime come as pointer from a lower-level function...

src/lib/common/clockFunctions.cpp
```diff
@@ -50,7 +50,7 @@ void clock_difftime(struct timespec* endTime, struct timespec* startTime, struct
 *
 * clock_addtime - 
 */
-void clock_addtime(struct timespec* accTime, struct timespec* diffTime)
+void clock_addtime(struct timespec* accTime, const struct timespec* diffTime)
```
fgalan: Similarly, maybe this could be?

```
void clock_addtime(struct timespec* accTime, const strunct timespec& diffTime)
```

kzangeli: Let's not mix C-pointers and C++ references, the call will look really ugly.
It is true a change like this would somehow add to the understanding of the function, but still.
I would keep it as is. IMO this extra 'const' is enough to show which is input/output.

src/lib/ngsiNotify/QueueStatistics.cpp
```diff
@@ -122,3 +130,38 @@ void QueueStatistics::incSentError()
 {
   __sync_fetch_and_add(&noOfNotificationsQueueSentError, 1);
 }
+/* ****************************************************************************
+*
+* getTimInQ -
+*/
+void QueueStatistics::getTimeInQ(char* buf, size_t bufLen)
+{
+  boost::mutex::scoped_lock lock(mtxTimeInQ);
```
fgalan: I understand that this is to lock the mutet until the current scope is destroyed (i.e. the getTimeInQ() function returns) so no explicit unlock statement is needed. Is that correct?

(This is specially interesting for functions which many exit point, we have several cases of them in mongoBackend and missing the unlock before `return` causes bug... one of them fixes few days ago: https://github.com/telefonicaid/fiware-orion/issues/1488)

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -455,10 +455,13 @@ std::string statisticsTreat
     out += TAG_ADD_COUNTER("subCacheRemovalFailures", noOfSubCacheRemovalFailures);
   }
 
-  if (noOfDroppedNotifications != -1)
   {
-    // Given that the noOfDroppedNotifications starts at -1, a +1 adjustement is needed
-    out += TAG_ADD_COUNTER("droppedNotifications", noOfDroppedNotifications + 1);
+    int noOfDroppedNotifications = __sync_fetch_and_add(&noOfDroppedNotifications,0);
```
fgalan: Nit: space between "," and "0" in ",0".

src/lib/common/SyncQOverflow.h
```diff
@@ -87,4 +88,15 @@ Data SyncQOverflow<Data>::pop()
...
 }
 
+/* ****************************************************************************
+*
+* SyncQOverflow<Data>::size -
+*/
+template <typename Data>
+size_t SyncQOverflow<Data>::size() const
+{
+  boost::mutex::scoped_lock lock(mtx);
```
fgalan: Do we really need the mutex here?

src/lib/serviceRoutines/statisticsTreat.cpp
crbrox: fixed db850b6

src/lib/common/SyncQOverflow.h
fgalan: After offline discussion, it seems so.

NTC

src/lib/common/clockFunctions.cpp
fgalan: NTC

src/lib/common/clockFunctions.cpp
fgalan: NTC

src/lib/ngsiNotify/QueueStatistics.cpp
fgalan: After internal discussion, it works as I mention.

NTC

fgalan:
LGTM

Issue #1367, opened 2015-10-15
Time measures in the broker itself


kzangeli:
The way we measure the time spent waiting for semaphores, we could "easily" implement for **blocks** of the code in the broker.
Typical blocks would be 
- jsonParse
- xmlParse
- jsonParseV2
- xmlRender
- jsonRender
- mongoBackend
- mongoSubCache (which is really a part of mongoBackend ...)

And probably many more.

As time measuring is time consuming, we would only perform these measurements if a CLI is set, just like we do with the semaphore waiting time (-mutexTimeStat).
To see the statistics, we'd just use the GET /statistics, just like with mutex-times.
No real reason to invent a new REST request for this (I wouldn't mind though: GET /profiling).

fgalan:
Not sure if I name this issue "profiling", which is a more wider topic. I will focus in the time measure, so probably "time measure in the context broker" maybe.

kzangeli:
Sure, just rename the issue :-)

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1502

PR #1502, opened 2015-11-13
Hardening/1367 time measures

kzangeli:
Time-profiling in statistics for a select number of operations.
Fixes issue #1367


doc/manuals/admin/cli.md
```diff
@@ -143,3 +143,4 @@ The list of available options is the following:
 
 -   **-maxConnections**. maximum number of simultaneous connections. Default value is "unlimited" (limited by max file descriptors of operating system).
 -   **-reqPoolSize**. size of thread pool for incoming connections. Default value is 0, meaning *no thread pool*.
+-   **-reqTimeStat. turn on request-time-measuring in run-time. By default it is turned off (turned on, the broker will be somewhat slower in execution).
```
fgalan: Mising `**` to close the bold stylt text.

fgalan: Given that I'm going to improve documentation in this PR, I think I can take care of this fix.

src/lib/common/clockFunctions.cpp
```diff
@@ -61,3 +61,21 @@ void clock_addtime(struct timespec* accTime, struct timespec* diffTime)
...
   }
 }
+
+
+
+/* ****************************************************************************
+*
+* clock_subtime - 
+*/
+void clock_subtime(struct timespec* subtrahend, struct timespec* minuend)
```
fgalan: It would be great to align the signature of this function with the sylte that will be used at the end for clockFunctions.cpp/h being discussed at https://github.com/telefonicaid/fiware-orion/pull/1501/files#r44779953. After lasts comments, I understand it should be:

```
void clock_subtime(struct timespec* subtrahend, const struct timespec* minuend)
```

src/lib/jsonParse/jsonParse.cpp
```diff
@@ -428,6 +432,13 @@ std::string jsonParse
   ptree              tree;
   ptree              subtree;
   std::string        path;
+  struct timespec    start;
+  struct timespec    end;
+
+  if (reqTimeStatistics)
```
fgalan: What about?

```
reqTimeStatistics && clock_gettime(CLOCK_REALTIME, &start);
```

(In general, as alternative to the multiple `if (reqTimeStatistics) {...}` structures that this PR introduces).

Not sure if that is a good idea, just to put the alternative on the table. First, because I haven't checked it it works :). Second, although it is more compact it could be considered less clear (for me is equivalently clear than the `if` but not sure others...).

src/lib/jsonParse/jsonParse.cpp
```diff
@@ -445,5 +456,11 @@ std::string jsonParse
     }
   }
 
+  if (reqTimeStatistics)
```
fgalan: The one-liner alternative probably don't fit very well in this point... so, probably it is not a good idea at the end.

src/lib/mongoBackend/connectionOperations.cpp
```diff
@@ -450,6 +473,7 @@ bool collectionCreateIndex
   std::string*        err
 )
 {
+  TIME_STAT_MONGO_WRITE_WAIT_START();
```
fgalan: Looking to the internal implementation of the `createIndex()` command:

```
> db.x.createIndex
function ( keys , options ){
    var o = this._indexSpec( keys, options );

    if ( this.getMongo().writeMode() == "commands" ) {
        delete o.ns; // ns is passed to the first element in the command.
        return this._db.runCommand({ createIndexes: this.getName(), indexes: [o] });
    }
    else if( this.getMongo().writeMode() == "compatibility" ) {
        // Use the downconversion machinery of the bulk api to do a safe write, report response as a
        // command response
        var result = this._db.getCollection( "system.indexes" ).insert( o , 0, true );

        if (result.hasWriteError() || result.hasWriteConcernError()) {
            var error = result.hasWriteError() ? result.getWriteError() :
                                                 result.getWriteConcernError();
            return { ok : 0.0, code : error.code, errmsg : error.errmsg };
        }
        else {
            return { ok : 1.0 };
        }
    }
    else {
        this._db.getCollection( "system.indexes" ).insert( o , 0, true );
    }
}
```

it is actually a command, so this whould use `_COMMAND_` macros.

src/lib/mongoBackend/connectionOperations.cpp
```diff
@@ -304,6 +319,7 @@ bool collectionInsert
   {
     connection->insert(col.c_str(), doc);
     releaseMongoConnection(connection);
+    TIME_STAT_MONGO_WRITE_WAIT_STOP();
```
fgalan: Maybe I'm missing something, but I understand that `TIME_STAT_MONGO_WRITE_WAIT_STOP()` should be also used after the `releaseMongoConnection(connection)` in the catch blocker where the program flow may end in the case of exception of conneciton insert.

I think the problem may happend in other functions in connectionOperations.cpp, as all them share the same try/catch structure.

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -67,7 +68,9 @@ std::string statisticsTreat
   std::string out     = "";
   std::string tag     = "orion";
   std::string indent  = "";
-  std::string indent2 = (ciP->outFormat == JSON)? indent + "    " : indent + "  ";
+  std::string indent2 = "  ";
+
+  LM_M(("In statisticsTreat"));
```
fgalan: I understand this LM_M should be removed.

src/lib/rest/rest.cpp
```diff
@@ -464,17 +466,54 @@ static void requestCompleted
...
+  //
+  // Flush this requests timing measures onto a global var to be read by "GET /statistics".
+  // Also, increment the accumulated measures.
+  if (reqTimeStatistics)
+  {
+    timeStatSemTake(__FUNCTION__, "updating statistics");
+
+    memcpy(&lastTimeStat, &threadLastTimeStat, sizeof(lastTimeStat));
+
+    // "Fix" mongoBackendTime
```
fgalan: I'd suggest a more elaborate description (we can forget what this "fix" mean in a couple of months and think it was due to some unknown bug or something like that), e.g.:

```
// Substract times waiting at mongo driver operation (in mongo[Read|Write|Command]WaitTime counters) so mongoBackendTime contains at the end the time passed in our logic, i.e. a kind of "self-time" for mongoBackend
```

src/lib/mongoBackend/connectionOperations.cpp
```diff
@@ -359,6 +377,7 @@ bool collectionUpdate
   {
     connection->update(col.c_str(), q, doc, upsert);
     releaseMongoConnection(connection);
+    TIME_STAT_MONGO_WRITE_WAIT_STOP();
```
fgalan: In the case a given operation at mongoBackend used more than one WRITE/READ/COMMAND operation... does the current logic accumulate the time of all them in the mongo[Write|Read|Command|WaitTime or it takes only the last one?

src/lib/serviceRoutines/deleteIndividualContextEntity.cpp
```diff
@@ -90,7 +93,10 @@ std::string deleteIndividualContextEntity
   }
 
   // 05. Cleanup and return result
+  TIME_STAT_RENDER_START();
```
fgalan: In order to avoid duplication (we have this same pattern in many places now), could a macro help to "wrap" the _START() and _STOP()? Something like this (probably not fully correct from a syntax point of view):

```
#define renderWithTimeStat(s, object, function, params)
TIME_STAT_RENDER_START();
s = object.function(params);
TIME_STAT_RENDER_STOP();
```

To be use like this:

```
renderWithTimeStat(answer, response, render, (ciP->outFormat, "", false, false))
```

(If render function is always the same, with name `render`, the macro definition could be simplified to `#define renderWithTimeStat(s, object, function, params)`)

src/lib/serviceRoutines/getAttributesForEntityType.cpp
```diff
@@ -56,12 +60,17 @@ std::string getAttributesForEntityType
 )
 {
   EntityTypeResponse  response;
-  std::string                   entityTypeName = compV[2];
+  std::string         entityTypeName = compV[2];
 
   response.statusCode.fill(SccOk);
+
+  TIME_STAT_MONGO_START();
```
fgalan: Similarly, could a macro help for mongoBackend stats? Eg:

```
#define mongoBackendWithTimeStat(function, params)
   TIME_STAT_RENDER_START();
   function(params);
   TIME_STAT_RENDER_STOP();
```
To be invoked:

```
mongoBackendWithTimeStat(mongoAttributesForEntityType,(entityTypeName, &response, ciP->tenant, ciP->servicePathV, ciP->uriParam))
```

src/lib/serviceRoutines/statisticsTreat.cpp
```diff
@@ -509,23 +517,37 @@ std::string statisticsTreat
...
+    out += TAG_ADD_INTEGER("noOfNotificationsQueueReject",    QueueStatistics::getReject(),    true);
+    out += TAG_ADD_INTEGER("noOfNotificationsQueueSentOK",    QueueStatistics::getSentOK(),    true);
+    out += TAG_ADD_INTEGER("noOfNotificationsQueueSentError", QueueStatistics::getSentError(), commaAfterThreadpool);
+  }
+
+
+  //
+  // Timing statistics ONLY if JSON
+  //
+  if (ciP->outFormat == JSON)
```
fgalan: Does this means that if statistics are rendered in XML timing statistics are not included?

Just to be sure for documentation purpose... In fact, probably we could adopt a more radical approach and enable statistics in JSON format only (let's take into account that XML is officially deprecated as request/response encoding since 0.23.0: https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/deprecated.md).

src/lib/serviceRoutinesV2/deleteEntity.cpp
```diff
@@ -91,28 +95,33 @@ std::string deleteEntity
...
+        oe.reasonPhrase      = "NotFound";
+        oe.details           = "Attribute not found";
+        ciP->httpStatusCode  = SccContextElementNotFound; // We don't want a 472
       }
       else
       {
-        oe.code = scc;
-        oe.reasonPhrase = sc.reasonPhrase;
+        oe.code          = scc;
+        oe.reasonPhrase  = sc.reasonPhrase;
```
fgalan: Actual changes above this line or just indentation?

src/lib/serviceRoutinesV2/getEntityAttributeValue.cpp
```diff
@@ -98,17 +109,24 @@ std::string getEntityAttributeValue
         answer = attribute.pcontextAttribute->compoundValueP->render(ciP, JSON, "");
         if (attribute.pcontextAttribute->compoundValueP->isObject())
         {
-            answer = "{" + answer + "}";
+          TIME_STAT_RENDER_START();
+          answer = "{" + answer + "}";
```
fgalan: Not fully sure but I would say that the `answer = "[" + answer + "]" and `answer = "{" + answer + "}"` parts should be left as the were and the TIME_STAT_RENDER_START/STOP be around the function doing the "hard work" to render, i.e. some lines above:

```
answer = attribute.pcontextAttribute->compoundValueP->render(ciP, JSON, "");
```

test/functionalTest/cases/0000_statistics_operation/statistics.test
```diff
@@ -63,22 +63,20 @@ Date: REGEX(.*)
...
-        "statisticsRequests": "2",
-        "subCacheInserts": "0",
-        "subCacheItems": "0",
-        "subCacheRefreshs": "1",
-        "subCacheRemoves": "0",
-        "subCacheUpdates": "0",
-        "uptime_in_secs": "REGEX(1?\d)",
-        "versionRequests": "REGEX((0|1))"
-    }
+    "measuring_interval_in_secs": "REGEX(1?\d)",
```
fgalan: Great to see that the useless "orion" field has been removed!

In fact, I think we can go even further structuring the content of the statistics section, but that would be a different PR.

NTC

src/lib/common/clockFunctions.cpp
kzangeli: Which is exactly the signature the function has right now ... :-)

NTC

src/lib/jsonParse/jsonParse.cpp
kzangeli: Not sure what you mean here ...
If I understand correctly, it seems like you're thinkiung of a typical SHELL construction.
Doesn't work like that in C

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: The reason not to put the 'end' statement in 'erroneous paths' of the program is that I thought those measurements might not be very interesting. The operation failed, so the time measured could be just anything ... 
There is no problem with the logic, the STOP macro isn't used so whatever was 'recorded' by the START macro is thrown away.

I don't have any problem in measuring erroneous paths as well, I just didn't find it very interesting ...

src/lib/serviceRoutines/deleteIndividualContextEntity.cpp
kzangeli: Of course we could do that, we just don't gain anything with it.
It only makes the source code a little more difficult to read.
I don't have a very strong opinion on this. If you really want a new macro for the three calls, sure.
Just let me know.

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: Not sure what you mean here.
This is very WYSIWYG, the measurements are *exactly* where the calls to the macros are in the code. The START macros measure the time for START and the STOP macros get the difference and records it. 

If you believe the macro calls should be elsewhere, just indicate where. This is your territory ...

src/lib/serviceRoutines/statisticsTreat.cpp
kzangeli: Yeah, I decided not to waste time on XML for this new feature.
If/When we rewrite the statistics, I also believe it's a good idea to do it ONLY for JSON.

src/lib/serviceRoutinesV2/getEntityAttributeValue.cpp
kzangeli: You are so right. I didn't see the call to the render method above.
Good catch!

src/lib/common/clockFunctions.cpp
fgalan: It isn't... `const` is missing before `struct timespec* minuend`.

src/lib/jsonParse/jsonParse.cpp
fgalan: NTC

src/lib/jsonParse/jsonParse.cpp
fgalan: NTC

src/lib/mongoBackend/connectionOperations.cpp
fgalan: I think is could be important.

src/lib/serviceRoutines/deleteIndividualContextEntity.cpp
fgalan: In mi opinion, readibility is similar ni both cases. Although `answer = response.render(ciP->outFormat, "", false, false)` is slighly more readable than `renderWithTimeStat(answer, response, render, (ciP->outFormat, "", false, false))` the TIME_STAT_RENDER_START() and TIME_STAT_RENDER_STOP() lines add readbility "noise".

Thus, being readibility more or less the same, I think the alternative of using a macro is safer. Otherwise, we can miss for example a  TIME_STAT_RENDER_STOP() when writting the code.

(BTW, I'm not sure that renderWithTimeStat is the best name... maybe you can find some better).

src/lib/mongoBackend/connectionOperations.cpp
fgalan: I'll try to clarify.

Let's imagine a given operation at mongoBackend that includes 1 query operation 2 cmd operations and 5 write operations. In that case, after processing the request, what would be set in "last" and accumulated in "acc" for mongoWriteTime? The time corresponding to the sum of the 5 write operations? Or the time of the fifth write operation?

src/lib/serviceRoutines/statisticsTreat.cpp
fgalan: NTC

src/lib/serviceRoutinesV2/getEntityAttributeValue.cpp
kzangeli: Fixed in dba0316

src/lib/rest/rest.cpp
kzangeli: Fixed in dba0316

src/lib/serviceRoutines/statisticsTreat.cpp
kzangeli: Fixed in dba0316

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: Fixed in dba0316

doc/manuals/admin/cli.md
kzangeli: The missing '**' fixed in dba0316

src/lib/common/clockFunctions.cpp
kzangeli: Ah, of course! :-)

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: Only the last one, the "part-operation" if you like.
That's for the "last" part.
In "acc", everything will be there, of course.

So, to get what you are looking for, right now you would have to do one 'GET /stat' before an operation, then another one after and do the math yourself.

These measurements aren't "intelligent", except for gathering the accumulated times.

src/lib/serviceRoutinesV2/deleteEntity.cpp
kzangeli: Those lines are just indentations.
Apart from:

    #include "common/statistics.h"
    #include "common/clockFunctions.h"

But that is pretty clear ...
If we ever get a "style guide checker" in the makefile and fix all our code to follow the style-guide, these style-guide changes will no longer be necessary. Until then, they will continue. I know it bothers, but try to see it as improvements.

Not doing these improvements (trying to maintain a list of all of them until after LGTM) ... it's simply not practical.
Like that these improvements will never be done.

NTC

src/lib/serviceRoutines/getAttributesForEntityType.cpp
kzangeli: Fixed in de32e1f

src/lib/serviceRoutines/deleteIndividualContextEntity.cpp
kzangeli: Fixed in de32e1f

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: Fixed in de32e1f

src/lib/common/clockFunctions.cpp
kzangeli: Fixed in de32e1f

src/lib/mongoBackend/connectionOperations.cpp
kzangeli: Fixed in e99e0e4.
As now the 'partial' measurements of "last" is accumulated until 'request completed' I understand this must be taken care of.

fgalan:
LGTM

PR #1507, opened 2015-11-17
Make notification simulation "orthogonal" to notificationMode 

crbrox:
Implements #1505 

Make notification simulation "orthogonal" to notificationMode

PR #1509, opened 2015-11-17
fixed a unittest

kzangeli:
Fixed a tiny little mistake (variable 'shading') that made a unittest fail


crbrox:
I think we have overlapped about this. It should be solved in #1507 , 
https://github.com/telefonicaid/fiware-orion/blob/feature/1505_orthogonal_simulated_notif/src/lib/serviceRoutines/statisticsTreat.cpp#L528

kzangeli:
OK, already fixed then

PR #1361, opened 2015-10-14
REFACTOR rewritten JSON representation section in NGSIv2 reference apib

fgalan:
Old "JSON entity representation" section has been rewritten, spliten in three subsections: "JSON entity representation", "JSON attribute representation" and "Direct access to attribute value" in order to match with the new JSON format (which implementation work is associated to issue #1259).

@jmcanterafonseca has been assigned as main reviewer of this, but all the others (@kzangeli, 
@mrutid and @antoniojoseln) who have take part in the looooong ;) discussions about this (both in skype and email) are welcome to contribute to this PR with comments.

PR #1511, opened 2015-11-18
FIX issue 1464 - Empty callback and metadata for registrations

jmcanterafonseca:
@fgalan @tobiasjacobs r?

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1124,11 +1124,13 @@ use cases of this functionality is the association of Context Providers.
 
 A context registration is represented by a JSON object with the following fields:
 
-+ `id`: Unique identifier assigned to the registration. Automatically created at creation time.
-+ `subject` : It s an object that describes the subject of the registration.
-+ `callback` : URL pointing to the service which is registered. In the case of a Context Provider
-corresponds to the URL of the provider service.
-+ `duration`: Duration of the registration in ISO8601 format. Default duration is infinite.
++ `id`       : Unique identifier assigned to the registration. Automatically generated at creation time.
++ `subject`  : It is an object that describes the subject of the registration.
++ `callback` : URL which denotes the end point of the registered service. It is optional, as it can be
+undefined under certain conditions, for instance, transient unavailability of the provider.
++ `metadata` : An optional JSON object which allows to associate extra metadata properties to the registration.
+It must follow the same representation conventions as context attribute metadata (See JSON Attribute Representation). 
++ `duration` : Duration of the registration in ISO8601 format. Default duration is infinite.
 
 A `subject` contains the following subfields:
 
@@ -1164,7 +1166,17 @@ Response:
                         "temperature"
                     ]
                 },
-                "callback": "http://localhost:1234",               
+                "callback": "http://weather.example.com/ngsi",
+                "metadata": {
+                    "providingService": {
+                       "value": "weather.example.com",
+                       "type": null
+                     },
+                    "providingAuthority": {
+                       "value": "AEMET - Spain",
+                       "type": null
+                     }
+                },
                 "duration": "PT1M"
             }
         ]
@@ -1194,6 +1206,11 @@ Response:
                 ]
             },
             "callback":  "http://localhost:1234",
+            "metadata" : {
+                "provider": {
+                   "value": "example.com"
+                }
+            },
             "duration": "PT1M"
         }
 


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1065,11 +1065,12 @@ use cases of this functionality is the association of Context Providers.
 
 A context registration is represented by a JSON object with the following fields:
 
-+ `id`: Unique identifier assigned to the registration. Automatically created at creation time.
-+ `subject` : It s an object that describes the subject of the registration.
-+ `callback` : URL pointing to the service which is registered. In the case of a Context Provider
-corresponds to the URL of the provider service.
-+ `duration`: Duration of the registration in ISO8601 format. Default duration is infinite.
++ `id`       : Unique identifier assigned to the registration. Automatically generated at creation time.
++ `subject`  : It s an object that describes the subject of the registration.
```
fgalan: It s -> It is

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1065,11 +1065,12 @@ use cases of this functionality is the association of Context Providers.
...
++ `callback` : URL which denotes the end point of the registered service. It can be undefined under certain conditions,
```
fgalan: I'd suggest to clearly state that the field is optional, e.g:

"URL which denotes the end point of the registered service. It is an optional field, e.g. it can be undefined under certain conditions, for instance, transient unavailability of the provider".

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: Or, even better:

  subject: field that describes the subject of the registration

short and concise ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1065,11 +1065,12 @@ use cases of this functionality is the association of Context Providers.
...
+for instance, transient unavailability of the provider.
++ `metadata` : An optional dictionary which allows to associate extra metadata to the registration.
```
fgalan: As far as I remember, registration metadata are of the same type that entity attribute metadata, so they have a name, type and value. Thus, description should be aligned with the one at https://github.com/telefonicaid/fiware-orion/pull/1361:

> metadata: a JSON object which represents attribute metadata. This property is optional in requests, meaning that the attribute doesn't have metadata. In responses, this property is set to `{}` if the attribute doesn't have metadata. Each property in this object represents a metadata element (which name is the name of the property) which value is
a JSON object composed of two properties: `value` and `type` (`value` and `type` following the same rules that the
ones described in the previous two bullets).

Probably is a good idea to merge first PR #1361, then the present PR, to ensure that both parts of the .apib are coherent.

jmcanterafonseca: do we really need such a complexity, having that for this use case seems to be overkilling

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I think the current wording is accurate as it states that subject is an object

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed at 53067a87a2e24ff5d63639ec4b9976e103cfbec1

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed at 53067a87a2e24ff5d63639ec4b9976e103cfbec1

kzangeli: 1.  "It s" id directly incorrect and has to be fixed.
2. "It is ..."  is not used anywhere else. Why here?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: It depends on whether we want to keep entities and registration metadata coherent (as FIWARE NGSIv1 and NGSI OMA do) or not. In my opinion, coherence is important (otherwise, users of the API has to remember that entity metadata use one approach, registration metadata use a different approach), so I would prefer that option.

However, maybe @tobiasjacobs is the one that could provide a better opinion here, as the supporter of the use case of metadata for registration is him :)

jmcanterafonseca:
@fgalan I've solved the metadata issue by aligning with attribute metadata. I think this now ready to land modulo your comments ;)

thanks

fgalan:
LGTM

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1194,6 +1206,11 @@ Response:
                 ]
             },
             "callback":  "http://localhost:1234",
+            "metadata" : {
+                "provider": {
+                   "value": "example.com"
```
fgalan: `type: null` should be included.

(Sorry, this is a post-LGTM comment... I should have looked closer before casting LGTM... :( )

jmcanterafonseca: at creation time, it is not necessary, isn't it?

fgalan: Correct. I miss that detail... NTC (LGTM stills valid! :)

PR #1513, opened 2015-11-18
Hardening/1368 cpp check issues in lmlib

kzangeli:
Fixed a few leaks that valgrind didn't see as the program never really gets there ...
All thanks to cppcheck.


fgalan:
LGTM

Issue #1368, opened 2015-10-15
Hardening with lint/cppcheck


kzangeli:
We should go over the entire source code with a static analyzer, such as lint or cppcheck.
Lots of potential bugs can be avoided this way.

**Effort: 15 man day**

fgalan:
Also include de Sonar alert in the mix ;)

fgalan:
Partially addressed in PR https://github.com/telefonicaid/fiware-orion/pull/1514

fgalan:
Partially addressed in PR https://github.com/telefonicaid/fiware-orion/pull/1515

fgalan:
Partially addressed in PR https://github.com/telefonicaid/fiware-orion/pull/1521

crbrox:
Command line used  currently
```
cppcheck --quiet --enable=warning,performance,portability  --inline-suppr -Isrc/lib/ -Isrc/app -DLM_ON  src/
```
desde el directorio raíz de Orion

fgalan:
(Should -> Nice2Have, as only jenkins-stuff and minor fixes remain, as far as I know).

fgalan:
Most of the code has been "cleaned". However the follwoing libreries remain:

* logMsg
* parseArg

fgalan:
Just for fun, executing cpplink at develop (at commit 9f5490f3008d4161379afe155741f4077578676f, i.e. just before releasing 0.26.0):

```
fermin@neodeb:~/src/fiware-orion$ cppcheck --quiet --enable=warning,performance,portability  --inline-suppr -Isrc/lib/ -Isrc/app -DLM_ON  src/
[src/app/contextBroker/contextBroker.cpp:1513]: (portability) scanf without field width limits can crash with huge input data on some versions of libc.
[src/lib/cache/subCache.cpp:66]: (performance) Variable 'entityId' is assigned in constructor body. Consider performing initialization in initialization list.
[src/lib/cache/subCache.cpp:67]: (performance) Variable 'entityType' is assigned in constructor body. Consider performing initialization in initialization list.
[src/lib/logMsg/logMsg.cpp:2729] -> [src/lib/logMsg/logMsg.cpp:2730]: (performance) Variable 'nada' is reassigned a value before the old one has been used.
[src/lib/logMsg/logMsg.cpp:2730] -> [src/lib/logMsg/logMsg.cpp:2731]: (performance) Variable 'nada' is reassigned a value before the old one has been used.
[...]

PR #1521, opened 2015-11-20
More cppcheck fixes

crbrox:
Related with https://github.com/telefonicaid/fiware-orion/issues/1368

### DIFF
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -95,7 +95,6 @@ bool isIPv6(const std::string& in)
 {
   size_t      pos;
   std::string partip;
-  std::string resu;
   std::string staux = in;
   int         cont  = 0;
 
@@ -104,7 +103,6 @@ bool isIPv6(const std::string& in)
   {
     cont++;
     partip = staux.substr(0, pos+1);
-    resu  += partip;
 
     if (checkGroupIPv6(partip) == false)
     {
--- a/src/lib/logMsg/logMsg.h
+++ b/src/lib/logMsg/logMsg.h
@@ -1760,7 +1760,7 @@ inline void LM_TRANSACTION_START(const char *keyword, const char *ip, int port, 
 *
 * LM_TRANSACTION_START_URL -
 */
-inline void LM_TRANSACTION_START_URL(const char *url)                                                                                  \
+inline void LM_TRANSACTION_START_URL(const char *url)
 {
   transactionIdSet();
   LM_I(("Starting transaction from %s", url));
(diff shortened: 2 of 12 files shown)


src/lib/logMsg/logMsg.h
```diff
@@ -1775,11 +1772,10 @@ do                                                              \
 *
 * LM_TRANSACTION_END -
 */
-#define LM_TRANSACTION_END()                              \
-do                                                        \
-{                                                         \
+inline void LM_TRANSACTION_END()                                                                                     \
+{
   LM_I(("Transaction ended"));                            \
   LM_TRANSACTION_RESET();                                 \
```
kzangeli: Here you forgot the backslashes at the end of the line

src/lib/rest/RestService.cpp
```diff
@@ -442,8 +442,7 @@ std::string restService(ConnectionInfo* ciP, RestService* serviceV)
     // A tenant string must not be longer than 50 characters and may only contain
     // underscores and alphanumeric characters.
     //
-    std::string result;
-    if ((ciP->tenant != "") && ((result = tenantCheck(ciP->tenant)) != "OK"))
+    if ((ciP->tenant != "") && (tenantCheck(ciP->tenant)) != "OK")
     {
       OrionError  error(SccBadRequest,
                         "tenant name not accepted - a tenant string must not be longer than " MAX_TENANT_NAME_LEN_STRING " characters"
```
kzangeli: This here seems like a mistake.
"result" should have been used, and not a fixed string (as detail for OrionError).
I just looked at tenantCheck() and (as I suspected) it has two ways of failing (invalid char + name too long).
Let's use this knowledge and fix this error, returning the correct error string (result).

src/lib/logMsg/logMsg.h
kzangeli: Ah, and also, please put in 'void' inside '()'   (undocumented style guide)

crbrox: Fixed eb662e3

src/lib/common/string.cpp
```diff
@@ -95,7 +95,6 @@ bool isIPv6(const std::string& in)
 {
   size_t      pos;
   std::string partip;
-  std::string resu;
```
fgalan: I understand that this is a case of variable not used, detected by cppcheck.

The code is supposed to be compiled with gcc flags to fail with error when variable not used is detected. I wonder why the compiler has not complaint about this...

Anyway, NTC, good catches.

kzangeli: Look a little further down.  It is added to and after that not used.
cppcheck seems a bit smarter than the compiler :-)

fgalan:
LGTM

src/lib/common/string.cpp
fgalan: > cppcheck seems a bit smarter than the compiler :-)

I'm scared... xD

kzangeli:
LGTM

PR #1524, opened 2015-11-20
ADD Location specification to the FIWARE-NGSIv2

jmcanterafonseca:


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,96 @@ Error list (HTTP response code in paranthesis):
...
+
+* *Simple Location Format*. It is meant as a very lightweight format that developers and users can quickly
+and easily add to their existing entities with little effort.
+
+* *GeoJSON*.  [GeoJSON](https://tools.ietf.org/html/draft-butler-geojson-06)
+is a geospatial data interchange format based on the JavaScript Object Notation (JSON). GeoJSON provides greater flexibility
+allowing the representation of complex geospatial features.
+
+Additionally, geospatial properties might be declared by means of [civic addresses](https://schema.org/PostalAddress),
+but in that particular case entity providers will have to implement their own geocoding procedures,
```
fgalan: "entity providers" -> "context producers managingt these entities"

("entity providers" is not a well-known term in NGSI parlance; "context producer" is)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,96 @@ Error list (HTTP response code in paranthesis):
...
+
+It is noteworthy that the Simple Location Format is not intended to represent complex positions on Earth surface.
+For instance, those applications which require to capture altitude coordinates will have to use GeoJSON as
+representation format for the geospatial properties of their entities. 
+
+A context attribute  which represents a location encoded using the Simple Location Format
+must conform to the following syntax:
+
+* The attribute type must be one of the following values: (`geo:point`, `geo:line`, `geo:box` or `geo:polygon`).
+* The attribute value must be a list of coordinates, using the *WGS84* datum,
```
fgalan: As with others in this section (GeoRSS, GeoJSON, etc.), a reference could be include for the term "WS84 datum".

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,96 @@ Error list (HTTP response code in paranthesis):
...
+    "type": "geo:box"
+  }
+}
+```
+
+### GeoJSON
+
+A context attribute which represents a location encoded using GeoJSON must conform to the following syntax:
+
+* The attribute type must be the MIME type assigned by IETF to the GeoJSON format. Currently 'application/vnd.geo+json'.
```
fgalan: "atribute type" -> "attribute NGSI type" (to avoid the ambiguety that "type" in the case of attribute may refer to NGSI types or JSON native types).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,96 @@ Error list (HTTP response code in paranthesis):
...
+must conform to the following syntax:
+
+* The attribute type must be one of the following values: (`geo:point`, `geo:line`, `geo:box` or `geo:polygon`).
+* The attribute value must be a list of coordinates, using the *WGS84* datum,
+with latitude and longitude units of decimal degrees. Such coordinate list will allow to encode the geometry
+specified by the `type` attribute and will be represented according to the specific rules defined below:
+
+  * Type `geo:point`:   the attribute value must contain a string containing a valid latitude-longitude pair, separated by a comma. 
+  * Type `geo:line`:    the attribute value must contain a string array of valid latitude-longitude pairs. There must be at least two pairs.
+  * Type `geo:polygon`: the attribute value must contain a string array of valid latitude-longitude pairs.
```
fgalan: Typically, implementations will require that polygon vertices be ordered to avoid "crossings" (e.g. [0,0], [0,2], [2,0], [2, 2] involves a cross). Thus I would suggest to specify that the verices should be in an order that avoid this kind of problems.

On the contrary, there is no need to have the last point be the same that the first one. It is an information easy to "complete" in the case the implementation needs it (as in the case of Orion in NGSIv1, where the polygon specificaiton in queries doesn't need the last point be the same that the first one but MongoDB does need, so Orion "complete" the information before interacting with MongoDB).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -355,23 +433,26 @@ Geographical queries are specified by means of the `geometry` and `coords` param
...
++ `geometry=circle;radius:4000`, entities inside a circle of the given radius (radius unit must be meters).
++ `geometry=circle;radius:4000;external`, entities outside a circle of the given radius (radius unit must be meters).
++ `geometry=polygon`, entities inside a polygon.
++ `geometry=polygon;external`, entities outside a polygon.
++ `geometry=box`, entities inside a bounding box.
++ `geometry=box;external`, entities outside a bounding box. 
+
+**coords** represents a a list of geographical coordinates as mandated by the Simple Location Format
+and according to the geometry specified.
+In the particular case of the `circle` geometry the `coords` parameter must include just one coordinate pair,
```
fgalan: Appart from circle, it should be described how `coords` should be for the other `geometry=` cases.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -355,23 +433,26 @@ Geographical queries are specified by means of the `geometry` and `coords` param
...
+which is the center of the circle.
+
+*Implementations are responsible for determining the entity attributes which convey geospatial properties*. Typically
```
fgalan: In order to get a complete picture of the geo localation support, I think that we need also to close this point in the context of this PR.

I think the simplest approach is to use an entity attribute. However, I don't like too much the idea of hardwiring a name, as it may conflict with client application semantics (e.g. the application may want to use "location" to mean a string-based postal address). I thinks is more orthogonal to use metadata to "mark" the choosen attribute.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: Although I'm sympathetic to the "autocomplete" feature for polygons, it seems specs like GeoJSON make it explicit the fact that the first and last point have to be equivalent. See https://tools.ietf.org/html/draft-butler-geojson-06#section-2.1.6

On the other hand I think crossing are not relevant with regard to polygon definition in the sense that what is important is the enclosed area regardless crossings, isn't it?

fgalan: This section is about "simple location format" so we don't need to follow the no-so-simple GeoJSON way here.

"Crosing vertices" situations may result in error in some implementations, so I think is safer to make it explicit in the API definition.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: Yes of course, geospatial properties will be conveyed by an entity attribute. But I don't think we need metadata for that purpose. Actually metadata will be used for stating other important parameters such as accuracy. The NGSI type itself can be used to mark the attribute (geo:point, etc.) or GeoJSON  and to recognize that it refers to a geographical coordinate. 

But I think there can be more than one attribute that express a GeoLocation and in that particular case for query resolution it might be needed to indicate what attribute must be taken for resolving geo-queries. 

>> Another ongoing issue is if metadata properties of attributes might contain as well geospatial data.

What I was referring here is to the following use case:

```json
{
   id: "C1234",
   name: {
     value: "Madrid"
   } ,
   temperature: {
      value: 34,
      metadata: {
         location: {
            type: "geo:point",
            value: "44.5, 2.6"
         }
      }
  }
}
```
Is that something we want to support? The original OMA NGSI spec talks about this possibility, but I don't think it is relevant because the same info can be captured using other entities and a regular attribute. What do you think?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: No, because in the other cases we refer to the Simple Location Format. geo:circle is a special case as it is not under the list of possible values for location type. Nonetheless I will state it more clearly

fgalan: Ah... ok.. I misses the "...as mandated by the Simple Location Format".

Good enough by the moment (although in the Simple Location Format we are describing the content of an attribute and here we are describing the content of an URL parameter...)

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed bea6aa5e648b29e15c26afd1ba8d9ecfd314e4b9

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed bea6aa5e648b29e15c26afd1ba8d9ecfd314e4b9

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed bea6aa5e648b29e15c26afd1ba8d9ecfd314e4b9

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: yes that's why I've fixed that at bea6aa5e648b29e15c26afd1ba8d9ecfd314e4b9

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: oh I surrender then on the cross-vertex thing :) , fixed at a1ff4f3

Wrt the polygon "autocomplete", I don't think it is a good idea as we can generate interoperability problems as typically GIS libraries assume that polygons have that characteristic and I would like to take the data as it is from Orion and then use to paint polygons on a map for instance. On the other hand, ensuring first and last coordinate pair to be equivalent is good from sanity check of data perspective. Additionally, our Simple Location Format has been inspired by GeoRSS Simple and in that case four points are required as well. http://www.georss.org/simple.html

I hope these reasons suffice to convince you :)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -355,23 +439,31 @@ Geographical queries are specified by means of the `geometry` and `coords` param
...
++ `geometry=polygon`, entities inside a polygon.
++ `geometry=polygon;external`, entities outside a polygon.
++ `geometry=box`, entities inside a bounding box.
++ `geometry=box;external`, entities outside a bounding box. 
+
+**coords** must be a string containing a comma separated list of pairs of geographical coordinates
+in accordance with the geometry specified and the rules mandated by the Simple Location Format:
+
+* `geometry=point`.   `coords` will contain a pair of WGS-84 geo-coordinates.
+* `geometry=line`.    `coords` will contain a list of pairs of WGS-84 geo-coordinates.
```
fgalan: To be cohrenet, `geometry=line` should be included in the list in the geometry section.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Well enough. NTC.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,102 @@ Error list (HTTP response code in paranthesis):
...
+with latitude and longitude units of decimal degrees. Such coordinate list will allow to encode the geometry
+specified by the `type` attribute and will be encoded according with the specific rules defined below:
+
+  * Type `geo:point`:   the attribute value must contain a string containing a valid latitude-longitude pair, separated by a comma. 
+  * Type `geo:line`:    the attribute value must contain a string array of valid latitude-longitude pairs. There must be at least two pairs.
+  * Type `geo:polygon`: the attribute value must contain a string array of valid latitude-longitude pairs.
+There must be at least four pairs, with the last being identical to the first (so a polygon has a minimum of three actual points).
+Coordinate pairs must be properly ordered so that the line segments that compose the polygon remain on the ​outer ​edge of the area defined.
+For instance, the following path, ```[0,0], [0,2], [2,0], [2, 2]```, is an example of an invalid polygon definition. 
+Implementations must raise an error when none of the former conditions are met by input data.
```
fgalan: Maybe the "must" should be changed to "may". Some chekcing could be complex for implementation and have a peformance penalty, so not sure if we should force to implementations to incur in such costs.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: geo:point seems to be a good way to mark attributes that are location as alternativa to "marks based on metadata". How would be the query to express which attribute to use to resolve a geo-query in the case of ambiguety?

Regarding metadata, I agree it is not a relevant use case, so I would suggest to limit geo-location to attribute themselves.

Finally, I think we should introduce in the API (not sure how) the idea that a given implementation could limit the number of attributes that can express a location, due to implementation reasons (e.g. each attribute may consume resources in the form of a DB index, thus an excesive number of geo-location attributes may have a significative penalty).

jmcanterafonseca: have a look at https://github.com/jmcanterafonseca/fiware-orion/commit/0ca7691cd98adbf7ae82a2b5fdb9065bc47b38ef

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/0ca7691cd98adbf7ae82a2b5fdb9065bc47b38ef

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: have a look at https://github.com/jmcanterafonseca/fiware-orion/commit/0ca7691cd98adbf7ae82a2b5fdb9065bc47b38ef

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -321,19 +413,19 @@ The list of operators (and the format of the values they use) is as follows:
       be used with attributes which contain dates or numbers. E.g. `temperature!=10..20` matches if
       the temperature is lower than 10 or greater than 20 (it will not match if the tempeature is 10 or the
       temperature is 20).
-+ **Greater than**: `&gt;`. The value is a single element, e.g. `temperature&gt;40`. It
++ **Greater than**: `>`. The value is a single element, e.g. `temperature&gt;40`. It
```
iariasleon: the example stil is temperature&gt;40 instead of temperature>40

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,110 @@ Error list (HTTP response code in paranthesis):
...
+  * Type `geo:polygon`: the attribute value must contain a string array of valid latitude-longitude pairs.
+There must be at least four pairs, with the last being identical to the first (so a polygon has a minimum of three actual points).
+Coordinate pairs should be properly ordered so that the line segments that compose the polygon remain on the ​outer ​edge of the area defined.
+For instance, the following path, ```[0,0], [0,2], [2,0], [2, 2]```, is an example of an invalid polygon definition. 
+Implementations should raise an error when none of the former conditions are met by input data. 
+  * Type `geo:box`:     A bounding box is a rectangular region, often used to define the extents of a map or
+a rough area of interest. A box is represented by a two-length string array of latitude-longitude pairs.
+The first pair is the lower corner, the second is the upper corner.
+
+The examples below illustrate the referred syntax:
```
fgalan: Given that circle is not supported, it should be great to include an explination (inclusing external links) justifying the reason.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -351,27 +443,54 @@ Unary statements are composed of a unary operator (either `+`or `-`) and are use
...
+must be supported:
+
++ `geometry=point`, entities located at exactly that point on the Earth surface.
++ `geometry=line`, entities located along the polygonal line defined.
++ `geometry=circle;radius:4000`, entities inside a circle of the given radius (radius unit must be meters).
++ `geometry=circle;radius:4000;external`, entities outside a circle of the given radius (radius unit must be meters).
++ `geometry=polygon`, entities inside a polygon.
++ `geometry=polygon;external`, entities outside a polygon.
++ `geometry=box`, entities inside a bounding box.
++ `geometry=box;external`, entities outside a bounding box.
```
fgalan: A richer set of modifier could be considered for circle, polygon and box, in particular:

* intersect: to mean intersection between the geometry defined by the query and the geometry in the entity attribute
* inside (or maybe it is better "within"?): to mean that the geometry in the entity attribute is inside the area defined by the query
* outside (instead of external): to mean that the gemoetry in the entity attribute is outside the area defined by the query

In addition, due to geo spacial query capabilities could be highly depedent on particular implementtation, the NGSIv2 API should implement a standard mechanism so if a given implementation is not able to solve a particular query type, an error could be reported to the client.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,110 @@ Error list (HTTP response code in paranthesis):
...
+Additionally, geospatial properties might be declared by means of [civic addresses](https://schema.org/PostalAddress),
+but in that particular case context producers will have to implement their own geocoding procedures,
+so that geographical queries can be resolved.
+
+Implementations are responsible for determining which entity attributes convey geospatial properties (by inspecting
+the NGSI type associated to attributes). Typically it will be an entity attribute named `location`,
+but nothing prevents use cases where an entity contains more than one geospatial attribute. For instance,
+locations specified at different granularity levels or provided by different location methods with different accuracy. 
+Nonetheless, it is noteworthy the fact that spatial properties
+need special indexes which can be under resource constraints imposed by backend databases. Thus, implementations may raise errors
```
fgalan: Typo: excedeeded.

Review sentence: "Thus, implementations may raise are excedeeded". It seems that something is missing... (implementation may raise an error if the number of geospatial attributes are exceeded"?).



In addition, I guess you refer to an

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,110 @@ Error list (HTTP response code in paranthesis):
...
+when spatial index limits are excedeeded. The recommended HTTP error code for those situations is ``413``.
```
fgalan: 413 Request entity too large (the text associated to the code would make it more clear).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -282,6 +270,110 @@ Error list (HTTP response code in paranthesis):
...
+
+* *GeoJSON*.  [GeoJSON](https://tools.ietf.org/html/draft-butler-geojson-06)
+is a geospatial data interchange format based on the JavaScript Object Notation (JSON). GeoJSON provides greater flexibility
+allowing the representation of complex geospatial features.
+
+Additionally, geospatial properties might be declared by means of [civic addresses](https://schema.org/PostalAddress),
+but in that particular case context producers will have to implement their own geocoding procedures,
+so that geographical queries can be resolved.
+
+Implementations are responsible for determining which entity attributes convey geospatial properties (by inspecting
```
fgalan: I understand you refer to "client implementation"... is that correct?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -351,27 +443,54 @@ Unary statements are composed of a unary operator (either `+`or `-`) and are use
...
+
+* `geometry=point`.   `coords` will contain a pair of WGS-84 geo-coordinates.
+* `geometry=line`.    `coords` will contain a list of pairs of WGS-84 geo-coordinates.
+* `geometry=polygon`. `coords` will be composed by at least four pairs of WGS-84 geo-coordinates.
+* `geometry=box`.     `coords` will be composed by two pairs of WGS-84 geo-coordinates. 
+* `geometry=circle`.  `coords`  must include just one coordinate pair,
+which will correspond to the center of the circle.
+
+When resolving geographical queries, through the Simple Query Language,
+implementations are responsible for determining which entity attribute
```
fgalan: server implementation?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/e9b8d2143b4d2008c99c25c53f3e50791b50c2de

wrt to capabilities to the API, I would disagree, due to the fact that introducing optional capabilities is an interoperability hole. Provided that both open source databases (Mongo, PostGIS, etc.) and commercial products support geo-queries as those specified here I think we would not need to introduce something new

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I have clarified it a bit at https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I have clarified it at https://github.com/jmcanterafonseca/fiware-orion/commit/e396fd3c6f10ad85e68dcdc0481dfbfd3624c397

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -321,6 +309,119 @@ Error list (HTTP response code in paranthesis):
...
+Two different syntaxes must be supported by compliant implementations: 
+
+* *Simple Location Format*. It is meant as a very lightweight format that developers and users can quickly
+and easily add to their existing entities with little effort.
+
+* *GeoJSON*.  [GeoJSON](https://tools.ietf.org/html/draft-butler-geojson-06)
+is a geospatial data interchange format based on the JavaScript Object Notation (JSON). GeoJSON provides greater flexibility
+allowing the representation of complex geospatial features.
+
+Additionally, geospatial properties might be declared by means of [civic addresses](https://schema.org/PostalAddress),
```
fgalan: Maybe this paragraph should be removed... It is not actually adding any "normative" aspect to the specification and its vageness could be confusing.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -321,6 +309,119 @@ Error list (HTTP response code in paranthesis):
...
+need special indexes which can be under resource constraints imposed by backend databases.
+Thus, implementations may raise errors when spatial index limits are exceeded.
+The recommended HTTP error code for those situations is ``413`` (Request entity too large).
+
+### Simple Location Format
+
+Simple Location Format supports basic geometries ( *point* , *line* , *box* , *polygon* ) and covers the typical
+use cases when encoding geographical locations. It has been inspired by [GeoRSS Simple](http://www.georss.org/simple.html).
+
+It is noteworthy that the Simple Location Format is not intended to represent complex positions on Earth surface.
```
fgalan: I think there would be many use cases willing to use simple location format to represent locations in Earth surface and this sentence could be discouraging for them. If the only problem is the altitude, I'd suggest to state it clearly (instead of speaking about "complex position" in general without define precisely what we understand by "complex" or, in other words, what is "simple" and what is "complex" in this context).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -321,6 +309,119 @@ Error list (HTTP response code in paranthesis):
...
+    "type": "geo:box"
+  }
+}
+```
+
+### GeoJSON
+
+A context attribute which represents a location encoded using GeoJSON must conform to the following syntax:
+
+* The NGSI type of the attribute must be the MIME type assigned by IETF to the GeoJSON format.
```
fgalan: A bit off topic in this PR, but in the context of the pararell discussion about allowed character sets, this means that "/" should be included in the set of supported characters in order to implement this functionality.

Let's take this into account in that discussion.

NTC (with regards to this PR).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Note that the line has two occurrences of the `&gt;` token. The first one has been fixed, but not the second (same in 3 other lines, below).

marcc-orange:
This proposition adds two formats to NGSI v2: a simple one and GeoJSON representation.

Do you plan on having the `options` parameters accept `geo:simple` or `geo:json` values to specify if responses should return either Simple or GeoJSON representation ? Or are they two "distinct" (incompatible) geospatial representations (meaning that the return format is the same as the submitted one) ?

Why not just use GeoJSON to replace the simple format: The GeoJSON types (`Point`, `Polygon`, ...) would be defined as special attribute types like the `date` type.
In this case, simple format would only be used in geo queries and GeoJSON for updates and responses.

Actual simple format (compact):
```
{
 "location":  "41.3763726, 2.186447514"
}
```
Proposed GeoJSON (compact):
```
{
  "location": [2.186447514, 41.3763726]
}
```
Actual simple format (canonical):
```
{
  "location": {
    "type": "geo:point"
    "value": "41.3763726, 2.186447514",
  }
}
```
Proposed GeoJSON (canonical):
```
{
  "location": {
    "type": "Point",
    "coordinates": [2.186447514, 41.3763726]
  }
}
```
This does not look like GeoJSON is more complex thant the simple format in this way.

Of course, this would force the "canonical" entity representation to use `coordinates` instead of `value` ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: can we keep it but mark it as an aside note?

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: yup :)

jmcanterafonseca:
@marcc-orange excellent comments.

Wrt having an option for forcing GeoJSON or Simple rendering, +1. ```options=geojson``` or ```options=geosimple``` could just work. If nothing specified I think we can default to one of them, but we should definitely avoid rendering entities using different formats, otherwise it would be a processing nightmare for client apps. 

Wrt dropping out the Simple Location Format, it is certainly an option but I would like to draw your attention that, such a format opens the door to different Coordinate Reference Systems. Unfortunately GeoJSON constraints us to EPSG:4326 (WGS84) (as per the last draft of IETF's GeoJSON). I don't know if you are aware of the INSPIRE EU directive which recommends different CRS, based on the ETRS89 datum. 

On the other hand, please take into account that  a GeoJSON object will be the value of an attribute, that means the possible renderings are:

Default mode (formerly known as canonical)

```
{
    location: {
      value: {
         type: "Point"
         coordinates: [1.5, 2.5]
      }
      type: "application/vnd.geo+json"
      metadata : { }
   }
}
```

keyValues mode:  
```
{
   location: {
      type: "Point"
      coordinates: [1.5, 2.5]
   }
}
```

to sum up, a GeoJSON object will correspond to the value of an NGSI attribute, i.e. we cannot mix up the GeoJSON type with the NGSI attribute type.

fgalan:
Regarding

> Do you plan on having the options parameters accept geo:simple or geo:json values to specify if responses should return either Simple or GeoJSON representation ? Or are they two "distinct" (incompatible) geospatial representations (meaning that the return format is the same as the submitted one) ?

and 

> Wrt having an option for forcing GeoJSON or Simple rendering, +1. options=geojson or options=geosimple could just work. If nothing specified I think we can default to one of them, but we should definitely avoid rendering entities using different formats, otherwise it would be a processing nightmare for client apps. 

I don't think we need such options. At the end, locations are attribute values. Thus, the client requesting the attributes will get them in the same format they were created/updated (following the general rule for every attribute in NGSI).

jmcanterafonseca:
@fgalan @marcc-orange 

can we discuss the location attributes rendering in a separate issue? It would allow us not to defer anymore this PR and land it, making a nice progress for NGSIv2.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: yup 

fixed

https://github.com/jmcanterafonseca/fiware-orion/commit/739e317f6260960f0c134bbe52a7c739a972c6dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/739e317f6260960f0c134bbe52a7c739a972c6dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed https://github.com/jmcanterafonseca/fiware-orion/commit/739e317f6260960f0c134bbe52a7c739a972c6dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
iariasleon: LGTM

fgalan:
LGTM

marcc-orange:
@jmcanterafonseca :
> Wrt dropping out the Simple Location Format, it is certainly an option but I would like to draw your attention that, such a format opens the door to different Coordinate Reference Systems. Unfortunately GeoJSON constraints us to EPSG:4326 (WGS84) (as per the last draft of IETF's GeoJSON). I don't know if you are aware of the INSPIRE EU directive which recommends different CRS, based on the ETRS89 datum

FYI GeoJSON does support other CRS [by adding an optional `crs` attribute to objects](http://geojson.org/geojson-spec.html#coordinate-reference-system-objects), WGS84 is just the default CRS.

Having two _nearly identical_ geospatial formats (Simple Location Format (SLF) and GeoJson) does not simplify the use of the NGSIv2 protocol (moreover requiring NGSIv2 clients SDKs and servers implementations to support both). Why not just ditch GeoJSON in favor of SLF if it brings nothing more ?

Issue #1475, opened 2015-11-05
[Reducing DB interactions at mongoBackend] Use csubs cache also for not pattern subscriptions with just one lookup operation


fgalan:
Currently, we are doing 1 query operation in csub collection *for each attribute in the CE* to search for not pattern subscriptions (pattern subscriptions are already managed by the cache), which is inefficient specially in the case of updates involving many attributes.

So, two actions have to be done:

1. Include not pattern subscriptions in the cache (addTriggeredSubscription_withCache need to be modified)
2. Base the implementation in just one lookup in the cache instead of one lookup per attribute

fgalan:
Item 1 implemented in PR #1525

fgalan:
Item 2 implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1530

PR #1525, opened 2015-11-20
Feature/1475 all subs in cache

kzangeli:
All subscriptions now in subscription cache.

Fixes part of #1475 (the part of including non-patterned subs in sub-cache)


src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -68,12 +69,21 @@ using namespace mongo;
...
-  regcomp(&entityIdPattern, idPattern.c_str(), 0);
+  entityId     = _entityId;
+  entityType   = _entityType;
+  isPattern    = (_isPattern == "true") || (_isPattern == "TRUE") || (_isPattern == "True");
 
-  entityType               = _entityType;
-  entityIdPatternToBeFreed = true;
+  if (isPattern)
+  {
+    regcomp(&entityIdPattern, _entityId.c_str(), 0);
```
fgalan: Regular expression is compiled just one time, in the moment the cache entry (in the isPattern=true case) is inserted. Is that correct?

(It seems so, but given that this is an important point, I prefer to ask to be 100% sure)

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -253,9 +253,9 @@ Date: REGEX(.*)
     "measuring_interval_in_secs": "REGEX(\d+)",
     "requestSemaphoreWaitingTime": "REGEX(\d+.\d+)",
     "statisticsRequests": "1",
-    "subCache": "REGEX([0-9a-f, ]{50})",
-    "subCacheInserts": "2",
-    "subCacheItems": "2",
+    "subCache": "REGEX([0-9a-f, ]{76})",
```
fgalan: This at the end should be a list `[...]`, probably in a per-tenant JSON object. Not relevant for this PR, anyway.

NTC

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -299,10 +299,10 @@ Date: REGEX(.*)
     "requestSemaphoreWaitingTime": "REGEX(\d+.\d+)",
     "statisticsRequests": "2",
     "subCache": "REGEX([0-9a-f, ]{50})",
-    "subCacheInserts": "2",
+    "subCacheInserts": "3",
```
fgalan: I think the +1 in this test is because the SHELL included a isPattern=false subscription that, with the new changes, now it taken into account in the cache. Is that correct?

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, just at creation time of the EntityInfo, right here in its constructor

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
```diff
@@ -32,21 +32,21 @@ accumulatorStart
 
 #
 # 01. Create subscription S1 - without pattern
-# 02. Check statistics, see NO entries in subCache
+# 02. Check statistics, see ONE entry in subCache
```
fgalan: Again, I understand this +1 is due to a isPattern=false subscription already created in SHELL. Is that correct?

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
```diff
@@ -422,12 +424,12 @@ Date: REGEX(.*)
     "jsonRequests": "5",
     "measuring_interval_in_secs": "REGEX(\d+)",
     "statisticsRequests": "5",
-    "subCache": "REGEX([0-9a-f]{24})",
-    "subCacheInserts": "REGEX((1|2))",
-    "subCacheItems": "REGEX((1|2))",
+    "subCache": "REGEX([0-9a-f, ]{76})",
+    "subCacheInserts": "REGEX((5|8))",
+    "subCacheItems": "3",
```
fgalan: You have colapsed an OR (in the REGEX) to a particular value (maybe the same in other places). This maybe make valgrind suite to cry, but I understand that it is difficult to adjust at this point...

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: NTC

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: Yes, there are LOTS of changes in the statistics. Now ALL subs are in the cache, and counted.  Before only the patterned were there ...

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
kzangeli: Correct

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
kzangeli: This thing here is a little bit of debugging, I would go further and REMOVE if from here and implement it in GET /subCache or something similar.

But, this is for the future

test/functionalTest/cases/0117_convop_using_standard_ops/putSubscriptionConvOp.test
```diff
@@ -228,6 +228,9 @@ echo
 echo
 
 
+# Avoid throttling (1 second) for the subscription
+sleep 2
```
kzangeli: Without this sleep, the coming update (11. Update E1/T1/A1) is ignored due to throttling.
This is how it MUST work, but it didn't before this PR.
So, this is actually a bug-fix in throttling.
I forgot about this, perhaps we should add this information in the CNR.

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -345,10 +345,10 @@ Date: REGEX(.*)
     "requestSemaphoreWaitingTime": "REGEX(\d+.\d+)",
     "statisticsRequests": "3",
     "subCache": "REGEX([0-9a-f]{24})",
-    "subCacheInserts": "REGEX((2|3))",
+    "subCacheInserts": "REGEX((3|4))",
```
kzangeli: A note about this "funny line here":  "subCacheInserts": "REGEX((3|4))"

When under valgrind, a subCacheRefresh may occur before this is measured and as (unfortunately) a subCacheRefresh uses the same Remove and Insert functions (and subCacheUpdate also), the statistics counters change because of the refresh.

This should probably be fixed when we change the statistics.

test/functionalTest/cases/0117_convop_using_standard_ops/putSubscriptionConvOp.test
fgalan: I agree: if it is a bugfix it should be mentioned in CNR.

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
kzangeli: Yes, I know ...
I haven't even run trhis under valgrind yet.
There will be a few errors, without a doubt.
As I said, this PR is a little premature.
Will be fixed

test/functionalTest/cases/0117_convop_using_standard_ops/putSubscriptionConvOp.test
kzangeli: Fixed in 8a7f479

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
kzangeli: Fixed in 8a7f479

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: NTC

test/functionalTest/cases/1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
fgalan: NTC

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: NTC

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval_long_expiration.test
```diff
@@ -72,7 +71,7 @@ payload='<?xml version="1.0"?>
     <attribute>temperature</attribute>
   </attributeList>
   <reference>http://127.0.0.1:'${LISTENER_PORT}'/notify</reference>
-  <duration>P99Y</duration>
+  <duration>PT11S</duration>
```
fgalan: The idea of the "_long_expiration.test" is to use a very very long duration. In fact, note the comment in the .test file: "#Subscribe for a LONG period of time (99 years - should give overflow in an int)".

Is this change to 11S really needed?

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
```diff
@@ -31,7 +31,7 @@ brokerStart CB 101,38 IPv4 -reqTimeStat
 --SHELL--
 
 #
-# 01. Initial call to GET /statistics (gives NO time stats as LAST /statistics is excluded)
+# 01. Pre-call to GET /statistics to give VALGRIND a possibility
```
fgalan: What does mean "to give VALGRIND a possibility"? It should be clarified in the comment...

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
```diff
@@ -287,26 +284,6 @@ echo
...
-    "subCacheItems": "0",
-    "subCacheRefreshs": "1",
-    "subCacheRemoves": "0",
-    "subCacheUpdates": "0",
-    "uptime_in_secs": "REGEX(\d+)",
-    "versionRequests": "0"
-}
-
-
 02. Second call to GET /statistics (gives time stat of FIRST request for /statistics)
```
fgalan: It is a bit werid a "REGEXPECT" section starting by step 2: user may think that something is wrong and some step is missing. Maybe a solution is to do `echo "01. Pre-call to GET /statistics to give VALGRIND a possibility"` even although the actual step is not shown due it uses `/dev/null` output redirection.

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval_long_expiration.test
kzangeli: It was the only way I managed to have this test not failing under valgrind.
The timing under valgrind is completely off ...

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
kzangeli: ok, I'll put this explanation:

The valgrind test suite uses a call to "GET /version" to check that the broker is alive. This request is reflected in the statistics so the first call to "GET /statistics" gets the "timing" stuff (as there are non-zero counters) when under valgrind. When running as a normal functest, there is no prevoius "GET /version" and thus no timing counters.
Now, we have no means to do REGEXPs about missing lines so with this call to "GET /statistics", a first request is forced and the second call will give "timing" counters, both as normal functest and when under valgrind.

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
kzangeli: ok, I can renumber them to start from 00

fgalan: Actually, if that is a "preparation" step, maybe it makes more sense to include it in the SHELL-INIT part?

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
fgalan: If the problem is related with timing counters, what about to just disble timing counters avoding `-requestTimeStat` thus making the test more deterministic?

kzangeli: Either that or sending that "hidden" request.
Or even better, make valgrindTestSuite and testHarness.sh behave in the exact same way

I picked the fastest option which was to send a hidden request ...

Fixed in 28ee59a

If you think this is not enough, I'd propose to create an issue about it.

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval_long_expiration.test
fgalan: I understand that the valgrind problems use to happen due to things run slowly under valgrind. Thus, I don't understand why reducing duration make valgrind work (it should be the opposite, shouldn't it?)...

Anyway, valgrind "adaptations" shouldn't go agains the test case purpose and, in this case, the purpose is to test with very long subscriptions as it was before the modifications.

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
kzangeli: Fixed in 28ee59a (just renumber - still  in --SHELL--)

I don't have a problem in moving it to SHELL-INIT if you prefer that

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_ontimeinterval_long_expiration.test
kzangeli: Fixed in 4b13b92

fgalan:
LGTM

PR #1535, opened 2015-11-24
Hardening/more cppcheck

crbrox:
@fgalan 
@kzangeli

### DIFF
--- a/src/lib/ngsi/ParseData.h
+++ b/src/lib/ngsi/ParseData.h
@@ -63,6 +63,7 @@
 */
 typedef struct RegisterContextData
 {
+  RegisterContextData(): crP(0), entityIdP(0), attributeP(0), attributeMetadataP(0), registrationMetadataP(0) {}
   RegisterContextRequest         res;
   ContextRegistration*           crP;
   EntityId*                      entityIdP;
@@ -88,12 +89,13 @@ typedef struct RegisterContextResponseData
 *
 * DiscoverContextAvailabilityData -
 */
-typedef struct DiscoverContextAvailabilityData
+struct DiscoverContextAvailabilityData
 {
+  DiscoverContextAvailabilityData(): entityIdP(0), scopeP(0) {}
   DiscoverContextAvailabilityRequest  res;
   EntityId*                           entityIdP;
   Scope*                              scopeP;
-} DiscoverContextAvailabilityData;
+};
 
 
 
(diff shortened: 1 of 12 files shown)


src/lib/ngsi/ParseData.h
```diff
@@ -237,15 +246,16 @@ typedef struct UpdateContextAvailabilitySubscriptionData
...
 {
+  UpdateContextData(): ceP(0), entityIdP(0), attributeP(0), contextMetadataP(0), domainMetadataP(0) {}
   UpdateContextRequest   res;
   ContextElement*        ceP;
   EntityId*              entityIdP;
   ContextAttribute*      attributeP;
   Metadata*              contextMetadataP;
   Metadata*              domainMetadataP;
-} UpdateContextData;
+};
```
kzangeli: Interesting with all these removal of typedefs here ...
These types are used later in the "mega struct" of ParseData.

I don't know if the compilers have changed, but in the old times, without using 'typedef', e.g.

    typedef struct XXX
    {
        ...
    } XXX;

you couldn't use the typename (without the "struct" keyword).
Has this changed in newer versions of gcc/"C compilers in general"?

src/lib/ngsi/ParseData.h
```diff
@@ -253,14 +263,15 @@ typedef struct UpdateContextData
 *
 * UpdateContextResponseData - 
 */
-typedef struct UpdateContextResponseData
+struct UpdateContextResponseData
 {
+  UpdateContextResponseData(): cerP(0), attributeP(0), metadataP(0), domainMetadataP(0) {}
```
kzangeli: Not that is changes anything (and it yields longer lines), but using NULL instead of 0 might be clearer.
I'm not sure what I prefer even myself ...

fgalan: I do prefer NULL, no make clear that the attribute referes to a point instead that to an `int`. However, not sure if NULL can be used in the initialization list...

kzangeli: NULL is 0, it's just a matter of a typecast.

    #define NULL ((void*) 0)

Sometimes char* is used.

The only problem with NULL is that the lines get longer.

fgalan: Yes, all we know that NULL is 0 for the machine :) That's not the point.

The point is that *humans*, when read `param(NULL)`, tend to think "oh! that param is a pointer and it is initialized to NULL" but when read `param(0)` tend to think "oh! that param is an integer and it is initialialized to 0". Of course, nothing precludes to use this is a different way (i.e. you could use 0 in a pointer param). In this sense, this is  a kind of documentation hint, but useful anyway.

crbrox: Fixed in 4175a0f

fgalan:
LGTM

kzangeli:
LGTM

Issue #1476, opened 2015-11-05
[Reducing DB interactions at mongoBackend] Avoid update csub when notifications are sent


fgalan:
Currently, each time a notification is sent, the csub is updated to increase `count` and set the new `lastNotification`. This is inefficent.

A better approach should be implemented. A couple of ideas:

* Not using at all `count` or `lastNotification` in DB (although it could be a problem considering the current NGSIv2 JSON for subscriptions, which takes into account both fields)
* Accumulate all the changes (maybe using the cache structures?) and consolidate in DB in a regular basic (maybe at cache refresh frecuency?). The drawback is the risk of data lost and eventual consistency in the case of Active-Active configurations.

fgalan:
Considering option 2 (accumulate and cosolidate at refresh time) a solution is to use one update() for each field:

To update count, being `<n>` the amount to sum to DB:

```
update({ _id: ObjectId(<subId>)}, {count: {$inc: <n>}})
```

To update lastNotified, being <lnTime> the last notification time:

```
update({ _id: ObjectId(<subId>), lastNotification: {$lt: <lnTime>}}, {lastNotification: <lnTime>}})
```

(Not sure if the above is 100% correct syntax)

Actually, the above is expressed in mongo query language. In our code, the collectionUpdate() function has to be used. The first argument of the update() above should be mapped to a BSONObj in the `q` argument, while the second argument should be mapped to a BSONObj in the `doc` argument. `upsert` should be set to false (it is assumed that the document in the csub collection already exists).

```
bool collectionUpdate
(
  const std::string&  col,
  const BSONObj&      q,
  const BSONObj&      doc,
  bool                upsert,
  std::string*        err
)
```

Of course 2 update() is worse than 1 update() :). However, I cannot figure out how to do this with just one operation. Maybe an important optimization would be to use a bulk update to update all the cache documents: https://docs.mongodb.org/manual/reference/method/Bulk.find.updateOne/#Bulk.find.updateOne (not essential for #1476 completion, as we can assume that refresh wouldn't be too frecuent in production cases, i.e. in the order of tens of senconds; an issue should be opened about it).

As additional comment `-subCacheIval 0` is no longer a good idea, as count/lastNotification should be sync at DB from time to time (to prevent that a CB crash lost too much information). In addition, orionExit() function should execute the refresh part to sync count/lastNotification fields. 

[...]

PR #1536, opened 2015-11-25
Feature/1476 count and last notification not in db

kzangeli:
Implements #1476 

No longer maintaining subscription 'count' and 'lastNotificationTime' in DB, but in sub-cache.
DB is just used for synchronizing.

### DIFF
--- a/src/lib/mongoBackend/mongoSubCache.cpp
+++ b/src/lib/mongoBackend/mongoSubCache.cpp
@@ -62,6 +62,7 @@
 #include "ngsi10/SubscribeContextRequest.h"
 
 using namespace mongo;
+using std::map;
 
 
 
@@ -922,7 +923,9 @@ void mongoSubCacheItemInsert
   const char*               subscriptionId,
   int64_t                   expirationTime,
   int64_t                   throttling,
-  Format                    notifyFormat
+  Format                    notifyFormat,
+  bool                      notificationDone,
+  int64_t                   lastNotificationTime
 )
 {
   //
@@ -967,11 +970,11 @@ void mongoSubCacheItemInsert
   cSubP->reference             = strdup(scrP->reference.get().c_str());
   cSubP->expirationTime        = expirationTime;
   cSubP->throttling            = throttling;
-  cSubP->lastNotificationTime  = 0;
+  cSubP->lastNotificationTime  = lastNotificationTime;
   cSubP->pendingNotifications  = 0;
   cSubP->notifyFormat          = notifyFormat;
   cSubP->next                  = NULL;
-  cSubP->count                 = 0;
+  cSubP->count                 = (notificationDone == true)? 1 : 0;
 
   LM_T(LmtMongoSubCache, ("inserting a new sub in cache (%s). lastNotifictionTime: %lu", cSubP->subscriptionId, cSubP->lastNotificationTime));
 
@@ -1213,12 +1216,16 @@ static void mongoSubCacheRefresh(std::string database)
 *
 * mongoSubCacheRefresh - 
 */
-void mongoSubCacheRefresh(void)
+void mongoSubCacheRefresh(bool semAlreadyTaken)
 {
   std::vector<std::string> databases;
 
   LM_T(LmtMongoSubCache, ("Refreshing subscription cache"));
-  cacheSemTake(__FUNCTION__, "subscription cache for refresh");
+
+  if (!semAlreadyTaken)
+  {
+    cacheSemTake(__FUNCTION__, "subscription cache for refresh");
+  }
 
 
   // Empty the cache
@@ -1278,7 +1285,10 @@ void mongoSubCacheRefresh(void)
   //
 #endif
 
-  cacheSemGive(__FUNCTION__, "subscription cache for refresh");
+  if (!semAlreadyTaken)
+  {
+    cacheSemGive(__FUNCTION__, "subscription cache for refresh");
+  }
 }
 
 
@@ -1290,17 +1300,9 @@ void mongoSubCacheRefresh(void)
 typedef struct CachedSubSaved
 {
   std::string   subscriptionId;
-  int64_t       lastNotificationTime;
-  int64_t       count;
+  long long     lastNotificationTime;
+  long long     count;
 } CachedSubSaved;
-
-
-
-/* ****************************************************************************
-*
-* savedSubV - 
-*/
-static std::map<std::string, CachedSubSaved*> savedSubV;
 
 
 
@@ -1308,7 +1310,7 @@ static std::map<std::string, CachedSubSaved*> savedSubV;
 *
 * mongoSubCacheSync - 
 *
-* 1. Save subscriptionId, lastNotificationTime, and count for all items in cache
+* 1. Save subscriptionId, lastNotificationTime, and count for all items in cache (savedSubV)
 * 2. Refresh cache (count set to 0)
 * 3. Compare lastNotificationTime in savedSubV with the new cache-contents and:
 *    3.1 Update cache-items where 'saved lastNotificationTime' > 'cached lastNotificationTime'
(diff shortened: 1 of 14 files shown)


src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1275,7 +1279,150 @@ void mongoSubCacheRefresh(void)
...
+    cSubP = cSubP->next;
+  }
+
+  LM_T(LmtMongoCacheSync, ("Pushed back %d items to savedSubV", savedSubV.size()));
+
+
+  //
+  // 2. Refresh cache (count set to 0)
+  //
+  mongoSubCacheRefresh(true);
```
fgalan: Maybe the semAlreadyTaken parameter is not really needed, as I understand the only method that call mongoSubCacheRefresh() is mongoSubCacheSync() (except during initialization, which is a moment in which semaphores are not needed.

Thus, mongoSubCacheRefresh() could be simplified to not use this parameter. Regarding the effective, cacheSemTake/cacheSemGive, not sure if the best place is inside mongoSubCacheRefresh() or as a "ward" when invoking mongoSubCacheRefresh, i.e. something like this at this pooint of code:

```
cacheSemTake(...);
mongoSubCacheRefresh();
cacheSemGive(...);
```

Inspired in other places, we we do:

```
   reqSemTake(__FUNCTION__, "mongoSubCacheStatisticsGet", SemReadOp, &reqSemTaken);
   mongoSubCacheStatisticsGet(&mscRefreshs, &mscInserts, &mscRemoves, &mscUpdates, &cacheItems, listBuffer, sizeof(listBuffer));
   reqSemGive(__FUNCTION__, "mongoSubCacheStatisticsGet", reqSemTaken);
```

(In that place I think the sem is wrong, but the idea is the same).

fgalan: In addition, looking to the code in mongoSubCacheRefresh function I have seen this invocation:

```
mongoSubCacheRefresh(databases[ix]);
```

Being database a vector of std::string.

This seems to be here from a previous PR, so I wonder how it worked (as before this PR, signature parameters for mongoSubCacheRefresh function as ` void`).

Either I'm missing something or it seem to be a important bug...

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1275,7 +1279,150 @@ void mongoSubCacheRefresh(void)
...
+      // Update lastNotificationTime
+      update = BSON("$set" << BSON(CSUB_LASTNOTIFICATION << cssP->lastNotificationTime));
+    }
+    else
+    {
+      cSubP = cSubP->next;
+      continue;
+    }
+
+    if (collectionUpdate(collection, query, update, false, &err) != true)
```
fgalan: Curent update strategy has the problem that, if a lastNotification update is update at DB by a sibling CB and that lastNotification is greater than the one that this code knows, you will override the lastNotification DB with an older one. This is avoided if you use a "$gt" in the query part (in that case, you need to split the update into two ones: one for count and another for lastNotification).

That's is explained in detail at https://github.com/telefonicaid/fiware-orion/issues/1476#issuecomment-158386171. Please have a look and it that's not clear, let's discuss at skype.

test/functionalTest/cases/0000_ngsi10_convenience/subscriptions_onchange.test
```diff
@@ -29,7 +29,6 @@ brokerStart CB
 accumulatorStart
```
fgalan: I'd suggest to get a copy of the original subscriptions_onchange.test file (e.g. subscriptions_onchange_nocache.test) in order to ensure that the -noCache functionaly still working in the old way (subscriptions_onchange_nocache.test would need an adjustment in the brokerStart line to use the -noCache option, of course).

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_onchange.test
```diff
@@ -30,7 +30,6 @@ accumulatorStart
```
fgalan: Same XXX-nocache.test idea here.

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptions_test.cpp
```diff
@@ -435,9 +435,9 @@ TEST(mongoUpdateContext_withOnchangeSubscriptions, Cond1_updateMatch)
     EXPECT_EQ(SccOk, ms);
 
     /* Check lastNotification */
```
fgalan: I see that 35 test were modified in order to fit to the new way. However, maybe to have a sample of them (e.g. 5 tests?) with cache disabled (it should be a matter of adjusting a global/library variable) is a good idea, as a "watchdog" to ensure that old functionality still working.

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Yes, you are missing something. If it were as you say, the code wouldn't even compile ...
Function overloading.  There are **two** functions called mongoSubCacheRefresh.

fgalan: Ok, now makes sense:

```
static void mongoSubCacheRefresh(std::string database)
```

(BTW, `const std::string& database` should be use there).

kzangeli: Yes definitely (about const &).
I tend to forget std::string is an object, I see it as a "string" and to me, a string is a char* ... :-(
So difficult to remember!!!

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Just to make sure you've correctly understood the strategy:

1. lastNotification taken from DB
2. compare to what's in cache
3. lastNotification written back to DB if cache value > what was found in DB.

What you are saying is that somebody could have modified the DB between (1) and (3).
We are talking microseconds here. 
I agree there is a possibility some other broker could update the DB between (1) and (3)
A very small one.

About splitting the update, that is already done. I just remove the combined 'if'

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in bc7065c

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed in bc7065c

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptions_test.cpp
kzangeli: Please create an issue about this

fgalan: I'd do a different thing, to avoid excesive "issue-cracy" ;): I would add as a comment to #1476. #1476 will not be closed until the `-noCache` funcionality would be "in good shape".

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_onchange.test
fgalan: We have decided to defer this to a short future (see https://github.com/telefonicaid/fiware-orion/issues/1476#issuecomment-159562257). However at least a subscription_sequence_onchange_nocache.test.DISABLED with the original content should be created in this PR, to ease the future work (otherwise, we would need to do some archeology work in develop branch).

test/functionalTest/cases/0000_ngsi10_convenience/subscriptions_onchange.test
fgalan: We have decided to defer this to a short future (see https://github.com/telefonicaid/fiware-orion/issues/1476#issuecomment-159562257). However at least a subscriptions_onchange.test.DISABLED with the original content should be created in this PR, to ease the future work (otherwise, we would need to do some archeology work in develop branch).

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptions_test.cpp
fgalan: Commented at https://github.com/telefonicaid/fiware-orion/issues/1476#issuecomment-159562257. NTC

test/functionalTest/cases/0000_ngsi10_subscriptions/subscription_sequence_onchange.test
kzangeli: Fixed in 9e3f41f

test/functionalTest/cases/0000_ngsi10_convenience/subscriptions_onchange.test
kzangeli: Fixed in 79fd151

fgalan:
LGTM

PR #1544, opened 2015-11-26
Hardening/1476 no cache

kzangeli:
noCache working again. Unittests and functests to test it keeps working,

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1534,7 +1534,10 @@ static bool processSubscriptions
                               BSON(CSUB_LASTNOTIFICATION << rightNow) <<
                               "$inc" << BSON(CSUB_COUNT << 1));
         
-        collectionUpdate(getSubscribeContextCollectionName(tenant), query, update, false, err);
+        if (collectionUpdate(getSubscribeContextCollectionName(tenant), query, update, false, err) == false)
+        {
+          ret = false;
+        }
       }
 
 
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1522,6 +1522,22 @@ static bool processSubscriptions
                                                  tenant,
                                                  xauthToken))
     {
+      long long rightNow = getCurrentTime();
+
+      //
+      // If broker running without subscription cache, put lastNotificationTime and count in DB 
+      //
+      if (mongoSubCacheActive == false)
+      {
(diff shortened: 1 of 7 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1522,6 +1522,22 @@ static bool processSubscriptions
...
+      // If broker running without subscription cache, put lastNotificationTime and count in DB 
+      //
+      if (mongoSubCacheActive == false)
+      {
+        BSONObj query  = BSON("_id" << OID(mapSubId));
+        BSONObj update = BSON("$set" <<
+                              BSON(CSUB_LASTNOTIFICATION << rightNow) <<
+                              "$inc" << BSON(CSUB_COUNT << 1));
+        
+        collectionUpdate(getSubscribeContextCollectionName(tenant), query, update, false, err);
```
fgalan: Should the err be processed in some way? How the old version (some PRs ago) were dealing with count/lastNotification updates at DB?

fgalan:
Cannot comment inline (as the comment is for a file not in the diff):

Avoid using the same names in the two almost equal unit test files, e.g:

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: collectionUpdate has its own error reporting, I didn't feel the need to add an extra error here.

There is not much to do if this fails ...

The only thing I could possible do is to not update the cache if collectionUpdate fails.
But ... why "destroy" the cache just because collectionUpdate failed?

kzangeli:
35fad6f fixes the problem with repeated names in unittests

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Makes sense... Anyway, I'm curious as the is working in (for example) 0.25.0. Checking...

fgalan: Old code in processSubscriptions()

```
    {
     ...
      if (collectionUpdate(getSubscribeContextCollectionName(tenant), query, update, false, err))
      {
        // Happy path
        // ...
      else
      {
        ret = false;
      }
    }
```

Thus I think the `ret = false` part in the case of fail should be recovered.

kzangeli: Ah, make the function return false ...
Sure, that's not a problem. Not sure that anybody looks at the return value though ... :-)

kzangeli: Fixed in a3f89b8

fgalan:
LGTM

PR #1556, opened 2015-11-27
FIX dup attributes updates problem

### LINKED ISSUE #908: Detect duplicated attributes in updateContext request
(Raised talking with @frbattid)

From 0.17.0 attributes are identified by name + ID. However, Orion doesn't check that in the same updateContext request, the same attribute is not repeated several times (if that occurs, then one of the attributes is chosen and the other ignored).

It would be great that Orion could detect that situation at parsing time and return a 4xx Bad Request error explaining the problem if duplicated attributes are found.

**Effort: 2 man day**

fgalan:
Fixes #908

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2470,6 +2470,24 @@ static bool contextElementPreconditionsCheck
   /* Getting the entity in the request (helpful in other places) */
   EntityId* enP = &ceP->entityId;
 
+  /* Checking there aren't duplicate attributes */
+  for (unsigned int ix = 0; ix < ceP->contextAttributeVector.size(); ++ix)
+  {
+    std::string name = ceP->contextAttributeVector.get(ix)->name;
+    std::string id   = ceP->contextAttributeVector.get(ix)->getId();
+    for (unsigned int jx = ix + 1; jx < ceP->contextAttributeVector.size(); ++jx)
+    {
+      if ((name == ceP->contextAttributeVector.get(jx)->name) && (id == ceP->contextAttributeVector.get(jx)->getId()))
+      {
+        ContextAttribute* ca = new ContextAttribute(ceP->contextAttributeVector.get(ix));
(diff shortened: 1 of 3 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2470,6 +2470,24 @@ static bool contextElementPreconditionsCheck
   /* Getting the entity in the request (helpful in other places) */
   EntityId* enP = &ceP->entityId;
 
+  /* Checking there aren't duplicate attributes */
```
fgalan: At the end, the checking is done at contextElementPreconditionsCheck() before starting processing the CE. Another alternative would be to keep record of the attributes while they are inserted in the toSet and toPush structures at processAttributeVector() but it is much complex and without a significant gain.

The getId() search makes this check expensive, but there isn't any good way of fighting agains that... except remove the ID functionality istelf, of course ;)

NTC (just an explanation of the fix)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2470,6 +2470,24 @@ static bool contextElementPreconditionsCheck
...
+  for (unsigned int ix = 0; ix < ceP->contextAttributeVector.size(); ++ix)
+  {
+    std::string name = ceP->contextAttributeVector.get(ix)->name;
+    std::string id   = ceP->contextAttributeVector.get(ix)->getId();
+    for (unsigned int jx = ix + 1; jx < ceP->contextAttributeVector.size(); ++jx)
```
fgalan: Note the initial value in the loop, starting in the position of the attribute being compared for duplicates. I understand that this optimizes the loop (as we can save many items) as we have "passed" cases already evaluated for duplication, but a second opinion would be welcome.

kzangeli: Of course, this is how you do it:

0: compared to 1-max
1: compared to 2-max  (has already been compared to 0)
2: compared to 3-max (has already been compred to 0 and 1)
etc

NTC

kzangeli:
LGTM

Issue #1405, opened 2015-10-23
mongo sub cache to own library


kzangeli:
The sub cache currently resides completely in the mongoBackend library.
However, part of the implementation of the cache has nothing to do with mongo and should be moved from mongoBackend. Either to a new library (perhaps 'cache'?) or to some existing one.

Function from sub-cache to stay in mongoBackend:

    Perhaps only one function:
    o mongoSubCacheRefresh(std::string database) (or the part of the function that retrieves subs from mongo)

    Then we have two functions that use BSONObj, but no direct contact with mongo:
    o mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub)?
    o mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub, const char* subscriptionId, const char* servicePath)?

I would extract the part of mongoSubCacheRefresh that actually queries mongo and put that single function in mongoBackend, everything else out to a new 'cache' library.
Not to forget, EntityInfo could/should have its own header and source file.

**Effort: 2 man day**

fgalan:
I'd suggest to move the old code to archive/ in the case we want to preserve at 0.25.0 releasing time.

kzangeli:
sure, why not

fgalan:
Code moved in PR https://github.com/telefonicaid/fiware-orion/pull/1434. Good enough for 0.25.0 (this issue could be moved to the next milestone).

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1557

Issue #1555, opened 2015-11-27
csub cache logic must ignore orion-XXXX dbs in the case of not using -multiservice


fgalan:
As far as I understand, the refresh csub cache logic reads all database following the `orion[-XXX]?` pattern no matter if `-multiservice` is enabled or not.

The right behaviour should be ingore `orion-XXX` dbs in the case of not using `-multiservice`. Or, in other words, if `-multiservice` is not in use, then only pay attention to the `orion` database.

(Not sure how to label this, but I think it should be "bug").

kzangeli:
Sure, this can be considered a bug.

The 'cache logic' depends on getOrionDatabases() to get the list of databases.
Now, if getOrionDatabases() always returns the entire list, perhaps it needs a new parameter?

     bool getOrionDatabases(std::vector<std::string>& dbs, bool allOrionDatabases) ?

Either that or the cache itself looks at the -multiservice flag and **not** call getOrionDatabases() if multi-service is off.

I don't have a strong opinion on how to solve the problem.

fgalan:
Is `getOrionDatabases()` getting into account `-multiservice`? In that case, maybe this is not an actual bug.

kzangeli:
The "problem" is that getOrionDatabases does not look at multiservice, it just returns all the databases (I just looked it up).

So, there are two solutions, either to make getOrionDatabases a little more intelligent (have it look at 'multiservice'), or the sub-cache logic stops calling getOrionDatabases() if multiservice is off

fgalan:
> So, there are two solutions, either to make getOrionDatabases a little more intelligent (have it look at 'multiservice'), or the sub-cache logic stops calling getOrionDatabases() if multiservice is off

Which solution was finally adopted in PR #1557 (the one that solves this issue as far as I understand)?

PR #1557, opened 2015-11-27
Hardening/1405 mongo sub cache to own library

kzangeli:
Fixes issue #1405 and #1555.
Subscription Cache to own library

Also 2 bugs fixed:
- fixed a 'bug' about putting subscriptions for tenants when tenants not is use (-multiservice flag is not ON).
- setting 'count' to ZERO after refreshing cache from DB


fgalan:
> fixed a 'bug' about putting subscriptions for tenants when tenants not is use (-multiservice flag is not ON).

You mean https://github.com/telefonicaid/fiware-orion/issues/1555? Or a different thing?

src/lib/cache/subCache.cpp
```diff
@@ -0,0 +1,1105 @@
...
+#else
+        snprintf(msg, sizeof(msg), "%s", cSubP->subscriptionId);
+#endif
+      //
+      // If "msg" and ", " has no room in the "list", then no list is shown.
+      // (the '+ 2' if for the string ", ")
+      //
+      if (strlen(msg) + 2 > bytesLeft)
+      {
+        snprintf(list, listSize, "too many subscriptions");
```
fgalan: Needs to mention this in CNR? I would depend if the item list functionality in cache statistics was already in 0.25.0...

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -1249,259 +414,43 @@ static void mongoSubCacheRefresh(const std::string& database)
...
-    else
-    {
-      cSubP = cSubP->next;
-      continue;
-    }
+    // Update lastNotificationTime
+    condition = BSON("_id" << OID(subId) << CSUB_LASTNOTIFICATION << BSON("$lt" << lastNotificationTime));
+    update    = BSON("$set" << BSON(CSUB_LASTNOTIFICATION << lastNotificationTime));
 
     if (collectionUpdate(collection, condition, update, false, &err) != true)
```
fgalan: I understand this PR fix also the problem we talk this morning and now we are a separate process (each one which its own collectionUpdate()) for count and lastNotification. Is that correct? (diff noise is hard in this part... :)

fgalan: (BTW I don't think this bugfix should be mentioned in CNR, as it was introduced during the 0.26.0 development cycle, i.e. it wasn't already at 0.25.0)

fgalan:
> setting 'count' to ZERO after refreshing cache from DB

This was bug present in 0.25.0 ?

src/lib/cache/subCache.cpp
fgalan: In addition, statistics.md should be adjusted to explain this, in the part which describes the `id` field in cache statistics.

src/lib/cache/subCache.cpp
```diff
@@ -0,0 +1,1105 @@
...
+  cSubP->tenant                = (tenant[0] == 0)? NULL : strdup(tenant);
+  cSubP->servicePath           = strdup(servicePath);
+  cSubP->subscriptionId        = strdup(subscriptionId);
+  cSubP->reference             = strdup(scrP->reference.get().c_str());
+  cSubP->expirationTime        = expirationTime;
+  cSubP->throttling            = throttling;
+  cSubP->lastNotificationTime  = lastNotificationTime;
+  cSubP->notifyFormat          = notifyFormat;
+  cSubP->next                  = NULL;
+  cSubP->count                 = (notificationDone == true)? 1 : 0;
```
kzangeli: This is a new item in the cache (after a subscription-creation). count is 1 or 0, depending on whether it provoked a notification or not

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -687,10 +101,10 @@ int mongoSubCacheItemInsert(const char* tenant, const BSONObj& sub)
   cSubP->throttling            = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLong(sub, CSUB_THROTTLING)       : -1;
   cSubP->expirationTime        = sub.hasField(CSUB_EXPIRATION)?       getIntOrLongFieldAsLong(sub, CSUB_EXPIRATION)       : 0;
   cSubP->lastNotificationTime  = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLong(sub, CSUB_LASTNOTIFICATION) : -1;
-  cSubP->count                 = sub.hasField(CSUB_COUNT)?            getIntOrLongFieldAsLong(sub, CSUB_COUNT)            : 0;
+  cSubP->count                 = 0;
```
kzangeli: Here is where I set count to ZERO, as this function is called during synch.
This is not 100% and a bit error prone.
I should probably remove the reset of count from here and loop over all subs in the cache as the last thing that is done after sync (in subCacheSync) ...

Opinions?

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: I am not 100% sure about what you are saying about "running in a separate process".
I have made no such modification.
There is one call to collectionUpdate for 'count' and another call to collectionUpdate for 'lastNotification', but, they both run in the same process ...

fgalan: "running in a separate process" -> "running in a separate block, one after other"

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: `cSubP->count = 0` is executed just after the accumulated count has been flushed into DB and wihouth the oportunity other thread update the cache count (given that I understand refresh thread has not give the cache sem yet). Is that correct?

kzangeli: That is correct.

The problem I see it that we are "deep inside a function" when this *important* reset takes place.
It would be clearer and less error prone if we moved the entire 'count-reset', outside the re-creation of sub-cache-items.   Clearer and ess error-prone but also slower, of course, as we'd have to go through the entire list of subscriptions again ...

I just want your opinion on that.
Keep it where it is (a bit hidden but fast), or make it more visible but slower ...

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: ok, that is how it is now, two separate calls to collectionUpdate

NTC, I believe

src/lib/cache/subCache.cpp
kzangeli: Fixed in bd1fcfc

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Finally I opted for the 'execution-time faster option', making sure all insert-functions do what they should. Documented in the header of each function, see bd1fcfc

src/lib/cache/subCache.cpp
fgalan: I understand that notificationDone is true when the situation is "notification sent upon subscription creation". In that case, looks good.

NTC

fgalan:
LGTM

PR #1559, opened 2015-11-27
Hardening/refactor contextattrvec constref

crbrox:
@fgalan @kzangeli

### DIFF
--- a/src/lib/ngsi/ContextAttributeVector.cpp
+++ b/src/lib/ngsi/ContextAttributeVector.cpp
@@ -311,13 +311,19 @@ ContextAttribute* ContextAttributeVector::get(unsigned int ix)
   return NULL;
 }
 
+const ContextAttribute* ContextAttributeVector::get(unsigned int ix) const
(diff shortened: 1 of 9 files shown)


fgalan:
LGTM

src/lib/ngsi/ContextAttributeVector.cpp
```diff
@@ -311,13 +311,19 @@ ContextAttribute* ContextAttributeVector::get(unsigned int ix)
   return NULL;
 }
 
+const ContextAttribute* ContextAttributeVector::get(unsigned int ix) const
+{
+  if (ix < vec.size())
+    return vec[ix];
+  return NULL;
+}
```
kzangeli: Instead of adding get() methods, I would remove them, and use only the operator[] instead.
The get() methods are old, we started to use operator[] instead a long time ago.

Perhaps, I've missed something about this ...

kzangeli:
LGTM

Issue #1558, opened 2015-11-27
Erratic behaviour with some DB populations


fgalan:
We have found at the end of 0.26.0 development cycle (i.e. 0.26.0 release candidates are on the way and the version is almost closed) some problems that makes CB behave in a weird way with some DB populations. We will use this issue to summarize what we know and possible research lines or potential solutions.

*Facts*

When the DB is populated in an specific way (ask to @fgalan for orion-evil-dump.tgz dump) CB behaves in a weird way. The behaviour depends if `-subCacheIval` is set to 0 (i.e. only one refresh is done, at startup) or a different value (i.e. refresh are done in a periodic way).

In the case of no periodic refresh, the CB seems to be stable (we have seen it work without fail during days at orion.lab) although some weird things occur:

* "Fatal Error" messages related with DB may appear in the log
* The item count at /cache/statistics doens't match the one it should be at DB (which may mean that the cache content is not correct)

In the case of periodic refresh, the CB crashes after a while (using `-subCacheIval 5` it uses to crash in less than 20 min in orion.lab). The core backtrace information shows exceptions (most of the time of assertion type) related with MongoDB C++ driver methods.

*Theories*

We don't know the actual cause of the problem yet, we only have some theories that need to be validaded.

One possible theory is related with incorrect usage of cursors due to thread safeness (more details here: http://stackoverflow.com/questions/33945987/thread-safeness-at-mongodb-c-driver-regarding-indirect-connection-usage-throug). If a cursor is corrupted that would explain why assertion in MongoDB C++ methods (such the get*Field family or more()) fail. It can also explain why the item count in the cache doesn't match the actual one at DB.

How to validate this theory:

* Hack CB code so cache use the request semaphore (quickest way seems to be hacking the implementation of cacheSemTake/cacheSemGive functions, to wait also at the request semaphore, appart from the semaphore properly protecting cache access at RAM) 
* Run CB with `-reqMutexPolicy all`. That will force all cursor usage in non-concurrent mode (as there aren't cursor usages outise mongoBackend and the "all" req mutex policy ensures that as much one thread at a time is executing code in the mongoBackend).

More info to check: how IoTAgent C++ is using cursors:

* https://github.com/telefonicaid/fiware-IoTAgent-Cplusplus/blob/develop/src/util/collection.cc
* https://github.com/telefonicaid/fiware-IoTAgent-Cplusplus/blob/develop/src/util/collection.h

*Another things that we could test*

[...]

PR #1564, opened 2015-11-30
Avoid concurrent usage of connections due to cursors

fgalan:
Implements #1558


src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -226,10 +226,15 @@ static bool addTriggeredSubscriptions
...
-  if (!collectionQuery(getSubscribeContextAvailabilityCollectionName(tenant), query, &cursor, &err))
+  TIME_STAT_MONGO_READ_WAIT_START();
+  DBClientBase* connection = getMongoConnection();
+  if (!collectionQuery(connection, getSubscribeContextAvailabilityCollectionName(tenant), query, &cursor, &err))
   {
+    TIME_STAT_MONGO_READ_WAIT_STOP();
+    releaseMongoConnection(connection);
     return false;
   }
+  TIME_STAT_MONGO_READ_WAIT_STOP();
```
fgalan: I have some doubts regarding the right moment to do the TIME_STAT_MONGO_READ_WAIT_STOP()

1. Just after the collectionQuery (as it is now)
2. After processing cursor results (i.e. it would be just after releaseMongoConnection() in line 277)

I preferred alternative 1, as alternative 2 is highly variable on the particular cursor processing, thus if we do in that way, then the mongoWait metric will be "mixed" with mongoBackend processing.

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -269,6 +274,7 @@ static bool addTriggeredSubscriptions
       subs.insert(std::pair<string, TriggeredSubscription*>(subIdStr, trigs));
     }
   }
+  releaseMongoConnection(connection);
```
fgalan: At this point, cursor has been exahusted and I undersand that is safe to release connection. Not fully sure... maybe kill() should be called in the cursor as an extra security measure (asked for clarifications in the comments to the answer in http://stackoverflow.com/questions/33945987/thread-safeness-at-mongodb-c-driver-regarding-indirect-connection-usage-throug/33977828#33977828)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2454,6 +2460,285 @@ static bool forwardsPending(UpdateContextResponse* upcrsP)
   return false;
 }
 
+/* ****************************************************************************
+*
+* updateEntity -
+*/
+static void updateEntity
```
fgalan: This big bunch of code was before in processContextElement() function. Thus, a side advantage of this PR is to get a better modularization.

NTC (just informative)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2645,12 +2936,8 @@ void processContextElement
...
   // FIXME P6: Once we allow for ServicePath to be modified, this loop must be looked at.
   //
-  int docs = 0;
-
-  // Used to accumulate error response information, checked at the end
-  bool         attributeAlreadyExistsError = false;
-  std::string  attributeAlreadyExistsList  = "[ ";
 
+  std::vector<BSONObj> results;
   while (moreSafe(cursor))
```
fgalan: In this case, element processing within while() was complex (involving a subsequent operation at mongo, i.e. an update()) so I addopted a two-phase approach:

1. In a first phase, all the results are accumulated in a vector of BSONObj, freeing connection after that.
2. The vector elements are processed in a loop.

I'm not sure if this two-phase approach should be adopted in all the cases... My concern is that the results of the cursor may be very long (in this particular case, they aren't: an update is not supposed to match many entities), so a two-phase apporach could be ineficient (i.e two passes and an the memory footprint of the intermediate BSONObj vector).

Opinions are welcome!

fgalan:
I have compiled an RPM on this branch and install in orion.lab. Now, the cache vs. DB count seems to be ok:

```
> db.csubs.count({"conditions.type": "ONCHANGE"})
227
> ^C
bye
[root@orion ~]# curl --header 'Accept: application/json' -s -S localhost:$CB_PORT/cache/statistics | python -mjson.tool
{
    "ids": "too many subscriptions",
    "inserts": 8172,
    "items": 227,
    "refresh": 36,
    "removes": 0,
    "updates": 0
}
```

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: Anything we can do to 'reset' the cursor
If there is anything more that we can do to clean the cursor for the next user, let's do it ...

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: This all depends on that we want to measure with MONGO_READ_WAIT.
If it is the time it takes to get a connection, then STOP should be right after the call to getMongoConnection.
I am not 100% sure on what we want to measure, so,

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2454,6 +2460,285 @@ static bool forwardsPending(UpdateContextResponse* upcrsP)
...
+  const std::string  idString          = "_id." ENT_ENTITY_ID;
+  const std::string  typeString        = "_id." ENT_ENTITY_TYPE;
+  const std::string  servicePathString = "_id." ENT_SERVICE_PATH;
+
+  BSONElement idField = getField(r, "_id");
+
+  std::string entityId    = getStringField(idField.embeddedObject(), ENT_ENTITY_ID);
+  std::string entityType  = getStringField(idField.embeddedObject(), ENT_ENTITY_TYPE);
+  std::string entitySPath = getStringField(idField.embeddedObject(), ENT_SERVICE_PATH);
+
```
kzangeli: align vars and '='s

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: So, you invented this 'two phased op' to not holding the connection for too long, right?
To me it seems like a good idea; connections will be held longer now, that we know about the problem with cursors.

About using this approach in all cases ... Yes, why not? But, as a separate issue about throughput. It's all about minimizing the time to wait for a mongo connection, right?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Always good to get smaller functions, however, the PR review gets more tricky, at least if there are changes inside the moved code.  Are there any changes in updateEntity vs what we had before?

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Indeed, not sure if that is fully right... As far as I see, there is a trade-off.

On the one side, the two-phase approach will release before the connection, so it would minimize the time waiting at it. On the other side, two-phase approach implies that some work is done twice versus the one-phase approach, which do all the work in the same pass.

Let's consider the following case: a queryContext that results in 1,000 elements (the maximum according to hard pagination limit). The one phase approach works in this way:

1. Take sem
2. Process 1,000 results from cursor
3. Release sem

The two-phase appraoch works this way:

1. Taken sem
2. Create a 1,000 BSONObj vector
3. Release sem
4. Process 1,000 results from vector

What would be the time from 1 to 3 in both cases? How compares the time in one-phase step 2 versus the sum of the times in two-phases 2 and 4? What about the impact of the 1,000 BSONObjt vector structure (x N, given that N threads at a time could be in step 4) compared with the cursor, which doesn't all the results at a time but gets them from MongoDB batch by batch throught the wiring protocol?

However, leaving performance appart. There is an good reason to use the two-phase approach: it would allow a thread-safe interface for all colletionQuerry() invokations. I mean, we could use the following alternative for collectionQuery():

```
extern bool collectionQuery
(
  const std::string&              col,
  const BSONObj&                  q,
  std::vector<BSONObj>*     results,
  std::string*                    err
);
```

and the connectionQuery() caller wouldn't need to deal with getMongoConnection()/releaseMongoConnection() or TIME_STAT_MONGO_READ_WAIT_START/STOP. In fact, the caller of that hypotecial collectionQuery() function wouldn't ever see the cursor. That's would lead to cleaner and safer code.

However, given that 0.26.0 closing is very near and that a decission on this needs to be analyzed carefully (there are many aspect to consider) I think is a good idea to spin off to a future issue (basically a copy-paste of this comment :)

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: I think that no cursor is reused in any place (i.e. the variable holding cursorr gets out of scope before that the cursor has the opportunity to be used again). However, I agree in do has much as possible to ensure this, even if that implies redundancy. Thus, I'd add a `cursor.kill()` statements just before releaseMongoConnection() calls.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: This block of three variables is aligned... Do you mean to align all from EntityId* to line 2494?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Sounds good to me. Close 0.26.0 asap and put this info in an issue, possibly for 0.27.0

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I was thinking about all eight of them, yes :-).
Just by adding an empty line (which perhaps should be avoided) doesn't mean that aligning vars stops being valid ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: No changes (except maybe variable declation reorder).

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: I have commented about this in stadistics.md (commit 5974d5f4658d64823329d6122c2a4a49fdfda636)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in 5974d5f4658d64823329d6122c2a4a49fdfda636

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: Fixed in 5974d5f4658d64823329d6122c2a4a49fdfda636... before I discovered that kill() is not avalabile in legacy-1.0.2, but in a newer driver version :(

Thus, I have prepared the code, so just removing a #if will make the code work with kill (see 2971d46538f8801f4468f2d8d0d288b872e6f261)

kzangeli:
LGTM

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: This is the issue just created: https://github.com/telefonicaid/fiware-orion/issues/1570

PR #1566, opened 2015-11-30
Refactor notifier-queue-workers. Exit leak

crbrox:
@fgalan @kzangeli

### DIFF
--- a/src/app/contextBroker/contextBroker.cpp
+++ b/src/app/contextBroker/contextBroker.cpp
@@ -1285,17 +1285,13 @@ static void contextBrokerInit(std::string dbPrefix, bool multitenant)
   /* If we use a queue for notifications, start worker threads */
   if (strcmp(notificationMode, "threadpool") == 0)
   {
-    LM_T(LmtNotifier,("Setting up queue and threads for notifications"));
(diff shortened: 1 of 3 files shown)


fgalan:
Manual test in a system in which previously the leak appeared, it seems to be solved:

```
fermin@neodeb:~/src/fiware-orion/test/valgrind$ ./valgrindTestSuite.sh -filter onchange_threadpool.test
lun nov 30 13:18:04 CET 2015
./valgrindTestSuite.sh: línea 218: cd: test/valgrind: No existe el fichero o el directorio
Test 001/1: 1449_notification_mode_cases/onchange_threadpool ........................................................................................ OK (12.83 seconds)
Total test time: 13.02 seconds (0 minutes)
```

Thus, LGTM

kzangeli:
An explanation would have helped ...
It seems QueueWorkers has been moved to reside inside QueueNotifier instead of outside.
Now, this is supposed to be a leak-fix. So,. where was the leak?  Where is the fix?
I guess avoiding to create a new QueueWorkers every time (and not deleting it) is the leak fix.

So, LGTM, I guess ...

fgalan:
I think (and checked with valgrind) that the leak was at (old code):

```
QueueWorkers*                       qws    = new QueueWorkers(pQueue, notificationThreadNum);
```

The memory at `qws` was never freed.

Issue #1552, opened 2015-11-27
Crash in GET /v2/subscription/<id> with ID not conforming OID syntax


fgalan:
ContextBroker generates a core, when receives a get single subscription NGSIv2 with inexistent subcriptionId

In stdout appears:

   terminate called after throwing an instance of 'mongo::AssertionException'
    what():  assertion src/mongo/bson/oid.cpp:133

ContextBroker was locally built with develop branch of 27 November, git_hash is:

```
[develop@localhost bin]$ curl 0.0.0.0:1026/version
<orion>
  <version>0.25.0-next</version>
  <uptime>0 d, 0 h, 0 m, 14 s</uptime>
  <git_hash>f5e62e5d26fdf96b63f44592f2fd321cf6ee924c</git_hash>
  <compile_time>Fri Nov 27 10:34:21 PST 2015</compile_time>
  <compiled_by>root</compiled_by>
  <compiled_in>localhost.localdomain</compiled_in>
</orion>
```

Core can be generated for example with curl:

curl -v -X GET http://0.0.0.0:1026/v2/subscriptions/inexistentid -H "Fiware-Service:service2" -H "Fiware-ServicePath:/subservice2"

Core stack is:

(gdb) where
#0  0x0000003ce2232625 in raise () from /lib64/libc.so.6
#1  0x0000003ce2233e05 in abort () from /lib64/libc.so.6
#2  0x0000003ceaabea7d in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib64/libstdc++.so.6
#3  0x0000003ceaabcbd6 in ?? () from /usr/lib64/libstdc++.so.6
#4  0x0000003ceaabcc03 in std::terminate() () from /usr/lib64/libstdc++.so.6
#5  0x0000003ceaabcd22 in __cxa_throw () from /usr/lib64/libstdc++.so.6
#6  0x00000000006b27ac in mongo::verifyFailed (expr=<value optimized out>, file=<value optimized out>, line=<value optimized out>)
    at src/mongo/util/assert_util.cpp:93
#7  0x000000000064a750 in mongo::OID::init (this=<value optimized out>, s=<value optimized out>) at src/mongo/bson/oid.cpp:133
#8  0x00000000005f45b9 in OID (sub=<value optimized out>, oe=0x7f0cd0aa49a0, idSub=<value optimized out>, uriParam=<value optimized out>, 
[...]

PR #1569, opened 2015-11-30
Check of Subscription ID before sending request to mongo layer

kzangeli:
Fixes issue #1552 

Checking subscription ID before letting it pass down to mongo layer, for GET "/v2/subscriptions/XXX"

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -22,4 +22,4 @@
 - Fix: no longer adding subscriptions from all tenants in the cache if broker isn't multitenant (Issue #1555)
 - Fix: resetting temporal counters at synching the subscription cache (Issue #1562)
 - Fix: using the text "too many sbuscriptions" at cache statistics operations in the case of too many results
-- Fix: crash due to subscription ID not conforming to supposed syntax (Issue #1552)
+- Fix: crash due to subscription ID not conforming to supposed syntax in request "GET /v2/subscriptions/XXX" (Issue #1552)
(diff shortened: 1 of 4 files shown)


src/lib/mongoBackend/mongoGetSubscriptions.cpp
```diff
@@ -260,14 +262,20 @@ void mongoGetSubscription
   const std::string&                  tenant
 )
 {
-  bool  reqSemTaken = false;
+  bool         reqSemTaken = false;
+  std::string  err;
+
+  if ((err = idCheck(idSub)) != "OK")
```
fgalan: I would suggest to use safeGetSubId() in the checking at mongoBackend layer due to two reasons:

1. Homogeneity (similar checkings in other parts of mongoBackend are using that method).
2. Using a different method to the one used at parsing stage. This kind of "HA" will mean that is some fail occur in the checking done by idCheck(), then safeGetSubId() could catch (or the opossite).

CHANGES_NEXT_RELEASE
```diff
@@ -22,3 +22,4 @@
 - Fix: no longer adding subscriptions from all tenants in the cache if broker isn't multitenant (Issue #1555)
 - Fix: resetting temporal counters at synching the subscription cache (Issue #1562)
 - Fix: using the text "too many sbuscriptions" at cache statistics operations in the case of too many results
+- Fix: crash due to subscription ID not conforming to supposed syntax (Issue #1552)
```
fgalan: - Fix: crash due to subscription ID not conforming to supposed syntax in GET /v2/subscriptions/<id> operation (Issue #1552)

The "in GET /v2/subscriptions/<id> operation" could be relevant given that, as far as I understand, this problem doesn't happen in similar operations of the NGSIv1 family.

kzangeli: a8c27f2

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: Fixed in a8c27f2

fgalan:
LGTM

Issue #1380, opened 2015-10-16
httpRequestSend return value


kzangeli:
The function httpRequestSend in rest/httpRequestSend.cpp returns a string, and if the value is "" it is an error. Also, if "error", it is an error, also ...

httpRequestSend should accept one more parameter and return the string there, while its return value should be 0 for OK and -X for any error

**Effort: 1 man day**

fgalan:
Fixed in PR #1576

PR #1576, opened 2015-12-01
Hardening/1380 http request send return value

kzangeli:
Changed the signature of httpRequestSend and httpRequestSendWithCurl, to return an integer and not a string. The string is now an output-parameter to the function instead of the return value.

Fixes issue #1380

### DIFF
--- a/test/functionalTest/cases/0880_timeout_for_forward_and_notifications/forward_that_times_out.test
+++ b/test/functionalTest/cases/0880_timeout_for_forward_and_notifications/forward_that_times_out.test
@@ -155,13 +155,14 @@ Date: REGEX(.*)
 02. Update/UPDATE E1/T1/A1 on CB (provoking forward to accumulator)
 ===================================================================
 HTTP/1.1 200 OK
-Content-Length: 94
+Content-Length: 147
 Content-Type: application/json
 Date: REGEX(.*)
 
 {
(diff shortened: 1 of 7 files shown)


test/functionalTest/cases/0880_timeout_for_forward_and_notifications/forward_that_times_out.test
```diff
@@ -155,13 +155,14 @@ Date: REGEX(.*)
...
-Content-Length: 94
+Content-Length: 147
 Content-Type: application/json
 Date: REGEX(.*)
 
 {
     "errorCode": {
-        "code": "404", 
+        "code": "404",
+        "details": "invalid context provider response",
```
fgalan: That means that now our errors are more descriptive in this case?

kzangeli: I got this as an extra, for 'filling in more info' for returning the text in httpRequestSend ...
Didn't bother to investigate why, as extra info doesn't really hurt

fgalan: Indeed, I think is an improvement.

NTC

fgalan:
LGTM

PR #1578, opened 2015-12-01
Hardening/refactor const reference

crbrox:


src/lib/ngsi/AttributeList.h
```diff
@@ -42,7 +42,7 @@ typedef struct AttributeList
 {
   std::vector<std::string>  attributeV;
 
-  void         fill(std::vector<std::string>);
+  void         fill(const std::vector<std::string>&);
```
kzangeli: The name of the parameter might not be mandatory but it's good to have it ...
I think ...
This is really a question for the style guide. Do we want to enforce 'named parameters' always ?
Opinions, please ...

crbrox: Fixed 50e1659

fgalan:
LGTM

kzangeli:
LGTM

Issue #1567, opened 2015-11-30
Forbidden character traces should be WARNINGS (not ERRORS)


fgalan:
Seen at the logs:

```
time=2015-11-30T12:24:43.761CET | lvl=ERROR | trans=1448882505-099-00000000711 | function=treat | comp=Orion | msg=xmlParse.cpp[131]: Found a forbidden value in 'tadata>'
time=2015-11-30T12:30:40.273CET | lvl=ERROR | trans=1448882505-099-00000002154 | function=forbiddenChars | comp=Orion | msg=forbiddenChars.cpp[63]: Bad Input (character '>')
time=2015-11-30T12:30:40.273CET | lvl=ERROR | trans=1448882505-099-00000002154 | function=treat | comp=Orion | msg=xmlParse.cpp[131]: Found a forbidden value in '
      tribute>'
time=2015-11-30T13:07:10.054CET | lvl=ERROR | trans=1448882505-099-00000011663 | function=forbiddenChars | comp=Orion | msg=forbiddenChars.cpp[63]: Bad Input (character '>')
time=2015-11-30T13:07:10.054CET | lvl=ERROR | trans=1448882505-099-00000011663 | function=treat | comp=Orion | msg=xmlParse.cpp[131]: Found a forbidden value in '
      ttribute>'
time=2015-11-30T13:13:55.932CET | lvl=ERROR | trans=1448882505-099-00000013631 | function=forbiddenChars | comp=Orion | msg=forbiddenChars.cpp[63]: Bad Input (character '>')
time=2015-11-30T13:13:55.932CET | lvl=ERROR | trans=1448882505-099-00000013631 | function=treat | comp=Orion | msg=xmlParse.cpp[131]: Found a forbidden value in '
      ttribute>'
```

As other case of bad user input, I understand this should use WARNING Bad Input (...) instead of ERROR

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/issues/1580

PR #1580, opened 2015-12-01
LM_W for forbidden characters

kzangeli:
Changed a few LM_E to LM_W for forbidden chars, added a lot  ...
Nothing in CNR ... 

Fixes issue #1567

### DIFF
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -103,9 +103,23 @@ std::string Entity::check(ConnectionInfo* ciP, RequestType requestType)
     return "No Entity ID";
   }
 
-  if (forbiddenChars(id.c_str()))         { return "Invalid characters in entity id";        }
-  if (forbiddenChars(type.c_str()))       { return "Invalid characters in entity type";      }
-  if (forbiddenChars(isPattern.c_str()))  { return "Invalid characters in entity isPattern"; }
+  if (forbiddenChars(id.c_str()))
+  {
+    LM_W(("Bad Input (found a forbidden character in the id of an entity"));
+    return "Invalid characters in entity id";
+  }
+
+  if (forbiddenChars(type.c_str()))
+  {
+    LM_W(("Bad Input (found a forbidden character in the type of an entity"));
+    return "Invalid characters in entity type";
+  }
+
+  if (forbiddenChars(isPattern.c_str()))
+  {
+    LM_W(("Bad Input (found a forbidden character in the pattern of an entity"));
+    return "Invalid characters in entity isPattern";
+  }
 
   return attributeVector.check(requestType, JSON, "", "", 0);
 }
--- a/src/lib/parse/forbiddenChars.cpp
+++ b/src/lib/parse/forbiddenChars.cpp
@@ -60,7 +60,7 @@ bool forbiddenChars(const char* s, const char* exceptions)
     case ';':
     case '(':
     case ')':
-      LM_E(("Bad Input (character '%c')", *s));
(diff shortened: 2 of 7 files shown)


src/lib/apiTypesV2/Entity.cpp
```diff
@@ -103,9 +103,23 @@ std::string Entity::check(ConnectionInfo* ciP, RequestType requestType)
     return "No Entity ID";
   }
 
-  if (forbiddenChars(id.c_str()))         { return "Invalid characters in entity id";        }
-  if (forbiddenChars(type.c_str()))       { return "Invalid characters in entity type";      }
-  if (forbiddenChars(isPattern.c_str()))  { return "Invalid characters in entity isPattern"; }
+  if (forbiddenChars(id.c_str()))
```
fgalan: Appart from changing from LM_E to LM_W, I wonder if with avoid "double logging", i.e. to avoid things like this:

``` 
time=2015-11-30T12:30:40.273CET | lvl=WARNING| trans=1448882505-099-00000002154 | function=forbiddenChars | comp=Orion | msg=forbiddenChars.cpp[63]: Bad Input (character '>')
time=2015-11-30T12:30:40.273CET | lvl=WARNING| trans=1448882505-099-00000002154 | function=treat | comp=Orion | msg=xmlParse.cpp[131]: Found a forbidden value in '
      tribute>'
```

which should be just:

```
time=2015-11-30T12:30:40.273CET | lvl=WARNING| trans=1448882505-099-00000002154 | function=forbiddenChars | comp=Orion | msg=forbiddenChars.cpp[63]: Bad Input (character '>')
```

kzangeli: To avoid "double/triple/quadruple logging" you need a whole framework, to push errors on an "error-stack" and finally present the entire error-set on a high level. Each level gives its part of the knowledge of the error.
We don't have anything like that in the broker and to implement it, it is pretty costy.

I am sure we have lots and lots of 'double logs' in the code.

For this specific error we could simply remove the calls to LM_W on one of the two levels.
Just pick which level to keep.

kzangeli: Fixed in 8afc5d4

Finally I removed the LM_W (which was actually an LM_E!!!) from the lower level (function forbiddenChars()) and I made sure **all** callers invoke  LM_W(("Bad Input ()")).
Advantage: more info on a higher level.
Inconvenient: **every** time forbiddenChars is invoked, the LM_W(("Bad Input ()")) must be invoked.

Doing it on a higher level is more 'risky' and more work, but the end-result is better.

fgalan: Of course, I'm referring to *this particular case* ;)

src/lib/parse/forbiddenChars.cpp
```diff
@@ -60,7 +60,7 @@ bool forbiddenChars(const char* s, const char* exceptions)
     case ';':
     case '(':
     case ')':
-      LM_E(("Bad Input (character '%c')", *s));
+      // LM_W(("Bad Input (character '%c')", *s));
```
fgalan: Leftover ?

kzangeli: Yeah, I left this one here (as LM_W, not LM_E) as it is still in the air whether "level 1" or "level 2" LM_Ws should removed to avoid "double logs"

fgalan:
LGTM

src/lib/parse/forbiddenChars.cpp
kzangeli: Fixed in a8020ee

Issue #1583, opened 2015-12-02
log improvements: Clearer log levels, aligned with "best practices"


fgalan:
Root issue: #1518 

Currently, we have `--silent` a kind of quick-fix to disable output (except errors). However, it would be clearer and more flexible to have a `-logLevel <level>` which, depending on the level:

* NONE: no log (a bit weird, but easy to include and may be useful for debugging in some cases)
* ERROR: only ERROR msgs are logged
* WARNING: WARNING and ERROR msgs are logged
* INFO: INFO, WARNING and ERROR msgs are logged
* DEBUG: DEBUG, INFO, WARNING and ERROR msgs are logged

`--silent` would be deprecated but maintained as an alias for `-logLevel ERROR`.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1590

PR #1590, opened 2015-12-02
Feature/1583 log levels

kzangeli:
Implements #1583 

New cli option --logLevel


CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,2 @@
 Add: servicePath header filtering in GET /v2/subscriptions operation (Issue #1557)
-
+Add: New command-line-option '--logLevel <level>', levels: NONE, ERROR, WARNING, INFO. DEBUG (Issue #1583)
```
fgalan: --logLevel -> -logLevel

(I guess that it also affects the code and cli.md)

doc/manuals/admin/cli.md
```diff
@@ -87,6 +87,12 @@ The list of available options is the following:
     contextBroker log file, instead of re-creating it when the
     broker starts.
 -   **--silent**. Suppress all log output except errors.
+-   **--logLevel**. Select logging level, supported levels:
+    - NONE    (suppress ALL log output, including errors),
+    - ERROR   (show only error messages),
+    - WARNING (show error and warning messages),
+    - INFO    (show error, warning and informational messages),
+    - DEBUG   (show ALL messages - this is the default setting).
```
fgalan: Operations suggest that default level be WARNING.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -377,6 +377,22 @@ PaArgument paArgs[] =
 
 /* ****************************************************************************
 *
+* validLogLevels - to pas to parweArgs library for validation of --logLevel
```
fgalan: to pas -> to pass
parweArgs -> parseArgs
--logLevel -> -logLevel

doc/manuals/admin/cli.md
```diff
@@ -87,6 +87,12 @@ The list of available options is the following:
     contextBroker log file, instead of re-creating it when the
     broker starts.
 -   **--silent**. Suppress all log output except errors.
```
fgalan: --silent should be deprecated, thus removed from the documentation and (if possible) hiden with -u and -U. A comment about this should be included in deprecated.md.

Until definitevely removed, --silent should be supported as equivlanet of `-logLevel ERROR`.

doc/manuals/admin/cli.md
kzangeli: Well, trace levels are off by default (unless you turn them on with -t).
Does Op want to remove the INFO by default?
We only have the INFO messages because Op asked us to ...
A bit confusing ...
However, I think it's a good idea to have WARNING as default, the only problem is that if somebody asks for  -t0-255  he/she will not get it, not unless he/she adds -logLevel DEBUG.
That I don't like.
Perhaps we should check traceLevels and if set, change logLevel to DEBUG?

doc/manuals/admin/cli.md
kzangeli: Fixed in 965c633

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in 965c633

doc/manuals/admin/cli.md
kzangeli: Fixed in 965c633

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 965c633

doc/manuals/deprecated.md
```diff
@@ -30,3 +30,5 @@ A list of deprecated features and the version in which they were deprecated foll
     the broker runs in "ConfMan mode".
 * Configuration Manager role (deprecated in 0.21.0, removed in 0.25.0)
 * Associations (deprecated in 0.21.0, removed in 0.25.0).
+* Deprecated command line argument in Orion 0.27.0:
```
fgalan: Move it to the first place (i.e. before the bullet taking about ONTIMEINTERVAL). The list follows a cronological-inverse ordering.

doc/manuals/admin/cli.md
```diff
@@ -86,7 +86,12 @@ The list of available options is the following:
 -   **-logAppend**. If used, the log lines are appended to the
     contextBroker log file, instead of re-creating it when the
     broker starts.
--   **--silent**. Suppress all log output except errors.
```
fgalan: What happen if --silent and -logLevel are used at the same time? In that case, two options:

* Exit in the same way we exit for concurrent usage of -t and -logLevel != DEBUG
* -logLevel sinlently overrides --silent

kzangeli: Fixed in a9daa7c  (exiting if both options are used)

doc/manuals/deprecated.md
kzangeli: Fixed in a9daa7c

fgalan:
LGTM

Issue #1593, opened 2015-12-03
Add srv=, subsrv=, from= fields in traces


fgalan:
Include the following fields in log traces, as per request from Operations.

* srv=
* subsrv=
* from=

fgalan:
New requierement:

* to=

fgalan:
This is the format that will be implementeed (ok with Operations):

```

time=2015-12-03T12:30:15.363CET | lvl=INFO | trans=1449141927-158-00000000003 | srv=pending | subsrv=pending | from=pending | function=LM_TRANSACTION_START | comp=Orion | msg=logMsg.h[1798]: Starting transaction from 192.168.206.1:58275/v1/contextEntities
time=2015-12-03T12:30:15.370CET | lvl=INFO | trans=1449141927-158-00000000003 | srv=valencia | subsrv=/Basuras | from=192.168.206.1 | function=collectionRangedQuery | comp=Orion | msg=connectionOperations.cpp[151]: Database Operation Successful (query: { query: { $or: [ { _id.id: /.*/ } ], _id.servicePath: { $in: [ /^/Basuras$/ ] } }, orderby: { creDate: 1 } })
time=2015-12-03T12:30:15.370CET | lvl=INFO | trans=1449141927-158-00000000003 | srv=valencia | subsrv=/Basuras | from=192.168.206.1 | function=collectionRangedQuery | comp=Orion | msg=connectionOperations.cpp[151]: Database Operation Successful (query: { query: { $or: [ { contextRegistration.entities.id: /.*/ }, { contextRegistration.entities: { $in: [] } }, { contextRegistration.entities.id: { $in: [] } } ], expiration: { $gt: 1449142215 }, servicePath: { $in: [ /^/Basuras$/ ] } }, orderby: { _id: 1 } })
time=2015-12-03T12:30:15.371CET | lvl=INFO | trans=1449141927-158-00000000003 | srv=valencia | subsrv=/Basuras | from=192.168.206.1 | function=collectionRangedQuery | comp=Orion | msg=connectionOperations.cpp[151]: Database Operation Successful (query: { query: { $or: [ { contextRegistration.entities.id: /.*/ }, { contextRegistration.entities: { $in: [] } }, { contextRegistration.entities.id: { $in: [] } } ], expiration: { $gt: 1449142215 }, servicePath: { $in: [ /^/Basuras$/ ] } }, orderby: { _id: 1 } })
[...]

PR #1596, opened 2015-12-03
Feature/1593 add fields to logs

fgalan:
Implements #1593


src/lib/logMsg/logMsg.cpp
```diff
@@ -343,6 +343,9 @@ int             inSigHandler      = 0;
 char*           progName;                   /* needed for messages (and by lmLib) */
 char            progNameV[512];             /* where to store progName            */
 __thread char   transactionId[64] = "N/A";
+__thread char   srv[50 + 1]       = "N/A";     // FIXME: Maybe we should use limits.h ?
+__thread char   subsrv[51*10 +1]  = "N/A";     // FIXME: Maybe we should use limits.h ?
+__thread char   from[4*4 + 1]     = "N/A";     // Based in XXX.XXX.XXX.XXX
```
kzangeli: srv ... I would prefer to write out the name "service", not especially long.
And, these names are very generic and need either to be static or have a prefix.
I guess static would be enough.

About 'from', it doesn't say much. Perhaps peerIp or clientIp would be more descriptive ?

About the default value, I'd leave it as is.
I really don't want any dependencies of common-lib in logMsg (which is a lower level library).

Now, about from[4*4 + 1]: just put 16, 4*4+1 is not even correct ...
Or, add a definition of NUMERICAL_IP_MAX_SIZE as 15 and then use NUMERICAL_IP_MAX_SIZE + 1.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1574,6 +1574,9 @@ static void notificationModeParse(char *notifModeArg, int *pQueueSize, int *pNum
 int main(int argC, char* argV[])
 {
   strncpy(transactionId, "N/A", sizeof(transactionId));
+  strncpy(srv,           "N/A", sizeof(srv));
+  strncpy(subsrv,        "N/A", sizeof(subsrv));
+  strncpy(from,          "N/A", sizeof(from));
```
kzangeli: It would be better to encapsulate the new LM-vars.
E.g.:
  lmServiceSet("N/A");
  lmSubServiceSet("N/A");
  lmClientIpSet("N/A");

Not sure what to call 'from'. 'from' is definitely not good enough.
clientIp, peerIp ... ?

src/lib/logMsg/logMsg.h
```diff
@@ -1793,13 +1802,77 @@ inline void LM_TRANSACTION_START(const char* keyword, const char* ip, int port,
...
+*/
+inline void LM_TRANSACTION_SET_SRV(const std::string& _srv)
+{
+  if (_srv != "")
+  {
+    snprintf(srv, sizeof(srv), _srv.c_str());
+  }
+  else
+  {
+    snprintf(srv, sizeof(srv), "<default>");
```
kzangeli: 'default' or N/A ?

src/lib/logMsg/logMsg.h
```diff
@@ -1793,13 +1802,77 @@ inline void LM_TRANSACTION_START(const char* keyword, const char* ip, int port,
 
 /* ****************************************************************************
 *
+* LM_TRANSACTION_SET_SRV -
+*/
+inline void LM_TRANSACTION_SET_SRV(const std::string& _srv)
```
kzangeli: I would really prefer to not use std::string here.
The lm-lib has gradually gone C++ but that is not at all my intention.
Please use a const char* as parameter.
Next project using lm-lib might be in C ...

src/lib/rest/rest.cpp
```diff
@@ -1041,6 +1042,16 @@ static int connectionTreat
       ciP->httpHeaders.servicePath = defaultServicePath(url, method);
     }
 
+    /* X-Forwared-For (used by a potential proxy on top of Orion) overrides ip */
+    if (ciP->httpHeaders.xforwardedFor == "")
+    {
+      LM_TRANSACTION_SET_FROM(ip);
```
kzangeli: LM_TRANSACTION_SET_FROM => lmClientIpSet    or similar.
No reason to 'pretend to be a macro'. This is just a normal function. Also, doesn't necessarily have anything to do with transactions.

fgalan: Ok. However, I would a like bit further and then rename all LM_TRANSACTION_* old-macro-new-inline-functions with names following the usual fucntion convention.

src/lib/logMsg/logMsg.cpp
fgalan: The names of the variables came after the names that Operaions have defined for the fields of the log.

```
srv= ... | subsrv= ... | from= ...
```

I think that having the same name will help to identify the purpose of each variable.

I'll move limits to limit.h

src/lib/logMsg/logMsg.h
fgalan: "N/A" is for places in which srv is not applicable (e.g. traces during Orion startup). "<default>" is for places where a srv makes sense but the client doesn't provide it (e.g. he/she doesn't sent the Fiware-Service header).

src/lib/logMsg/logMsg.h
fgalan: Ok, no problem.

src/lib/rest/rest.cpp
kzangeli: It's a mess right now, let's look it over later. We could create an issue about it and I fix it some day ...

fgalan: Not so much... just 5 functions as much. Very easy to change with "Replace All" ;)

src/lib/logMsg/logMsg.cpp
kzangeli: I understand your point, however, logMsg is a generic library.
srv ... server?
In contextBroker.cpp we would have the string "srv=SERVICE", which is OK, as it is specific for the broker, but lmLib ... is not.  We need to discuss this

fgalan: Ok, no problem.

src/lib/logMsg/logMsg.h
kzangeli: ok, NTC

src/lib/logMsg/logMsg.cpp
fgalan: Fixed in 7b4270c

src/lib/logMsg/logMsg.h
fgalan: 7b4270c

src/lib/rest/rest.cpp
fgalan: Fixed in 7b4270c

src/lib/common/statistics.h
```diff
@@ -304,13 +304,6 @@ extern int noOfRtUnsubscribeContextResponse;
 extern int noOfRtSubscribeResponse;
 extern int noOfRtSubscribeError;
 
-/*
```
fgalan: Old leftover, spoted in this PR.

NTC

test/functionalTest/cases/0000_cli/bool_option_with_value.test
```diff
@@ -79,6 +79,6 @@ Usage: contextBroker  [option '-U' (extended usage)]
                       [option '-statCounters' (enable request/notification counters statistics)]
                       [option '-statSemWait' (enable semaphore waiting time statistics)]
                       [option '-statTiming' (enable request-time-measuring statistics)]
-                      [option '-statNotifQueue' (enalle thread pool notifications queue statistics)]
+                      [option '-statNotifQueue' (enable thread pool notifications queue statistics)]
```
fgalan: Typo spoted during this PR.

NTC

kzangeli: Nice!

NTC

src/app/contextBroker/contextBroker.cpp
fgalan: Fixed in 79cd1ff

kzangeli:
LGTM

fgalan:
Ready to merge... when jenkins finished.

PR #1597, opened 2015-12-03
hardening/1583_log_levels_tracelevel_problem_in_functests

kzangeli:
Change for Orion to use LogLevelDebug for LM_T instead of its 'correct' level LogLevelTrace.
This was a bug introduced in my previous PR #1590

### DIFF
--- a/src/lib/logMsg/logMsg.h
+++ b/src/lib/logMsg/logMsg.h
@@ -981,11 +981,14 @@ do                                                                       \
 /* ****************************************************************************
 *
 * LM_T - log trace message
+*
+* FIXME: temporal change, just for Orion contextBroker, use LogLevelDebug for LM_T
+*        instead of its correct level LogLevelTrace.
 */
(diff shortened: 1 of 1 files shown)


src/lib/logMsg/logMsg.h
```diff
@@ -981,11 +981,14 @@ do                                                                       \
 /* ****************************************************************************
 *
 * LM_T - log trace message
+*
+* FIXME: temporal change, just for Orion contextBroker, use LogLevelDebug for LM_T
```
fgalan: You mean temporal in the context of logMsg library but permanent in the cotext of Oiron. Is that correct?

kzangeli: That is correct

fgalan: Just curiosity... :)

NTC

fgalan:
LGTM

Issue #1619, opened 2015-12-11
test harness: repeat failing test 3 times before concluding it has actually failed


fgalan:
Implement the following procedure:

* Try a first time
* If the first time fails, wait for N seconds and retry
* If the second time fails, wait for 2*N seconds and retry
* If the third time fails, it is an actual fail

The time printed in the output will be the one of the success execution (or the time of the last execution in the case of fail). In addition, in the case of sucess in the 2nd or 3rd trial, it could be added to output.

```
297/519: 0627_fix_wgs84_string/legacy_wgs84.test ...................................................................... 04 seconds
298/519: 0633_new_name_for_json_object_rendering/convop_attributes_as_json_objects.test ............................... 01 seconds (2nd trial)
299/519: 0640_attribute_name_unique_key_field/notify_type_change.test ................................................. 03 seconds (3rd trial) 
```

(Not sure if we have already an issue about this).

**Effort: 2 man day**

fgalan:
Implemented in PR #1620

PR #1620, opened 2015-12-14
Three tries for failing func tests

kzangeli:
Implements issue #1619

### DIFF
--- a/test/functionalTest/testHarness.sh
+++ b/test/functionalTest/testHarness.sh
@@ -24,6 +24,9 @@
 
 date
 testStartTime=$(date +%s.%2N)
+MAX_TRIES=3
+
+
 
 # ------------------------------------------------------------------------------
 #
@@ -507,6 +510,7 @@ function partExecute()
   what=$1
   path=$2
   forcedDie=$3
+  __tryNo=$4
 
   vMsg Executing $what part for $path
   dirname=$(dirname $path)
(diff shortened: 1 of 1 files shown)


test/functionalTest/testHarness.sh
```diff
@@ -640,6 +666,7 @@ function runTest()
   fileCreation $path $filename
   if [ "$toBeStopped" == "yes" ]
   then
+    runTestStatus="stopped2"
```
fgalan: What "stopped2" means?

test/functionalTest/testHarness.sh
```diff
@@ -731,22 +764,54 @@ do
...
+      else
+        printf "Running test %03d/%d: %s\n" "$testNo" "$noOfTests" "$testFile"
+      fi
+
+      runTest $testFile $tryNo
+      if [ $shellResult == "0" ]
+      then
+        if [ $tryNo != 1 ]
+        then
+          echo "OK"
```
fgalan: Is the trial number printed in the case of 2nd and 3rd trial as part as the output line corresponding to the test? Maybe it can be interred from the logic script, but it is a bit hard to follow to me...

kzangeli: Yes - as you say.
The first 'informative line' is as before while the second and consecutive kines get an extra "(intent X).

Look up line 781:

    if [ $tryNo != 1 ]
      then
      tryNoInfo="(intent $tryNo)"

test/functionalTest/testHarness.sh
kzangeli: "stopped2" and not "stopped", as "stopped" has already been used.
I don't even use these values. Just that it is != "OK"

test/functionalTest/testHarness.sh
fgalan: NTC

test/functionalTest/testHarness.sh
fgalan: NTC

fgalan:
LGTM

Issue #1617, opened 2015-12-11
Support float values in duration seconds


fgalan:
Currently the parse8601() supports:

```
P3Y1M1DT1H1M11S
```

but not:

```
P0Y0M0DT0H59M58.157S
```

The problem seems to be at `58.157`.

The solution is to adjust parse8601() to process float values with secound (rounding up/down to the nearest integer value), including corresponding unit tests to cover this case.

(Note that althought the standard allow float values in other fields, https://en.wikipedia.org/wiki/ISO_8601#Durations, this issue is just about supporting floats in seconds. We don't think float number for other fields is a general use case -never seem in last 3 years)

fgalan:
Related with question: http://stackoverflow.com/questions/34206712/error-in-subscribecontext (answer may need edition after this issue completion).

fgalan:
Implemented in PR #1624

PR #1624, opened 2015-12-14
Bug/1617 seconds as flots in duration strings

kzangeli:
Fixed issue #1617

### DIFF
--- a/src/lib/common/globals.cpp
+++ b/src/lib/common/globals.cpp
@@ -24,6 +24,7 @@
 */
 #include <time.h>
 #include <stdint.h>
+#include <math.h>
 
 #include <string>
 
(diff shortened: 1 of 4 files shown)


src/lib/common/globals.cpp
```diff
@@ -340,9 +340,37 @@ int64_t parse8601(const std::string& s)
...
+        // To round e.g. 4.51 to 5, we split 4.51 in its integer part and its decimal part:
+        //   secs    = 4.51
+        //   intPart = 4
+        //   rest    = 0.51
+        //
+        // After this simple operation is done, we just need to add ONE to 'intPart' if
+        // 'rest' >= 0.5 ('greater than' or just 'greater'?)
+        // and if 'rest' < 0.5, it is just thrown away.
+        //
+        float secs    = atof(start);
```
fgalan: What about of just use the round() function (www.gnu.org/software/libc/manual/html_node/Rounding-Functions.html), e.g:

```
value = (int) round(start);
```

Much shorter (just one statement) and clearer code, in my opinion.

kzangeli: Fixed in 4b74fc6

fgalan:
LGTM, but not merging yet (some modifications will be included in the .md as additional commits)

Issue #810, opened 2015-03-04
Implement operator[] in all vector-like classes and get rid of all get() occurrences


fgalan:
Some classes already implement the operator[] but we have to do an overall refactor of this.

**Effort: 3 man day**

kzangeli:
Also, the older method _**get()**_ should be removed.

Actually, if we start with removing those 'get-methods', the compiler will tell us wherever a modification is necessary.

fgalan:
I understand that we are chosing [] (vs get() or at()) because we found [] shorter and easier to understand.

fgalan:
Fixed in PR #1633

PR #1633, opened 2015-12-16
Changed from get()method to operator[]

raconte:
Implements #810

It has been modified the get method --> operator [].

There are some cases that haven´t been modified :
For example :
In fileRestrictionString.cpp
std::string RestrictionString::get(void --> this cases has also to be changed ?


src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -135,7 +135,7 @@ static bool addTriggeredSubscriptions
   BSONArrayBuilder attrA;
   for (unsigned int ix = 0; ix < cr.contextRegistrationAttributeVector.size(); ++ix)
   {
-    ContextRegistrationAttribute* craP = cr.contextRegistrationAttributeVector.get(ix);
+    ContextRegistrationAttribute* craP = cr.contextRegistrationAttributeVector.operator[](ix);
```
kzangeli: This might work, but it's not what we want.
These calls (all of them) should be made like this:

    cr.contextRegistrationAttributeVector[ix];

Otherwise, we don't really gain anything on the change ...

src/lib/ngsi/ContextElementResponseVector.cpp
```diff
@@ -144,14 +146,15 @@ ContextElementResponse* ContextElementResponseVector::get(unsigned int ix)
...
 /* ****************************************************************************
 *
 * ContextElementResponseVector::size -
 */
 unsigned int ContextElementResponseVector::size(void)
 {
+    
   return vec.size();
+
 }
```
kzangeli: unnecessary empty lines

src/lib/ngsi/ContextElementVector.cpp
```diff
@@ -116,15 +116,21 @@ void ContextElementVector::release(void)
...
 {
-  return vec[ix];
+    if (ix < vec.size())
+    {
+      return vec[ix];
+    }
+    else
+    {
+      return NULL;
+    }
```
kzangeli: unnecessary lines:

    else
     {
         return NULL;
     }

It's enough with:

    return NULL;

src/lib/serviceRoutines/getAttributeValueInstance.cpp
```diff
@@ -123,13 +123,13 @@ std::string getAttributeValueInstance
...
-        response.contextAttributeVector.push_back(cerP->contextElement.contextAttributeVector.get(i));
-      }
-    }
+    //for (int i = 0; i < cerP->contextElement.contextAttributeVector.size(); i++)
+   // {
+      //if (cerP->contextElement.vec[i])->getId() == metaIdValue)
+      //{
+        //response.contextAttributeVector.push_back(cerP->contextElement.vec[i]);
+      //}
+   // }
```
kzangeli: Not sure why this part has been commented out ...

src/lib/serviceRoutines/getAttributeValueInstance.cpp
```diff
@@ -56,7 +56,7 @@
 *   - !exist=entity::type
 *   - attributesFormat=object?
 *
-* 0. Take care of URI params
+* 0. Take care of URI par/home/raconte/fiware-orion/src/lib/serviceRoutines/getAttributeValueInstance.cppams
```
fgalan: This seems to be a unwanted copy-paste ;)

kzangeli:
LGTM

PR #1643, opened 2016-01-04
FIX: Centos 7 build

jmrobles:


src/app/contextBroker/CMakeLists.txt
```diff
@@ -88,6 +88,9 @@ ELSEIF(${DISTRO} STREQUAL "openSUSE_13.1")
 ELSEIF(${DISTRO} MATCHES "CentOS_6.*")
   MESSAGE("contextBroker: CentOS DISTRO: '${DISTRO}'")
   TARGET_LINK_LIBRARIES(contextBroker ${STATIC_LIBS} -lmicrohttpd -lmongoclient -lboost_thread-mt -lboost_filesystem-mt -lboost_system-mt -lboost_regex-mt)
+ELSEIF(${DISTRO} MATCHES "CentOS_release")
```
fgalan: I should have expected something like this here:

```
ELSEIF(${DISTRO} MATCHES "CentOS_7.*")
```

If `CentOS_release` is the one which works, something makes me think that the procedure to fill DISTRO variable (based on the `osDistro.sh` script) is not working...

Could you tell me the result of the following commands in your CentOS 7.x box, please?

* `cat /etc/redhat-release`
* `scripts/build/osDistro.sh` (from the fiware-orion repository root)

fgalan:
Thanks for the contribution! Please, find my comments above.

src/app/contextBroker/CMakeLists.txt
jmrobles: Yes, me too :)

But with "CentOS_7.*" did not work and the console displays "CentOS_release" as version.
-> cat /etc/redhat-release
```
CentOS Linux release 7.2.1511 (Core) 
```
-> scripts/build/osDistro.sh
```
CentOS_release
``` 

With this failover code it's solved:
```
centos_distro=$(cat /etc/redhat-release 2> /dev/null | awk '{print $3}') 
if [ $centos_distro == "release" ]; then
        centos_distro=$(cat /etc/redhat-release 2> /dev/null | awk '{print $4}')
fi
```

fgalan: Looking to a CentOS 6.6 system, the content of the /etc/redhat-release file is as follows:

```
CentOS release 6.6 (Final)
```

So I understand that, at least in some systems, CentOS 7.x has introduced the `Linux` token.

scripts/build/osDistro.sh
```diff
@@ -22,8 +22,10 @@ ubuntu_distro=$(grep DISTRIB_DESCRIPTION= /etc/lsb-release 2>/dev/null | cut -d\
 debian_distro=$(cat /etc/debian_version 2> /dev/null)
 suse_distro=$(cat /etc/SuSE-release 2> /dev/null | grep SUSE | cut -d ' ' -f 1-2 | sed 's/ /_/g')
 centos_distro=$(cat /etc/redhat-release 2> /dev/null | awk '{print $3}')
+if [ $centos_distro == "release" ]; then
```
fgalan: Good fix. However, I'd suggest to add some explanation about this, as it may be a bit criptic. For example, just above this line:

```
# In some cases (e.g. CentOS 7.x) we have found that the /etc/redhat-release content use the following pattern:
#
# CentOS Linux release 7.2.1511 (Core) 
# 
# Instead the one in previous versions:
#
# CentOS release 6.6 (Final)
#
# The check in the next line ensure that the distro version is correctly detected in any case.
```

Or a similar wording.

kzangeli:
LGTM

fgalan:
LGTM. Merging.

Thanks again for the contribution!

PR #1646, opened 2016-01-07
task/1259_part7_1_new_parser_for_v2_entities_options_keyValues

kzangeli:
New parse method for entities, using URI param options=keyValues

### DIFF
--- a/src/lib/jsonParseV2/parseContextAttribute.cpp
+++ b/src/lib/jsonParseV2/parseContextAttribute.cpp
@@ -190,7 +190,7 @@ std::string parseContextAttribute(ConnectionInfo* ciP, const Value::ConstMemberI
     //
     // If the Object contains "value", then it is considered an object, not a compound
     //
-    if (iter->value.HasMember("value"))
+    if ((iter->value.HasMember("value")) && (ciP->uriParamOptions["keyValues"] == false))
     {
       r = parseContextAttributeObject(iter->value, caP, keyValues);
       if (r != "OK")
--- a/src/lib/jsonParseV2/parseContextAttribute.cpp
+++ b/src/lib/jsonParseV2/parseContextAttribute.cpp
@@ -44,7 +44,7 @@ using namespace rapidjson;
 *
 * parseContextAttributeObject - 
 */
-static std::string parseContextAttributeObject(const Value& start, ContextAttribute* caP)
+static std::string parseContextAttributeObject(const Value& start, ContextAttribute* caP, bool keyValues = false)
 {
   // valueTypeNone will be overridden inside the 'for' block in case the attribute has an actual value
   caP->valueType = orion::ValueTypeNone;
@@ -121,10 +121,53 @@ static std::string parseContextAttributeObject(const Value& start, ContextAttrib
         return r;
       }
     }
-    else  // ERROR
+    else if (keyValues == false) // ERROR
     {
       LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
       return "unrecognized property for context attribute";
+    }
+    else  // Anything captured here is a metadata, and options=keyValues has been set
+    {
+      Metadata* mdP = new Metadata();
+
+      mdP->name = name;
+
+      if (type == "String")
+      {
+        mdP->valueType   = orion::ValueTypeString;
+        mdP->stringValue = iter->value.GetString();
+      }
+      else if (type == "Number")
+      {
+        mdP->valueType   = orion::ValueTypeNumber;
+        mdP->numberValue = iter->value.GetDouble();
+      }
+      else if (type == "True")
+      {
+        mdP->valueType   = orion::ValueTypeBoolean;
+        mdP->boolValue   = true;
+      }
+      else if (type == "False")
+      {
+        mdP->valueType   = orion::ValueTypeBoolean;
+        mdP->boolValue   = false;
+      }
+      else if (type == "Null")
+      {
+        mdP->valueType   = orion::ValueTypeNone;
+      }
+      else if ((type == "Array") || (type == "Object"))
+      {
+        alarmMgr.badInput(clientIp, "Compound values not supported for metadata");
+        return "json error in metadata value";
+      }
+      else
+      {
+        alarmMgr.badInput(clientIp, "Invalid value for metadata");
+        return "json error in metadata value";
+      }
+
+      caP->metadataVector.push_back(mdP);
     }
   }
 
@@ -139,8 +182,10 @@ static std::string parseContextAttributeObject(const Value& start, ContextAttrib
 */
 std::string parseContextAttribute(ConnectionInfo* ciP, const Value::ConstMemberIterator& iter, ContextAttribute* caP)
 {
-  std::string name   = iter->name.GetString();
-  std::string type   = jsonParseTypeNames[iter->value.GetType()];
+  std::string name      = iter->name.GetString();
+  std::string type      = jsonParseTypeNames[iter->value.GetType()];
+  bool        keyValues = ciP->uriParamOptions["keyValues"];
+
   
   caP->name = name;
 
@@ -190,7 +235,7 @@ std::string parseContextAttribute(ConnectionInfo* ciP, const Value::ConstMemberI
     //
     if (iter->value.HasMember("value"))
     {
-      r = parseContextAttributeObject(iter->value, caP);
+      r = parseContextAttributeObject(iter->value, caP, keyValues);
       if (r != "OK")
       {
         alarmMgr.badInput(clientIp, "JSON parse error in ContextAttribute::Object");
@@ -229,7 +274,8 @@ std::string parseContextAttribute(ConnectionInfo* ciP, const Value::ConstMemberI
 */
 std::string parseContextAttribute(ConnectionInfo* ciP, ContextAttribute* caP)
 {
-  Document document;
+  bool      keyValues = ciP->uriParamOptions["keyValues"];
+  Document  document;
 
   document.Parse(ciP->payload);
 
@@ -254,7 +300,7 @@ std::string parseContextAttribute(ConnectionInfo* ciP, ContextAttribute* caP)
     return oe.render(ciP, "");
   }
 
-  std::string  r = parseContextAttributeObject(document, caP);
+  std::string  r = parseContextAttributeObject(document, caP, keyValues);
   if (r != "OK")
   {
     OrionError oe(SccBadRequest, r);
--- a/src/lib/jsonParseV2/parseContextAttribute.cpp
+++ b/src/lib/jsonParseV2/parseContextAttribute.cpp
@@ -121,53 +121,10 @@ static std::string parseContextAttributeObject(const Value& start, ContextAttrib
         return r;
       }
     }
-    else if (keyValues == false) // ERROR
+    else // ERROR
     {
       LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
       return "unrecognized property for context attribute";
-    }
-    else  // Anything captured here is a metadata, and options=keyValues has been set
-    {
-      Metadata* mdP = new Metadata();
-
-      mdP->name = name;
-
-      if (type == "String")
-      {
-        mdP->valueType   = orion::ValueTypeString;
-        mdP->stringValue = iter->value.GetString();
-      }
-      else if (type == "Number")
-      {
-        mdP->valueType   = orion::ValueTypeNumber;
-        mdP->numberValue = iter->value.GetDouble();
-      }
-      else if (type == "True")
-      {
-        mdP->valueType   = orion::ValueTypeBoolean;
-        mdP->boolValue   = true;
-      }
-      else if (type == "False")
-      {
-        mdP->valueType   = orion::ValueTypeBoolean;
-        mdP->boolValue   = false;
-      }
-      else if (type == "Null")
-      {
-        mdP->valueType   = orion::ValueTypeNone;
-      }
-      else if ((type == "Array") || (type == "Object"))
-      {
-        alarmMgr.badInput(clientIp, "Compound values not supported for metadata");
-        return "json error in metadata value";
-      }
-      else
-      {
-        alarmMgr.badInput(clientIp, "Invalid value for metadata");
-        return "json error in metadata value";
-      }
-
-      caP->metadataVector.push_back(mdP);
     }
   }
 
--- a/src/lib/jsonParseV2/parseEntity.cpp
+++ b/src/lib/jsonParseV2/parseEntity.cpp
@@ -51,7 +51,9 @@ using namespace rapidjson;
 * the service routine postEntity.
 *
 * Also, if the URI param 'options' includes the value 'keyValues', then the
-* parse changes to accept 'unknown fields' as attributes inside entities.
+* parse changes for compound values of attributes. If the value is a JSON object
+* then there is no looking inside to find the 'value' field, but the attribute is
+* always treated as a compound attribute.
 * 
 */
 std::string parseEntity(ConnectionInfo* ciP, Entity* eP, bool eidInURL)
--- a/src/lib/jsonParseV2/parseEntity.cpp
+++ b/src/lib/jsonParseV2/parseEntity.cpp
@@ -49,6 +49,10 @@ using namespace rapidjson;
 *
 * In the case of /v2/entities/<eid>, the entityId of 'Entity* eP' is set in
 * the service routine postEntity.
+*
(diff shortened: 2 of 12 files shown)


test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
```diff
@@ -0,0 +1,215 @@
...
+#         "type": "MT1
+#       }
+#     }
+#   }
+# }
+#
+# In keyValues mode, there are no metadata for attributes. If you need that, use the "normalized" method.
+# 
+#
+# An example just for Curiosity ...
```
fgalan: Curiosity -> curiosity ?

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
```diff
@@ -0,0 +1,215 @@
...
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# We now have two different ways of parsing "POST /v2/entities":
+#
+# 1. The 'old way', or "Normalized", which is the default parsing method:
```
fgalan: The SHELL preamble comments are a very good piece of documentation from a "pragmatic usage of the API" point of view. Maybe the text should be reused in some part of the documentation, e.g. the NGSIv2 cookbook.

src/lib/jsonParseV2/parseContextAttribute.cpp
```diff
@@ -121,11 +121,54 @@ static std::string parseContextAttributeObject(const Value& start, ContextAttrib
         return r;
       }
     }
-    else  // ERROR
+    else if (keyValues == false) // ERROR
     {
       LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
       return "unrecognized property for context attribute";
     }
+    else  // Anything captured here is a metadata, and options=keyValues has been set
```
fgalan: Not sure of understand the comment "Anything captured here is a metadata, and options=keyValues has been set" as is keyValues is set, then no metadata should be captured... maybe I'm missing something. Could you elaborate, please?

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -49,6 +49,10 @@ using namespace rapidjson;
 *
 * In the case of /v2/entities/<eid>, the entityId of 'Entity* eP' is set in
 * the service routine postEntity.
+*
+* Also, if the URI param 'options' includes the value 'keyValues', then the
```
fgalan: The part "or metadata inside attribute" is not correct, as keyValues only allows to "compact" attribute values, but not metadata. Or maybe I'm not understanding correctly this comment.

test/functionalTest/cases/1187_empty_attribute_in_create/empty_attribute_in_create.test
```diff
@@ -120,11 +120,6 @@ Date: REGEX(.*)
         "type": null,
         "value": ""
     },
-    "A2": {
```
fgalan: I understand that the "side effect" which causes A2 not being shown, should be fixed.

src/lib/jsonParseV2/parseContextAttribute.cpp
kzangeli: We are inside an attribute here (and keyValues is *true*), so if a field named "m1" is encountered, then it is not an error, but a metadata (without keyValues true) it would be an error.

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Maybe I'm the one that has misunderstood. 
I am treating 'metadatas inside attributes' just like 'attributes inside entities'.

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
kzangeli: sure :-)

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
kzangeli: With some minor changes this chunk of text could be part of out documentation, yes.  
Tell me where exactly and I'll put it in.

test/functionalTest/cases/1187_empty_attribute_in_create/empty_attribute_in_create.test
kzangeli: Actually, there has been a bug fix.
We talked about this ...
Before, the broker signaled error and created A2 (in the previous step).
Now, the same error is signaled but A2 is not created.

This is better than before.

Even better would be to NOT signal error and to create A2, but that is a bigger fix and not included in this PR.

fgalan: You are right... Actually, the fix is correct.

Maybe it would be clerarer chaning a little bit the step descriptions. In particular:

03. POST /v2/entities to create E1/A2="", using json object

to

03. POST /v2/entities to create E1/A2="", using json object (fails, as entity already exists)

And:

04. GET /v2/entities/E1 to see A2==""

to

04. GET /v2/entities/E1 to not see A2

kzangeli: Of course. The 'descriptive texts' haven't been modified. just a lapsus on my behalf.

src/lib/jsonParseV2/parseContextAttribute.cpp
kzangeli: After a short discussion on skype I have realized my mistake.
I still think 'the old way' I'm afraid ...
Will be fixed.

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Yes, I am definitely the one that has misunderstood. To be fixed.

test/functionalTest/cases/1187_empty_attribute_in_create/empty_attribute_in_create.test
kzangeli: Fixed in 47a8a42

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Fixed in 47a8a42

src/lib/jsonParseV2/parseContextAttribute.cpp
kzangeli: Fixed in 47a8a42

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
kzangeli: Fixed in 47a8a42

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_with_keyvalues.test
fgalan: In order not to block this PR, I'd suggest to do it in a separate PR.

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -49,6 +49,9 @@ using namespace rapidjson;
...
+* parse changes to accept 'unknown fields' as attributes inside entities.
```
fgalan: "Also, if the URI param 'options' includes the value 'keyValues', then the parse changes to accept 'unknown fields' as attributes inside entities".

Not sure if the comment is fully correct... I mean, in no-keyValues mode, the attributes are also unkown fields (the difference is that the value of such unknown fields follows a given pattern). Maybe I don't understanding correctly...

kzangeli: True.
I guess I was thinking about the way I wanted to implement the metadatas when I wrote this.

test/functionalTest/cases/0979_patch_v2_entities_eid/patch_v2_entities_eid.test
```diff
@@ -35,7 +35,8 @@ brokerStart CB
 # 03. GET /v2/entities/E1
 # 04. PATCH /v2/entities/E1 attr1=4
 # 05. GET /v2/entities/E1 - see attr1=4 AND attr2=2 (untouched)
-# 06. PATCH /v2/entities/E1 attr3=6 - error, as attr3 doesn't exist
+# 06. PATCH /v2/entities/E1 attr3=6
```
fgalan: I think this should be rolled-back... PATCH cannot be used to add new attributes, so if attr3 doesn't previously exist, then it should lead to an error.

test/functionalTest/cases/0979_patch_v2_entities_eid/patch_v2_entities_eid.test
```diff
@@ -35,7 +35,8 @@ brokerStart CB
...
+# 07. GET /v2/entities/E1 - see attr1, attr2 and attr3
```
fgalan: attr3 will not be shown, after rollback (probably the step 07 could be removed, to revert this .test to its original form).

test/functionalTest/cases/1187_empty_attribute_in_create/empty_attribute_in_create.test
```diff
@@ -32,8 +32,8 @@ brokerStart CB
 #
 # 01. POST /v2/entities to create E1/A1=""
 # 02. GET /v2/entities/E1 to see A1==""
-# 03. POST /v2/entities to create E1/A2="", using json object
-# 04. 04. GET /v2/entities/E1 to see A2==""
+# 03. POST /v2/entities to create E1/A2="", using json object (fails, as entity already exists)
```
fgalan: This should be roll-backed... According to specification, POST on /v2/entities/E1 will lead to append/update attributes. Similarly, step 03 should be also rollbacked. Probably, leaving this .test in its original form will work.

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
```diff
@@ -0,0 +1,272 @@
...
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# 01. POST /v2/entities, creating E1
+# 02. PUT /v2/entities/E1?options=keyValues, adding A1 to E1
+# 03. GET /v2/entities/E1
+# 04. PATCH /v2/entities/E1?options=keyValues, adding A2 to E1
```
fgalan: It should be POST (PATCH is not for adding)

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
```diff
@@ -0,0 +1,272 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# 01. POST /v2/entities, creating E1
+# 02. PUT /v2/entities/E1?options=keyValues, adding A1 to E1
```
fgalan: Maybe this PUT could be changed to PATCH (PUT will also work, but it is properly tested in "replace" semantics in step 6).

test/functionalTest/cases/1260_v2_update_patch_error_description/update_patch_error_description.test
```diff
@@ -88,15 +88,10 @@ Date: REGEX(.*)
 
 03. PATCH /v2/entities/Eno (not found)
```
fgalan: I think this test should be rollbacked.

test/functionalTest/cases/1278_patch_unknown_attribute/patch_unknown_attribute.test
```diff
@@ -31,8 +31,9 @@ brokerStart CB
 
 #
 # 01. POST /v2/entities to create E1/T1/A1=0
-# 02. PATCH /v2/entities/E1 (A2 unknown)
-
+# 02. PATCH /v2/entities/E1 (to create attr A2)
```
fgalan: I think this .test should be rollbacked.

test/functionalTest/cases/1187_empty_attribute_in_create/empty_attribute_in_create.test
kzangeli: Except there is no POST on /v2/entities/E1.
There is a second POST on /v2/entities (with E1 in payload).

I understand a second POST /v2/entities with the same *id* in the payload should give an error ...

kzangeli: Discussed on skype. All OK

NTC

src/lib/jsonParseV2/parseEntity.cpp
kzangeli: Fixed in a5f5402

test/functionalTest/cases/0979_patch_v2_entities_eid/patch_v2_entities_eid.test
kzangeli: Fixed in a5f5402

test/functionalTest/cases/0979_patch_v2_entities_eid/patch_v2_entities_eid.test
kzangeli: Fixed in a5f5402

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
kzangeli: Fixed in a5f5402

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
kzangeli: Fixed in a5f5402

test/functionalTest/cases/1260_v2_update_patch_error_description/update_patch_error_description.test
kzangeli: Fixed in a5f5402

test/functionalTest/cases/1278_patch_unknown_attribute/patch_unknown_attribute.test
kzangeli: Fixed in a5f5402

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2921,6 +2921,11 @@ void processContextElement
         buildGeneralErrorResponse(ceP, NULL, responseP, SccInvalidModification, "Already Exists");
         return;
     }
+    else if ((checkEntityExistance == true) && (action == "APPEND") && (entitiesNumber == 0))
```
fgalan: Why don't just `checkEntityExistance` instead of `checkEntityExistance == true`? ;)

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
```diff
@@ -0,0 +1,274 @@
...
+POST /v2/entities with key-values
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# 01. POST /v2/entities, creating E1
```
fgalan: Maybe PATCH testing should be added to this sequence?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Matter of opinion. I prefer to have the '== true' there.
If you want to, we decide a way to do this and we add it to the style guide.
I don't have an especially strong opinion.
I prefer adding the == true for clarity but I know most people dislike it.

test/functionalTest/cases/1259_new_json_for_v2/post_v2_entities_eid_with_keyvalues.test
kzangeli: Fixed in 2d3d4cd

fgalan:
LGTM

Issue #1652, opened 2016-01-11
Incorrect detection of wrong URLs at notification time


fgalan:
It seems that Orion (as in 0.26.1) allows create subscriptions which reference is port 0 (ilegal, as port ranges from 1 to 65535):

```
time=2016-01-11T12:29:20.204CET | lvl=INFO | trans=1452511738-622-00000000337 | function=LM_TRANSACTION_START | comp=Orion | msg=logMsg.h[1792]: Starting transaction to foo.bar.net:0/ngsi10/notifyContext
time=2016-01-11T12:29:20.204CET | lvl=ERROR | trans=1452511738-622-00000000337 | function=httpRequestSendWithCurl | comp=Orion | msg=httpRequestSend.cpp[220]: Runtime Error (port is ZERO)
time=2016-01-11T12:29:20.204CET | lvl=INFO | trans=1452511738-622-00000000337 | function=LM_TRANSACTION_END | comp=Orion | msg=logMsg.h[1815]: Transaction ended
```

Thus, probably reference should be checked at subscription creation/update time, in order to detect ilegal port (and maybe other URL checkings).

fgalan:
Taking into account the content of the DB:

```
"reference" : "http://fiwarenr.eu-gb.mybluemix.net/ngsi10/notifyContext/dev-fixed",
```

The proble is not related with ilegal port at subscription time, but with incorrectly setting the default to 80.

fgalan:
After more research, the problem is due the following strings are not detected as wrong URL:

```
http://http://XXX:1234/path
http://XXX:/path
```

Adjusting parseUrl() solves the issue. I'll take care.

fgalan:
Fixed in PR #1653

PR #1653, opened 2016-01-12
FIX pasrseUrl() function

fgalan:
Fix bug #1652


scripts/build/osDistro.sh
```diff
@@ -27,7 +27,8 @@ centos_distro=$(cat /etc/redhat-release 2> /dev/null | awk '{print $3}')
 # Instead the one in previous versions:
 # CentOS release 6.6 (Final)
 # The check in the next line ensure that the distro version is correctly detected in any case.
-if [ $centos_distro == "release" ]; then
+if [ "$centos_distro" = "release" ]
```
fgalan: It seems that `==` don't work with `sh` (it works with `bash`). See http://stackoverflow.com/questions/10849297/compare-a-string-in-unix#comment14133719_10849357.

NTC

kzangeli:
LGTM (but merge with develop needed)

PR #1656, opened 2016-01-12
Made attribute string values being accepted without citation marks

kzangeli:


fgalan:
Looking to the NGSIv2 specification, at the end this PR is not needed.

PR #1659, opened 2016-01-12
op=append no longer in use. options=append is used in its place

kzangeli:
op=append no longer in use. options=append is used in its place


fgalan:
LGTM

PR #1674, opened 2016-01-14
Bug/1661 invalid subscriptions

### LINKED ISSUE #1661: UpdateContext wrongly gets Bad Input due to invalid subscription
It seems we accept subscriptions with invalid IP/port in *reference*. The error isn't discovered until it is time to send a notification (due to a ContextUpdate) and a BadInput is issued to the alarm manager.
Of course, the BadInput is signaled for the innocent ContextUpdate instead of the all but innocent Subscribe. This must be changed,

Two things to do here:

* Change the BadInput to RuntimeError for invalid subscriptions detected during ContextUpdates.
* Detect the invalid subscription data within the subscription request, respond with error and issue a BadInput for the 'subscription creator'

kzangeli:
Fixes #1661 

Detecting subscriptions with invalid reference/callback, both at subscription creation time (raising a Bad Input alarm), and at Notification time (sending an error to log file about Runtime Error).

### DIFF
--- a/src/lib/ngsiNotify/Notifier.cpp
+++ b/src/lib/ngsiNotify/Notifier.cpp
@@ -104,12 +104,11 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
     int          port;
     std::string  uriPath;
     std::string  protocol;
-    
+    std::string  initialUrl = url;
+
     if (!parseUrl(url, host, port, uriPath, protocol))
     {
-      std::string details = std::string("sending NotifyContextRequest: malformed URL: '") + url + "'";
-      alarmMgr.badInput(clientIp, details);
-
+      LM_E(("Runtime Error (not sending NotifyContextRequest: malformed URL: '%s')", initialUrl.c_str()));
       return;
     }
 
--- a/src/lib/ngsiNotify/Notifier.cpp
+++ b/src/lib/ngsiNotify/Notifier.cpp
@@ -104,11 +104,10 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
     int          port;
     std::string  uriPath;
     std::string  protocol;
-    std::string  initialUrl = url;
 
     if (!parseUrl(url, host, port, uriPath, protocol))
     {
-      LM_E(("Runtime Error (not sending NotifyContextRequest: malformed URL: '%s')", initialUrl.c_str()));
+      LM_E(("Runtime Error (not sending NotifyContextRequest: malformed URL: '%s')", url.c_str()));
       return;
     }
 
--- a/test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error.test
+++ b/test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error.test
@@ -153,7 +153,7 @@ echo
 
 echo "06. Check log-file to make sure the Bad Input alarm is reset"
 echo "============================================================"
-grep "BadInput" /tmp/contextBroker.log | cut -d\| -f 9-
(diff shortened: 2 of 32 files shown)


src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -104,12 +104,11 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
     int          port;
     std::string  uriPath;
     std::string  protocol;
-    
+    std::string  initialUrl = url;
```
fgalan: url is passed to parseUrl() as const string& so... why an extra variable initialUrl is needed here?

test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error.test
```diff
@@ -0,0 +1,295 @@
...
+HTTP/1.1 201 Created
+Content-Length: 0
+Location: /v2/subscriptions/REGEX([0-9a-f]{24})
+Date: REGEX(.*)
+
+
+
+06. Check log-file to make sure the Bad Input alarm is reset
+============================================================
+ msg=AlarmManager.cpp[213]: Raising alarm BadInput 0.0.0.0: Invalid URL
```
fgalan: This is a little weak... any code shift will modify this number. Why don't apply the same command that to step 4?

test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error_api_v1.test
```diff
@@ -0,0 +1,305 @@
...
+    "subscribeResponse": {
+        "duration": "P1M",
+        "subscriptionId": "REGEX([0-9a-f]{24})"
+    }
+}
+
+
+06. Check log-file to make sure the Bad Input alarm is reset
+============================================================
+ msg=AlarmManager.cpp[213]: Raising alarm BadInput 0.0.0.0: Invalid URL
```
fgalan: Same here...

test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error.test
kzangeli: Yes, of course.
I thought about removing those line numbers but with so much 'turbulence' in this PR, it seems  I forgot ...

src/lib/ngsiNotify/Notifier.cpp
kzangeli: I took for granted that url is destroyed in parseUrl.
Perhaps it isn't ...

If it is not destroyed, then there is no need to copy it, of course.

kzangeli: OK,

parseUrl calls stringSplit that does:

    int stringSplit(const std::string& in, char delimiter, std::vector<std::string>& outV)
    {
      char* s          = strdup(in.c_str());

There is my copy.
initialUrl to be removed.

test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error_api_v1.test
kzangeli: Fixed in 6080cf6

test/functionalTest/cases/1661_bad_input_on_notification_error/bad_input_on_notification_error.test
kzangeli: Fixed in 6080cf6

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 6080cf6

fgalan:
LGTM

Issue #1316, opened 2015-10-05
API v2: Create subscription


fgalan:
Implement the following operation in in v2:

```
POST /v2/subscriptions
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
Once NGSIv2 subscription get fully implemented ("fully" meaning the posibility of implement filters using the `condition` field) the answer to question http://stackoverflow.com/questions/33669300/fiware-orion-context-broker-restrictions-on-subscriptions/33752490#33752490 should be properly edited.

fgalan:
The answer to the following question at SOF has to be edited upon this issue completion: http://stackoverflow.com/questions/34136801/orion-context-broker-v2-subscription-issue/34187804#34187804

fgalan:
Basic creation implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1650. However, this issue should be kept opened until actual expression evaluation (at least for `q`) gets implemented.

fgalan:
Subscription q expression evalution upon update entity implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1687

fgalan:
Initial notification taking into account filters implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1691

fgalan:
At the present moment, the tasks that remain to close this issue are related with cache. In particular, to take into account q, georel, geometry and coords in the cache entries (adjust addTriggeredSubscription_withCache needed, look for `FIXME P10 #1316` in the code).

Moreover, it could be a good idea not storing `q` as string but as a class, in order to parse once time (i.e. the moment in which the cache entry is populated) intead of each time a new update context arrives and the expression is evaluated from the cache entry. The same idea could be extended to the other three: georel, geometry and coord (not currently used but they will be in the near future).

**No need of new test harness**. Just look for `# FIXME P10: replace with just 'brokerStart CB' once cache part gets implemented` inside cases/1316_XX .test and adjust `brokerStart` line consecuently.

fgalan:
Finished in PR #1704. The task about using objects instead of string for expression has been defered to a tech debt issue: https://github.com/telefonicaid/fiware-orion/issues/1705

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests

[...]

Issue #350, opened 2014-05-07
 ONANYCHANGE subscriptions


fgalan:
NGSI standard ONCHANGE subscription always need a list of "triggering" attributes (see section 5.5.10 in the OMA spec). Thus, it is not possible to specify that the subscriber wants to be notified each time any of the attributes of the entities change without a previous knowledge of the list of attributes of the entity. This is a limitation on some scenarios.

This issue is about to enhance the ContextBroker, including a new type of subscriptions for changes in any attribute (let's name them ONANYCHANGE). In this case the condValue element is not required (this fits in NGSI spec, as this element is optional in subscribeContextRequests according to OMA spec).

Regarding implementation, it would be a matter of tuning the query done to MongoDB in the addTriggeredSubscriptions() function at mongoBackend. It also will involve changes in the processConditionVector(), taking into account the new subscription change (and using the corresponding query for that).

ImanolBarba:
+1

alexferreiros:
+1

goofyahead:
+1

dmoranj:
+1

frbattid:
+1

fgalan:
The following answer at SOF should be edited when this issue gets implemented: http://stackoverflow.com/questions/31826459/fiware-context-broker-can-we-subscribe-to-all-the-context-updates/31831730?noredirect=1#comment53804184_31831730

silva96:
:+1:

c-castillo:
+1

dimh:
+1

diogonunes92:
+1

fgalan:
Implemented (finally!) in PR https://github.com/telefonicaid/fiware-orion/pull/1679. Once that PR gets merged, it will be available in develop branch. In addition, it would be available in the next official release (0.27.0), which is expected by the end of January / beginning of February.

goofyahead:
Yay! :dancer:

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
LGTM

PR #1679, opened 2016-01-15
ONANYCHANGE subscriptions

fgalan:
This PR stated as preparation for expression `q` filter in NGSIv2 subscriptions (thus the name of the branch, referring #1316) but at the end has become an implementation of ONANYCHANGE subscriptions, i.e. #350 (functionality that it is also needed for NGSIv2 subscriptions).

(Commit history is a bit weird, maybe due to a not-so-good rebase, but the diff agains develop seems to be ok anyway).

### DIFF
--- a/doc/manuals/user/walkthrough_apiv1.md
+++ b/doc/manuals/user/walkthrough_apiv1.md
@@ -1020,6 +1020,8 @@ Let's examine in detail the different elements included in the payload:
     notified of the value of *temperature* each time the value of
     *pressure* changes may not be too useful. The example is chosen this
     way only to show the enormous flexibility of subscriptions.
+-   You can leave empty the condValue list (or even omit) to make notification
+    triggers whenever any entity attribute (no matter which name) changes.
 -   The throttling element is used to specify a minimum
     inter-notification arrival time. So, setting throttling to 5 seconds
     as in the example above, makes a notification not to be sent
--- a/doc/manuals/user/walkthrough_apiv1.md
+++ b/doc/manuals/user/walkthrough_apiv1.md
@@ -1020,8 +1020,8 @@ Let's examine in detail the different elements included in the payload:
     notified of the value of *temperature* each time the value of
     *pressure* changes may not be too useful. The example is chosen this
     way only to show the enormous flexibility of subscriptions.
--   You can leave empty the condValue list (or even omit) to make notification
-    triggers whenever any entity attribute (no matter which name) changes.
+-   You can leave the condValue list empty (or even omit it) to make a notification
+    trigger on any entity attribute change (regardless of the name of the attribute).
 -   The throttling element is used to specify a minimum
     inter-notification arrival time. So, setting throttling to 5 seconds
     as in the example above, makes a notification not to be sent
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1512,7 +1512,7 @@ void processOntimeIntervalCondition(const std::string& subId, int interval, cons
 *
 * matchExpression
 */
-static bool matchExpresion(ContextElementResponse* cerP, const std::string& q)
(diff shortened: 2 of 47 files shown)


test/functionalTest/cases/0350_onanychange_subscriptions/onanychange_empty_cond_values.test
```diff
@@ -0,0 +1,422 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: The 4 tests included in this PR are almost identical. Thus, you can review just one and check in the others just the following bits:

* brokerStart  line uses -noCache in the proper files
* That condValues is `[]` or has been omitted in the proper files.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1485,6 +1507,26 @@ void processOntimeIntervalCondition(const std::string& subId, int interval, cons
 }
 
 
+
+/* ****************************************************************************
+*
+* matchExpression
+*/
+static bool matchExpresion(ContextElementResponse* cerP, const std::string& q)
```
kzangeli: matchExpression   (one 's' missing)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1966,6 +1966,63 @@ bool isCondValueInContextElementResponse(ConditionValueList* condValues, Context
 }
 
 
+
+/* ****************************************************************************
+*
+* someEmptyCondValue -
+*
+* This logic would be MUCH simpler in the case conditions was a single field instead of a vector.
```
kzangeli: Is this something we are considering?
Otherwise, the comment is not very interesting ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in 72c5ebc

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Yes, why don't? In the future, when ONTIMEINTERVAL gets removed the data model for csubs document could be simplified, to avoid the conditions vector. Thus, I think the comment is relevant and should be kept.

kzangeli: ok, sure.

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1206,6 +1206,10 @@ static bool addTriggeredSubscriptions_withCache
                                                            aList,
                                                            cSubP->subscriptionId,
                                                            cSubP->tenant);
+
+    // FIXME P10 #1316: expression needs to be stored in the cache
```
kzangeli: P10 ...
Is this something you marked to not forget and then forgot?
Looks important enough to be included in this PR ...   ?

fgalan: I think it is better to be included in a next PR.

doc/manuals/user/walkthrough_apiv1.md
```diff
@@ -1020,6 +1020,8 @@ Let's examine in detail the different elements included in the payload:
     notified of the value of *temperature* each time the value of
     *pressure* changes may not be too useful. The example is chosen this
     way only to show the enormous flexibility of subscriptions.
+-   You can leave empty the condValue list (or even omit) to make notification
```
kzangeli: You can leave the condValue list empty (or even omit it) to make a notification trigger on any entity attribute change (regardless of the name of the attribute)

fgalan: Fixed in e3d4339

src/lib/mongoBackend/MongoCommonUpdate.cpp

kzangeli:
LGTM

PR #1691, opened 2016-01-19
Subscription q filter evaluation for initial notification in both create and update subscriptions in NGSIv2

fgalan:
Implementing part of #1316 (and in some way, a missing part of #1358).

@kzangeli 
@crbrox


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2347,6 +2354,13 @@ bool processOnChangeConditionForSubscription
   Restriction                   res;
   ContextElementResponseVector  rawCerV;
 
+  /* Include scopes for entitiesQuery() if q filter is not empty */
+  if (qFilter != "")
+  {
+    Scope* sc = new Scope(SCOPE_TYPE_SIMPLE_QUERY, qFilter);
+    res.scopeVector.push_back(sc);
+  }
```
fgalan: I'm not explicitely releaseing the memory allocated with the `new` in this block, but I understand that as part of the Restriction release process is automatically freed. Or maybe I'm wrong?

(Anyway, check-leak jenkins jobs will warn if I'm wrong)

src/lib/mongoBackend/dbConstants.h
```diff
@@ -76,11 +76,11 @@
 #define CSUB_CONDITIONS         "conditions"
 #define CSUB_CONDITIONS_TYPE    "type"
 #define CSUB_CONDITIONS_VALUE   "value"
-#define CSUB_CONDITIONS_EXPR    "expression"
-#define CSUB_CONDITIONS_Q       "q"
-#define CSUB_CONDITIONS_GEOM    "geometry"
-#define CSUB_CONDITIONS_COORDS  "coords"
-#define CSUB_CONDITIONS_GEOREL  "georel"
+#define CSUB_EXPR               "expression"
```
fgalan: I have take the opportunity of this PR to do a global renaming for names closer to how the fields are in the DB model.

NTC

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -245,8 +245,36 @@ HttpStatusCode mongoUpdateContextSubscription
...
+         attrL = subToAttributeList(sub);
+       }
+       else // v2
+       {
+         // In v2 entities and attribute are updatable (as part of subject or notification) wo
+         // we have to check in order to know if we get the attribute from the request or from
+         // the subscription
+         if (requestP->entityIdVector.size() > 0)
+         {
+           enV = requestP->entityIdVector;
```
fgalan: Double-free potential problem? I mean, enV releasing and requestP releasing end releasing the same piece of allocated dynamic memory.

Let's see how test harness behave regarding this (double-free will cause segfault in some causes, so some tests should faild).

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: It should be released automagically

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2338,7 +2344,8 @@ bool processOnChangeConditionForSubscription
   Format                           format,
   const std::string&               tenant,
   const std::string&               xauthToken,
-  const std::vector<std::string>&  servicePathV
+  const std::vector<std::string>&  servicePathV,
+  const std::string&               qFilter
 )
 {
   // FIXME P10: we are using dummy scope at the moment, until subscription scopes get implemented
```
fgalan: Probably moment has come to remove this pretty old FIXME :)

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: The question is what exactly the C++ compiler does with this command.
My guess is that the std::vector with all its items are created and set to point to the same EntityIds.
If you use enV->release, those EntityIds will be freed and we could have a double-free problem.
If you do not call enV->release there should be no problem.

fgalan: I'm not explicitely releasing (as part of this PR) but I guess it could be done as part of the releasing of contaning classes in parts of the code not touched by this PR.

By the moment 295/565 in test harness without any fail...

fgalan: 565/565 all ok

kzangeli:
LGTM

crbrox:
LGTM

Issue #1585, opened 2015-12-02
(Only doc adjustments remain) log improvements: Log summary mode


fgalan:
Root issue: #1518 

It has been found that the current way of logging each transaction, based on Operations requirements (see https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/admin/logs.md), is too verbose in the case of high load. E.g. it has been found in some tests that with around 4000 TPS load, around 450 MB of log is created in 1.5 minutes.

Thus, the CB could implement a "summary mode" (e.g. using the `-logSummary <period>`, being `<period>` the summary generation frequency): 

```
time=... | lvl=SUMMARY | trans=N/A | serv=N/A | subServ=N/A | function=... | comp=Orion | msg=Transactions: 2132321 (new: 242)
time=... | lvl=SUMMARY | trans=N/A | serv=N/A | subServ=N/A | function=... | comp=Orion | msg=DB status: ok (created: 1, removed: 1)
time=... | lvl=SUMMARY | trans=N/A | serv=N/A | subServ=N/A | function=... | comp=Orion | msg=Bad input active alarms: 10 (created: 2, removed: 4)
time=... | lvl=SUMMARY | trans=N/A | serv=N/A | subServ=N/A | function=... | comp=Orion | msg=Notification failure active alarms: 5 (created: 5, removed: 10)

```

Where new/created/removed is the "delta" since last summary traces. Line by line:
- Transactions: the number of total transaction carried out so far is printed, along with the delta (in the "new" field) since the last report (the thread in charge of summary printing should have a thread variable to know the last printed value, so delta can be calculated).
- DB status. Just printing either "ok" or "alarmed" (or "failed"?) in the case the DB alarm is active (AlarmManager should provide a method to check this), along with the number of times DB was alarmed/released since the last summary.
- Bad input and notification failure alarms: the number of current alarms in the system of each type (AlarmManager should provide a method to get this information), along with the number of alarms created and disabled since the last summary (AlarmManger should implemented internal counters for created/removed alarms in each type, which could be retrieved from the thread in charge of printing summary and a reset method to reset the counters).

Note this traces use a new level, named SUMMARY.

fgalan:
Let's try to refine the format a little bit more. Each time the sumary thread wake ups it prints 3 lines in a new level called SUMMARY, in the following way:

```
[...]

PR #1696, opened 2016-01-20
Feature/1585 log summary

kzangeli:
Fixes issue #1585 

You will see in the tunctest REGEXPECT that more than *one* BadInput is reported for some of the errors provoked. I didn't know that before seeing the results of the functest.
This is perhaps not ideal, but it has nothing to do with the Log Summary.
If we want to fix it, it should be in another PR.
Now, about the func test, it is a bit simple. Perhaps a better functest should be provided.
These questions is what PR reviews are for ... :-)

@fgalan  
@crbrox


CHANGES_NEXT_RELEASE
```diff
@@ -16,3 +16,4 @@
 - Add: subscriptions triggered by modifications in any attribute without explicitly list them (aka ONANYCHANGE) (Issue #350)
 - Add: 'q' expression evaluation in NGSIv2 subscription (at subscription creation/update and update context times) (Issue #1316 and #1658)
 - Fix: subscription service path is lost at update subscription time (Issue #1693)
+- Add: Log summary (Issue #1585)
```
fgalan: I'd suggest to mention the name of the CLI we are using for this, maybe:

```
Add: Enable log summary traces using -logSummary CLI (Issue #1585)
```

src/app/contextBroker/contextBroker.cpp
```diff
@@ -271,6 +274,7 @@ bool            statNotifQueue;
 #define PIDPATH             _i "/tmp/contextBroker.pid"
 #define IP_ALL              _i "0.0.0.0"
 #define LOCALHOST           _i "localhost"
+#define MAX_PERIOD          (3600 * 24 * 31)  // Once a month ...
```
fgalan: MAX_PERIOD -> MAX_SUMMARY_TRACE_PERIOD (or maybe it is used for some other thing?)

In addition, if there is a maximum value for `-logSummary`, it should be mentioned in the cli.md file.

src/lib/logMsg/logMsg.cpp
```diff
@@ -971,7 +971,7 @@ static char* timeGet(int index, char* line, int lineSize)
 }
 
 
-
+#if 0
```
fgalan: Leftover?

CHANGES_NEXT_RELEASE
kzangeli: ok.
I was never really satisfied with that line. Also not in cli.md.
The problem is that you both turn on log summary and set the interval with one single option.

src/lib/logSummary/logSummary.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include "alarmMgr/alarmMgr.h"
+#include "logSummary/logSummary.h"
+
+
+
+/* ****************************************************************************
+*
+* Global vars
+*/
+static sem_t               lsSem;
```
fgalan: The same way this sem_t is named lsSem, maybe the sem_t for alarm manager should be named amSem.

src/lib/logSummary/logSummary.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+  if (logSummaryOn == false)
+  {
+    return;
+  }
+
+  sem_wait(&lsSem);
+  ++transactionCounter;
+  ++transactionsSinceLastSummary;
+  sem_post(&lsSem);
```
fgalan: I would propose a different strategy:

* logSummary keeps only an internal counter, named `transactionCounterForSummary` (or a better name) initialized to 0
* each time the thread wakes up to print, you get the current value of `transaction` (I refer to the transaction counter, a static int defined in transactionIdSet). Given that this is a read operation, you don't need to wait on transaction semaphoere (i.e. the one managed with transSemTake/transSemGive) for this.
* You use for the trace log the difference `transaction - transactionCounterForSummary`
* Update the `transactionCounterForSummary` with `transaction` for the next round
* Sleep until next wake up.

I think this will lead to a simpler code, as there is no need of using transactionCounterIncrement() from rest.cpp and httpRequestSend.cpp

In addition, probably there isn't a need for lsSem, as no thread (appart from the one managing log summary, of course) will touch `transactionCounterForSummary`.

(Atually, is not so simple in step 3... the difference could be negative, in the case of overflow, but it is easy to adjust in that case).

test/functionalTest/cases/1585_log_summary/log_summary.test
```diff
@@ -0,0 +1,202 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
```
fgalan: I think this .test is enough for an automated testing of this feature (I think that the real test of this will not be in the automatic test, but after actual usage of it).

Regarding "overbadinputing", I'd suggest to add a "#FIXME xx" in the header of the test (where 'xx' is the number of issue about that).

src/app/contextBroker/contextBroker.cpp
kzangeli: MAX_SUMMARY_TRACE_PERIOD ...
That is just TOO long.
It is only used in the parseargs vector, and only once.

What I could do is to define ONE_MONTH and use it.
Yes, I like that idea

src/lib/logMsg/logMsg.cpp
kzangeli: I removed the old LM_S ('S' standing for timeStamp) to make room for a new LM_S ('S' standing for Summary), as the broker doesn't use LM_S anyway.

However, for the future of the logMsg library, I prefer to keep the function timeStampGet, thus the #if 0/#endif

src/lib/logSummary/logSummary.cpp
kzangeli: Nope.
This is a global variable, the other one is a member of the class.
There is no reason to prefix a class member.
Actually as *lsSem* is static, there is no reason to prefix it either, it just hurts my eyes to see a global variable (even though it is static) without prefix.

src/app/contextBroker/contextBroker.cpp
fgalan: What about ONE_MOTH_PERIOD? ONE_MONTH could be misleaded with other things meaning one month (e.g. a string).

src/lib/logMsg/logMsg.cpp
fgalan: In that case, I'd suggest to include a comment about this #if 0 is legal.

src/lib/logSummary/logSummary.cpp
fgalan: NTC

src/app/contextBroker/contextBroker.cpp
kzangeli: Not sure what is won with that, but OK.
ONE_MONTH_PERIOD is it :-)

CHANGES_NEXT_RELEASE
kzangeli: Fixed in c858a32

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in c858a32

src/lib/logMsg/logMsg.cpp
kzangeli: Fixed in c858a32

test/functionalTest/cases/1585_log_summary/log_summary.test
kzangeli: Fixed in c858a32

src/lib/alarmMgr/AlarmManager.h
```diff
@@ -38,26 +40,44 @@ class AlarmManager
...
   int                         badInputLogSampling;
+  sem_t                       sem;
 
 public:
   AlarmManager();
   AlarmManager(int _notificationErrorLogSampling, int _badInputLogSampling);
 
+  int  init(void);
+
+  int  semInit(void);
```
fgalan: Used only internally? Maybe they should be in the `private` scope.

src/lib/common/globals.cpp
```diff
@@ -55,6 +55,34 @@ bool                   timingStatistics     = false;
...
 
 
+
+/* ****************************************************************************
+*
+* transactionIdGet - 
+*
+* Unless readonly, add one to the transactionId and return it.
+* If readonly - just return the current transactionId.
+* If the counter has gone 'round-the-corner', return this info in the output
```
fgalan: L65-L66 to be removed (now obsolete comment)

src/lib/logSummary/logSummary.cpp
```diff
@@ -0,0 +1,143 @@
...
+    long long diff;
+
+    sleep(period);
+
+    transactionsNow     = transactionIdGet();
+
+    // Has transactionId gone 'round-the-corner'?
+    if (transactionsNow < transactionsAtLastSummary)
+    {
+      diff = 0x7FFFFFFF - transactionsAtLastSummary + transactionsNow;
```
fgalan: As dicussed, 0x7FFFFFFF should be moved to a define,

In addition, note that transactionsNow, transactionsAtLastSummary, diff, etc. are "long long", while transactionId is (signed) int. What is 0x7FFFFFFF? Max "long long" or max (signed) int?

kzangeli: transactionId is an int - 32 bits => 0x7FFFFFFF is max value (for signed int).
A long long (64 bits): 0x7FFFFFFFFFFFFFFF  (one '7' and 15 'F's)

src/lib/alarmMgr/AlarmManager.h
kzangeli: sem methods could probably go in the "private sector" :-)
init() is used outside.   I'll test moving the sem methods

src/lib/logSummary/logSummary.cpp
kzangeli: Fixed in b525e5f
Finally, INT_MAX from /usr/include/limits.h is used.

src/lib/common/globals.cpp
kzangeli: Fixed in b525e5f

src/lib/alarmMgr/AlarmManager.h
kzangeli: Fixed in b525e5f
Only semInit can be private

fgalan:
LGTM

PR #1698, opened 2016-01-20
Syntax for id fields in NGSIv2 specification

fgalan:
To be checked by:

@mrutid
@jmcanterafonseca
@joser
@antoniojoseln


mrutid:
OK

PR #1701, opened 2016-01-21
FIX ops on /v2/entities to take options=keyValues correctly into account

fgalan:
@kzangeli


test/functionalTest/cases/1444_err_msg_duplicate_attr/create_entity_dup_attr.test
```diff
@@ -31,6 +31,7 @@ brokerStart CB
 
 #
 # 01. POST /v2/entities, creating first entity with id E1
+# FIXME P10: obsolete preamble? steps need update?
```
fgalan: What do you think?

test/functionalTest/cases/0987_put_attributes/put_entity_error.test
```diff
@@ -55,8 +55,9 @@ echo
 
 echo "03. Create E1"
 echo "============="
+# FIXME P10: This creation is not legal! Maybe related with issue #1682
```
fgalan: I have spotted this (and other similar ones) when modifying the .test... The point is that `POST /v2/entities/E1` can be used to append or update entity attributes, but not for creating new attributes. I'd review (and eventually fix in a PR) after this PR.

NTC

kzangeli: Yeah, I believe I've seen in the code as well that /v2/entities/{eid} is able to create entities.
Probably a remnant from the old way of working.

test/functionalTest/cases/1444_err_msg_duplicate_attr/create_entity_dup_attr.test
kzangeli: This whole test seems unfinished. Lots of mistakes ...
It is called
  GET Entity V2 Too Many Entities

So, I imagine the idea for it is to create two entities with the same ID but different types and then do a GET /v2/entities/{ID} and see an error.

Easy to fix.
The test in its current state is pretty useless.

For a new PR ?

fgalan: Let's keep the FIXME and complete in a next PR.

NTC

kzangeli:
LGTM

PR #1706, opened 2016-01-21
FIX POST /v1/entities/{id} not allowed to create entities

fgalan:
@kzangeli 
@crbrox

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2935,7 +2935,7 @@ void processContextElement
   std::map<std::string, std::string>&  uriParams,   // FIXME P7: we need this to implement "restriction-based" filters
   const std::string&                   xauthToken,
   const std::string&                   apiVersion,
-  bool                                 checkEntityExistance
+  const std::string&                   ngsiv2Flavour
 )
 {
   /* Check preconditions */
@@ -2998,9 +2998,7 @@ void processContextElement
 
   auto_ptr<DBClientCursor> cursor;
 
-  // This block is to avoid that several entities with the same ID get updated at the same time, which is
-  // not allowed in NGSIv2. The if is needed, as multiple update has been allowed in NGSIv1 (maybe without
-  // thinking too much about it, but NGSIv1 behaviour now will break backward compatibility)
+  // Several checkings related with NGSIv2
   if (apiVersion == "v2")
   {
     unsigned long long entitiesNumber;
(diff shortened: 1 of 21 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2935,7 +2935,7 @@ void processContextElement
   std::map<std::string, std::string>&  uriParams,   // FIXME P7: we need this to implement "restriction-based" filters
   const std::string&                   xauthToken,
   const std::string&                   apiVersion,
-  bool                                 checkEntityExistance
+  const std::string&                   ngsiv2Flavour
```
fgalan: At the end, checkEntityExistance was a bit limitted to achieve specific functionality at mongoBackend update for some update cases, so I opted for thiw "flavour" thing.

NTC

crbrox: Maybe a enum type could be "safer" than a string ...

fgalan: Fixed in d23cb2e

kzangeli: Actually, there is a RequestType, created in the beginning of the project that I think does just this.
Common code that needs to behave slightly differently depending on the Type of Request

kzangeli: Here: src/lib/ngsi/Request.h

typedef enum RequestType ...

fgalan: The idea is that flavours are not directly linked with request types, but with adjustable behaviours at mongoBackend, i.e. it could happen that the same operation could involve different flavour in the future or that two operations use the same flavour. Let's see how it evolves in the future but, by the moment, I prefer to have them as a separate thing.

Having said that, I don't like this flavour-based approach (in the same way I don't like the `apiVersion` branching in mongoBackend) but it is a "necessary evil" to survive without a complete re-work of the mongoBackend while NGSIv1 coexist with NGSIv2. In a situation in which NGSIv2 will be the only API version and the data types match the API, this is going to be removed.

kzangeli: In that case ...

NTC

kzangeli:
LGTM

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Maybe we can can add is a FIXME message on Ngiv2Flavour enum about the possibility of using Request enum in the future, depending on how the flavour things evolved in the future.

fgalan: Comment added in 78a456d6bf0d2d610b490e0ae9a9fb276fe7f209

crbrox:
LGTM

PR #1721, opened 2016-01-26
Fixed leaks

kzangeli:
valgrind test suite without leaks for 0.70 release.
As I was inside parseSubscription.cpp fixing leaks, I fixed style guide also.
I know ... sorry.
Only three important changes there, I will mark them for the reviewers

### DIFF
--- a/src/lib/jsonParseV2/parseSubscription.cpp
+++ b/src/lib/jsonParseV2/parseSubscription.cpp
@@ -34,10 +34,11 @@
 #include "ngsi/ParseData.h"
 #include "ngsi/Request.h"
 #include "jsonParseV2/jsonParseTypeNames.h"
-
+#include "jsonParseV2/jsonRequestTreat.h"
 #include "jsonParseV2/parseSubscription.h"
 
 using namespace rapidjson;
+
 
 
 /* Prototypes */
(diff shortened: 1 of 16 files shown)


src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -48,40 +49,43 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
...
+  Document                  document;
+  SubscribeContextRequest*  destination;
 
   document.Parse(ciP->payload);
 
   if (update)
   {
-    destination = &parseDataP->ucsr.res;
+    destination     = &parseDataP->ucsr.res;
+    releaseP->ucsrP = &parseDataP->ucsr.res;
```
kzangeli: releaseP->ucsrP is set to point to &parseDataP->ucsr.res, so that it will be released after usage

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -48,40 +49,43 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
...
   }
   else
   {
-    destination = &parseDataP->scr.res;
+    destination    = &parseDataP->scr.res;
+    releaseP->scrP = &parseDataP->scr.res;
```
kzangeli: releaseP->scrP is set to point to &parseDataP->scr.res, so that it will be released after usage

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -256,9 +265,11 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
 
       return oe.render(ciP, "");
     }
-    EntityId* eiP         = new EntityId();
```
kzangeli: I moved this 'new EntityId' down, to avoid to call 'delete' in every 'return caused by error'

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -313,10 +323,12 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
       }
     }
 
+    EntityId*  eiP = new EntityId(id, type, isPattern);
```
kzangeli: Here iot is now.
[ I created the var 'type' to keep the entity type ]

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1664,13 +1662,10 @@ static bool processSubscriptions
...
     }
-
-    /* Release object created dynamically (including the value in the map created by addTriggeredSubscriptions */
-    trigs->attrL.release();
-    delete trigs;
   }
 
-  subs.clear();
+  releaseTriggeredSubscriptions(subs);
+
```
kzangeli: What you see above is the fix to the 'strange' leak about triggered subscriptions.
Instead of deleting/releasing inside the loop, I call releaseTriggeredSubscriptions after the loop.
Don't ask me why the old way stopped working. I don't have a clue.
What I **do** know is that the new way works :-)

fgalan:
Hard to see the actual changes mixed with check style fixes... but if this achieves a clean `make valgrind`, I guess that LGTM.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Not sure if releaseTriggeredSubscriptions() does more things than just delete, but if this fixes leaks and no harness test is affected in the usual `make test`, then we can assume that everything is ok with the change (I guess).

crbrox:
LGTM

Issue #1720, opened 2016-01-26
[BUG] in GET /v2/entities using "id" query param, is used as regular expression


iariasleon:
We have only 5 entities with several ids:
 ```
   "type": "car"
   "id": "vehicle_0",  "vehicle_1",  "vehicle_2",  "vehicle_3",  "vehicle_4",
```
And in GET Request using `id=vehicle` query param, it should returns an empty list. Now returns 2 entities because `offset=3` is used. 

#### Create Request (five times, replace "id" by vehicle_0, ..., vehicle_4)
```
POST http://localhost:1026/v2/entities
 Content-Type: application/json
 Fiware-Service: test_happy_path
 Fiware-ServicePath: /test
 {"speed_1": {"type": "km/h", "value": "89", "metadata": {"very_hot_0": {"type": "alarm", "value": "hot"}}}, "speed_0": {"type": "km/h", "value": "89", "metadata": {"very_hot_0": {"type": "alarm", "value": "hot"}}}, "type": "car", "id": "vehicle_N"}
```

#### Get Request
GET http://qa-orion-fe-02:1026/v2/entities?type=car&limit=9&**id=vehicle**&offset=3
```
 Content-Type: application/json
 Fiware-Service: test_happy_path
 Fiware-ServicePath: /test
```

##### Get Response 
```
 http code: 200
 content-length: 681
 content-type: application/json
 [
    {"id":"vehicle_3","type":"car","speed_0":{"type":"km/h","value":"89","metadata":{"very_hot_0":{"type":"alarm","value":"hot"}}},"speed_1":{"type":"km/h","value":"89","metadata":{"very_hot_0":{"type":"alarm","value":"hot"}}}},
    {"id":"vehicle_4","type":"car","speed_0":{"type":"km/h","value":"89","metadata":{"very_hot_0":{"type":"alarm","value":"hot"}}},"speed_1":{"type":"km/h","value":"89","metadata":{"very_hot_0":{"type":"alarm","value":"hot"}}}}
 ]

```

##### Expected Response 
```
[0]
```

**Comment by fgalan** [comment]
monboBackend could be related with this. I'm asigned to me to research on it.

**Comment by fgalan** [comment]
Fixed in PR #1727. Please @iariasleon have a look in order to see if this can be closed.

**Comment by iariasleon** [comment]
Re-tested and continues the same error

**Comment by iariasleon** [comment]
LGTM
```
http code: 200
 content-length: 2
 content-type: application/json
[...]

Issue #1428, opened 2015-10-27
[BUG] in /v2/entities/<id>/attrs/<attr_name> with errors found in incoming JSON buffer


iariasleon:
These datasets are replaced by attribute value in update request without doble quote ""

#### Datasets:
```
| rwerwer          |
| True             |
| TRUE             |
| False            |
| FALSE            |
| 34r              |
| 5_34             |
| ["a", "b"        |
| ["a" "b"]        |
| "a", "b"]        |
| ["a" "b"}        |
| {"a": "b"        |
| {"a" "b"}        |
| "a": "b"}        |
```

#### Create request:
```
POST http://localhost:1026/v2/entities
Content-Type: application/json
Fiware-Service: test_update_attribute_value_special
Fiware-ServicePath: /test
{"temperature_2": "45", "temperature_0": "45", "temperature_1": "45", "type": "house", "id": "room_13"}
```

#### Update request:
```
PUT http://localhost:1026/v2/entities/room_13/attrs/temperature_0
Content-Type: application/json
Fiware-Service: test_update_attribute_value_special
Fiware-ServicePath: /test
{ "value": {"a" "b"} }
```

#### Update response:
```
http code: 400
{    
     "error":"BadRequest",
     "description":"Errors found in incoming JSON buffer"
}
```

#### Expected:
```
http code: 400
{    
     "error":"ParseError",
     "description":"Errors found in incoming JSON buffer"
}
```

raconte:
Proved with other examples, and continuos returning the same error "BadRequest" instead of "ParseError".

raconte:
#### Create entity
```
(curl localhost:1026/v2/entities --header Content-Type:application/json -d @- | python -mjson.tool)<< EOF
{
	"temperature_2": {
		"value": 40
	},
	"temperature_1": {
		"value": 40
	},
	"temperature_0": {
		"value": 40
	},
	"type": "house",
	"id": "room_13"
}
EOF
```

#### PUT /v2/entities/room_13/attrs/temperature_0
```
(curl localhost:1026/v2/entities/room13/attrs/temperature --header Content-Type:application/json -X PUT -d @- | python -mjson.tool )<< EOF 
{
	"value": {
		"a"
		"b"
	}
}
EOF
```

#### Returned
```
{
    "description": "Errors found in incoming JSON buffer",
    "error": "BadRequest"
}
```

"BadRequest" instead of "ParseError".

fgalan:
[...]

Issue #1286, opened 2015-09-24
NGSIv2 error generation rules


fgalan:
NGSIv2 error reporting is as follows:

`(<HTTP code> - <HTTP error text> { "error": "<payload error text>", ... })`
- JSON payload error (pe. the JSON misses a `;`, or `}`, etc.): **400 - Bad Request {"error": "ParseError", ...}**
- Request error which only depend on the request itself (i.e. they don't depend on the CB state), either in the URL parameters (e.g. `coords` format is invalid) or in the payload _(note1)_ (e.g. a `id` is missing in a create entity operation): **400 - Bad Request {"error": "BadRequest, ... }**
- Ambiguety due to the request may refer to several resources, e.g. updating an entity providing only its ID with several entities existing in the DB with the same ID (but different types). **409 - Conflict {"error": "TooManyResults", ...}**
- The resource identified by the request (e.g. entity, subscription) is not found: **404 - Not Found {"error": "NotFound", ...}**
- Errors due to the request + state combination but not exclusively from the request (e.g. POST with `op=append` on an existing attribute) _(note2)_: **422 - Unprocessable Entity {"error": "Unprocesable", ...}**
- HTTP layer errors
  - **411 - Length Required {"error": "LengthRequired", ... }**
  - **413 - Unsupported Media Type {"error": "RequestEntityTooLarge", ... }**
  - **415 - Payload Too Large {"error": "UnsupportedMediaType", ... }**
  - Maybe other may arose (e.g. Method Not Allowed), but they are quite straighforward as they derive from the standard HTTP protocol operation.

_Note1:_ Except the ones due to JSON parser error, already considered in the previous bullet.
_Note2:_ Except the request + status conditions that lead to 404 or 409 errors described in the previous bullets.

This issue will be open while NGSIv2 is still on development, in order to check at the end that all the error match the above rules. In additon, the Error Responses in the .apib document should be edited consecuently.

fgalan:
The description in the issue body has been adapted and include in NGSIv2 specification in PR https://github.com/telefonicaid/fiware-orion/pull/3226

PR #1727, opened 2016-01-27
FIX issues related with NGSIv2 refactor (#1720 and #1428)

fgalan:
Fixes:

* #1720 
* #1428

No CNR entry needed (already covered by the entry about #1259)

@crbrox 
@kzangeli


src/lib/jsonParseV2/parseAttributeValue.cpp
```diff
@@ -49,7 +49,8 @@ std::string parseAttributeValue(ConnectionInfo* ciP, ContextAttribute* caP)
   if (document.HasParseError())
   {
     alarmMgr.badInput(clientIp, "JSON parse error");
-    oe.fill(SccBadRequest, "Errors found in incoming JSON buffer");
+    oe.reasonPhrase = ERROR_STRING_PARSERROR;
```
fgalan: This could be done with a new fill() method, but not sure if just one case (this one) justifies the creation of the new method...

src/lib/rest/HttpStatusCode.h
```diff
@@ -28,6 +28,8 @@
 * Author: developer
 */
 
+#define ERROR_STRING_PARSERROR  "ParseError"
```
fgalan: Not sure if this is the best place for this #define. I have put in HttpStatusCode.h because in some sense it is related with responses. 

Alternatives places are welcome!

test/functionalTest/cases/1259_new_json_for_v2/put_attribute_value.test
```diff
@@ -307,7 +307,7 @@ Date: REGEX(.*)
 
 {
     "description": "Errors found in incoming JSON buffer",
-    "error": "BadRequest"
```
fgalan: This is step 2, which execution is:

```
echo '02. PUT /v2/entities/E1/attrs/A1/value with erroneous JSON payload'
echo '=================================================================='
payload='{ "A1" }'
orionCurl --url /v2/entities/E1/attrs/A1/value --payload "$payload" -X PUT --json
echo
echo
```

It is not exactly the same one that @raconte uses at https://github.com/telefonicaid/fiware-orion/issues/1428#issuecomment-173985143 but it should work...

src/lib/jsonParseV2/parseContextAttribute.cpp
```diff
@@ -258,7 +258,7 @@ std::string parseContextAttribute(ConnectionInfo* ciP, ContextAttribute* caP)
 
   if (document.HasParseError())
   {
-    OrionError oe(SccBadRequest, "Errors found in incoming JSON buffer");
+    OrionError oe(ERROR_STRING_PARSERROR, "Errors found in incoming JSON buffer");
```
fgalan: I have a global search to use "ParserError" in all cases where "Errors found in incoming JSON buffer" are reported (as it has to be, according to the error rules defined at https://github.com/telefonicaid/fiware-orion/issues/1286)

test/functionalTest/cases/1720_get_entities_by_id/get_entities_by_id.test
```diff
@@ -0,0 +1,166 @@
...
+Get entities by id parameter
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+# The bug that this test covers was that step 4 returned vehicle1,
+# vehicle2 and vehicle 3 entities. Steps 5 and eyond are hardening of
```
crbrox: beyond ?

src/lib/jsonParseV2/parseAttributeValue.cpp
kzangeli: Or simply use the old fill() but with this new define:

    oe.fill(SccBadRequest, ERROR_STRING_PARSERROR);

Isn't that what you want?

BTW, what about oe.code?  Not calling fill, just setting the reasonPhrase might not be enough ...

test/functionalTest/cases/1720_get_entities_by_id/get_entities_by_id.test
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/2f0eba0415804da243b299027b415404142470b9

src/lib/jsonParseV2/parseAttributeValue.cpp
fgalan: Doesn't work that way:

```
OrionError::OrionError(HttpStatusCode _code, const std::string& _details)
{
  code          = _code;
  reasonPhrase  = httpStatusCodeString(code);
  details       = _details;
}
```

The reasonPhrase  is automatically derived from SccBadRequest (i.e. something like "Bad Request"), but in this case I want to use a different one. And I don't what details be "ERROR_STRING_PARSERROR", but "Errors found in incoming JSON buffer".

kzangeli: Yeah, I know. We "fixed" the text.
However, you still have no oe.code. You might need one.

To "fix" your problem, we could add a third parameter to the constructor, and default it to NULL:

    OrionError::OrionError(HttpStatusCode _code, const std::string& _details, const char* reason = NULL)
    {
        code          = _code;
        reasonPhrase  = (reason == NULL) httpStatusCodeString(code) : reason;
        details       = _details;
    }

src/lib/rest/OrionError.cpp
```diff
@@ -60,12 +60,12 @@ OrionError::OrionError(HttpStatusCode _code, const std::string& _details)
...
 */
-OrionError::OrionError(HttpStatusCode _code, std::string& _details)
+OrionError::OrionError(const std::string _reasonPhrase, const std::string& _details)
 {
-  code          = _code;
-  reasonPhrase  = httpStatusCodeString(code);
+  code          = SccNone;
+  reasonPhrase  = _reasonPhrase;
   details       = _details;
 }
```
kzangeli: So, you remove the code, setting it no SccNone always?

Not sure this is a good idea ...

fgalan: I agree. Probably what is a but idea is a `OrionError::OrionError(const std::string _reasonPhrase, const std::string& _details)` constructor, without code. I'll have a second pass to OrionError constructors, trying to fix this.

src/lib/jsonParseV2/parseAttributeValue.cpp
fgalan: Fixed in fa4ed7edfa832f045fe123acfe0ceab117c1485c.

src/lib/rest/HttpStatusCode.h
fgalan: NTC (I guess)

test/functionalTest/cases/1259_new_json_for_v2/put_attribute_value.test
fgalan: NTC

src/lib/jsonParseV2/parseContextAttribute.cpp
fgalan: NTC (just informative)

src/lib/rest/OrionError.cpp
fgalan: Fixed in fa4ed7edfa832f045fe123acfe0ceab117c1485c

src/lib/rest/HttpStatusCode.h
kzangeli: Yes,  NTC

kzangeli:
LGTM

PR #1723, opened 2016-01-27
Fixed some minor errors with bad format JSON

raconte:
Fixed some errors with the JSON format in fiware-ngsiv2-reference.apib file

PR #1730, opened 2016-01-27
FIX JSON payloads in NGSIv2

fgalan:
Almost the same than https://github.com/telefonicaid/fiware-orion/pull/1723, this time checking with local apiary client that the final result renders.


kzangeli:
LGTM

Issue #1592, opened 2015-12-02
[BUG] in GET /v2/entities with "q" query parameter returns wrong results


iariasleon:
#### Create Requests
```
POST http://localhost:1026/v2/entities
Fiware-Service: qp_with_binary
Fiware-ServicePath: /test
{"id": "room_6", "my_value": 75}
```

```
POST http://localhost:1026/v2/entities
Fiware-Service: qp_with_binary
Fiware-ServicePath: /test
{"id": "room_5", "my_value": 45}
```

#### Mongo doc:
```
{ "_id" : { "id" : "room_5", "servicePath" : "/test" }, "attrNames" : [ "my_value" ], "attrs" : { "my_value" : { "type" : "", "creDate" : 1449084032, "modDate" : 1449084032, "value" : 45 } }, "creDate" : 1449084032, "modDate" : 1449084032 }
{ "_id" : { "id" : "room_6", "servicePath" : "/test" }, "attrNames" : [ "my_value" ], "attrs" : { "my_value" : { "type" : "", "creDate" : 1449084042, "modDate" : 1449084042, "value" : 75 } }, "creDate" : 1449084042, "modDate" : 1449084042 }
```

#### Get Request
```
GEThttp://localhost:1026/v2/entities?q=my_value>=
Fiware-Service: qp_with_type
Fiware-ServicePath: /test
```

#### Dataset
```
q=my_value>=
q=my_value>
q=my_value!=
```

#### Get Response
```
Status: 200 OK
[{"id":"room_5","my_value":45.000000},{"id":"room_6","my_value":75.000000}]
```

kzangeli:
It might be that the parser of 'q' has a stupid bug doing that >= is interpreted as BIGGER THAN '=', instead of 'BIGGER OR EQUAL TO' ...

iariasleon:
I think that if a binary expression did not have the three elements (attribute, operator and value) in that order, the parser should return an error, the same for the unary expressions, which should have two elements (operator and type or attribute) and determine if the unary operator is allowed only in front or behind or in front no matter, if so it should be in the documentation.

raconte:
1.-Create entity 
(curl localhost:1026/v2/entities -s -S --header Content-Type:application/json --header 'Accept: application/json' -d @- | python -mjson.tool)<< EOF
{
  "type": "Apartment",
  "id": "Floor3",
  "temperature": {
    "value": 24.4
  }
}  
EOF
 
[...]

Issue #1705, opened 2016-01-21
Storing q, georel, gemotry, coord as objects instead of string in csubs cache entries


fgalan:
(Spin-off from https://github.com/telefonicaid/fiware-orion/issues/1316#issuecomment-172894582)

it could be a good idea not storing `q` as string but as a object (of a class to be defined) in the cache entries. Thus, parsing will take place only one time (i.e. the moment in which the cache entry is populated) instead of each time a new update context arrives and the expression from the cache entry has to be evaluated. 

The same idea could be extended to the other three: georel, geometry and coord (not currently used but they will be in the near future).

Probably this issue should be completed once georel, geometry and coords got implemented for subscriptions filters.

**Effort: 2 man day**

fgalan:
We have realized (see https://github.com/telefonicaid/fiware-orion/pull/1735/files#r51107410) that this task is needed not only from a point of view of cache efficiency: it is also needed to detect errors in the query string.

Thus, the q string -> ocjects parsing should be done at the parsing layers, passing the information to mongoBackend in some way (probably as a new field in the corresponding NGSI class object).

This should be "must" tech debt. Chaning.

fgalan:
q part of the issue implemented in PR #1900

fgalan:
Part about geo-stuff pending.

PR #1735, opened 2016-01-28
bug/1592_query_with_q_op_greater_or_equal

kzangeli:
Fixed issue #1592 
1592: return error for q with GTE without right-hand.
q with LTE without right-hand also fixed

Added two 'future functests', for issues 1587 and 1589, to be fully incorporated once the issue 1675 is fixed.


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1282,9 +1294,9 @@ static void addBsonFilter
   }
   else
   {
-    std::string details = std::string("unknown query operator: ") + opr;
+    std::string details = std::string("unknown query operator: '") + opr + "'";
```
fgalan: Is this `details` used in some way in the response payload to client? Note that `'` is a forbidden char....

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1092,7 +1092,7 @@ static bool matchFilter
 * Add a BSON filter based on a "q token".
 *
 */
-static void addBsonFilter
+static bool addBsonFilter
```
fgalan: This solution sufices for the case of q expressions in the GET /v2/entities. However, q expressions in subscriptions will be also problematic. 

This fact support the idea of do filter parsing at early stages, both for GET /v2/entities and subscriptions, and have the data as objects (i.e. https://github.com/telefonicaid/fiware-orion/issues/1705)

NTC (just a thought)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1282,9 +1294,9 @@ static void addBsonFilter
...
   else
   {
-    std::string details = std::string("unknown query operator: ") + opr;
+    std::string details = std::string("unknown query operator: '") + opr + "'";
     alarmMgr.badInput(clientIp, details);
-    pushBackFilter = false;
+    return false;
   }
 
   if (pushBackFilter)
```
fgalan: Given the early-return which formarly corresponds to `pushBackFilter = false`, this if could be removed, replaced by just: 

```
filters.push_back(f);
```

As side effect, the `pushBackFilter` variable is no longer needed and can be completely removed from this function.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1647,7 +1665,16 @@ bool entitiesQuery
     }
     else if (sco->type == SCOPE_TYPE_SIMPLE_QUERY)
     {
-      qStringFilters(sco->value, filters);
+      if (qStringFilters(sco->value, filters) != true)
```
fgalan: Would be useful to add a `FIXME #1705` above this line to explain that this checking will be removed once that issue will be addressed?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: As far as I know, it is not used. Might be used in the future though.
Change for /  ?

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I only changed this part, I believe there are more instances of **pushBackFilter = false**.
I can fix this though if you want.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Sure

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I think there aren't more instances. If that is correct, please go ahead with the fix.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Ok.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I was wrong, it was in only one place.
Fixed in next commit.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in f2b6eb3

test/functionalTest/cases/1587_q_query_has_type/q_query_has_type.test
```diff
@@ -0,0 +1,192 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: Although I undesrtand that having the test is working as it is now due to some steps are commented, I would suggest to rename it to .pending1675.DISABLED in order to make explicit when this test could be "recovered".

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in f2b6eb3

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in f2b6eb3

test/functionalTest/cases/1587_q_query_has_type/q_query_has_type.test
kzangeli: Fixed in b069933

fgalan:

crbrox:
LGTM

Issue #1607, opened 2015-12-09
[BUG] in GET /v2/entities with "q" query parameter using <=, >=, >, < operators do not return error


iariasleon:
in GET /v2/entities with "q" query parameter using <=, >=, >, < operators and a range do not return error.

`<=` and `<` operators behave like `!=`
`>=` and `>` operators behave like `==`

#### Create request:
```
POST http://localhost:1026/v2/entities
Fiware-Service: happy_path
Fiware-ServicePath: /test
{"id": "room_1", "attr": 45}
```

#### Get requests
```
GET http://qa-orion-fe-02:1026/v2/entities?q=attr>=40..47
```
```
GET http://qa-orion-fe-02:1026/v2/entities?q=attr<=46..47
```
```
GET http://qa-orion-fe-02:1026/v2/entities?q=attr>40..47
```
```
GET http://qa-orion-fe-02:1026/v2/entities?q=attr<46..47
```

#### Responses
```
Status: 200 OK
[{"id":"room_1","attr":45.000000}]
```

#### Expected
```
http code: 400 Bad Request
{ 
   "error":  "BadRequest",                                                                 
    "description": "<, <=, > and >= operators are not allowed with a range in query parameter"
}
```

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1736. Please @iariasleon have a look to to check if it is actually fixed.

iariasleon:
LGTM
```
http code: 400
date: Wed, 03 Feb 2016 11:49:17 GMT
content-length: 63
content-type: application/json
{"error":"BadRequest","description":"invalid query expression"}
```

PR #1736, opened 2016-01-28
Bug/1607 query with q and ranges but invalid operator

kzangeli:
Fixed issue #1607.
Sanity checks in qStringFilters, specifically, checking that if the filter contains "..", i.e. it is a range, then either '!=' or '==' must be present as well.

Not much of a check, but it works for the simple cases ...


fgalan:

crbrox:
LGTM

PR #1738, opened 2016-01-28
Fixed all of issue 1592

kzangeli:
Fixes issue #1592   (for not 3 but for 8 operators, all ops that we support :-))
Checking ALL operations for missing right-hand-value.


fgalan:
LGTM

crbrox:
LGTM

PR #1739, opened 2016-01-28
FIX update compounds with dots

### LINKED ISSUE #1712: Support JSON values for metadata
As the issue title describes.

fgalan:
Fixes #1712 

@kzangeli 
@crbrox


src/lib/mongoBackend/dbFieldEncoding.cpp
```diff
@@ -41,8 +41,9 @@
 * Replace:
 *   . => =
 */
-std::string dbDotEncode(std::string s)
+std::string dbDotEncode(const std::string& _s)
```
fgalan: Trying to enforce `const std::string&` as coding style specifies... However, I have needed the `std::string s = s_` ugly statement...

I have also tried `const std::string&` as retun value, but I haven't able to make it work. I have tried the same way `getDbPrefix()` but it doesn't work (maybe in that function it works due to it returns a static string).

In addition, I think it would be a good idea to make this an inline function (moving it to .h). That would make the .cpp be an empty file, but I understand it should be a problem.

Opinions on the above?

kzangeli: I tried to return "const std::string&" once and I didn't have any problem.
Perhaps it depends on how the return value is received ...

About inlining the funcion, sure.
I imagine the return/param types aren't that important anymore if he function in inline.

src/lib/mongoBackend/dbFieldEncoding.cpp
```diff
@@ -41,8 +41,9 @@
...
 {
+  std::string s = _s; // replace cannot be used in const std::string&
   std::replace(s.begin(), s.end(), ESCAPE_1_DECODED, ESCAPE_1_ENCODED);
   return s;
 }
```
kzangeli: Let's see, this is a function that modifies the parameter, so the parameter should be sent using its pointer, I mean, real pointer, not a C++ reference

fgalan: Not sure if I get the point... using `const std::string& _s` is a way of ensuring that `s` cannot be modified back to the caller. That's is the reason of the copy done by `s = s_`.

I mean, I don't see the need of using a pointer when `const std:string&` suffices the case (as far as I understand).

kzangeli: It depends, of course. 
Do we want to do the replacement in the very string or do we want to do it in a copy?

fgalan: Copy + modify + return.

(Note that previous versions of this function also return the modified string... probably previous design wasn't correct, given it was somewhere "in the middle" of both possiblities; what I wonder is how it could work before :).

kzangeli: The old version has a copy on the stack. That version was modified and returned

test/functionalTest/cases/1217_update_compound_with_dots/update_compound_with_dots.test
```diff
@@ -0,0 +1,110 @@
...
+
+
+echo "02. Update entitiy attribute value with dots in keys"
+echo "===================================================="
+payload='{
+  "temperature_2": {
+    "rt.ty": "5678"
+  }
+}'
+orionCurl --url '/v2/entities/apartment_3?options=keyValues' --payload "$payload" --json
```
kzangeli: temperature_2 is a compound here, so the check of "dots in keys" might not be complete (compound values are 'special').
Why not add a new attribute called "rt.ty" ?

src/lib/mongoBackend/dbFieldEncoding.cpp
kzangeli: Fixed in 8000009

kzangeli:
LGTM

crbrox:
LGTM

PR #1742, opened 2016-01-29
fixed error in addBsonFilter 'right-hand-value-empty-check

kzangeli:
Stupid mistake made during my last PR. Now fixed


crbrox:
LGTM

PR #1757, opened 2016-02-03
Fix #1387

### LINKED ISSUE #1387: [BUG] in update request /v2/entities/<ID>/attrs/<ATTR_NAME> with two entities with the same id
in update request `/v2/entities/<ID>/attrs/<ATTR_NAME>` with two entities with the same id
Ex:
#### create request Nº1
```
POST http://localhost:1026/v2/entities
Fiware-Service: test_update_more_entities
Fiware-ServicePath: /test
{"temperature_2": "34", "type": "house", "id": "room"}
```

#### create request Nº2
```
POST http://localhost:1026/v2/entities
Fiware-Service: test_update_more_entities
Fiware-ServicePath: /test
{"temperature": "67", "type": "home", "id": "room"}
```

####  Update by id and attr name 
```
PUT http://localhost:1026/v2/entities/room/attrs/temperature
Fiware-Service: test_update_more_entities
Fiware-ServicePath: /test
{"value": "80"}
```
####  Response
```
 http code: 204
 content-length: 0
```
And both entites are updated in mongo 
```
{ "_id" : { "id" : "room", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "value" : "80", "type" : "", "creDate" : 1445334741, "modDate" : 1445334741}}, "creDate" : 1445334741, "modDate" : 1445334741 }
{ "_id" : { "id" : "room", "type" : "home", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "value" : "80", "type" : "", "creDate" : 1445334741, "modDate" : 1445334741}}, "creDate" : 1445334741, "modDate" : 1445334741 }
```

#### Expected
```
http code: 409 Conflict
{ 
   "error":  "TooManyResults",                                                                 
    "description": "There is more than one entity that match the update. Please refine your query."
}
```

crbrox:
@kzangeli 
@fgalan 

fixes #1387

### DIFF
--- a/src/lib/serviceRoutinesV2/putEntityAttributeValue.cpp
+++ b/src/lib/serviceRoutinesV2/putEntityAttributeValue.cpp
@@ -26,6 +26,7 @@
 #include <vector>
 
 #include "common/statistics.h"
+#include "common/errorMessages.h"
 
 #include "rest/ConnectionInfo.h"
 #include "ngsi/ParseData.h"
@@ -83,7 +84,7 @@ std::string putEntityAttributeValue
 
   if (ciP->httpStatusCode == SccConflict)
   {
-    ErrorCode   ec("TooManyResults", "There is more than one entity that match the update. Please refine your query.");
+    ErrorCode   ec("TooManyResults", MORE_MATCHING_ENT);
     std::string answer;
 
     TIMED_RENDER(answer = ec.toJson(true));
--- a/src/lib/serviceRoutinesV2/putEntityAttributeValue.cpp
+++ b/src/lib/serviceRoutinesV2/putEntityAttributeValue.cpp
@@ -25,6 +25,8 @@
 #include <string>
 #include <vector>
 
+#include "common/statistics.h"
+
 #include "rest/ConnectionInfo.h"
 #include "ngsi/ParseData.h"
 #include "rest/EntityTypeInfo.h"
@@ -79,6 +81,15 @@ std::string putEntityAttributeValue
     }
   }
 
+  if (ciP->httpStatusCode == SccConflict)
+  {
+    ErrorCode   ec("TooManyResults", "There is more than one entity that match the update. Please refine your query.");
+    std::string answer;
+
+    TIMED_RENDER(answer = ec.toJson(true));
+
+    return answer;
+  }
 
   // 04. Prepare HTTP headers
   if ((ciP->httpStatusCode == SccOk) || (ciP->httpStatusCode == SccNone))
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -7,3 +7,5 @@
 - Fix: syntax change in string query 'q' for exist and not-exist (Issue #1751)
 - Fix: sanity check for string query 'q' - detect 'left-hand-side missing' (Issue #1754)
 - Fix: more sanity checks for string query 'q' (q empty, parts of 'q' empty - parts of 'q' are separated by ';')
+- Fix: error message when updating attribute value for two entities with same id (Issue #1387)
+ 
--- a/src/lib/apiTypesV2/Attribute.cpp
+++ b/src/lib/apiTypesV2/Attribute.cpp
@@ -26,6 +26,7 @@
 #include <vector>
 
 #include "common/tag.h"
+#include "common/errorMessages.h"
 #include "apiTypesV2/Attribute.h"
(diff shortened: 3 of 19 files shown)


fgalan:

src/lib/serviceRoutinesV2/putEntityAttributeValue.cpp
```diff
@@ -79,6 +81,15 @@ std::string putEntityAttributeValue
     }
   }
 
+  if (ciP->httpStatusCode == SccConflict)
+  {
+    ErrorCode   ec("TooManyResults", "There is more than one entity that match the update. Please refine your query.");
```
kzangeli: match the update ...
match the criteria?

Or perhaps make it simpler:

    More than one matching entity. Please refine your query.

Opinions?

fgalan: I think we have this text in other places around the code. If that is correct, I'd use a similar wording (except if these have also "typos" in the text).

kzangeli: This is what we have right now:

    getEntityAttributeValue.cpp: "There is more than one entity with that id - please refine your query"
    Entity.cpp:                  "There is more than one entity with that id. Refine your query."
    Attribute.cpp:               "There is more than one entity with that id. Refine your query."

I still prefer "More than one matching entity".

A decision to be made ...

fgalan: I like

"There is more than one entity with that id. Refine your query."

In addition, to avoid having "diperssion" in text messages, maybe it makes sense to have a #define for this (not sure in which .h file... maybe we could start a "textMessages.h" file in the common library.

kzangeli: Desperate intent of persuasion ... :-):
- What if the matching is on a pattern?
- What if the matching is per id **and type** of the entity?

"that id" would no longer be 100% valid.
"More than one matching entity" covers all the possibilities, apart from being shorter and more "elegant" ...

fgalan: Should I understand that you suggest?

"More than one matching entity. Please refine your query"

Ok for me. 

For me, more than the particular wording is to make sure the same text is used in all places where it makes sense and the best way of ensuring that is using a #define. That way, if we change our mind and decide to change, we only need to change in one place (i.e. the #define value).

crbrox: The text is copy&pasted from original issue #1387 
Not sure if it is taken from spec ...

fgalan: As far as I remember, spec only specifies the "error" text (in this case "TooManyResults") but "description" is up to the each particular NGSIv2 implementation.

Let's talk about it by skype next daily meeting.

kzangeli: More 'similar  texts':

    putEntity.cpp:                 OrionError orionError(SccConflict, "There is more than one entity that match the update. Please refine your query.");
    patchEntity.cpp:               OrionError orionError(SccConflict, "There is more than one entity that match the update. Please refine your query.");
    ContextAttributeResponse.cpp:  alarmMgr.badInput(clientIp, "more than one context element found in this query - selecting the first one");
    ContextElementResponse.cpp:    alarmMgr.badInput(clientIp, "more than one context element found the this query - selecting the first one");
    MongoCommonUpdate.cpp:         buildGeneralErrorResponse(ceP, NULL, responseP, SccConflict, "There is more than one entity that match the update. Please refine your query.");

fgalan: Ok with the 1st, 2nd and 5th.

However, 3rd and 4th should stay as they are now in my opinion, given they are not responses towards the user but an "assment" of what CB is doing internally to solve a conflictive situation.

crbrox: Hopefully fixed in ffb32cff54319862dc42a2773d6dff8c0204c0a7

kzangeli:
LGTM

fgalan:
LGTM after-new-changes

Issue #1306, opened 2015-09-30
IsDouble + string2double


kzangeli:
Now that we have a possible solution to the problem about assuring a string is a 'double', let's add a few functions to the 'common' lib:

    bool isDouble(char*);
    bool string2double(char* s, double* dP);  // return FALSE if 's' does NOT contain a double

Also, all uses of atof, strtod and similar functions must be looked up and made use these new functions instead of whatever they are using right now.

Already known cases:
Case 1: PUT .../attribute/value   with Content-Type text   (See PR #1299)
Case 2: Fermin had a similar problem not long ago - there we probably can use these new functions as well
Case X: any other uses of atof, strtod etc ...

**Effort: 1 man day**

fgalan:
Fixed in PR #1762

PR #1762, opened 2016-02-04
Hardening/1306 string2double

kzangeli:
Issue #1306 
Fixed a bug in str2double (setting errno to 0).
Replaces most of the calls to atof, atoF and strtod with calls to str2double, thus checking validity of numbers.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -6,3 +6,4 @@
 - Fix: syntax change in string query 'q' for exist and not-exist (Issue #1751)
 - Fix: sanity check for string query 'q' - detect 'left-hand-side missing' (Issue #1754)
 - Fix: more sanity checks for string query 'q' (q empty, parts of 'q' empty - parts of 'q' are separated by ';')
+- Fix: sanity checks for numbers (Issue #1306)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -8,5 +8,5 @@
 - Fix: sanity check for string query 'q' - detect 'left-hand-side missing' (Issue #1754)
 - Fix: more sanity checks for string query 'q' (q empty, parts of 'q' empty - parts of 'q' are separated by ';')
 - Fix: error message when updating attribute value for two entities with same id (Issue #1387)
-- Add: sanity checks for numbers (Issue #1306)
+- Hardening: sanity checks for numbers (Issue #1306)
 - Fix: bug causing false not-a-number when it really is a valid number (very very rarely)
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -417,6 +417,8 @@ bool string2coords(const std::string& s, double& latitude, double& longitude)
   char*  comma;
   char*  number1;
   char*  number2;
+  double newLatitude;
+  double newLongitude;
 
   cP    = wsStrip(cP);
 
@@ -434,7 +436,7 @@ bool string2coords(const std::string& s, double& latitude, double& longitude)
   number1 = wsStrip(number1);
   number2 = wsStrip(number2);
 
-  if ((str2double(number1, &latitude) == false) || (latitude > 90) || (latitude < -90))
+  if ((str2double(number1, &newLatitude) == false) || (newLatitude > 90) || (newLatitude < -90))
   {
     std::string details = std::string("bad latitude value in coordinate string '") + initial + "'";
     alarmMgr.badInput(clientIp, details);
@@ -443,7 +445,7 @@ bool string2coords(const std::string& s, double& latitude, double& longitude)
     return false;
   }
 
-  if ((str2double(number2, &longitude) == false) || (longitude > 180) || (longitude < -180))
+  if ((str2double(number2, &newLongitude) == false) || (newLongitude > 180) || (newLongitude < -180))
   {
     std::string details = std::string("bad longitude value in coordinate string '") + initial + "'";
     alarmMgr.badInput(clientIp, details);
@@ -451,6 +453,9 @@ bool string2coords(const std::string& s, double& latitude, double& longitude)
     free(initial);
     return false;
   }
+
+  latitude  = newLatitude;
+  longitude = newLongitude;
 
   free(initial);
   return true;
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -417,7 +417,6 @@ bool string2coords(const std::string& s, double& latitude, double& longitude)
   char*  comma;
   char*  number1;
   char*  number2;
-  bool   ret = true;
 
   cP    = wsStrip(cP);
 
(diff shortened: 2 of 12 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -6,3 +6,4 @@
 - Fix: syntax change in string query 'q' for exist and not-exist (Issue #1751)
 - Fix: sanity check for string query 'q' - detect 'left-hand-side missing' (Issue #1754)
 - Fix: more sanity checks for string query 'q' (q empty, parts of 'q' empty - parts of 'q' are separated by ';')
+- Fix: sanity checks for numbers (Issue #1306)
```
fgalan: > Fixed a bug in str2double (setting errno to 0).

If a bug has been solved, the probably it should be mentioned in CNR...

fgalan:
> Replaces most of the calls to atof, atoF and strtod with calls to str2double, thus checking validity of numbers.

Why don't not "all calls" instead of "most of the calls" ?

CHANGES_NEXT_RELEASE
kzangeli: It's un unknown bug, but sure, why not?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1767,11 +1865,15 @@ bool entitiesQuery
       // FIXME P4: Once issue #1705 is implemented, this check can be removed.
       if (qStringFilters(sco->value, filters) != true)
       {
-        if (badInputP)
+        if (badInputP)  // Bad Input reported by higher level
         {
           *badInputP = true;
           *err       = "invalid query expression";
         }
+        else  // There is a certain risk that this error is reported twice ...
```
fgalan: Could you elaborate in the risk (in the code comment)?

src/lib/common/string.cpp
```diff
@@ -774,11 +750,12 @@ std::string servicePathCheck(const char* servicePath)
 * It IS a double, but this function will say it is not, as it actually is not a valid double for
 * this computer as the computer cannot represent it as the C builtin type 'double' ... OK!
 */
-bool str2double(char* s, double* dP)
+bool str2double(const char* s, double* dP)
 {
-  char*   rest = s;
+  char*   rest = (char*) s;
```
crbrox: I think this works with 
```char *rest = NULL; ```
avoiding casting out const-ness

**Comment by kzangeli** [comment]
In some places it just doesn't seem necessary.
Look at this method, for example:

    double Point::longitude(void) const
    {
        return atof(_longitude.c_str());
    }

Before _longitude is accepted in the Point object, it is checked. No more checks are really necessary. 
Also, it I would replace atof for str2double in this method, the return type could no longer be 'double', it would have to be 'bool'. And all calling places must be modified, of course. A "big" change without anything gained.

In total there are four calls left to atof (atoF and strtod are no longer used, except str2double that calls strtod, of course);

Three methods like Point::longitude and one more call to atof, in the function parse8601.
That function is not 100% perfect exactly.

We might add an issue to fix it.

File: src/lib/common/string.cpp (hunk 3 above)
Comment (kzangeli) [inline src/lib/common/string.cpp]: But ...   I need **rest** to point to **s**, not NULL  ...

File: src/lib/mongoBackend/MongoGlobal.cpp (hunk 2 above)
Comment (kzangeli) [inline src/lib/mongoBackend/MongoGlobal.cpp]: Well, every time we call alarmMgr.badInput, there is a risk that it is already called (either in a lower level, or in a higher level).
And, if it isn't called right now, a call could easily be added in the future.

This with maintaining only **one** badInput call per Bad Input is pretty difficult without any kind of tool for error-stacking (or whatever it is called).

Only that, nothing more.

**Comment by fgalan** [comment]
If only atof call remains in 4 places, then let's add some comment like:

```
// intentional usage of atof(), as str2double() is not appropiated here
```

or a similar wording.

What I want to avoid is that due to a (human) memory leak after some time (months), somebody looks to this atof() calls and thinks "wait... we have that str2double thing... what we are not using it here?" and have to pass again the "mind process" to reach the same conclusion.

File: src/lib/mongoBackend/MongoGlobal.cpp [hunk 4]
```diff
@@ -1767,11 +1865,15 @@ bool entitiesQuery
       // FIXME P4: Once issue #1705 is implemented, this check can be removed.
       if (qStringFilters(sco->value, filters) != true)
       {
-        if (badInputP)
+        if (badInputP)  // Bad Input reported by higher level
```
fgalan: I guess that badInputP is a pointer. In that case, I'd suggest to be more explicit:

```
if (badInpuP != NULL)
```

src/lib/common/string.cpp
crbrox: AFAICS, is an out (write-only) parameter for ```strtod```, the initial value should not be taken into account and the value after the call is what is needed/used in the following statements

kzangeli: Ah, you might be right. Let's look closer at it ...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1767,11 +1865,15 @@ bool entitiesQuery
...
         {
           *badInputP = true;
           *err       = "invalid query expression";
         }
+        else
```
fgalan: Why this else brach has been added? I mean, it don't see the relation with the atof() -> str2double() refactoring...

Maybe is due to some .test get broken?

kzangeli: Nope, I simply thought a call to badInput was in order

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: This seems like a question for our style guide.
Very similar to comparisons with booleans.

I believe we said "up to each user".
If so, perhaps it should be remove it as we did with '++' ... ?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I understand that in that case, the caller would manage to set a not-NULL badInputP and manage the alarm itself.

If this else is not needed at the end, I'd suggest to remove it, for not adding extra entropy not related with the fix that the PR is addressing. At the end, any potential imprevement/fix in the bad input stuff will be covered by the issue https://github.com/telefonicaid/fiware-orion/issues/1763

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 9c7c1a9 (when merging with develop)

src/lib/common/string.cpp
kzangeli: Carlos was right, again :-), changed to NULL in d620cab

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Within that "up to each user" context that was a suggestion :)

I mean, for me is more clear to have `if (badInpuP != NULL)`. In general in my opinion, `if (foo)` should be used only if `foo` is a boolean.

It is a matter of personal taste of course. If you prefer stay as it is now, then NTC this thread comment.

CHANGES_NEXT_RELEASE
```diff
@@ -8,4 +8,5 @@
 - Fix: sanity check for string query 'q' - detect 'left-hand-side missing' (Issue #1754)
 - Fix: more sanity checks for string query 'q' (q empty, parts of 'q' empty - parts of 'q' are separated by ';')
 - Fix: error message when updating attribute value for two entities with same id (Issue #1387)
- 
+- Add: sanity checks for numbers (Issue #1306)
```
fgalan: Maybe this should be "Hardening:" ? 

("Add:" is typically used for new features)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: NTC, then

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: OK, I'll remove it. No probs

CHANGES_NEXT_RELEASE
kzangeli: Fixed in e99b892

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in e99b892

fgalan:

crbrox:
LGTM

Issue #990, opened 2015-06-11
API v2: Get a single attribute (by entity type and ID)


fgalan:
Implement the following operation in in v2:

```
GET /v2/entities/type/<entityType>/id/<entityId>/attrs/<attrName>
```

Same operation that https://github.com/telefonicaid/fiware-orion/issues/989, but taking into account entityType also in the URL.

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
At the end, this is going to be implemented using `?type=<entityType>`.

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/1768

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
LGTM 

See:
```
test/acceptance/behave/components/ngsiv2/attributes/get_attribute_data/get_attribute_data.feature
```

Issue #972, opened 2015-06-10
API v2: Get entity attributes (by entity ID and type)


fgalan:
Same as #971 but with the following URL:

```
GET /v2/entities/type/<entityType>/id/<entityId>?attrs=<attrName>
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
At the end, this is going to be implemented using `?type=<entityType>`.

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/1768

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
LGTM
```
test/acceptance/behave/components/ngsiv2/entities/retrieve_entity/retrieve_entity_queries_parameters.feature
```

Issue #951, opened 2015-06-03
API v2: Get a particular entity (by type and ID)


fgalan:
Same as https://github.com/telefonicaid/fiware-orion/issues/950 but with the following URL:

```
GET /v2/entities/type/<entityType>/id/<entityId>
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
At the end, this is going to be implemented using `?type=<entityType>`.

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/1768

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

fgalan:
Issues in milestone QATestPending are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeles with QATestPending label in the case thinks change and we need to recover it.

PR #1768, opened 2016-02-05
Add type param for getting entities in v2

crbrox:
Add type query parameter for GET entity/entities (with attribute / attribute value)

#951 #972 #990  #998

@fgalan 
@kzangeli


kzangeli:
LGTM

fgalan:
LGTM

Issue #1038, opened 2015-06-26
geo:point attribute type support [+ geo:line, geo:box and geo:polygon]


fgalan:
(Originally included as comment on https://github.com/telefonicaid/fiware-orion/issues/947)

Support the special processing of geo:point, i.e. string format checking and using it to match geo-queries.

fgalan:
Partially implemented for geo:point in PR https://github.com/telefonicaid/fiware-orion/pull/1771. The other "geo:things" are pending.

fgalan:
Completed in PR https://github.com/telefonicaid/fiware-orion/pull/1982

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1982. Moved to QATestPending while the acceptance test to cover this functionality gets implemented.

iariasleon:
LGTM
issue associated: #2426

PR #1771, opened 2016-02-05
ADD geo:point support for NGSIv2 keeping compatibility with NGSIv1 geo-location approach

fgalan:
Implements #1038 

@crbrox 
@kzangeli

### DIFF
--- a/src/lib/common/globals.h
+++ b/src/lib/common/globals.h
@@ -41,6 +41,23 @@
 #define FIWARE_LOCATION             "FIWARE::Location"
 #define FIWARE_LOCATION_DEPRECATED  "FIWARE_Location"   // Deprecated (but still supported) in Orion 0.16.0
 
+#define EARTH_RADIUS_METERS     6371000
+
+#define LOCATION_WGS84          "WGS84"
+#define LOCATION_WGS84_LEGACY   "WSG84"    /* We fixed the right string at 0.17.0, but the old one needs to be mantained */
+
+
+
+/* ****************************************************************************
+*
+* "geo:" types
+*/
+#define GEO_POINT    "geo:point"
+#define GEO_LINE     "geo:line"
+#define GEO_BOX      "geo:box"
+#define GEO_POLYGON  "geo:polygon"
+
+
 
 /* ****************************************************************************
 *
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1078,29 +1078,31 @@ static bool processLocation
 
     std::string location = caP->getLocation(apiVersion);
 
-    if (location.length() > 0)
+    if (location.length() == 0)
     {
-      if ((location != LOCATION_WGS84) && (location != LOCATION_WGS84_LEGACY))
-      {
-        *errDetail = "only WGS84 are supported, found: " + location;
-        return false;
-      }
-
-      if (locAttr.length() > 0)
-      {
-        *errDetail = "You cannot use more than one location attribute "
-                     "when creating an entity [see Orion user manual]";
-        return false;
-      }
-
-      if (!string2coords(caP->stringValue, coordLat, coordLong))
-      {
-        *errDetail = "coordinate format error [see Orion user manual]: " + caP->stringValue;
-        return false;
-      }
-
-      locAttr = caP->name;
+      continue;
     }
+
+    if ((location != LOCATION_WGS84) && (location != LOCATION_WGS84_LEGACY))
+    {
+      *errDetail = "only WGS84 are supported, found: " + location;
+      return false;
+    }
+
+    if (locAttr.length() > 0)
+    {
+      *errDetail = "You cannot use more than one location attribute "
+                     "when creating an entity [see Orion user manual]";
+      return false;
+    }
+
+    if (!string2coords(caP->stringValue, coordLat, coordLong))
+    {
+      *errDetail = "coordinate format error [see Orion user manual]: " + caP->stringValue;
+      return false;
+    }
+
+    locAttr = caP->name;
   }
 
   return true;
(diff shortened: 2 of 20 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1074,35 +1076,30 @@ static bool processLocation
   {
     const ContextAttribute* caP = caV[ix];
 
-    for (unsigned jx = 0; jx < caP->metadataVector.size(); ++jx)
```
fgalan: Hard to follow in diff mode. I'd recommend to have a look to the final code of this function directly.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1992,28 +1990,74 @@ static bool updateContextAttributeItem
   // given that the logic in NGSIv2 for specifying location attributes is gogint to change
   // (the best moment to address this FIXME is probably once NGSIv1 has been deprecated and
   // removed from code)
-  if (targetAttr->getLocation().length() > 0 && targetAttr->name != locAttr)
```
fgalan: The location stuff in this function has been heavily re-written. Hard to follow in diff mode. I'd recommend to have a look to the final code of this part directly.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2046,14 +2092,46 @@ static bool appendContextAttributeItem
...
   if (legalIdUsage(attrs, targetAttr))
   {
-    appendAttribute(attrs, toSet, toPush, targetAttr, actualUpdate, apiVersion);
+    bool actualAppend = appendAttribute(attrs, toSet, toPush, targetAttr, actualUpdate, apiVersion);
     entityModified = actualUpdate || entityModified;
     updateAttrInNotifyCer(notifyCerP, targetAttr);
 
     /* Check aspects related with location */
-    if (targetAttr->getLocation().length() > 0)
+    std::string locationString = targetAttr->getLocation(apiVersion);
```
fgalan: The location stuff in this function has been heavily re-written. Hard to follow in diff mode. I'd recommend to have a look to the final code of this part directly.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2695,15 +2786,11 @@ static void updateEntity
 
   if (r.hasField(ENT_LOCATION))
   {
-    //
```
fgalan: I think this comment was deprecated, thus I have removed.

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
```diff
@@ -207,6 +207,7 @@ TEST(mongoUpdateContextGeoRequest, newEntityLocAttribute)
     EXPECT_FALSE(a2.hasField("creDate"));
     EXPECT_FALSE(a2.hasField("modDate"));
     EXPECT_STREQ("A1", ent.getObjectField("location").getStringField("attrName"));
+    ASSERT_TRUE(ent.hasField("location"));
```
fgalan: The ASEERT_TRUE statement introduced in this test are a hardening measure for this test: it becomes a segfault in a test error.

NTC

kzangeli: Why a segfault?
Pretty much the reason for having asserts is to **avoid** segfaults, and get a cemprehensive error message instead ...
Is it not possible (in a not too difficult way) to avoid having a segfault?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -897,12 +896,13 @@ static bool updateAttribute
 * original value of the attribute was different than the one used in the update (this is
 * important for ONCHANGE notifications). Otherwise it is false
 *
-* (Previous versions of this function used the return value for that, but we have
-* modified to make in similar to updateAttribute()
+* In addition, return value is as follows:
+* - true: there was an actual append change
+* - false: there was an append-as-update change
 *
```
kzangeli: Perhaps true and false aren't the best return values.
Why not create an enum for it?

E.g.

    type enum MongoUpdateStatus
    {
      MusNormalAppendChange,
      MusAppendAsUpdateChange
    } MongoUpdateStatus;

    static MongoUpdateStatus appendAttribute(...)

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
fgalan: What I mean is tha using that  ASSERT_TRUE test execution is safer. For example, if the "location" field is missing and we don't use ASSERT_TRUE then next line `EXPECT_EQ(2, coordX(ent))` will end in segfault, given that `coordX()` function assumes that the field exists in the entity.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: We extensively use the true/false return pattern in many functions in the mongoBackend. Typically, you only need to distinguish between two states, so althoug your solution is probably smarter, I think that it involves extra work (not only in this function, but the others using the similar pattern) with little gain.

Thus, I think this should be NTC

src/lib/common/globals.h
```diff
@@ -41,6 +41,23 @@
...
+
+
+/* ****************************************************************************
+*
+* "geo:" types
+*/
+#define GEO_POINT    "geo:point"
+#define GEO_LINE     "geo:line"
+#define GEO_BOX      "geo:box"
+#define GEO_POLYGON  "geo:polygon"
```
kzangeli: FIWARE::Location  vs geo:point ...
One colon or two?
We should probably be consistent ... Right?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Unnecessarily cryptic, but it's your code.
If you want it like that, who am I to argue? :-)

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I looked at the three "difficult-diff" functions and all ok. Code that is ease to follow.
However, there are two cases where you could increase readibility.
Case 1: do a 'continue' instead of a big if:
processLocation:

     if (location.length() > 0)
    {
    }

better:

    if (location.length() > 0)
    {
        continue;
    }

Then in  appendContextAttributeItem the entire function is inside a "if (legalUsage())" (or something similar).
If could really gain on starting with  "if (!legalUsage()) { ...   return }" instead.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1992,28 +1990,74 @@ static bool updateContextAttributeItem
...
+          " - entity: [" + eP->toString() + "]" +
+          " - offending attribute: " + targetAttr->toString() +
+          " - only WGS84 is supported for location, found: [" + targetAttr->getLocation() + "]");
+    alarmMgr.badInput(clientIp, "only WGS84 is supported for location");
     return false;
   }
 
-  if (locAttr == targetAttr->name)
+  /* Case 1: update attribute from no-location -> location. There are 2 sub-cases */
+  if (locationString.length() > 0)
```
kzangeli: This might be the **if (locationString.length() > 0)** I mentioned

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2046,14 +2092,46 @@ static bool appendContextAttributeItem
 {
   if (legalIdUsage(attrs, targetAttr))
```
kzangeli: Here is the call to legalIdUsage I was talking about.

src/lib/serviceRoutinesV2/postEntities.cpp
```diff
@@ -120,6 +120,17 @@ std::string postEntities
 
     TIMED_RENDER(answer = oe.render(ciP, ""));
   }
+  else if (rhttpcode == SccInvalidParameter)
+  {
+    // This v1 -> v2 error conversion is a little crazy (for example, the same
+    // SccInvalidParameter at v1 level could match two different error conditions
+    // at v2 level, making harder de logic). However, I'm afraid that we need
```
kzangeli: making harder de logic ... :-)

making the logic harder

src/lib/serviceRoutinesV2/postEntity.cpp
```diff
@@ -89,14 +89,24 @@ std::string postEntity
     if ((upcrsP->contextElementResponseVector[ix]->statusCode.code != SccOk) &&
         (upcrsP->contextElementResponseVector[ix]->statusCode.code != SccNone))
     {
+      if (upcrsP->contextElementResponseVector[ix]->statusCode.code == SccInvalidParameter)
+      {
+        // Ugly v1 -> v2 error conversion here :(
+        OrionError error(SccRequestEntityTooLarge, "NoResourcesAvailable", "No more than one geo-location attribute allowed");
```
kzangeli: This is a little bit strange.
You see a statusCode.code == SccInvalidParameter and directly know that it is a "No more than one geo-location attribute allowed"?
I'm not saying it's wrong, just a little strange and possibly 'dangerous', what if the function that marked SccInvalidParameter is changed and gets more than one reason to flag SccInvalidParameter?
Adding a comment with a NOTE (or even a FIXME), would help.

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
kzangeli: OK, that's what I'm talking about :-)
Assert to **avoid** segfaults.
Misunderstanding,

NTC

src/lib/common/globals.h
fgalan: FIWARE::Location is legacy (NGSIv1). geo:point is the one defined in the NGSIv2 specification.

src/lib/serviceRoutinesV2/postEntity.cpp
fgalan: This is related with a similar comment in this PR:

```
    // This v1 -> v2 error conversion is a little crazy (for example, the same
    // SccInvalidParameter at v1 level could match two different error conditions
    // at v2 level, making the logic harder). However, I'm afraid that we need
    // to live with this while NGSIv1 and NGSIv2 coexists :(
```

I will add the same comment also here.

I agree with you: it is dangerous. But empirical evidence (make test) provides relative security.

Hopefully, all this will be "normalized" once NGSIv1 gets deprecated.

src/lib/common/globals.h
kzangeli: ok, reason good enough.

NTC

src/lib/serviceRoutinesV2/postEntity.cpp
kzangeli: ok, I agree to the above

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/a8d0650bc335b0e735f4c57a87096bbbede7507b

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/a8d0650bc335b0e735f4c57a87096bbbede7507b

src/lib/serviceRoutinesV2/postEntities.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/a8d0650bc335b0e735f4c57a87096bbbede7507b

src/lib/serviceRoutinesV2/postEntity.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/a8d0650bc335b0e735f4c57a87096bbbede7507b

crbrox:
LGTM

src/lib/mongoBackend/MongoCommonUpdate.cpp

kzangeli:
LGTM

Issue #992, opened 2015-06-11
API v2: Update single attribute (by entity type and ID)


fgalan:
Implement the following operation in in v2:

```
PUT /v2/entities/type/<entityType>/id/<entityId>/attrs/<attrName>
```

Same operation that https://github.com/telefonicaid/fiware-orion/issues/991, but taking into account entityType also in the URL.

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
At the end, this is going to be implemented using `?type=<entityType>`.

crbrox:
closed by #1779 

@iariasleon

fgalan:
"backlog" issues are closed once the PR that implement them is created/merged (only "bug" issues are passed to QA in opened status for final verfication).

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
LGTM 

See:
```
test/acceptance/behave/components/ngsiv2/attributes/update_attribute_data/update_attribute_data_part4.feature
```

Issue #988, opened 2015-06-11
API v2: Replace attributes (by entity type and ID)


fgalan:
Implement the following operation in in v2:

```
PUT /v2/entities/type/<entityType>/id/<entityId>
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
At the end, this is going to be implemented using `?type=<entityType>`.

crbrox:
closed by #1779 

@iariasleon

iariasleon:
reopened and pending to tests by @iariasleon

fgalan:
"backlog" issues are closed once the PR that implement them is created/merged (only "bug" issues are passed to QA in opened status for final verfication).

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests: `PUT - /v2/entities/<entity_id>?type=<entityType>`

fgalan:
Issues in milestone QATestPending are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeles with QATestPending label in the case thinks change and we need to recover it.

PR #1779, opened 2016-02-09
Add type param for PUT entity in v2

crbrox:
@kzangeli 
@fgalan 

 Add: type param for PUT entity in v2

#988 #992 #1000

### DIFF
--- a/test/functionalTest/cases/0988_put_with_param_type/0988_put_with_param_type.test
+++ b/test/functionalTest/cases/0988_put_with_param_type/0988_put_with_param_type.test
@@ -1,0 +1,257 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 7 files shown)


test/functionalTest/cases/0988_put_with_param_type/0988_put_with_param_type.test
```diff
@@ -0,0 +1,257 @@
...
+    "id": "E1",
+    "type": "T2",
+    "A1":{
+         "value": "V1 A1 E1/T2"
+    },
+    "A2":{
+         "value": "V1 A2 E1/T2"
+    }
+}'
+orionCurl --url /v2/entities --payload "$payload" --json
```
kzangeli: What happens if this payload (where type is set to T2) is sent with URI param type=T5 ?

Hint: 400 Bad Request (discussion in NGSIv2 needed to decide the behavior) ...

With the current implementation I believe the URI param silently overwrites the value in the payload.

kzangeli: Awaiting a decision in this matter we leave this *as is* for possible future modification.

NTC

kzangeli:
LGTM

crbrox:
CNR fixed at 687aed1

Issue #1774, opened 2016-02-08
[BUG] In v1/updateContext if servicePath exceed the max length allowed


iariasleon:
Similar to https://github.com/telefonicaid/fiware-orion/issues/1423

The servicePath allowed ten levels with 50 char each one, but now if we use 50 or more in several levels, ex: /tester1testcom/12345678901234567890123456789012, it is returned an error. Reference: https://github.com/telefonicaid/fiware-orion/issues/1764.

#### Append Request
```
POST http://localhost:1026/v1/updateContext
Fiware-Service: default
Fiware-ServicePath: /tester1testcom/12345678901234567890123456789012
Content-Type: application/json
{
    "contextElements": [
        {
            "type": "House",
            "isPattern": "false",
            "id": "Room",
            "attributes": [
            {
                "name": "temperature",
                "type": "float",
                "value": "23"
            }
            ]
        }
    ],
    "updateAction": "APPEND"
}
```

#### Append response
```
Status: 200: OK
<updateContextResponse>
  <contextResponseList>
    <contextElementResponse>
      <contextElement>
        <entityId type="House" isPattern="false">
          <id>Room</id>
        </entityId>
        <contextAttributeList>
          <contextAttribute>
            <name>temperature</name>
            <type>float</type>
            <contextValue></contextValue>
          </contextAttribute>
        </contextAttributeList>
      </contextElement>
      <statusCode>
        <code>200</code>
        <reasonPhrase>OK</reasonPhrase>
      </statusCode>
    </contextElementResponse>
  </contextResponseList>
</updateContextResponse>
```

If we try to use the same previous request, returns an error, but if we reduce the `service-Path` length, works perfectly.

#### Append response
```
Status: 200: OK
<updateContextResponse>
  <contextResponseList>
    <contextElementResponse>
      <contextElement>
        <entityId type="House" isPattern="false">
          <id>Room</id>
        </entityId>
        <contextAttributeList>
          <contextAttribute>
            <name>temperature</name>
            <type>float</type>
            <contextValue></contextValue>
          </contextAttribute>
        </contextAttributeList>
      </contextElement>
      <statusCode>
        <code>472</code>
        <reasonPhrase>request parameter is invalid/not allowed</reasonPhrase>
[...]

Issue #1423, opened 2015-10-27
[BUG] with some service-path header value, nothing is returned from db


iariasleon:
### Datasets:
```
      | /serv1/serv2/serv3/serv4/serv5/serv6/serv7/serv8/serv9/serv10 |
      | max length allowed                                            | where max length = 50 chars
      | max length allowed and ten levels                             |
```

##### Create request
```
 POST http://localhost:1026/v2/entities
 Content-Type: application/json
 Fiware-Service: test_update_service_path
 Fiware-ServicePath: /EhZa1vtv49pfnQa8ffF_N0Bf4Er7skJA6nvA5tSQzsxCyC6Fv_/Gj1FrYs09OED9Sz4ECimZI1JCITymjv2YYXZBSK3WDjStdcppq/HBXz8BmEJ3RZNH2XIyy9_OnKU8LAmHw2DByAln6DaxgKSkt3HC/rIuun2LhD_Ed4mf98XHU2_V0Pe2RGRsjyP9iAXHv7cuywSdxnT/nJf4oSd9tynFa_IwIEZSqvC6BEeup_qeZyfFaXAD5oO5Yx13Gj/9SASSborWAojYtX0Y10Oy0_2uITVnJglWNXj9svGDyrh1H1iM8/b1zRgnH378rW3jcp7mRe5TQfY_WZvvqMNQud0B8yjjt8EJwbsN/nl1dEI0ApCf1f0YBMOl_jj6uO5P2wAwaliIBu8eul1ORE4Xed7/4IQmeTFEeXv0GDxyZE9TIPWmk3vUFiEkd4kz0XHfUWbbOQSjGG/n5BcG296qVRG8x1z3SgG2usNcvednOyfM1BmFbm89zU3SgRKMj
 {"temperature_2": {"type": "celcius", "value": "34", "very_hot_1": {"type": "alarm", "value": "hot"}, "very_hot_0": {"type": "alarm", "value": "hot"}}, "temperature_0": {"type": "celcius", "value": "34", "very_hot_1": {"type": "alarm", "value": "hot"}, "very_hot_0": {"type": "alarm", "value": "hot"}}, "temperature_1": {"type": "celcius", "value": "34", "very_hot_1": {"type": "alarm", "value": "hot"}, "very_hot_0": {"type": "alarm", "value": "hot"}}, "type": "house", "id": "room"}
```

##### docs in mongodb
```
use orion-test_update_service_path
> db.entities.find()
[...]

PR #1780, opened 2016-02-10
Fix error with long servicepath component

### LINKED ISSUE #1764: Updating entity attributes fails when mongo key becomes too long
The following call creates an entity in a service path:
```
POST /v2/entities/test
Fiware-Service: default
Fiware-ServicePath: /tester1testcom/e2dbaecbcccccccccccccccccccccccc

{
    "type": "Patient",
    "key": "value"
}
```

If the call is sent again, this should update the attribute 'key'. However, Sending the call a second time returns HTTP statuscode 472, with a mongodb error of a duplicate key insertion.

Also, when using PUT calls to update attributes in this entity, orion reports the entity cannot be found.
The entity was inserted successfully, because a GET call to v2/entities/test returns the correct data.

This leads me to believe the POST/PUT code uses another query to test for the existance of the entity. I assume it fails to find this entity and proceeds by running an insert query, which fails.

By trial and error, I discovered that this behaviour occurs as soon as the mongo _id property has a certain length. In the example call, if we remove one character from either the service path, the entity id or the entity type, orion behaves as expected.

Our service path respects the restrictions of a maximum of 50 characters per level and a maximum of 10 levels.

Also, the mongo maximum index key length is 1024 bytes, which is well above the length of the _id value. Since the entities are successfully inserted into the mongodb database, I don't think the issue is located somewhere in the orion entity update logic rather than in mongodb

This was tested on orion v0.26.1 and mongo v2.6.11

crbrox:
It should fix #1764 #1774 and (re-opened) #1423 

@kzangeli 
@fgalan

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -3,7 +3,7 @@
 - Fix: '+' supported in entity ids and names in URLs (Issue #1675)
 - Fix: libmicrohttpd 0.9.48 included in contextBroker as static lib (previous Orion versions used 0.9.22 as dynamic library) (Issue #1675)
 - Fix: list of attribute names in URI param 'type' (Issue #1749)
-- Fix: long servicepath when getting in NGSIv2 (#1423)
+- Fix: long servicepath component in NGSIv2 (Issue #1423, #1764, #1774)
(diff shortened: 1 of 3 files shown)


kzangeli:
LGTM

Issue #1782, opened 2016-02-10
[BUG] in update v2/entities/<id> does not validate forbidden chars in url


iariasleon:
in `PATCH v2/entities/<id>` does not validate forbidden chars in url and it continues to looking for the id in mongoDB.

##### Datasets
```
      | house_?             |
      | house_&             |
      | house_/             |
      | house_#             |
      | my house            |
```
 
##### Update request
```
PATCH http://localhost:1026/v2/entities/house_&
Content-Type: application/json
Fiware-Service: test_update_entity_id
Fiware-ServicePath: /test
 {"temperature": {"value": "80"}}
```

##### Update response
```
http code: 404
date: Wed, 10 Feb 2016 09:59:21 GMT
connection: Keep-Alive
content-type: application/json
content-length: 61
{"error":"NotFound","description":"No context element found"}
```

##### Expected response
```
http code: 400
date: Wed, 10 Feb 2016 09:52:36 GMT
connection: Keep-Alive
content-type: application/json
content-length: 63
{"error":"BadRequest","description":"invalid character in URI"}
```

##### log messages
```
time=2016-02-10T10:59:21.201CET | lvl=DEBUG | trans=1455098360-858-00000000002 | srv=test_update_entity_id | subsrv=/test | from=10.95.233.37 | function=composeDatabaseName | comp=Orion | msg=MongoGlobal.cpp[419]: database name composed: 'orion-test_update_entity_id'
time=2016-02-10T10:59:21.201CET | lvl=DEBUG | trans=1455098360-858-00000000002 | srv=test_update_entity_id | subsrv=/test | from=10.95.233.37 | function=collectionRangedQuery | comp=Orion | msg=connectionOperations.cpp[134]: query() in 'orion-test_update_entity_id.registrations' collection limit=0, offset=0: '{ query: { $or: [ { contextRegistration.entities: { $in: [] } }, { contextRegistration.entities.id: { $in: [ "house_&" ] } } ], expiration: { $gt: 1455098361 }, servicePath: { $in: [ /^/test$/ ] } }, orderby: { _id: 1 } }'
[...]

PR #1787, opened 2016-02-11
Fix bug #1782, forbidden chars in ID PATCH v2

crbrox:
Fix bug #1782, forbidden chars in ID PATCH v2

@kzangeli 
@fgalan

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -16,3 +16,5 @@
 - Add: type param for POST entity in v2 (Issue #982, #984)
 - Add: support for geo:point type as a way of specifying location attribute in NGSIv2 (Issue #1038)
 - Add: type param for PUT entity in v2 (Issue #988, #992, #1000)
+- Fix: not detecting forbidden chars in ID for PATH v2 (Issue #1782)
+
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -16,5 +16,5 @@
 - Add: type param for POST entity in v2 (Issue #982, #984)
 - Add: support for geo:point type as a way of specifying location attribute in NGSIv2 (Issue #1038)
 - Add: type param for PUT entity in v2 (Issue #988, #992, #1000)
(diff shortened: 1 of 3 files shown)


src/lib/serviceRoutinesV2/patchEntity.cpp
```diff
@@ -71,6 +71,12 @@ std::string patchEntity
   eP->id = compV[2];
   eP->type = ciP->uriParam["type"];
 
+  if (forbiddenIdChars(ciP->apiVersion, eP->id.c_str() , NULL))
+  {
+    OrionError oe(SccBadRequest, "invalid character in URI");
+    return oe.render(ciP, "");
+  }
+
```
kzangeli: This seems like something that should be done for **all** v2 requests (perhaps v1 also),
not only PATCH /v2/entities/<eid> ...
Should the check be on a higher level?
Somewhere in rest.cpp ...

kzangeli: Ah, it's only about the ID, not the whole path ...

NTC

kzangeli:
LGTM

src/lib/serviceRoutinesV2/patchEntity.cpp
crbrox: I'll check  other 'paths' with ID for this point. If any is found, we can create an issue for all them and solve them in the same PR  ...

CHANGES_NEXT_RELEASE
```diff
@@ -16,3 +16,5 @@
 - Add: type param for POST entity in v2 (Issue #982, #984)
 - Add: support for geo:point type as a way of specifying location attribute in NGSIv2 (Issue #1038)
 - Add: type param for PUT entity in v2 (Issue #988, #992, #1000)
+- Fix: not detecting forbidden chars in ID for PATH v2 (Issue #1782)
```
fgalan: in ID -> in entityID ?
PATH -> PATCH ?

crbrox: Fixed 297de23

fgalan:
LGTM

Issue #1677, opened 2016-01-15
APIv2: align geo-location queries with last changes in spec


fgalan:
PR https://github.com/telefonicaid/fiware-orion/pull/1524 introduced some changes in the way geolocation queries work regarding the first implementation (e.g. new `georel` parameter). This issue is about modifying Orion in order to align with that new approach.

fgalan:
As with former NGSIv2 geo-query parameters, the idea is to translate then to a NGSIv1 FIWARE::Location scope, then process that FIWARE::Location at mongoBackend to render the query. As (positive) side effect, NGSIv1 geoquery functionality will be enhanced.

In fact, let's use "FIWARE::Location::NGSIv2". Trying to "fit" NGSIv2 geoqueries in the original FIWARE::Location containers can be hard.

```
            {
                "type": "FIWARE::Location::NGSIv2",
                "value": {
                    "georel": [ <the parsed georel: near, coveredBy, intersects, equals, disjoint; one element in each vector> ]
                    "geometry": "<point|line|polygon|box>",
                   "coords": [ <array of tuples with coordinates> ]
                }
            }
```

Note that "georel" couldn't be a simple property, due to the near operation has minRadious and maxRadius that we need also to take into account.

Probably a new class to hold this information would be needed as interface between serivceRoutines and mongoBackend.

fgalan:
Scope/areas implemented in PR https://github.com/telefonicaid/fiware-orion/issues/1677. Next, mongoBackend processing the new types is needed.

fgalan:
mongoBackend part (along with some adjust in other layers) in PR https://github.com/telefonicaid/fiware-orion/pull/1801

What is missing yet to close this issue is the processing of FIWARE::Location::NGSIv2 scopes as part of the payload of /v1/queryContext

fgalan:
The branch to complete the work is *feature/1677_ngsi2_geoqueries_as_payload_scope*. It currently includes doc changes and .test (only the SHELL part... the REGEXEPCT part would need to be adjusted to the NGSIv1 format).

fgalan:
Finished (the last part regarding NGSIv1) in PR https://github.com/telefonicaid/fiware-orion/pull/1828

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests are created in its feature associated

iariasleon:
E2E covered in:
[...]

PR #1798, opened 2016-02-15
Feature/1677 geo location for api v2

kzangeli:
Fixes part of the issue #1677.

The coords for polygons has changed in v2.
New rules: last point must be equal to first point, and the number of points must be >= 4.
These two rules are implemented but *ifdeffed out*. Perhaps this should be fixed within this PR.
Also, there is a comment or two about more detailed 'details' in errors.
Both these (two rules + details) give errors in functests and are disabled in this PR.

### DIFF
--- a/test/valgrind/valgrindTestSuite.sh
+++ b/test/valgrind/valgrindTestSuite.sh
@@ -156,7 +156,7 @@ dryrun=off
 leakTest=off
 dryLeaks=off
 fromIx=0
-ixLists=""
+ixList=""
 
 vMsg "parsing options"
 while [ "$#" != 0 ]
--- a/src/lib/common/limits.h
+++ b/src/lib/common/limits.h
@@ -62,8 +62,9 @@
 * Others -
 *
 */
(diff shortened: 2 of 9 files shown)


test/valgrind/valgrindTestSuite.sh
```diff
@@ -156,7 +156,7 @@ dryrun=off
 leakTest=off
 dryLeaks=off
 fromIx=0
-ixLists=""
+ixList=""
```
fgalan: Interesting fix... valgrind suite was broken before fixing this?

kzangeli: My guess is that the variable ixList was empty by default ...
Perhaps all shell variables are, and these initializations are not necessary ... ?

fgalan: NTC

fgalan:
Speed-LGTM (this PR is actually part of an ongoing work in #1677 to be finished y me)

PR #1801, opened 2016-02-16
Feature/1677 geo location

kzangeli:
Added some stuff missing for geo-location in URI params #1677

functest running ...


src/lib/rest/rest.cpp
```diff
@@ -243,7 +243,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   {
     containsForbiddenChars = forbiddenChars(val, ";");
   }
-  else if ((key != "q") && (key != "idPattern"))
+  else if ((key != "q") && (key != "idPattern") && (key != "geometry") && (key != "georel"))
```
fgalan: Note that geometry has its own "if clause" some lines above (around 238).

src/lib/ngsi/Scope.cpp
```diff
@@ -308,6 +316,21 @@ int Scope::fill
     return -1;
   }
 
+
+  //
+  // Georel 'near' only valid for a Point geometry
```
fgalan: Actually, near could work for not point geometries (it depend on the underlaying DB engine to solve the query). However, we could keep this restriction as it is by the moment and remove it once all the other things related with geo-queries are working.

src/lib/ngsi/Scope.cpp
```diff
@@ -511,64 +534,76 @@ void Scope::present(const std::string& indent, int ix)
...
-	  indent.c_str(), 
-	  circle.center.longitudeString().c_str()));
-    LM_T(LmtPresent, ("%s    Latitude:   %s", 
-	  indent.c_str(), 
-	  circle.center.latitudeString().c_str()));
-    LM_T(LmtPresent, ("%s    Inverted:   %s", 
-	  indent.c_str(), 
-	  circle.invertedString().c_str()));
+    LM_T(LmtPresent, ("%s    Radius:     %s",     indent.c_str(), circle.radiusString().c_str()));
+    LM_T(LmtPresent, ("%s    Longitude:  %s",     indent.c_str(), circle.center.longitudeString().c_str()));
```
fgalan: Indent.

src/lib/rest/rest.cpp
kzangeli: ok, this was already taken care of.
Didn't see that ...

src/lib/ngsi/Scope.cpp
kzangeli: Not sure what you mean here ...
It is indented:

    else if (areaType == orion::CircleType)
    {
      LM_T(LmtPresent, ("%s  FI-WARE Circle Area:", indent.c_str()));
      LM_T(LmtPresent, ("%s    Radius:     %s",     indent.c_str(), circle.radiusString().c_str()));
      LM_T(LmtPresent, ("%s    Longitude:  %s",     indent.c_str(), circle.center.longitudeString().c_str()));
      LM_T(LmtPresent, ("%s    Latitude:   %s",     indent.c_str(), circle.center.latitudeString().c_str()));
      LM_T(LmtPresent, ("%s    Inverted:   %s",     indent.c_str(), circle.invertedString().c_str()));
    }

Radius/Longitude etc all have two spaces more than 'FI-WARE Circle Area'.

src/lib/rest/rest.cpp
kzangeli: Fixed in 6bcd211 (georel treated as geometry)

kzangeli:
Finally got some time to run functests on this branch ...
Had a few failures, now fixed (01ec8be)

Complete 'make test' running (again) right now

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -299,7 +299,7 @@ Fields used as identifiers in the NGSIv2 API follow special rules regarding allo
 
 The rules are:
 
-* Allowed characters are the ones in the plain ASCII set except the following ones: whitespace, `&`, ?`, `/` and `#`.
+* Allowed characters are the ones in the plain ASCII set except the following ones: whitespace, `&`, `?`, `/` and `#`.
```
fgalan: Not actually related with this PR, but a small typo known time ago now being fixed. NTC

src/lib/ngsi/Scope.cpp
fgalan: Maybe github is missleading me, but I think I have spotted a one-whitespace misaling. However, not important. NTC

CHANGES_NEXT_RELEASE
```diff
@@ -20,3 +20,5 @@
 - Add: detect forbidden chars in entity ids and attr names in URI (Issue #1793)
 - Add: dates support in attribute values (Issue #1038)
 - Add: implemented new operation: POST /v2/op/query (Issue #1080)
+- Fix: NGSIv2 URI param 'georel' (along with 'geometry' and 'coords') proper support (Issue #1677)
+- Add: scope FIWARE::Location::NGSIv2 to allow using NGSIv2 geo-queries also with NGSIv1 (Issue #1677)
```
fgalan: This last line is not actually part of this PR, but soon to be added.

NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -898,6 +898,108 @@ static bool processAreaScope(const Scope* scoP, BSONObj &areaQuery)
...
+    {
+      orion::Point* p = scoP->line.pointList[ix];
+      ps.append(BSON_ARRAY(p->longitude() << p->latitude()));
+    }
+    geometry = BSON("type" << "LineString" << "coordinates" << ps.arr());
+  }
+  else if (scoP->areaType == orion::BoxType)
+  {
+    BSONArrayBuilder ps;
+    ps.append(BSON_ARRAY(scoP->box.lowerLeft.longitude()  << scoP->box.lowerLeft.latitude()));
```
fgalan: In order this to work, lowerLeft has to be the *real* lower left corner and upperRight has to be the *real* upper right corner. Is the fill() method at Scope class ensuring this?

src/lib/ngsi/Scope.cpp
```diff
@@ -308,6 +306,46 @@ int Scope::fill
...
+     *
+     *  { $err: "Can't canonicalize query: BadValue $within not supported with provided geometry:
+     *    { $geoWithin: { $geometry: { type: "Point", coordinates: [ [ 5.0...", code: 17287 }
+     */
+
+    *errorStringP = "point geometry cannot be used with coveredBy georel";
+    return -1;
+  }
+
+  if ((geometry.areaType != "point") && (georel.type == "near"))
```
fgalan: I removed the checking, but at the end it seems that MongoDB is not supporting $near with geometries different from Point, so I'm introducing it again (although probably not exactly the same way...).

NTC

src/lib/ngsi/Scope.cpp
```diff
@@ -308,6 +306,46 @@ int Scope::fill
     return -1;
   }
 
+  //pointVectorRelease(pointV); ??
```
fgalan: Not sure if the pointV.clear() and pointVectorRelease(pointV) statements are ok along all the fill() method. @kzangeli , as original author of the fill() method, could you have a look to be sure, so we can early remove any potential leak?

fgalan:
Ready for review!

@kzangeli 
@crbrox

src/lib/ngsi/Scope.cpp
kzangeli: Looks good, excpeto for one thing - the call to pointV.clear() on line 310 seems unnecessary. Not that it hurts but ...
The idea is to 'inherit the points from the vector **pointV** to the geometry objects.
After inheriting, the points cannot be freed (pointVectorRelease cannot be called), but the original vector, pointV, must be destroyed.
In case of errors, the entire vector with its points must be destroyed, so pointVectorRelease must be called apart from destroying the vector pointV itself.

Also, the comment on line 309 should be removed.
We **cannot** call pointVectorRelease, as that would free all the points stored in the geometry objects.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: As discussed over skype - no, there is no check/fix for that right now. Must be implemented.

src/lib/ngsi/Scope.cpp
```diff
@@ -97,16 +97,24 @@ int Scope::fill
...
 
+  if (apiVersion == "v1")
+  {
+    type = FIWARE_LOCATION;
+  }
+  else
+  {
+    type = FIWARE_LOCATION_V2;
+  }
+
```
kzangeli: I think I wrote these 8 lines ..

Now I wish I had done it like this, on **one** line:

    type = (apiVersion == "v1")? FIWARE_LOCATION : FIWARE_LOCATION_V2:

src/lib/ngsi/Scope.cpp
```diff
@@ -308,6 +306,46 @@ int Scope::fill
...
+  // Check invalid combinations
+  if ((geometry.areaType == "line") && (georel.type == "coveredBy"))
+  {
+    /* It seems that MongoDB 3.2 doesn't support this kind of queries, we get this error:
+     *
+     *  { $err: "Can't canonicalize query: BadValue $within not supported with provided geometry:
+     *    { $geoWithin: { $geometry: { type: "LineString", coordinates: [ [ 5.0...", code: 17287 }
+     */
+
+    *errorStringP = "line geometry cannot be used with coveredBy georel";
```
fgalan: Should we call pointVectorRelease(pointV); and pointV.clear() before return -1 ?

kzangeli:
I just realized a thing I've forgotten to implement - the release of the area objects.
Look at the method Scope::release():

    *
    * release -
    */
    void Scope::release(void)
    {
      polygon.release();
    }

We need to call the release() methods also for **point, line, and box** (and circle?), and create the methods if they don't exist.

src/lib/ngsi/Scope.cpp
kzangeli: Probably, the release methods will take care of pointVectorRelease, as all points have been inherited already. pointV.clear is already done, so nothing should be necessary down here.
Actually, it would be dangerous to call pointVectorRelease, as Scope::release when called will do a second free and that normally leads to SIGSEGV.

The best solution would be to move these checks up - to before the point vector is created.

src/lib/ngsi/Scope.cpp
fgalan: At the end, it seems that MongoDB only allows points. Thus, the check stills making sense, although it has been implemented in a different way at the end.

NTC

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/92aaa61d837588d35039bc4af28a98af1cdfbba9

src/lib/ngsi/Scope.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/92aaa61d837588d35039bc4af28a98af1cdfbba9

src/lib/ngsi/Scope.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/92aaa61d837588d35039bc4af28a98af1cdfbba9, although a second review to the fill() function will be a good idea to ensure we don't add leaks on purpose.

src/lib/ngsi/Scope.cpp
```diff
@@ -281,7 +309,31 @@ int Scope::fill
       pointV.clear();
       return -1;
     }
-    box.fill(pointV[0], pointV[1]);
+
+    // Check that points are different an not aligned (either horizontally or vertically)
```
kzangeli: type: an => and

src/lib/ngsi/Scope.cpp
```diff
@@ -281,7 +309,31 @@ int Scope::fill
...
+    if ((pointV[0]->latitude() == pointV[1]->latitude())  || (pointV[0]->longitude() == pointV[1]->longitude()))
+    {
+      *errorStringP = "box coordiantes are not defining an actual box";
```
kzangeli: typo: coordiantes => coordinates

src/lib/ngsi/Scope.cpp
```diff
@@ -281,7 +309,31 @@ int Scope::fill
...
+
+    // Lower left: smaller lat and long, upper right: greater lat and long
+    Point ll;
+    ll.latitudeSet(minLat);
+    ll.longitudeSet(minLon);
+    Point ur;
+    ur.latitudeSet(maxLat);
+    ur.longitudeSet(maxLon);
+    box.fill(&ll, &ur);
+
```
kzangeli: So, like this, the points can come in any order. New points for lowerLeft and upperRight are 'invented' if necessary. Right?

fgalan: Yes, that's the point. As long as the point are different ant not aligned, any combination will work.

src/lib/ngsi/Scope.cpp
```diff
@@ -578,6 +647,8 @@ void Scope::present(const std::string& indent, int ix)
 * release -
 */
 void Scope::release(void)
-{
+{  
+  // note that georel, circle, box, point don't use dynamic memory, so they dont' need release methods
```
kzangeli: typo: dont' => don't

src/lib/ngsi/Scope.cpp
kzangeli: Nice. 

NTC

src/lib/ngsi/Scope.cpp
fgalan: Fixed in 367b713

src/lib/ngsi/Scope.cpp
fgalan: Fixed in 367b713

src/lib/ngsi/Scope.cpp
fgalan: Fixed in 367b713

src/lib/ngsi/Scope.cpp
fgalan: NTC

kzangeli:
LGTM

crbrox:
LGTM

PR #1802, opened 2016-02-16
Some of the erroneous functests fixed

kzangeli:
14 functests fixed.
They need to be verified still ...

The remaining 55 functests added by @fgalan in commit 1e06cbe


kzangeli:
LGTM

PR #1806, opened 2016-02-17
Fix unchecked dereference

### LINKED ISSUE #1805: Segmentation fault with option without value in query string
```curl localhost:1026/v2/entities/E1?a```

### PR DESCRIPTION
**Pull request by crbrox**
Fix #1805 

@kzangeli 
@fgalan

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -20,3 +20,4 @@
 - Add: detect forbidden chars in entity ids and attr names in URI (Issue #1793)
 - Add: dates support in attribute values (Issue #1038)
 - Add: implemented new operation: POST /v2/op/query (Issue #1080)
+- Fix: bug with option without value in query string (Issue #1805)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -20,4 +20,4 @@
 - Add: detect forbidden chars in entity ids and attr names in URI (Issue #1793)
 - Add: dates support in attribute values (Issue #1038)
 - Add: implemented new operation: POST /v2/op/query (Issue #1080)
-- Fix: bug with option without value in query string (Issue #1805)
+- Fix: segfault caused by parameter without value in query string (Issue #1805)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/lib/rest/rest.cpp [hunk 1]
```diff
@@ -91,7 +91,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
   std::string      key   = ckey;
   std::string      value = (val == NULL)? "" : val;
 
-  if (*val == 0)
+  if (val == NULL || *val == 0)
   {
     OrionError error(SccBadRequest, std::string("Empty right-hand-side for URI param /") + ckey + "/");
```
kzangeli: Not sure this the error that we want to respond with in the case of 'no right hand side nor equal-sign'.
Would this be more of a "URI parameter /xxx/ not supported"?

E.g.

    if (val == NULL)
    {
      OrionError error(SccBadRequest, std::string("invalid URI param /") + ckey + "/");
￼      ...
    }

Open for discussion ...

CHANGES_NEXT_RELEASE
```diff
@@ -20,3 +20,4 @@
 - Add: detect forbidden chars in entity ids and attr names in URI (Issue #1793)
 - Add: dates support in attribute values (Issue #1038)
 - Add: implemented new operation: POST /v2/op/query (Issue #1080)
+- Fix: bug with option without value in query string (Issue #1805)
```
fgalan: Suggestion:

"Fix: segfault caused by parameter without value in query string (Issue #1805)"

(The bug was severe, so its effect -segfault- should be clearly described)

crbrox: Fixed 4f29a0c0aa647e9a07cfa178bdb5071a90ee5273

src/lib/rest/rest.cpp
kzangeli: NTC, after skype discussion

fgalan:
LGTM (passing the ball to @kzangeli )

kzangeli:
LGTM

Issue #1715, opened 2016-01-25
API v2: "batch" update


fgalan:
Implement the following operation in in v2:

```
POST /v2/op/update
```

See detailed description in the NGSIv2 APIary reference at http://telefonicaid.github.io/fiware-orion/api/v2/.

fgalan:
This operation should implement delete patterns of entities, as has been suggested for NSGIv1 in this separate issue: https://github.com/telefonicaid/fiware-orion/issues/1645

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1812

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests

iariasleon:
LGTM 
See  https://github.com/telefonicaid/fiware-orion/tree/master/test/acceptance/behave/components/ngsiv2/batch_operations/batch_op_update

PR #1812, opened 2016-02-19
Feature/1715 batch update

kzangeli:
Do not merge, there is still a problem to address

Fixes issue #1715, POST /v2/op/update

About the problem still to address, see the long FIXME in MongoCommonUpdate.cpp.


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3369,6 +3369,39 @@ void processContextElement
...
+    //   To me, this is a bit confusing.
+    //   Also, if the broker is to work this way, for the case of POST /v2/op/update, we will need complex responses, to be
+    //   able to tell the user which attributes were updated and which were not ...
+    //
+    //   This fix fixes part 5 of the functest v1_append_strict.test, while it destroys part 7 ...
+    //   A smarter solution is necessary (no db change and error returned? :-))
+    //
+    //   [ to mend part 7 of v1_append_strict.test, we could always remember if anything was mopdified in the db and in such case NOT
+    //     applying this new 'fix', deleting the last contextElementResponse ]
+    //
```
fgalan: This best effort behaviour is aligned with what NGSIv2 specification says:

> If append is used (i.e. strict append semantics): all the attributes in the payload not previously existing in the entity are appended. In addition to that, in the case some of the attributes in the payload already exist in the entity, an error is returned.

Not sure right now how to fix this...

fgalan: Fixed in a4640a0

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -527,7 +528,10 @@ std::string postUpdateContext
     ciP->httpStatusCode = httpStatusCode;
   }
 
-  foundAndNotFoundAttributeSeparation(upcrsP, upcrP, ciP);
+  if (ciP->apiVersion == "v1") // FIXME PR: v2 has no forwaring yet - I imagine this function is not necessary for v2 ...
```
fgalan: FIXME PR? 'R' should be number from 1 to 10?

test/functionalTest/cases/1715_batch_update/batch_update_APPEND.test
```diff
@@ -0,0 +1,314 @@
...
+dbInit CB
+brokerStart CB 0
+
+--SHELL--
+
+#
+# 01. Create E1/T1/A1=1, using POST /v2/op/update with APPEND
+# 02. GET /v2/entities/E1 to see E1/T1/A1=1
+# 03. Add an attribute A1=3 using POST /v2/op/update with APPEND_STRICT, see error
+# 04. GET /v2/entities/E1 to see E1/T1/A1=1 (part-test 03 did NOT modify A1)
```
fgalan: We typically use "step" term instead of "part-test" term.

test/functionalTest/cases/1715_batch_update/batch_update_errors.test
```diff
@@ -0,0 +1,273 @@
...
+
+{
+    "description": "attribute must be a JSON object",
+    "error": "BadRequest"
+}
+
+
+--TEARDOWN--
+brokerStop CB
+dbDrop CB
```
fgalan: As far as I checked (maybe I have missed something) all the /op/update cases use only one entity in the payload. It would be interesesting to add some test using more than one (take into account that the "value" of update batch is to modify/add/delete several entities in a single operation, for only-one-entity operation, the usual REST operations could be used).

test/functionalTest/cases/1716_bug_in_v2_entities_without_options_key_values/bug_in_v2_entities_without_options_key_values.test
```diff
@@ -58,12 +58,12 @@ echo
...
 =============================================================================================
 HTTP/1.1 400 Bad Request
-Content-Length: 56
+Content-Length: 70
 Content-Type: application/json
 Date: REGEX(.*)
 
 {
-    "description": "not a JSON object",
+    "description": "attribute must be a JSON object",
```
fgalan: (A little bit offtopic in this PR but...)

Maybe a more descriptive text could be:

"attribute must be a JSON object if keyValues option is not used"

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: I used the 'R' to indicate that this FIXME should be resolved within the PR ... :-)

fgalan: I tend to use P11 for the same :)

Thus, I understand that this will be fixed in this PR (and a "Fixed in XXX" comment added to this thread).

kzangeli: Fixed

test/functionalTest/cases/1715_batch_update/batch_update_APPEND.test
kzangeli: Fixed

test/functionalTest/cases/1715_batch_update/batch_update_errors.test
kzangeli: New functest with many entities in each operation - 24efc3f

test/functionalTest/cases/1716_bug_in_v2_entities_without_options_key_values/bug_in_v2_entities_without_options_key_values.test
kzangeli: Fixed in 677bb08

test/functionalTest/cases/1715_batch_update/batch_update_many.test
```diff
@@ -0,0 +1,523 @@
...
+      "A3": 3
+    },
+    {
+      "id": "E4",
+      "type": "T4",
+      "A4": 4
+    }
+  ]
+}'
+orionCurl --url '/v2/op/update?options=keyValues' --payload "$payload" --json
```
fgalan: Looking to NGSIv2 .apib, options=keyValues is not included in the /op/update operation. Please, add it (hint: copy-paste from an operation that already has it in the .apib).

kzangeli:
Actually, it responds either with 201 or 204
Not sure how this should be expressed in apiary.

201 for append and 204 for the other **actionTypes**

fgalan:
LGTM (passing the ball to @crbrox for final LGTM + merge + delete)

crbrox:
LGTM

Issue #1666, opened 2016-01-14
Websockets module: include header in input envelope


fgalan:
(Parent issue: #1181)

The JSON used as “input envelope” (composed of url, verb and payload) should also include the information that currently goes in HTTP headers in the REST API.

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/1814

PR #1814, opened 2016-02-19
Add headers in input envelope

fortizc:
Solve issue #1666

### DIFF
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -68,8 +68,8 @@ void ws_parser_parse(const char *msg,
 
   if (doc.HasMember("headers"))
   {
-    std::map<std::string, std::string *>::iterator it = head.supportedAttributes.begin();
-    while (it != head.supportedAttributes.end())
+    std::map<std::string, std::string *>::iterator it = head.supportedHeader.begin();
+    while (it != head.supportedHeader.end())
     {
       const char *value = doc["headers"][it->first.c_str()].GetString();
       *(it->second) = value ? std::string(value) : std::string();
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -23,7 +23,7 @@
 * Author: Felipe Ortiz
 */
 
-#include "parser.h"
+#include <map>
 
 #include "rapidjson/document.h"
 #include "rapidjson/stringbuffer.h"
@@ -31,10 +31,7 @@
 
 #include "logMsg/logMsg.h"
 #include "rest/HttpHeaders.h"
-
-
-#include <map>
-
+#include "parser.h"
 
 void ws_parser_parse(const char *msg,
                      std::string &url, std::string &verb,
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -33,9 +33,14 @@
 #include "rest/HttpHeaders.h"
 #include "parser.h"
 
-void ws_parser_parse(const char *msg,
-                     std::string &url, std::string &verb,
-                     std::string &payload, HttpHeaders &head)
+void ws_parser_parse
+(
+    const char*   msg,
+    std::string&  url,
+    std::string&  verb,
+    std::string&  payload,
+    HttpHeaders&  head
+)
 {
   rapidjson::Document doc;
 
@@ -76,5 +81,7 @@ void ws_parser_parse(const char *msg,
     head.gotHeaders = true;
   }
   else
+  {
     head.gotHeaders = false;
+  }
 }
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -70,8 +70,8 @@ void ws_parser_parse
 
   if (doc.HasMember("headers"))
   {
-    std::map<std::string, std::string *>::iterator it = head.supportedHeader.begin();
-    while (it != head.supportedHeader.end())
+    std::map<std::string, std::string *>::iterator it = head.headerMap.begin();
+    while (it != head.headerMap.end())
     {
       const char *value = doc["headers"][it->first.c_str()].GetString();
       *(it->second) = value ? std::string(value) : std::string();
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -30,10 +30,19 @@
 #include "rapidjson/writer.h"
 
 #include "logMsg/logMsg.h"
+#include "rest/HttpHeaders.h"
 
-void ws_parser_parse(std::string &url, std::string &verb, std::string &payload, const char *msg)
+
+#include <map>
+
+
+void ws_parser_parse(const char *msg,
+                     std::string &url, std::string &verb,
+                     std::string &payload, HttpHeaders &head)
 {
   rapidjson::Document doc;
+
+
   doc.Parse(msg);
 
   if (!doc.IsObject())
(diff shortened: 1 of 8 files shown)


src/lib/rest/HttpHeaders.cpp
```diff
@@ -0,0 +1,43 @@
...
+*
+* Author: Felipe Ortiz
+*/
+
+HttpHeaders::HttpHeaders()
+  : gotHeaders(false)
+  , servicePathReceived(false)
+  , contentLength(0)
+{
+  supportedAttributes["userAgent"] = &userAgent;
```
fgalan: Not sure about this, but "attribute" has a strong semantic in Orion, meaning a NGSI attribute. In this case, however, it seems it refers to a header name. Thus... what about using `supportedHeader` ?

fortizc: Ok, no problem

kzangeli: About the same topic ... if it wasn't 'supported' if wouldn't be in the map ...
I would propose a shorter name, for example 'header' or headerV, to indicate it's a vector (which it is not :-D ).

And, please align the 'equal signs', makes it a lot easier to read.

src/lib/orion_websocket/parser.cpp
```diff
@@ -30,10 +30,19 @@
 #include "rapidjson/writer.h"
 
 #include "logMsg/logMsg.h"
+#include "rest/HttpHeaders.h"
 
-void ws_parser_parse(std::string &url, std::string &verb, std::string &payload, const char *msg)
+
+#include <map>
```
kzangeli: Style guide ...

Include order:
  1. C include files
  2. C++ include files
  3. Third party software
  4. Our own headers (starting with headers from other libs and then the headers for *this* lib)

Not my idea, but I kind of agree ...

src/lib/orion_websocket/parser.cpp
```diff
@@ -30,10 +30,19 @@
...
+
+
+void ws_parser_parse(const char *msg,
+                     std::string &url, std::string &verb,
+                     std::string &payload, HttpHeaders &head)
 {
```
kzangeli: More style guide:

With long arg lists we normally (me, always) lay them out like this:

    void ws_parser_parse
    (
      const char*   msg,
      std::string&  url,
      std::string&  verb,
      std::string&  payload,
      HttpHeaders&  head
    )
    {
      ...
    }

It's easier to read that way.

Note that '\*' and '&' are 'glued' to the **type**, not the *variable*.
This makes a lot of sense as these electors change the type (*this* is my idea :-))
Also note that there are two spaces between type and variable name, not one.
This is due to some type need more than one word, e.g:

    struct timeval  tv;

All style guide stuff ... we have a lot of rules ...

src/lib/orion_websocket/parser.cpp
```diff
@@ -56,4 +65,19 @@ void ws_parser_parse(std::string &url, std::string &verb, std::string &payload,
...
+    {
+      const char *value = doc["headers"][it->first.c_str()].GetString();
+      *(it->second) = value ? std::string(value) : std::string();
+      ++it;
+    }
+
+    head.gotHeaders = true;
+  }
+  else
+    head.gotHeaders = false;
```
kzangeli: Another style rule we have (Fermins idea), is to always use brackets '{}', even with one line blocks.

src/lib/orion_websocket/ws.cpp
```diff
@@ -54,12 +54,15 @@ typedef struct
...
   char *message;
+  char *request;
+  int index;
 }data;
 
-static int wscllbck(libwebsocket_context *ctx,
-                    libwebsocket *ws,
-                    libwebsocket_callback_reasons reason,
-                    void *user, void *in, size_t len)
+static int wscllbck(lws * ws,
```
kzangeli: To save two chars ...
wscllbck => wsCallback

right?

src/lib/orion_websocket/ws.cpp
```diff
@@ -97,7 +102,7 @@ static int wscllbck(libwebsocket_context *ctx,
       unsigned char *p = &buff[LWS_SEND_BUFFER_PRE_PADDING];
 
       sprintf((char*)p, "%s", dat->message);
-      libwebsocket_write(ws, p, strlen(dat->message), LWS_WRITE_TEXT);
+      lws_write(ws, p, strlen(dat->message), LWS_WRITE_TEXT);
```
kzangeli: This I like, lots of bytes 'saved' here.
I would call it wsWrite though.
Matter of opinion.

NTC

src/lib/orion_websocket/ws.cpp
```diff
@@ -105,9 +110,26 @@ static int wscllbck(libwebsocket_context *ctx,
...
-      ConnectionInfo *ci = connection_manager_get(dat->cid, message);
-      dat->message = strdup((restService(ci, orionServices)).c_str());
-      libwebsocket_callback_on_writable(ctx, ws);
+      const size_t remaining = lws_remaining_packet_payload(ws);
+      const size_t msg_size = strlen(message);
+
+      if (!dat->request)
+        dat->request = (char *) malloc (msg_size + remaining);
+
+      memcpy(&dat->request[dat->index], message, msg_size);
```
kzangeli: Normally you would include the terminating ZERO of a string.
If not, strange things can happen ...

[ I say this as msg_size comes from strlen(message) ]

fortizc: If you see the line 124 the ZERO is inserted at the end. In this place of code I'm not using a null character because the message is not totally received.

kzangeli: ok, I understand now. The first time (if (!dat->request)), you allocate room for the entire message, knowing the total size thanks to lws_remaining_packet_payload.
Then, in consequent calls you just add to dat->request until all the message is received which is when dat->request is set to NULL.

NTC

src/lib/rest/HttpHeaders.cpp
fortizc: In this case I prefer something like headerMap, sounds more "right"

src/lib/orion_websocket/parser.cpp
fortizc: Fix in b395ca796083f02a07fba6ba969abcffe9f96c51

src/lib/orion_websocket/parser.cpp
fortizc: fix in 903a149549f0204243c8a7664ceaa675ebad2003

src/lib/orion_websocket/parser.cpp
fortizc: fixed 903a149549f0204243c8a7664ceaa675ebad2003

src/lib/orion_websocket/ws.cpp
fortizc: fixed a81900d989514f8e38063ae9356583ca23ec282f

src/lib/orion_websocket/parser.cpp
fortizc: Remember, for all old code without standards I have a task in this sprint to fix that ;)

src/lib/orion_websocket/parser.cpp
kzangeli: Ah, and another one from the style guide ... :-)
Always include one directory in the inclusion:

    #include "orion_websocket/parser.h"

Also not my idea, but it makes a lot of sense ...

src/lib/rest/HttpHeaders.cpp
kzangeli: headerMap sounds good

src/lib/orion_websocket/parser.cpp
kzangeli: ok, I'll stop complaining about style-guide until that PR then ;-)

src/lib/rest/HttpHeaders.cpp
fgalan: +1 headerMap

fgalan: (or maybe header**s**Map?)

kzangeli: headerMap sounds much better.

Compare:
  shoe store - more than *one* pair of shoes are sold there,
  stamp collection - not only *one* stamp, I hope :-)
  I could go on ...

'header' is also good, and shorter ...

fortizc: Fixed b4e5554a72150d7b505633756fead3fdb71525d0

fgalan:
LGTM (passing the ball to @kzangeli for final LGTM + merge)

kzangeli:
LGTM

Issue #1103, opened 2015-07-26
Sorting capabilities in list entity operation [pending: orderBy=$near]


fgalan:
Eg:

`http://broker.example.org/entities?type=Room&q=temperature<=22&sort=seatNumber`

The parameter could be also used with a sequence of attributes, e.g `sort=seatNumber,temperature`. ASC and DESC order are specified with a `+` (default) and `-` prefix.

`sort=+seatNumber,-temperature`

Feasibility assessment with Mongo shell: https://gist.github.com/fgalan/efd63ef261fa33d91784

fgalan:
Sorting by `creDate` and `modDate` (definitive names pending) should be also possible.

fgalan:
SOF answers that need to be edited when this issue gets completed: http://stackoverflow.com/questions/32703011/change-order-of-orion-response/32712248#32712248

fgalan:
Take into account talso sort by "nearest point". This PR may be useful: https://github.com/telefonicaid/fiware-orion/pull/1651.

fgalan:
SOF answer that need to be edited when this issue gets completed: http://stackoverflow.com/questions/35172155/orion-context-broker-nearest-to-point-query

fgalan:
Code implementation completed in PR https://github.com/telefonicaid/fiware-orion/pull/1815. However, let's keep this issue opened while doc for NGSIv2 get reviewed (and eventually updated) and the questions in SOF edited.

fgalan:
At the end, the name of the URI parameter (to use in the documentation) is `orderBy`

fgalan:
Doc modifications describing `orderBy` implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1818. However, let's keep this issue opened until `orderBy=$near` gets implemented (issue name will be adjusted).

fgalan:
* orderBy=geo:proximity implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1824
* orderBy=dateCreation and orderBy=dateModification implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1829

fgalan:
QA Acceptance Test pending. Reopened and asigned to QATestPending milestone.

iariasleon:
pending tests are created in its feature associated

fgalan:
[...]

PR #1815, opened 2016-02-19
ADD sorted URI param in NGSIv2 queries

fgalan:
Implements #1103 

@kzangeli 
@crbrox

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -1882,7 +1882,8 @@ bool entitiesQuery
   bool*                            limitReached,
   long long*                       countP,
   bool*                            badInputP,
-  const std::string                apiVersion
+  const std::string&               sorted,
+  const std::string&               apiVersion
 )
 {
 
@@ -2004,7 +2005,34 @@ bool entitiesQuery
   auto_ptr<DBClientCursor>  cursor;
   Query                     query(finalQuery.obj());
 
-  query.sort(BSON(ENT_CREATION_DATE << 1));
+  if (sorted == "")
+  {
+    query.sort(BSON(ENT_CREATION_DATE << 1));
+  }
+  else
+  {
+    std::vector<std::string>  sortedV;
+    int components = stringSplit(sorted, ',', sortedV);
+    BSONObjBuilder sortOrder;
+    for (int ix = 0; ix < components; ix++)
+    {
+      std::string sortToken;
+      int sortDirection;
+      if (sortedV[ix][0] == '!')
+      {
+        // reverse
+        sortToken = sortedV[ix].substr(1);
+        sortDirection = -1;
+      }
+      else
+      {
+        sortToken = sortedV[ix];
+        sortDirection = 1;
+      }
+      sortOrder.append(std::string(ENT_ATTRS) + "." + sortToken + "." + ENT_ATTRS_VALUE, sortDirection);
+    }
+    query.sort(sortOrder.obj());
+  }
 
   TIME_STAT_MONGO_READ_WAIT_START();
   DBClientBase* connection = getMongoConnection();
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -1882,7 +1882,7 @@ bool entitiesQuery
   bool*                            limitReached,
   long long*                       countP,
   bool*                            badInputP,
-  const std::string&               sorted,
+  const std::string&               sortOrderList,
   const std::string&               apiVersion
 )
 {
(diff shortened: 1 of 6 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2004,7 +2005,34 @@ bool entitiesQuery
...
+      }
+      else
+      {
+        sortToken = sortedV[ix];
+        sortDirection = 1;
+      }
+      sortOrder.append(std::string(ENT_ATTRS) + "." + sortToken + "." + ENT_ATTRS_VALUE, sortDirection);
+    }
+    query.sort(sortOrder.obj());
+  }
```
kzangeli: Wow, not one single empty line here ...
I see just a mass of characters, might only be me having these problems ... :-)

Style guide:
* Indent local variables, and add an empty lines after the local var declarations.
* After blocks it is a good idea to add an empty line (before sortOrder.append(... ) and query.sort(sortOrder...)
* Indent '='s

This might just be my opinion, but I really think the code gets easier to understand like this ...

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2004,7 +2005,34 @@ bool entitiesQuery
   auto_ptr<DBClientCursor>  cursor;
   Query                     query(finalQuery.obj());
 
-  query.sort(BSON(ENT_CREATION_DATE << 1));
+  if (sorted == "")
```
kzangeli: Not sure I understand this ...
sorted is a string, and if its empty then *query.sort()* is called and if it contains **anything at all**. then it is a vector of sortTokens?

First I thought it was some kind of boolean variable, telling us that it was either sorted or not.
Perhaps sortOrderList would be a better name?

src/lib/rest/uriParamNames.h
```diff
@@ -45,6 +45,7 @@
 #define URI_PARAM_ATTRIBUTE_FORMAT        "attributeFormat"
 #define URI_PARAM_OPTIONS                 "options"
 #define URI_PARAM_TYPE                    "type"
+#define URI_PARAM_SORTED                  "sorted"
```
kzangeli: sorted ...   what happened to *sortBy* ?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: `sorted == ""` means that the `?sorted` parameter was not spedificed, i.e. CB should use default ordering which is based on creation date.

Or maybe I'm missing something in your point...

src/lib/rest/uriParamNames.h
fgalan: Looking https://github.com/telefonicaid/fiware-orion/issues/1103 in detail, it tells about `sort=`.

Not sure what is the best word... `sort`, `sorted`, `sortBy` all are on the table. What do you prefer? @jmcanterafonseca feedback is also valuable.

fgalan: Which word are using other REST APIs for a similar feature?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: All them are good suggestion. More "styled" code after commit b682597.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yeah, I understood all this ... after a while (a long one :-)).

The name is very confusing, that's all.
Had it been called sortOrderList, then it would have been easier

src/lib/rest/uriParamNames.h
kzangeli: Adjectives are normally used for booleans, like *occupied*, or *empty*.
You don't expect them to carry any value.

I vote for sortBy or orderBy, both excellent choices.
ORDER BY is used in SQL and is probably familiar to a lot of our users.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Nice!
Thank you :-)

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Thus... NTC?

src/lib/rest/uriParamNames.h
fgalan: +1 to `orderBy`

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Up to you, your code ... :-)

I would change the name of the variable.

NTC, if you wish :-)

fgalan: Rename done in 22bacb2

src/lib/rest/uriParamNames.h
fgalan: Fixed in 4d0797b

kzangeli:
LGTM

crbrox:
LGTM

PR #1823, opened 2016-02-22
FIX add libwebscocket installation instructions

fgalan:
**DON'T MERGE YET**. What is pending:

* Check the procedure actually works in CentOS 6.x (by the moment it has been tested only in Debian)
* Sync in Dockerfile

### DIFF
--- a/doc/manuals/admin/build_source.md
+++ b/doc/manuals/admin/build_source.md
@@ -11,6 +11,7 @@ The Orion Context Broker uses the following libraries as build dependencies:
 * libcurl: 7.19.7
 * Mongo Driver: legacy-1.0.7 (from source)
 * rapidjson: 1.0.2
+* libwebsockets: 1.6.3 (from source)
 * gtest (only for `make unit_test` building target): 1.5 (from sources)
 * gmock (only for `make unit_test` building target): 1.5 (from sources)
 
@@ -41,6 +42,17 @@ commands that require root privilege):
         wget https://github.com/miloyip/rapidjson/archive/v1.0.2.tar.gz
         tar xfvz v1.0.2.tar.gz
         sudo mv rapidjson-1.0.2/include/rapidjson/ /usr/local/include
+
+* Install libwebsocket from sources:
+
+        wget https://github.com/warmcat/libwebsockets/archive/v1.6.3.tar.gz
+        tar xfvz v1.6.3.tar.gz
+        cd libwebsockets-1.6.3/
+        mkdir build
+        cd build
+        cmake ..          # default cmake sets /usr/local as installation path
+        make
+        sudo make install
 
 * Install libmicrohttpd from sources (the `./configure` command below shows the recommended build configuration to get minimum library footprint, but if you are an advanced user, you can configure as you prefer)
 


doc/manuals/admin/build_source.md
```diff
@@ -42,6 +43,17 @@ commands that require root privilege):
...
+
+        wget https://github.com/warmcat/libwebsockets/archive/v1.6.3.tar.gz
+        tar xfvz v1.6.3.tar.gz
+        cd libwebsockets-1.6.3/
+        mkdir build
+        cd build
+        cmake ..          # default cmake sets /usr/local as installation path
+        make
+        sudo make install
+
```
fgalan: The procedure seems to be ok to compile in CentOS 6.x. However, it seems that (differently from Debian) /usr/local/lib is not included in the libary loading path so running compiled binary may result in

```
contextBroker: error while loading shared libraries: libwebsockets.so.6: cannot open shared object file: No such file or directory
```

The solution is described in http://serverfault.com/questions/372987/centos-usr-local-lib-system-wide-ld-library-path. However, it is not clear to me how to fit this in the RPM (including libwebsocket.so.6 as part of the CB RPM? In which directory? Create an ad hoc package RPM for the libwebsocket library and link packages through dependencies?)

fgalan: One possible solution to both problems is using static library. The contextBroker binary will be bigger, but probably it is not a problem...

kzangeli: Yeah, I think I also prefer a static lib ...
Easier for our users, as there will never be any problems with version mismatches.
It would be nice to see the difference is size though ... (after 'strip')

fgalan: Thus, a PR fixing CMakeList.txt to use static library should be done. @fortizc, could you manage that, please?

fortizc: Ok, no problem

fgalan:
I have added https://github.com/telefonicaid/fiware-orion/pull/1823/commits/90e6cea9cc8167bfa1bcc3d25358f9a8f8e02db9 commit to this PR to integrate websockets.md.

The work in this PR once @fortizc finished with changing CMakeList.txt to use static linking.

doc/manuals/admin/build_source.md
fortizc: I'm not really convinced with this... do you try use DCMAKE_INSTALL_PREFIX:PATH option?
Something like that
```
wget https://github.com/warmcat/libwebsockets/archive/v1.6.3.tar.gz
tar xfvz v1.6.3.tar.gz
cd libwebsockets-1.6.3/
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX:PATH=/MY_PREFERED_PATH ..
make
sudo make install
```

fgalan: As long as MY_PREFERED_PATH would be `/usr/local`, it's ok. However, given that it seems that `cmake` uses that installation path by default, I dont' see the point in making the procedure more complex by the introduction of this `-DCMAKE_INSTALL_PREFIX` variable.

To be discussed at skype.

kzangeli: If used, please add an 'R': PREFERRED

fgalan: After some discussion at skype, we propose to finally use:

```
wget https://github.com/warmcat/libwebsockets/archive/v1.6.3.tar.gz
tar xfvz v1.6.3.tar.gz
cd libwebsockets-1.6.3/
mkdir build
cd build
cmake  # use -DCMAKE_INSTALL_PREFIX:PATH=/your/path in order to override /usr/local as installation path
make
sudo make install
```

fgalan:
This PR cannot be considered stale. It shows a way of implementing websocket in Orion and, in that sense, it is useful.

kzangeli:
Why not give it a special label, e,g,  "Future", close the PR and keep a link to it in its issue?
That would solve the problem.

fgalan:
> Why not give it a special label, e,g, "Future", close the PR and keep a link to it in its issue?

I agree in the first part (define an specifica label for this kind of PRs) but not in the second (close the PR). As long as github doesn't charge us for the number of open PRs ;) it is prefereable to have then alive as open PR.

Issue #1830, opened 2016-02-23
APiv2: make string "none" default type for entity, attribute and metadata.


fgalan:
Changes in .apib and code required.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/1835

PR #1835, opened 2016-02-24
ADD default type "none" for NGSIv2

fgalan:
Implements #1830

@crbrox 
@kzangeli 
@jmcanterafonseca (nice to have)
@iariasleon (to make him aware of this, as the changes in this PR may impact behave tests)


test/functionalTest/cases/1587_q_query_has_type/q_query_has_type.test
```diff
@@ -1,186 +0,0 @@
-# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: The `+type` doesn't make sense as in NGSIv2 all entities will have type (there is a restriction on IDs for a minimum length of 1 chraacter). Thus, this .test is removed.

NTC (just informative).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -302,6 +301,7 @@ The rules are:
 
 * Allowed characters are the ones in the plain ASCII set except the following ones: whitespace, `&`, `?`, `/` and `#`.
 * Maximum field length is 256 characters.
+* Minimum field length is 1 character.
```
fgalan: This new restriction has not been yet implemented. I have created this issue for not forgetting about it: https://github.com/telefonicaid/fiware-orion/issues/1836

NTC

crbrox:
LGTM

kzangeli:
LGTM

jmcanterafonseca:
sorry for arriving late, but

why do we need 'none' ? why it cannot be just the ```null``` value of JSON?

fgalan:
Entity/attribute/metadata type (as other IDs, see "Field syntax restrictions") needs to be a string (as string constrains need to be checked, e.g. allowed charset and length), so `null` would not be valid. The string `"null"` would be valid, but it would add confussion as it can be misslead with the `null` native type, so `"none"` seems to be the best choice.

Note also that if we allow `null` in types, the query:

```
GET /v2/entities/E?type=null
```

would be ambiguous, as it may refer to entities with string type `"null"` or entities with default type `null`.

However, note that `null` is supported in attribute values (and eventually it would be supported also in metadata value) as these fields are not constrained by the restrictions that apply to ID fields.

jmcanterafonseca:
couldn't be as easy as saying types are allowed to be null which means that no entity type is provided but  if type is given a string value that string value cannot be the empty string and has to have a length of least one character? 

Sorry guys but I don't like the 'none' thing ... it introduces anomalies in the API which can be solved more elegantly

fgalan:
Having a field that sometimes is a string, sometimes isn't, with different rules depending if if is a string or not it is a greater anomaly. It is very confusing for users. I don't see any lack of eleangtly, but the oposite: having homogeneous rules (i.e. types are always strings) is elegant.

The bottom line is that types *are not* allowed to be `null`. That breaks ID field rules, along with some operations (e.g. `GET /v2/types`, as `null` cannot be use a JSON key because JSON keys are always strings).

mrutid:
Identifiers MUST be as homogeneous and concrete as posible, since they are used for Operational purposes. Type is a first class identifier and must follow the rules defined for CB identifiers. I would accept some kind of soft sugar to allow some "no type scenarios" in external interfaces. But it is important that everything has a Type no matter if the type is "void", we need to avoid chaotic and fuzzy scenarios (that typically appear in real life asociated to null / empty types in CB).

jmcanterafonseca:
the use of 'none' is going to introduce anomalies in the future, particularly JSON-LD renderings, apart from being weird as JSON offers null for these purposes

fgalan:
Is the problem is the exact string `"none"` please do a new suggestion for a *string* for default types which, at the same time matches the rules for IDs defined at "Field syntax restriction" section at http://telefonicaid.github.io/fiware-orion/api/v2, so we could evaluate the impact of the change. If that string has 4 characters, the better ;) (otherwise, changes in Content-Length of around ~100 .test files would be needed, as you can see in the present PR in the "Files" tab).

PR #1856, opened 2016-02-26
Updated test in order to work under harness and valgrind mode

raconte:
After trying different timings the test now are working .It has been modified in order to work test under harness and valgrind mode.So, now in my virtual machine with Ubuntu are working.


fgalan:
LGTM (this PR is safe to be merged even when development is frozen, as it doesn't touch actual source code).

test/functionalTest/cases/1585_log_summary/log_summary.test
```diff
@@ -25,8 +25,16 @@ Log Summary
...
+ then
+  export BROKER_AWAIT_SLEEP_TIME=6
+  brokerStart CB 0 IPv4 -logSummary 3 
+  unset BROKER_AWAIT_SLEEP_TIME
+  sleep 7
+  valgrindSleep 8
+else 
+  brokerStart CB 0 IPv4 -logSummary 3 
+fi
 --SHELL--
```
kzangeli: Strange mix here between "$FUNC_TEST_RUNNING_UNDER_VALGRIND" == "true" and calls to valgrindSleep ...

valgrindSleep does this:

    if [ "$FUNC_TEST_RUNNING_UNDER_VALGRIND" == "true" ]
    then
      sleep $1
    fi

So, if we already know that $FUNC_TEST_RUNNING_UNDER_VALGRIND is true, we can call sleep right away.

So, 

    sleep 7
    valgrindSleep 8

should be changed for 

    sleep 15

Or even better, use 

    valgrindSleep 15

after the if/else/fi.
And, as its safe to UNSET a variable, we can move around that part also ...
That way the code gets easier to read at the same time:


    if [ "$FUNC_TEST_RUNNING_UNDER_VALGRIND" == "true" ]
    then
        export BROKER_AWAIT_SLEEP_TIME=6
    fi

    brokerStart CB 0 IPv4 -logSummary 3 
    unset BROKER_AWAIT_SLEEP_TIME
    valgrindSleep 15

Same thing valid for more cases in this PR.

kzangeli: Fixed in 1faaa2c

kzangeli:
The problem with these two tests wasn't timing, but that the valgrind test suite sends a "GET /version" to the broker to make sure the broker is a live.
This "GET /version" is a transaction and it 'destroys' the statistics for the LOG SUMMARY.

What was needed was to modify the expected number of transactions to be 5 and not 4 when under valgrind, 6 and not 5 + 7 and not 6.
That, and some problems encountered with python REGEX used right next to parenthesis.
The parenthesis has to be modified to some other character (sed was used to convert () to underscores).

Now the two tests work. Both under valgrins and as 'normal functests'

PR #1860, opened 2016-02-29
Bug/1852 escaped slashes in json v1 payload

### LINKED ISSUE #1852: "JSON parse error" with valid JSON using escaped slashes
Hi!

The context broker throws an error (`400 Bad Request` - `JSON Parse Error`) if the JSON string contains escaped slashes, i.e., contains the sequence `\/` within any string. This is valid JSON, so it should be parsed normally.

For example, this is the JSON object I'm sending to the broker (the `reference` property is what is causing the parse error):

```
POST /v1/subscribeContext
```
```javascript
{
    "entities": [
        {
            "type": "Device",
            "isPattern": "false",
            "id": "1234"
        }
    ],
    "attributes": [
        "movement"
    ],
    "reference": "http:\/\/localhost:1028\/accumulate",
    "duration": "P1M",
    "notifyConditions": [
        {
            "type": "ONCHANGE",
            "condValues": [
                "movement"
            ]
        }
    ],
    "throttling": "PT1M"
}
```

This JSON was generated using the [`json_encode`](http://php.net/manual/en/function.json-encode.php)'s PHP function. I know this function accepts a parameter to avoid scaped slashes, but in this case I'm using a library which wraps the call to the JSON-encoding function internally and I don't want to modify that library.

Could the server be fixed in order to accept that valid JSON? Thanks!

kzangeli:
Fixes #1852

In JSON v1 payload, changing backslash-slash for just slash - this way a JSON parse error is avoided.


test/functionalTest/cases/1852_escaped_slashes_in_json_v1_payload/escaped_slashes_in_json_v1_payload.test
```diff
@@ -0,0 +1,85 @@
...
+Get entities by id parameter
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# 01. Create a subscription with escaped slashes in reference
```
fgalan: I would add `GET /v2/subscriptions` as second step in order to check that subscription was created correctly.

kzangeli: Sure

kzangeli: Fixed in 9fca38c

fgalan:

crbrox:
LGTM

Issue #1862, opened 2016-02-29
XML support removal


fgalan:
This is a big work that needs to be addressed step by step:

1. "Hack" at rest/network layer to remove XML support. All the functional test failing will be temporarly moved to a `cases_xml/` subdirectory, keeping the same directory case structure. At the end of this step, XML parsing/render code will still be there, but as dead code. This could be done in one PR.
2. Remove dead code at parsing/render stages. During the process, the "hacks" donde in step 1 will become in proper solutions. This may take several PRs, ensuring that unit test / functional test keep working after each one.
3. Once the code is ok again, look to the .test files in  `cases_xml/`. Some of the .test could be just removed (typically, the ones with a counterpart .test based in JSON), other will need adaptation (just removing XML steps) or complete rewritting (i.e. rewritte the test with the same steps but in JSON). This may take also several PRs.

The target is at least to end steps 1 and 2 by the Orion 1.0.0 release date. Step 3 could be large and posponed to post-1.0.0 releases.

As a consecuence of step 1, a drop in coverage will occur, as a consecuence of dead code (as reference, coverage at 0.28.0 releasing date is around 78.2%). The gap will be reduced upon finishing step 2 and definitively closed when .test covering missing functionality will be migrated back during step 3.

**Effort: 10 man day** (mainly due to step 3)

kzangeli:
There are more things to consider when we remove XML.
Constructions were made just to be able to support both XML and JSON. Look for example at the "Format format", sent to all old render and check methods.
We will find things like this also to be removed when we stop supporting XML.

fgalan:
Yes, you are right. I see this the kind of thing you mention as "spin-off" of the main task about removal parsing/rendering code in step 2. I hope to easily find that stuff as we go.

fgalan:
Step 1 completed in PR #1864. The case_xml/ directory holds 202 .test files.

```
fermin@neodeb:~/src/fiware-orion$ find test/functionalTest/cases_xml | grep '.test' | wc -l
202
```

fgalan:
Part of step 2 done in PR https://github.com/telefonicaid/fiware-orion/pull/1866.

As result, some XML-based unit test without JSON countepart has been DISABLED, while we decide what to do with them.

fgalan:
After merging step 1 PR, coverage has drop 75.8%. Around 3.5% drop is not too much...

fgalan:
[...]

PR #1864, opened 2016-02-29
XML removal (part 1)

fgalan:
Completed step 1 in issue #1862


src/lib/rest/OrionError.cpp
```diff
@@ -161,8 +161,8 @@ std::string OrionError::render(ConnectionInfo* ciP, const std::string& _indent)
 
   if (format == NOFORMAT)
   {
-    // Default format is XML for "v1"
-    format = XML;
+    // Default format is JSON
+    format = JSON;
   }
```
kzangeli: I understand this entire **if** will be removed finally ...
'format' is no longer interesting as a variable. It is always JSON.

Wait, perhaps I'm too quick here ... We support common text also in some operations ...

test/functionalTest/harnessFunctions.sh
```diff
@@ -1242,13 +1246,17 @@ function coapCurl()
     then
       if [ "$_outFormat" == application/xml ] || [ "$_outFormat" == "" ]
```
kzangeli: Why not remove also the XML part of this 'double' if ?

kzangeli:
So, I understand this PR is just a small step (big in functests) in the complete XML removal. Still lots of stuff, in the source code,  to remove.

src/lib/rest/OrionError.cpp
fgalan: Yes, it seems we are still using format in the NGSIv2 for text rendering. From getEntityAttributeValue() function:

```
  bool         text       = (ciP->uriParamOptions["options"] == true || ciP->outFormat == TEXT);
```

(Side-note: the `ciP->uriParamOptions["options"] == true` part should be removed. It corresponds to an older version of the NGSIv2, no longer supported)

test/functionalTest/harnessFunctions.sh
fgalan: Because 0000_deprecated_checkings/xml_support.test needs to attemp use XML. Maybe the same could be done in a simpler way, but not in the scope of this PR (that's why I have leave the FIXME P10 here).

fgalan:
> So, I understand this PR is just a small step (big in functests) in the complete XML removal. Still lots of stuff, in the source code, to remove.

That's correct. This is the minimum viable PR to say "Orion is no longer supporting XML interactions" ensuring at the same time that `make test` will not break. But dead code and "disabled" test need to be properly addressed (in sucesive PRs).

src/lib/rest/OrionError.cpp
kzangeli: ok,

NTC

test/functionalTest/harnessFunctions.sh
kzangeli: ok,

NTC

kzangeli:
LGTM

crbrox:
LGTM

PR #1866, opened 2016-03-01
XML removal (part 2) 

fgalan:
Part of step 2 in #1862 

@kzangeli 
@crbrox 

All test harness and unit tests pass (see some commenta about unit test in the PR).


fgalan:
Unfortunatelly, the PR is so long that unit test changes are not shown. However, to summarize, the changes in unit test have been donde in test related with ngsi, ngsi9, ngsi10 and serviceRoutines. These changes are of two kinds:

* For these cases with a JSON test countepart (I have looked only test name, e.g. if I have "bar_xml" then I have searched for "foo_json", but without looking test body in detail to see if it is an exact copy) has been found, then the XML counterpart has been removed.

* For these cases without a JSON test counterpart (*), the test has been DISABLED. You can search these cases greping for "FIXME P5 #1862" in the code (have a look, please, as the PR is not showing that part).

Thus, an open question remains... what to do with these DISABLED test? One alternative would be to write the JSON countepart, but that which be costly. In addition, we should take into account that these functionality in ngsi, ngsi10, ngsi9 and serviceRoutin unit test is currently covered by test test harness. 

Thus, the easy solution could be removed these, assuming that they are already covered. However, let's hold that decission, in order to check that (i.e. the coverage metric at the end of the XML removal process should be equal or higher than the value before starting).

Opinions?

kzangeli:
LGTM

crbrox:
LGTM

Issue #1668, opened 2016-01-14
Websockets module: include status code and headers in output messages


fgalan:
(Parent issue: #1181)

Currently the output that the client gets is composed only for the equivalent to the payload in the REST API. However, response messages could also include other information pieces (specially in NGSIv2), in particular: response code and headers.

Maybe using a JSON envelope approach (as for input messages) could be also useful in this case.

fortizc:
Something like this?

```javascript
{
   "headers": {
      "Fiware-Service": "myServ",
      "Fiware-ServicePath": "/test2",
      "X-Auth-Token": <...>
   },
   "message": {
      "error":"NotFound",
      "description":"The requested entity has not been found. Check type and id"
   }
}
```

fgalan:
I understand that is using the same structure of input messages, I mean, input messages also use "headers" and "message" field. In that case, it looks good! :)

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1878.

PR #1878, opened 2016-03-03
Output Message and parameters processor

fortizc:
Fix issue #1668 and add a parameter processor

### DIFF
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -120,8 +120,15 @@ const char *ws_parser_message
   }
   writer.EndObject();
 
+  if (statusCode < 100 || statusCode > 599) // Code is not a valid HTTP status code
+    statusCode = 0;
+
   const char *headers = buff.GetString();
-  char *json = (char *) malloc(strlen(tmpl) + strlen(headers) + msg.size() + 1);
+
+  /* Sum all lengths, discard string format specifier (-6)
+   * and add status code (3) and the final null character
+   */
+  char *json = (char *) malloc(strlen(tmpl) - 6 + strlen(headers) + msg.size() + 3 + 1);
   sprintf(json, tmpl, headers, msg.c_str(), statusCode);
 
   return json;
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -87,6 +87,16 @@ void ws_parser_parse
   {
     head.gotHeaders = false;
   }
+
+  if (doc.HasMember("params") && doc["params"].IsObject())
+  {
+    rapidjson::Value::ConstMemberIterator it = doc["params"].MemberBegin();
+    while (it != doc["params"].MemberEnd())
+    {
+      uriParamGet(ciP, it->name.GetString(), it->value.GetString());
+      ++it;
+    }
+  }
 }
 
 const char *ws_parser_message
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -93,7 +93,7 @@ void ws_parser_parse
     rapidjson::Value::ConstMemberIterator it = doc["params"].MemberBegin();
     while (it != doc["params"].MemberEnd())
     {
-      uriParamGet(ciP, it->name.GetString(), it->value.GetString());
+      uriParamPush(ciP, it->name.GetString(), it->value.GetString());
       ++it;
     }
   }
--- a/src/lib/orion_websocket/parser.cpp
+++ b/src/lib/orion_websocket/parser.cpp
@@ -31,15 +31,18 @@
 
 #include "logMsg/logMsg.h"
 #include "rest/HttpHeaders.h"
+#include "rest/ConnectionInfo.h"
+#include "rest/rest.h"
 #include "parser.h"
 
 void ws_parser_parse
 (
-    const char*   msg,
-    std::string&  url,
-    std::string&  verb,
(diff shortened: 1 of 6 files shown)


fgalan:

src/lib/orion_websocket/parser.cpp
```diff
@@ -84,4 +87,42 @@ void ws_parser_parse
...
+  {
+    writer.Key(it->first.c_str());
+    writer.String(it->second->c_str());
+    ++it;
+  }
+  writer.EndObject();
+
+  const char *headers = buff.GetString();
+  char *json = (char *) malloc(strlen(tmpl) + strlen(headers) + msg.size() + 1);
+  sprintf(json, tmpl, headers, msg.c_str(), statusCode);
```
kzangeli: The 'json' buffer is not big enough here - you may write over it, causing memory corruption.

More correct size:

    strlen(tmpl) - 6 + strlen(headers) + msg.size() + 10 + 1;

    // The '6' comes from strlen of %s + %s + %d (removed from total len of 'tmpl')
    // The '10' is the maximum number of bytes in a string a 32 bit integer may need.
    // The '1' is of course cor the null termination of the string.

Also, if you use snprintf instead of sprintf any possible memory corruption is avoided. The entire string may not get inside, but at least there will never be any memory corruption - a lot easier to troubleshoot ...

src/lib/rest/rest.cpp
```diff
@@ -265,6 +265,15 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
...
 
+/* ****************************************************************************
+*
+* uriParamGet -
+*/
+int uriParamGet(ConnectionInfo *ciP, const char* ckey, const char* val)
+{
+  return uriArgumentGet((void *) ciP, (MHD_ValueKind)0, ckey, val);
+}
+
```
kzangeli: The name of this function is a little funny ...
The original name, **uriArgumentGet** came from 'getting uri param from MHD', but this function is more like 'pushing param to ConnectionInfo', right?

Perhaps we should rename **uriArgumentGet** to **uriArgumentReceive** and **uriParamGet** to **uriParamPush** ?

src/lib/orion_websocket/parser.cpp
fortizc: Yes, my mistake, it will be fixed. But status code has a maximum of 3 digits maybe the best choice is: `strlen(tmpl) - 6 + strlen(headers) + msg.size() + 3 + 1` 
and use a IF statement to discard any code when code < 100 && code > 599

kzangeli: That is also true :-)

src/lib/rest/rest.cpp
fortizc: Fixed in c5a9b0a99a474ff8c3255c9a2d482f402c28bc27 and 61d094cc2ffa707f196ba056586d7ece6821d69b

src/lib/orion_websocket/parser.cpp
fortizc: Fixed in 082f19edaafd04c9dfd85416083acc5252bc9360

Note: it must be code < 100 || code > 599

src/lib/orion_websocket/parser.cpp
```diff
@@ -84,4 +87,49 @@ void ws_parser_parse
...
+  while (it != head.headerMap.end())
+  {
+    writer.Key(it->first.c_str());
+    writer.String(it->second->c_str());
+    ++it;
+  }
+  writer.EndObject();
+
+  if (statusCode < 100 || statusCode > 599) // Code is not a valid HTTP status code
+    statusCode = 0;
```
kzangeli: 0 or 500 ?
This is a code that we actually send back to clients, right?
As 0 is not (afaik) a valid http status code, we can't rally use it ...
Perhaps 500 is a better choice here ?

kzangeli:
Apart from 0/500 'discussion', LGTM

src/lib/orion_websocket/parser.cpp
fortizc: Yes, maybe a 0 isn't the best choice, but 500 can be confused. What if the code come wrong (for any unknown reason) but the operation was successfully?

fgalan: As long as I unerstand (tell me if I'm wrong), `statusCode` is the response code of the operation. I don't understand why we are doing any manypulation... why don't just sent whatever code has arrived from the lower layer? Note that in the REST API, that is the code we will be sending, isn't it?

kzangeli: If the code comes wrong for some unknown reason, then it's a bug in the broker. Right?
500 seems like a good code for that.

#fgalan: this code is set **only** if the code is invalid, otherwise it goes thru as is.

fgalan: I agress that it should be a 500 Internal Error in that case.

However, probably this is not the right place for this checking and it should be moved to the lower layer (rest?) so we can take advantage of it for both the REST and the WS APIs.

fortizc: Fixed 02ded57bbc9e4413c44ff3e4e7549bda70e2d02b

src/lib/orion_websocket/parser.cpp
fgalan: What about to check that at the rest layer as I mention in my previous comment?

kzangeli:
LGTM (and handed back to @fgalan)

PR #1880, opened 2016-03-07
Strange fix to make functests work under valgrind in jenkins

kzangeli:


iariasleon:
LGTM

test/functionalTest/harnessFunctions.sh
```diff
@@ -278,6 +278,7 @@ function brokerStartAwait
     if [ "$?" == "0" ]
     then
       vMsg The orion context broker has started, listening on port $port
+      echo "Broker started after $loopNo checks" >> /tmp/brokerStart
```
kzangeli: This here is the list that 'fixes' the problem.
Hard to believe, right?

Also, even if the echo is done to /dev/null, the 'fix' still works.
Now, at least some info is gathered in /tmp/brokerStart (emptied in test/functionalTest/testHarness.sh)

NTC

fgalan: A comment should be added above this line explaining that if this line is removed then problems may occur due to unknown causes. Otherwise, we could forgot about this fix and, passed some time, removed this appartently "harmless" echo line.

test/functionalTest/harnessFunctions.sh
```diff
@@ -289,9 +290,10 @@ function brokerStartAwait
 
   sleep .5
 
-  # Check that CB started fine
-  curl -s localhost:${port}/version | grep version > /dev/null
+  # Check that CB started
+  curl -s localhost:${port}/version | grep version >> /tmp/brokerStart
```
fgalan: Is this related with the fix?

test/functionalTest/testHarness.sh
```diff
@@ -25,6 +25,7 @@
 date
 testStartTime=$(date +%s.%2N)
 MAX_TRIES=${CB_MAX_TRIES:-3}
+echo $testStartTime > /tmp/brokerStart
```
fgalan: Is this related with the fix?

kzangeli: Yes, I need to remove the file /tmp/brokerStart somewhere, as only '>>' is used.
If I don't do that, the file would grow and grow and grow ...

test/functionalTest/harnessFunctions.sh
kzangeli: Yes, just some more info in /tmp/brokerStart now that we have the file ...
Without this change I would have had to add some other line to indicate that the broker actually responds.

test/functionalTest/harnessFunctions.sh
kzangeli: True, it would be good to have this strange thing explained

test/functionalTest/harnessFunctions.sh
fgalan: I'd suggest to rename the file to something more meanifull taking into accont its content, e.g.

/tmp/brokerStart.counter

PR #1897, opened 2016-03-11
XML removal (part 6)

fgalan:
Part of issue #1862 

@kzangeli 
@crbrox

### DIFF
--- a/src/lib/jsonParse/jsonRequest.cpp
+++ b/src/lib/jsonParse/jsonRequest.cpp
@@ -182,10 +182,11 @@ std::string jsonTreat
 
 
   //
-  // If the payload is empty, the XML parsing library does an assert
-  // and the broker dies. I don't know about the JSON library, but just in case ...
+  // FIXME P4 #1862:
   //
-  // FIXME P4 #1862: check if we really need this now that Orion only support JSON
+  // This check comes from the old XML days, as the the XML parsing library did an assert
+  // and the broker died. We need to test what happen with the JSON library.
+  // If JSON library is "safer" with that regards, the check should be removed.
   //
   // 'OK' is returned as there is no error to send a request without payload.
   //
--- a/src/lib/jsonParse/jsonRequest.cpp
+++ b/src/lib/jsonParse/jsonRequest.cpp
@@ -184,7 +184,9 @@ std::string jsonTreat
   //
   // If the payload is empty, the XML parsing library does an assert
   // and the broker dies. I don't know about the JSON library, but just in case ...
-  // 
+  //
+  // FIXME P4 #1862: check if we really need this now that Orion only support JSON
+  //
   // 'OK' is returned as there is no error to send a request without payload.
   //
   if ((content == NULL) || (*content == 0))
--- a/src/lib/jsonParse/jsonUpdateContextAttributeRequest.cpp
+++ b/src/lib/jsonParse/jsonUpdateContextAttributeRequest.cpp
@@ -63,6 +63,9 @@ static std::string attributeValue(const std::string& path, const std::string& va
   //       save the 'typeFromXmlAttribute', a ContextAttribute has been added to 
   //       UpdateContextAttributeData.
   //
+  // FIXME P4 #1862: typeFromXmlAttribute was removed. The comment abov is obsolete and maybe
+  // also part of the implementation
+  //
   reqData->lastContextAttribute   = &reqData->upcar.attribute;
   reqData->upcar.res.valueType    = orion::ValueTypeString;
   reqData->upcar.res.contextValue = value;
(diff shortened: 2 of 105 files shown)


src/lib/jsonParse/jsonRequest.cpp
```diff
@@ -184,7 +184,9 @@ std::string jsonTreat
   //
   // If the payload is empty, the XML parsing library does an assert
   // and the broker dies. I don't know about the JSON library, but just in case ...
-  // 
+  //
+  // FIXME P4 #1862: check if we really need this now that Orion only support JSON
```
fgalan: What do you think, @kzangeli ?

src/lib/jsonParse/jsonUpdateContextAttributeRequest.cpp
```diff
@@ -63,6 +63,9 @@ static std::string attributeValue(const std::string& path, const std::string& va
   //       save the 'typeFromXmlAttribute', a ContextAttribute has been added to 
   //       UpdateContextAttributeData.
   //
+  // FIXME P4 #1862: typeFromXmlAttribute was removed. The comment abov is obsolete and maybe
```
fgalan: @kzangeli I'd need the new version of the above comment (or maybe it can be removed)?

src/lib/parse/CompoundValueNode.h
```diff
@@ -43,7 +43,7 @@ namespace orion
 * The class fields:
 * -------------------------------------------------------------------------------
 * o name         When parsing an XML payload, each node in the tree has a tag.
-*                The name of the node is taken from the tag-name.
+*                The name of the node is taken from the tag-name. (FIXME P4 #1862: reference to XML in this comment is obsolete)
```
fgalan: @kzangeli I'd need the new version of the comment for the name description.

src/lib/parse/compoundValue.cpp
```diff
@@ -162,6 +157,8 @@ void compoundValueEnd(ConnectionInfo* ciP, ParseData* parseDataP)
   // latest contextAttribute, i.e. the attribute whose 'contextValue' is the
   // owner of this compound value tree.
   //
+  // FIXME P4 #1862: obsolete reference to XML in the above comment?
```
fgalan: @kzangeli I'd need the new version of the above comment (or maybe it can be removed)?

src/lib/jsonParse/jsonRequest.cpp
kzangeli: This has nothing to do with XML.
If we want to remove the check, we'd have to test it.
What does the old json parser do if it is sent an empty payload (or NULL) ?
If it doesn't crash (as the XML parser), and the broker works just likr without the check, then it's safe to remove the check

src/lib/jsonParse/jsonUpdateContextAttributeRequest.cpp
kzangeli: Perhaps/Probably the entire 'lastContextAttribute mechanism' can be removed.
We need to see exactly where this pointer is used.

src/lib/parse/CompoundValueNode.h
kzangeli: Suggestion:

    * o name         Note that not all nodes have a name, not in JSON payloads, nor when
    *                getting the info from mongo BSON.

src/lib/parse/compoundValue.cpp
kzangeli: I looked into the whole lastContextAttribute mechanism.
It is very much needed, in the JSON v1 parse.
JSON v2 parse is completely different and no lastContextAttribute pointer is needed.

So, just change **XML** for **JSON V1** in the comment.

src/lib/rest/HttpStatusCode.h
```diff
@@ -48,7 +48,7 @@ typedef enum HttpStatusCode
   SccConflict               = 409,
   SccLengthRequired         = 411,   // Content-Length header missing
   SccRequestEntityTooLarge  = 413,   // Request Entity Too Large - over 1Mb of payload
-  SccUnsupportedMediaType   = 415,   // Unsupported Media Type (only support application/xml and application/json)
+  SccUnsupportedMediaType   = 415,   // Unsupported Media Type (only support and application/json)
```
kzangeli: only support plain/text and application/json ?

fgalan: Fixed in 64bf1a1

src/lib/jsonParse/jsonRequest.cpp
fgalan: Fixed in 64bf1a1

src/lib/jsonParse/jsonUpdateContextAttributeRequest.cpp
fgalan: Fixed (at tested with unit teset and ft) in 64bf1a1

src/lib/parse/CompoundValueNode.h
fgalan: Fixed in 64bf1a1

src/lib/parse/compoundValue.cpp
fgalan: Fixed in 64bf1a1

kzangeli:
LGTM

crbrox:
LGTM

PR #1900, opened 2016-03-11
Hardening/1705 csub cache objects

kzangeli:
Solves the main part of issue #1705 - the q-string-filter part.
Geometry stuff will have to come later, this PR is big enough as is.
Important func-test for error handling checks included as well as a pretty unnecessary functest for normal function - older functests already cover normal operation ...


CHANGES_NEXT_RELEASE
```diff
@@ -7,6 +7,5 @@
 - Add: new value for URI param 'options': unique - like 'values', but avoiding duplicates (Issue #1050)
 - Add: check min length for IDs in v2 (Issue #1836)
 - Fix: PATCH /v2/entities correct interpretaion of null value and missing type in payload
-. Fix: PATCH /v2/entities missing value field (Issue #1789)
-
-
+- Fix: PATCH /v2/entities missing value field (Issue #1789)
+- Fix: parsing 'q-string-filter' in early stages and storing resulting objects for mongo-backend to access (q-part of Issue #1705)
```
fgalan: Not sure if this should be in the CNR... the change is transparant from the point of view of the user, without functinoal impact or fixing any existing problem.

Maybe "Hardening: XXX" should be a better entry category

(In addition, we typically use "mongoBackend" instead of "mongo-backend" in CNR entries)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1682,7 +1688,27 @@ static bool processSubscriptions
     }
 
     /* Check 2: expression (q) */
-    if (!matchExpression(notifyCerP, trigs->expression.q))
+    if (noCache == true)
```
fgalan: Not sure if I understand this correctly... It seems that TriggeredSubscription uses both `expression` (with string) and `stringFilterP` (parsed object). The later is used directly in the cache case, the former is used in the no-cache case, with an "on the fly" expression -> StringFilter conversion step. Is that correct?

Why don't simply TriggeredSubscription (and the logic in this part of the code) removing the `expression` field and creating the objects pinted by stringFilterP at TriggeringSubscription construction/setting time (if I'm remembering correctly, in the addTrriggeredSubscptions() function).

src/lib/mongoBackend/TriggeredSubscription.cpp
```diff
@@ -97,9 +99,14 @@ TriggeredSubscription::~TriggeredSubscription()
...
 */
-void TriggeredSubscription::fillExpression (const std::string& q, const std::string& georel, const std::string& geometry, const std::string& coords)
+void TriggeredSubscription::fillExpression
+(
+  const std::string&  q,
+  const std::string&  georel,
+  const std::string&  geometry,
+  const std::string&  coords
+)
 {
```
fgalan: Related with my previous comment, this fillExpression() method should build the object pointed by stringFilterP (and the release() method should release the memory allocated here).

src/lib/cache/subCache.cpp
```diff
@@ -720,6 +720,16 @@ void subCacheItemInsert
   cSubP->expression.coords     = coords;
   cSubP->expression.georel     = georel;
 
+  std::string errorString;
+  if (q != "")
+  {
+    if (cSubP->expression.stringFilter.parse(q.c_str(), &errorString) == false)
+    {
+      LM_E(("Runtime Error (error parsing q-string '%s': %s)", q.c_str(), errorString.c_str()));
```
fgalan: Where is the bad input associated to a wrong q string in subsciption creation/update being managed?

src/lib/rest/StringFilter.h
```diff
@@ -0,0 +1,138 @@
...
+
+class ContextElementResponse;
+/* ****************************************************************************
+*
+* StringFilter - 
+*/
+class StringFilter
+{
+public:
+  std::vector<StringFilterItem>  filters;
```
fgalan: Some comments explaining this filed should be added. For example, what is plainString? The string representing the q filter? Regarding mongoFilters, it should be clarified that this field needs the call to mongoFilterPopulated() to be used. Do we need to call parse() before calling mongoFilterPopulate() and/or match()?

I thing adding some hint covering these topics will help if we later need to use or mantain this stuff.

src/lib/mongoBackend/TriggeredSubscription.cpp
```diff
@@ -121,3 +128,14 @@ std::string TriggeredSubscription::toString(const std::string& delimiter)
...
   return ss.str();
 }
+
+
+
+/* ****************************************************************************
+*
+* TriggeredSubscription::stringFilterSet - 
+*/
+void TriggeredSubscription::stringFilterSet(StringFilter* _stringFilterP)
```
fgalan: I understand this function is used when the string filter is populated into triggered subscription from cache. Is that correct?

src/lib/rest/rest.cpp
```diff
@@ -193,6 +195,23 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
       ciP->uriParamTypes.push_back(val);
     }
   }
+  else if (key == URI_PARAM_Q)
```
fgalan: Not sure of understanding this piece of code...

As far as I undertand, q filter is procesed and stored in a thread var stringFilterP. However, not sure were that variable is used after this parsing step...

src/lib/rest/StringFilter.cpp
```diff
@@ -0,0 +1,1101 @@
+/*
+*
+* Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This class now encapsulated all the complexity related with filtering. Good!

However, now old qStringFilters() and children functions in mongoBackend should disapear in favour of this new class. Otherwise, we would have duplicated code for the same functionality (with the aggravating circumstance that this is a particularly complex piece of code).

CHANGES_NEXT_RELEASE
kzangeli: Correct, the user should not care about these changes, seeing the broker as a 'black box'.
I'll remove the line from CNR.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: The pre-parsed string-filter is in the sub cache.
Without sub-cache, the string-filter will have to be parsed every time, with the 'q-string' obtained from mongo.

Now, if the parse is done here or at creation time of TriggeredSubscription  ... not so important, I'd say.

src/lib/cache/subCache.cpp
kzangeli: StringFilter::parse takes care of all error handling.
And, true, it is not a 'Runtime Error' but a 'Bad Input' error.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: The idea behin TriggeredSubscriptions is to have there all the information that, at processSubscription() time, the logic will need to send notifications. Thus, I think that to be coherent with that idea, the parsing should be done at that time.

src/lib/cache/subCache.cpp
fgalan: In addition, is the alarm logic "tied" with this error condition detection? I mean, would an error in q filter at subscription creation/updating time to raise a bad input alarm?

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: The pointer is simply obtained from the cache

src/lib/rest/rest.cpp
kzangeli: It is used in rest.cpp and RestService.cpp, this PR contains all its uses, of course

src/lib/rest/StringFilter.cpp
kzangeli: Yes, that is true.
With all the problems I had, I forgot to remove the old stuff

src/lib/rest/rest.cpp
fgalan: Could you elaborate a little bit more, pls? I don't "link" this piece of code with an actual usage in the mongoQuery logic...

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Sure, but ... it will complicate things a little.
Creating it just-in-time and then removing it is also not a bad idea.
Unless it has to be created more than once, of course. That would be a mistake.

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: I changed this in the "mem leak branch", not to use a pointer but a StringFilter object.
If moved here I will do the same.
TriggeredSubscription will have a StringFilter pointer, pointing to the sub.cache and a StringFilter object, used when the cache is not used.

Actually, I got the impression that lots of info is stored in both cache and in TriggeredSubscription. The whole mechanish should be looked over.

TriggeredSubscription is older than subCache and should probably have changed when the subCache was implemented.

src/lib/cache/subCache.cpp
kzangeli: Yes, of course. Bad Input aren't "LM_W" anymore, but calls to alarmMgr.badInput()

src/lib/mongoBackend/TriggeredSubscription.cpp
fgalan: > TriggeredSubscription is older than subCache and should probably have changed when the subCache was implemented.

However, note that we also have a `-noCache` way of working, so probably keep TriggeredSubscription idependent from the csub cache is a good idea.

src/lib/mongoBackend/TriggeredSubscription.cpp
fgalan: If this is set with the cache pointer and not actually copied, are we protected from concurrency problems? For example, one update context thread is ussing a TriggeredSubscription while a unsubscribe cache thread is deleting the same csub entry where the pointed StringFilter object lives.

src/lib/rest/rest.cpp
kzangeli: A subscription request comes in, the q-string is parsed and stored in the subCache.
Or, what is taken from mongo at startup (subscriptions) are parsed and stored in the cache.
TriggeredSubscriptions are esat to point to the parsed 'q expression' and the match method and mongoFilters vector is used by mongoBackend.

fgalan: Note that the Q_URI_PARAM is not used in subscriptions, but in queries. In subscriptions, que q param comes "encapsulated" inside `expression` field (see http://telefonicaid.github.io/fiware-orion/api/v2/ for details). In the case of queries Q_URI_PARAM is used to create an interim Scope (of type FIWARE::StringFilter) which is used in the mongoQueryContext() invocation at mongoBackend.

Thus, for the case of queries, maybe the parsing should be delayed to the scope generation time and the StringFilter object passed as a new attribute in the Scope class, as the current area fields for FIWARE::Location and FIWARE::Location::NGSIv2 filters. Probably the need for a thread var would disapear in this case (and maybe the associated leak), in the same way we haven't needed thread vars for area related stuff.

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 1987bf1

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: Yes, with noCache we have no choice but to parse the qStringFilter every time.
The change I did in TriggeredSubscription was simply to add a pointer to the parsed qStringFilter, pointing to the subCache.  When noCache is used, this pointer is NULL

src/lib/cache/subCache.cpp
kzangeli: Fixed in ca922d2

src/lib/mongoBackend/TriggeredSubscription.cpp
fgalan: In the case of noCache, I understand tha the StringFilter has to be build from the DB content. In the DB, the q filter string is stored in some place I don't remember right know :) but checking DB model documentation it should be easy to locate.

src/lib/rest/StringFilter.h
kzangeli: Fixed in f976caf

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in c194c34

src/lib/rest/rest.cpp
kzangeli: The commit 94f0b76 implements passing of the string filter to mongoQueryContext.
Before, ewhat happended was that the 'qString' was parsed upon request arrival, and then not used. Instead of using the parsed qString (stringFilterP), que original string was added as a scope and then the parsing was done again. It worked, of course, but not the way it was meant.

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: That is true, this could be a problem.
Two ways of fixing this:
1. sem-protect the operation (we probably already have a semaphore tu use, one protecting the subCache
2. cloning the StringFilter instead of pointing to it. Not valid, as we'd need to sem-protect the cloning anyways ...

I'll look into subCache and see if we have a semaphore there that I can use here.

kzangeli: New subCache strategy in ce05047

src/lib/cache/subCache.cpp
kzangeli: Fixed in ca922d2

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 5fb1e62

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: Yes, this is already implemented

NTC

src/lib/apiTypesV2/SubscriptionExpression.h
```diff
@@ -40,8 +43,9 @@ struct SubscriptionExpression
   std::string               geometry;
   std::string               coords;
   std::string               georel;
+
+  StringFilter              stringFilter;
```
fgalan: The `std::string q` field reamins in this class. I guess that it is due that the "string nature of q" is needed for some operations (e.g. persiting in DB). Maybe a comment about this could be useful (at a first glance, it may be a bit confusing have stringFilter in addition to q).

Or maybe I'm being too paranoid with documentation stuff :). If you find it self-evident, ignore this comment with a NTC.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -52,10 +52,10 @@
 #include "mongoBackend/connectionOperations.h"
 #include "mongoBackend/TriggeredSubscription.h"
 #include "cache/subCache.h"
-
+#include "rest/StringFilter.h"
 #include "ngsi/Scope.h"
 #include "rest/uriParamNames.h"
-
+#include "rest/StringFilter.h"
```
fgalan: Twice the same include? I guess this is a typo...

(Offtopic: it is a pity C/C++ doesn't have any mechanism to detect this redundancy... :( )

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -67,6 +67,14 @@ using namespace mongo;
 
 /* ****************************************************************************
 *
+* stringFilterP - 
+*/
+extern __thread StringFilter* stringFilterP;
```
fgalan: This is really needed or a leftover of the first version of the PR?

As far as I understand, StringFilter usage is "local":

* In the case of queries it is generated as part of the Scope passed to mongoBackend
* In the case of subscriptions creation/update, it is generated upon csub entry creation (except in the case of not using csub cache, in which case nothing is done regarding StringFilter)
* In the case of updated, StringFilter is retrieved from the cache (in the case of using csub cache) or generated on-the-fly from the q string stored in DB (in the case of not using csub cache).

The situation is similar to fields modeling areas, which are included as part of the Scope in the case of queries. Why in this case (StringFilter) we need a thread var and in that (area stuff) case we don't?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1619,21 +1643,21 @@ static bool processOnChangeConditionForUpdateContext
 }
 
 
-
+#if 0
```
fgalan: Leftover

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1509,7 +1513,27 @@ static bool addTriggeredSubscriptions_noCache
...
+            delete stringFilterP;
+          
+            LM_E(("Runtime Error (%s)", err.c_str()));
+            TIME_STAT_MONGO_READ_WAIT_STOP();
+            releaseMongoConnection(connection);
+            return false;
+          }
+          else
+          {
+            trigs->stringFilterSet(stringFilterP);
```
fgalan: Maybe the else branch is also a leftover of the first PR version? Why don't just?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3150,6 +3173,8 @@ void processContextElement
     return; // Error already in responseP
   } 
 
+  cacheSemTake(__FUNCTION__, "Processing Context Element");
```
fgalan: Thinking in the new location of the semaphores, let's take into account that processContextElement is a bit bunch of code, almost all the logic from updating each CE is there.

So, while update context is being processes other update context will be blocked. It is like going back to the old `-reqMutextPolicy all` days. Even worse: I understand that the cache update thread and subscription/update subscription operation will also block...

Not sure what to do regarding this. Maybe we should talk at skype.

(Sorry, probably I should have detected this when you told me about the changes in the sempahore strategy this morning).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -81,6 +82,14 @@ static bool                 multitenant;
 
 /* ****************************************************************************
 *
+* stringFilterP - 
+*/
+extern __thread StringFilter* stringFilterP;
```
fgalan: Same comments regarding the need of thread var for this.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -875,934 +881,9 @@ static bool processAreaScopeV2(const Scope* scoP, BSONObj &areaQuery)
 #define OPR_NOT_EXIST "NOT EXIST"
 
 
-
```
fgalan: Great to see all this stuff removed, now moved to the new classes :)

However, what about the #define above?

```
#define OPR_EXIST     "EXISTS"
#define OPR_NOT_EXIST "NOT EXIST"
```

Maybe the should be also removed?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1974,19 +1055,19 @@ bool entitiesQuery
         }
       }
     }
-    else if (sco->type == SCOPE_TYPE_SIMPLE_QUERY)
+    else if (sco->type == SCOPE_TYPE_SIMPLE_QUERY)  // Used for string filters in V1
     {
-      // FIXME P4: Once issue #1705 is implemented, this check can be removed.
-      if (qStringFilters(sco->value, filters) != true)
+      stringFilterP = new StringFilter();
+      if (stringFilterP->parse(sco->value.c_str(), err) == false)
```
fgalan: If we add a StringFilter to the Scope class I understand that we will save doing parsing at mongoBackend level. Similar case to geo-things, that are parsed at getEntities() invoking the Scope fill() method:

```
if (scopeP->fill(ciP->apiVersion, geometry, coords, georel, &errorString) != 0)
```

(As far as I have checked, Scope.cpp/h haven't been modified in this PR... maybe I'm wrong, it seems that github has changed its web interface and now it is more difficult to see...)

src/lib/mongoBackend/MongoGlobal.h
```diff
@@ -258,7 +258,8 @@ extern bool entitiesQuery
   const std::string&               sortOrderList  = "",
   bool                             includeCreDate = false,
   bool                             includeModDate = false,
-  const std::string&               apiVersion     = "v1"
+  const std::string&               apiVersion     = "v1",
+  StringFilter*                    stringFilterP  = NULL
```
fgalan: If we move StringFilter to Scope, we don't need this new paramter (it will come "inside" `const Restriction& res`).

src/lib/mongoBackend/TriggeredSubscription.h
```diff
@@ -78,9 +80,10 @@ class TriggeredSubscription
 
   ~TriggeredSubscription();
 
-  void fillExpression(const std::string& q, const std::string& georel, const std::string& geometry, const std::string& coords);
-
-  std::string toString(const std::string& delimiter);
+  // FIXME P5: This method will cease to exist once geo-stuff is implemented the same way StringFilter was implemented (for Issue #1705)
```
fgalan: Do you mean by adding georelFilterSet(), geometryFilterSet() and so?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -279,7 +279,8 @@ HttpStatusCode mongoQueryContext
   std::map<std::string, std::string>&  uriParams,
   std::map<std::string, bool>&         options,
   long long*                           countP,
-  const std::string&                   apiVersion
+  const std::string&                   apiVersion,
+  StringFilter*                        stringFilterP
```
fgalan: If we move StringFilter to Scope, we don't need this new paramter (it will come "inside" `QueryContextRequest* requestP`).

src/lib/rest/StringFilter.h
```diff
@@ -0,0 +1,159 @@
...
+class ContextElementResponse;
+/* ****************************************************************************
+*
+* StringFilter - 
+*
+* FIELDS
+*   filters       vector of filter-items, see StringFilterItem comment header
+*   mongoFilters  the filter-items translated to filters that can be used by mongo
+*                 Note that the method 'mongoFilterPopulate' must be called to translate
+*                 'filters' to 'mongoFilters'
```
fgalan: Good comment. However, what about using a similar one for match(), e.g:

"Note that the method 'match' needs to be called after 'parse'"

(Or maybe I'm wrong and it is not really needed... :)

src/lib/rest/rest.cpp
```diff
@@ -74,6 +75,7 @@ static struct sockaddr_in        sad;
 static struct sockaddr_in6       sad_v6;
 __thread char                    static_buffer[STATIC_BUFFER_SIZE + 1];
 __thread char                    clientIp[IP_LENGTH_MAX + 1];
+__thread StringFilter*           stringFilterP        = NULL;
```
fgalan: Same commenta about thread vars here.

src/lib/rest/rest.cpp
```diff
@@ -193,6 +195,23 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
       ciP->uriParamTypes.push_back(val);
     }
   }
+  else if (key == URI_PARAM_Q)
+  {
+    if (*val != 0)
```
fgalan: I think we don't need this parsing if at the end we move this to Scope (covering the query context case). Or maybe in the case of things already in Scope (like georel= ) we are doing also parsing a URI param level?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -46,6 +46,8 @@
 #include "serviceRoutines/postQueryContext.h"
 #include "jsonParse/jsonRequest.h"
 
+extern __thread StringFilter*  stringFilterP;
```
fgalan: Same commenta about thread var.

src/lib/rest/rest.cpp
fgalan: In addition, just to check that everything is covered... where the parsing error for create/update subscriptions (where q is in the payload) in the case of wrong q expression?

(Maybe it is self-evident in the PR, but it is a bit long and a hint will be great to locate that piece of code :)

src/lib/apiTypesV2/SubscriptionExpression.h
kzangeli: Yes, this 'q' cannot be removed I'm afraid (I tried :-)).
A short comment to be added the the function header.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: It is not C/C++, but the C preprocessor.
And cpp is pretty stupid ...
However there is the good old #ifndef/#define/#endif   or, even better, #pragma once  to avoid double inclusion.

To include the file twice is a mistake of course and removed in the next commit.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: These '__thread StringFilter* stringFilterP' are part of this PR.
Due to leakage (depending on the thread mechanism of MHD), detected in the coming PR, this variable is no longer a thread variable but a part of ConnectionInfo, but that is for the next PR, not this one.

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, this was a test that worked and the function could be removed.
I forgot to completely remove it.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: I don't fully understand this comment.

if (q != "") => if we have a string filter that is not empty

After that, the parse of the string filter fails, and as this 'q' came from the database, it is not a 'Bad Input' but an error in the broker accepting it in the first place.
This will hopefully never ever occur.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yeah, I know that the 'sem-lock-span' has grown a lot.
To be protected, it actually has to grow.
However, we might be able to narrow it down a little.
Let's talk ...

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: As responded earlier, must be here for this PR. Change comes in the next PR.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Those should be removed also, of course.

kzangeli: Fixed in f8ba716

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in f8ba716

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in f8ba716

src/lib/apiTypesV2/SubscriptionExpression.h
kzangeli: Fixed in f8ba716

src/lib/mongoBackend/TriggeredSubscription.h
kzangeli: 'q' has been removed, the other params will be removed also and the function will no lopnger make sense and will be removed.

Now, if we use georelFilterSet() etc or some other mechanism, that I cannot answer to right now. When this part of issue 1705 is implemented (or planned in detail) I will be able to answer that question.

NTC

src/lib/rest/StringFilter.h
kzangeli: Sure, the methods can get their short explanation as well

src/lib/rest/rest.cpp
kzangeli: Already responded.

NTC

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: NTC

src/lib/rest/rest.cpp
kzangeli: I can't really see any reason to move the parse part to Scope ...
What if in the future we want the q-string-filter in an operation that doesn't have Scope in the payload ?

About create/update subscriptions, I can't say I know.
Will look it up tomorrow.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: What about merging the branch that you have prepared for the next PR in the branch of the current PR? That should "fix" the thread var stuff an ease reviewing all in the same "context". Otherwise, we should take note of these inter-PR comments and verify in the new PR, i.e. we have to remember that some bits in the current PR need to be checked in the next one (complex for the reviewer).

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: I undertant that the effect is probably the same, as stringFilterP is NULL at this point. However, it less explicit and thus more complex, i.e. if I land in that part of the code and see `trigs->stringFilterSet(stringFilterP)` I ineed to look where that `stringFilterP` came from.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Yes, I think this should be improved. Note that csub cache was created to improve efficiency... but now we have the paradox that due to csub cache (more precicely, to the semaphores of csub cache) we are loosing any advantage that csub cache may bring!

src/lib/rest/rest.cpp
fgalan: > I can't really see any reason to move the parse part to Scope ...

The main reason is uniformity/homogeneity. From an external looker or a newcomer to the code it is difficult to explain why some filters are implemented as part of Scope (in particular the ones related with geo-stuff) and others (the one related with q) are implemented in a different way.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, I was thinking about doing that also.
Let's do it

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Valgrind branch merged in 9a3c661

src/lib/rest/rest.cpp
kzangeli: Valgrind branch merged in 9a3c661

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Valgrind branch merged in 9a3c661

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Valgrind branch merged in 9a3c661

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: No, why do you say that stringFilterP is NULL?
Look up 13 lines and you find:

    StringFilter* stringFilterP = new StringFilter();

This is the string filter of the 'plain string' in DB

src/lib/rest/StringFilter.h
kzangeli: Fixed in 31090cf

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: But 

```
StringFilter* stringFilterP = new StringFilter();
```

If within the no-else branch of the if. All my comment refers to the else branch of the if...

fgalan: Wait... there are two nested if. Re-evalutaing...

fgalan: Ok, all seems to be ok. Sorry for the noise.

NTC

kzangeli: ok :-) 
I knew there had to be some misunderstanding here.

kzangeli:
In the commit b10bc64, the parsing of URI param 'q' has been moved to service-routines and the StringFilter to the class 'Scope'.
This change is not 100% implemented, as 20 functests are still failing:

     0876_entity_dates/filter_datemodified_sub.test
     0876_entity_dates/filter_entity_dates.test
     0967_v2_query_op/illegal_chars_in_url_param.test
     0967_v2_query_op/q_in_uri_param_to_scope.test
     1039_ngsiv2_date_support/dates_in_q_uri_param.test
     1039_ngsiv2_date_support/dates_in_q_uri_param_lists.test
     1039_ngsiv2_date_support/dates_in_q_uri_param_ranges.test
     1080_batch_query_context/batch_query_context_scopes.test
     1306_string2double/string2double_not_a_number.test
     1316_create_subscription/post_subs_expr_initial_notification.test
     1589_query_entities_with_specific_attributes/query_entities_with_specific_attributes.test
     1592_query_with_q_op_greater_or_equal/attribute_exists_or_not.test
     1592_query_with_q_op_greater_or_equal/query_with_all_q_ops_without_right_hand_side.test
     1592_query_with_q_op_greater_or_equal/query_with_q_op_greater_or_equal.test
     1592_query_with_q_op_greater_or_equal/query_with_q_op_less_or_equal.test
     1594_query_with_q_and_equals_true/query_with_q_and_equals_true.test
     1595_query_with_q_and_string_values/query_with_q_and_string_values.test
     1607_query_with_q_and_ranges_but_invalid_operator/query_with_q_and_ranges_but_invalid_operator.test
     1658_v2_patch_subscription/patch_subs_expr_initial_notification.test
     1705_csub_cache_objects/csub_cache_objects_errors.test

Also unit tests got problems but this is fixed already.
MongoGlobal no longer parses the q-string in scopes and the unit tests attacked mongoQueryContext directly. Calls to Scope::stringFilter.parse() were added to fix the problem.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1331,13 +1329,12 @@ static bool addTriggeredSubscriptions_withCache
                                                            cSubP->subscriptionId,
                                                            cSubP->tenant);
 
-    sub->fillExpression(cSubP->expression.q, cSubP->expression.geometry, cSubP->expression.coords, cSubP->expression.georel);
-
+    sub->fillExpression(cSubP->expression.geometry, cSubP->expression.coords, cSubP->expression.georel);
+    sub->stringFilterSet(&cSubP->expression.stringFilter);  // FIXME PR: Instead of just referencing stringFilter, clone the whole thing
```
fgalan: I understand that the comment is obsolete, isn't it?

src/lib/serviceRoutinesV2/getEntities.cpp
```diff
@@ -175,25 +176,48 @@ std::string getEntities
       scopeP->release();
       delete scopeP;
 
+      TIMED_RENDER(out = oe.render(ciP, ""));
```
fgalan: This line seems to be a duplicate of 174... Do we need to render twice?

src/lib/rest/ConnectionInfo.h
```diff
@@ -160,6 +164,7 @@ class ConnectionInfo
   std::string                ip;
   std::string                apiVersion;
   RequestType                requestType;
+  StringFilter*              stringFilterP;
```
fgalan: After skype dicussion it seems this field is not needed. Try to remove it and let's see were it crashes :)

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1884,10 +953,10 @@ bool entitiesQuery
   const std::string&               sortOrderList,
   bool                             includeCreDate,
   bool                             includeModDate,
-  const std::string&               apiVersion
+  const std::string&               apiVersion,
+  StringFilter*                    stringFilterP
```
fgalan: As far as I understand, now the string filters are provides as Scopes (within ghte `res` parameter). Thus, probably this is a leftover of an previous version along the history of this PR.

src/lib/serviceRoutinesV2/getEntities.cpp
kzangeli: Of course not.
Cut 'n paste error, I imagine.

Fixed in e782813

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Of course.
Fixed in e782813

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1974,27 +1043,28 @@ bool entitiesQuery
         }
       }
     }
-    else if (sco->type == SCOPE_TYPE_SIMPLE_QUERY)
+    else if (scopeP->type == SCOPE_TYPE_SIMPLE_QUERY)
     {
-      // FIXME P4: Once issue #1705 is implemented, this check can be removed.
-      if (qStringFilters(sco->value, filters) != true)
+      for (unsigned int ix = 0; ix < scopeP->stringFilter.mongoFilters.size(); ++ix)
```
fgalan: Where the parse() (that is needed before using the mongoFitlers) is done for this Scope? At parsing stage?

src/lib/mongoBackend/TriggeredSubscription.cpp
```diff
@@ -117,7 +120,18 @@ std::string TriggeredSubscription::toString(const std::string& delimiter)
...
+
+
+
+/* ****************************************************************************
+*
+* TriggeredSubscription::stringFilterSet - 
+*/
+void TriggeredSubscription::stringFilterSet(StringFilter* stringFilterP)
+{
+  stringFilter = *stringFilterP;
```
fgalan: I understand that this method was the one thought for copying... but maybe I'm wrong.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Thread var is still there...

src/lib/serviceRoutines/postSubscribeContext.cpp
```diff
@@ -82,7 +84,47 @@ std::string postSubscribeContext
     return answer;
   }
 
-  TIMED_MONGO(ciP->httpStatusCode = mongoSubscribeContext(&parseDataP->scr.res, &scr, ciP->tenant, ciP->uriParam, ciP->httpHeaders.xauthToken, ciP->servicePathV));
+
+  //
+  // String filter in URI param 'q' ?
```
fgalan: NGSIv1 subscriptions doesn't support q. Note that q as URI param has been only defined for `GET /v2/entities` operation. Thus I guess that all this new code should be rolled-back.

src/lib/serviceRoutinesV2/postSubscriptions.cpp
```diff
@@ -62,7 +62,48 @@ extern std::string postSubscriptions
     return answer;
   }
 
-  TIMED_MONGO(ciP->httpStatusCode = mongoSubscribeContext(&parseDataP->scr.res, &scr, ciP->tenant, ciP->uriParam, ciP->httpHeaders.xauthToken, ciP->servicePathV));
+
+  //
+  // String filter in URI param 'q' ?
+  // What if it was also in NotifyConditionVector::NotifyCondition::Expression ... ?
+  //
+  if (ciP->uriParam[URI_PARAM_Q] != "")
```
fgalan: q parameter works only for `GET /v2/entities`. In the case of subscriptions the q parameter is part of the payload, already parsed by parsing stage (I guess) and put inside a SubscriptionExpression. Thus, I think that no special processing has to be done in serviceRoutinesV2/postSubscriptions.cpp.

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: Yeah, this is a 'C++ miracle'. I tried it and it actually works!

    stringFilter = *stringFilterP;

That's a object copy, like a memcpy(&stringFilter, stringFilterP), but including the vectors inside stringFilterP.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2726,14 +1796,9 @@ bool processOnChangeConditionForSubscription
   Restriction                   res;
   ContextElementResponseVector  rawCerV;
 
-  /* Include scopes for entitiesQuery() if q filter is not empty */
```
fgalan: The right way of doing this is using "Scope inyection", i.e. recover the deleted code, as you have done in other placess. You would need to do the parse() step in order to "prepare" it for entitiesQuery().

In addition, that would allow to avoid stringFilterP and to go back to the old call to entitiesQuery()... That `NULL, NULL, NULL, "", false, false, "v1") is very hard to follow:)

src/lib/mongoBackend/TriggeredSubscription.cpp
fgalan: I'd suggest a comment above this line:

```
// This is an object copy, like memcpy(&stringFilter, stringFilterP), but including the vectors inside stringFilterP.
```

Probably for C++ people more skilled than me is obvious, but... I think it doesn't hurt.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: entitiesQuery is called from a number of places. Difficult to respond exactly to this question.
But yes, if the q-string is in the payload, the parsing is done in the JSON parsing stage.
If URI param (GET /v2/entities), then the parse is done in the service routine.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yes, this has been modified, the string filter is now passed via Scope

Fixed in some commit, March 22, I think ...

src/lib/serviceRoutinesV2/postSubscriptions.cpp
kzangeli: The usage of ciP->uriParam[URI_PARAM_Q] everywhere except in GET /v2/entities

src/lib/serviceRoutines/postSubscribeContext.cpp
kzangeli: Fixed

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: Fixed in 85774e1

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: The string filter is now always passed as a Scope
Fixed in many of the latest commits

src/lib/rest/ConnectionInfo.h
kzangeli: Yup, it has been removed and finally everything seems to work.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Last remaining `__thread stringFilterP` removed in d9fdaec

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: semaphore problem fixed by making a copy of the string filter - in 6e5b741

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: There are no more instances of `__thread StringFilter`

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: StringFilter added to Scope, in one of the commits quite some time ago...

src/lib/mongoBackend/MongoGlobal.h
kzangeli: StringFilter is now always passed inside a Scope

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: StringFilter is now always passed inside a Scope

src/lib/rest/rest.cpp
kzangeli: Fixed ... many different commits

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Not anymore :-)  	d9fdaec

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -110,16 +111,14 @@ std::string Condition::toJson()
...
+  JsonHelper jhe;
 
-    jh.addRaw("expression", jhe.str());
-  }
+  jhe.addString("q", this->expression.q);
+  jhe.addString("geometry", this->expression.geometry);
+  jhe.addString("coords", this->expression.coords);
+  jhe.addString("georel", this->expression.georel);
+  
+  jh.addRaw("expression", jhe.str());
```
fgalan: Just to be sure... only indent changes here? (That case mark this as NTC)

src/lib/apiTypesV2/SubscriptionExpression.h
```diff
@@ -27,21 +27,30 @@
...
+* NOTE
+*   This struct contains both 'q' and stringFilter.
+*   q is the 'plain string' of the stringFilter and it is used in parsing of V2 Subscriptions
+*   and when the q-string is read from the database.
 */
 struct SubscriptionExpression
 {
-  SubscriptionExpression(): isSet(false) {}
-
+SubscriptionExpression(): isSet(false) { }
```
fgalan: Not sure but... member methods in a class should be indent in 2 spaces, shouldn't they?

src/lib/cache/subCache.cpp
```diff
@@ -705,20 +707,25 @@ void subCacheItemInsert
...
-  cSubP->throttling            = throttling;
-  cSubP->lastNotificationTime  = lastNotificationTime;
-  cSubP->notifyFormat          = notifyFormat;
-  cSubP->next                  = NULL;
-  cSubP->count                 = (notificationDone == true)? 1 : 0;
-  cSubP->expression.q          = q;
-  cSubP->expression.geometry   = geometry;
-  cSubP->expression.coords     = coords;
-  cSubP->expression.georel     = georel;
+  cSubP->tenant                  = (tenant[0] == 0)? NULL : strdup(tenant);
```
fgalan: Only indent changes here?

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Makes sense. NTC

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2718,26 +1779,18 @@ bool processOnChangeConditionForSubscription
   const std::string&               tenant,
   const std::string&               xauthToken,
   const std::vector<std::string>&  servicePathV,
-  const std::string&               qFilter
+  Restriction*                     resP
```
fgalan: I think (not sure) that the convenion for read-only parameter was to use const X&. Thus, why don't use the following for this?

```
const Restriction& res
```

(In some cases, it is difficult for the caller to get the data as const X& and a pointer is better. However, please consider if it possible to do it in "the right way". If not possible, mark this as NTC).

src/lib/apiTypesV2/Subscription.cpp
kzangeli: Yeah, almost ...
I removed an unnecessary block also - that's why the indentation changed

src/lib/apiTypesV2/SubscriptionExpression.h
kzangeli: Strange, it is not indented at all ...
And am empty space inside {} ...

Fixing it now!

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -149,7 +168,7 @@ HttpStatusCode mongoSubscribeContext
                                              tenant,
                                              xauthToken,
                                              servicePathV,
-                                             requestP->expression.q);
+                                             &restriction);
```
fgalan: After looking this, I guess that the `const Restriction& rest` alternative is possible.

src/lib/apiTypesV2/SubscriptionExpression.h
kzangeli: Now I understand why.
My emacs sees it as a label due to the ":", and it indented it that way.
My fault for not paying attention.

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -133,11 +133,30 @@ HttpStatusCode mongoSubscribeContext
...
     sub.append(CSUB_ATTRS, attrs.arr());
 
+
+    //
+    // StringFilter in Scope?
+    //
+    Restriction    restriction;
+    StringFilter*  stringFilterP = NULL;
+
+    for (unsigned int ix = 0; ix < requestP->restriction.scopeVector.size(); ++ix)
```
fgalan: If I'm understanding correctly, you are processing `requestP->restriction` to copy Scopes to a local `restriction` variable. Why don't just passing the `requestP->restriction` (either as pointer or as const X& as told in other comments) in the processConditionVector() call?

src/lib/cache/subCache.cpp
kzangeli: Yes, apart for the adding of if (stringFilterP != NULL) ... further down.
To tell you the truth I don't know why the indentation has been changed.
Probably at some time I had another assignment in the list, that required the new indentation.
Yes!   I had 

    cSubP->expression.stringFilter = *stringFilterP;

But as often stringFilterP is NULL, I got hundreds of sigsegvs in functests due to that line ... :-)

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -190,6 +209,11 @@ HttpStatusCode mongoSubscribeContext
 
     LM_T(LmtSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
 
+    //
+    // Any Scope of type SCOPE_TYPE_SIMPLE_QUERY in restriction.scopeVector?
```
fgalan: Not sure if understanding this comment or if it a leftover... Maybe it need some clarification.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I tried - `const Restriction&` was my first attempt, of course.
But, the compiler didn't let me so I had to use pointers instead ...

NTC

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -231,6 +231,25 @@ HttpStatusCode mongoUpdateContextSubscription
...
   }
 
+
+  //
+  // StringFilter in Scope?
+  //
+  Restriction    restriction;
+  StringFilter*  stringFilterP = NULL;
+
+  for (unsigned int ix = 0; ix < requestP->restriction.scopeVector.size(); ++ix)
```
fgalan: Same comment as in https://github.com/telefonicaid/fiware-orion/pull/1900/files#r57232596

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Probably a leftover, yes.
I moved this part to before the call to ... processAttributeVector?

src/lib/serviceRoutines/postSubscribeContext.cpp
```diff
@@ -51,6 +52,7 @@
 *
 * URI parameters
 *   - notifyFormat=XXX    (used by mongoBackend)
```
fgalan: notifyFormat is no longer used after removing XML. However, not so important for this PR. It can be removed within the "Remove XML (part X)" serie.

NTC

src/lib/serviceRoutinesV2/patchSubscription.cpp
```diff
@@ -65,6 +65,10 @@ std::string patchSubscription
   // 'Fill In' UpdateContextSubscriptionRequest
   parseDataP->ucsr.res.subscriptionId.set(subscriptionId);
 
+  // If a string-filter is present, it is parsed in 
+  // jsonParseV2/parseSubscription.cpp, function parseNotifyConditionVector() and
+  // the resulting StringFilter object resides in XXX
```
fgalan: "XXX" is correct in this comment or it has to be replaced by an actual word?

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: True, that should work too, and removes code ans is faster in run-time.
I will look into it but I really think you're right

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Yeah, it might look like that. But it is not.
The problem is not here, but inside processConditionVector where it is sent to another function.

src/lib/serviceRoutinesV2/postBatchQuery.cpp
```diff
@@ -62,12 +65,42 @@ std::string postBatchQuery
...
   BatchQuery*           bqP  = &parseDataP->bq.res;
   QueryContextRequest*  qcrP = &parseDataP->qcr.res;
+  std::string           q    = ciP->uriParam[URI_PARAM_Q];
   Entities              entities;
   std::string           answer;
 
   qcrP->fill(bqP);
   bqP->release();  // qcrP just 'took over' the data from bqP, bqP no longer needed
 
+  //
```
fgalan: Batch query (i.e. "POST /v2/op/query") doesn't allow `q` param. The parameters supported by this operation (see http://telefonicaid.github.io/fiware-orion/api/v2/) are limit, offset, orderBy and options.

src/lib/serviceRoutinesV2/patchSubscription.cpp
kzangeli: Unfinished comment.
XXX to be replaced by the real info.

src/lib/apiTypesV2/Subscription.cpp
fgalan: NTC

src/lib/cache/subCache.cpp
fgalan: NTC

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed.

src/lib/mongoBackend/MongoGlobal.h
fgalan: Fixed.

src/lib/mongoBackend/mongoQueryContext.cpp
fgalan: Fixed.

src/lib/rest/ConnectionInfo.h
fgalan: Fixed.

src/lib/serviceRoutinesV2/postSubscriptions.cpp
fgalan: Not in every place... "POST /v2/op/query" is yet using q= and it shuldn't. However, that is addressed in a separate comment, so this one ends with NTC.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2718,26 +1779,18 @@ bool processOnChangeConditionForSubscription
...
-    res.scopeVector.push_back(sc);
-  }
-
-  if (!entitiesQuery(enV, attrL, res, &rawCerV, &err, true, tenant, servicePathV))
+  if (!entitiesQuery(enV, attrL, *resP, &rawCerV, &err, true, tenant, servicePathV))
   {
     ncr.contextElementResponseVector.release();
     rawCerV.release();
-    res.release();
+    resP->release();
```
fgalan: `resP` is provides by the caller of processOnChangeConditionForSubscription() functions. For processOnChangeConditionForSubscription() this is like a read-only parameter.

Thus, processOnChangeConditionForSubscription() should be a good fellow :) and not releasing this (as processOnChangeConditionForSubscription() doesn't know if it caller will need resP this for something else).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2760,11 +1813,11 @@ bool processOnChangeConditionForSubscription
       ContextElementResponseVector  allCerV;
       AttributeList                 emptyList;
 
-      if (!entitiesQuery(enV, emptyList, res, &rawCerV, &err, false, tenant, servicePathV))
+      if (!entitiesQuery(enV, emptyList, *resP, &rawCerV, &err, false, tenant, servicePathV))
       {
         rawCerV.release();
         ncr.contextElementResponseVector.release();
-        res.release();
+        resP->release();
```
fgalan: Same comment here.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2779,7 +1832,7 @@ bool processOnChangeConditionForSubscription
         getNotifier()->sendNotifyContextRequest(&ncr, notifyUrl, tenant, xauthToken, format);
         allCerV.release();
         ncr.contextElementResponseVector.release();
-        res.release();
+        resP->release();
```
fgalan: Same comment here (last comment about this... but if you search/grep for `resP->release()` you will catch them all).

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
```diff
@@ -42,23 +42,23 @@ accumulatorStart
...
 
-echo "01. At t_0: Create subscription with q: dateCreated==t_2..t_4"
+echo "01. At t_0: Create subscription with q: dateCreated==t_2..t_6"
 echo "============================================================="
 payload='
 {
     "subject": {
         "entities": [
             {
-                "idPaatern": "E.*",
```
fgalan: idPaatern? How this was working before the change? Maybe because fields different from id or idPattern is interpreted as "no field" and, in sequence, "no field" is intepreted as ".*" pattern?

This case passed unnoticed until now supports the (already taken decission) of being explicit regarding id/idPattern and don't allow implicit defaults ;)

NTC

test/functionalTest/cases/1306_string2double/string2double_not_a_number.test
```diff
@@ -148,12 +148,12 @@ echo
...
 ============================
 HTTP/1.1 400 Bad Request
-Content-Length: 63
+Content-Length: 103
 Content-Type: application/json
 Date: REGEX(.*)
 
 {
-    "description": "invalid query expression",
+    "description": "values of type /String/ not supported for operator /GreaterThan/",
```
fgalan: This is going to change in the future (see https://github.com/telefonicaid/fiware-orion/issues/1945).

NTC (just to prepare for the future :)

test/functionalTest/cases/1705_csub_cache_objects/csub_cache_objects.test
```diff
@@ -0,0 +1,57 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: As "remark note": a .test is missing in this PR (we have been discussing about it at skype today but I'm unable right now of remembering the topic of the .test :)

test/functionalTest/cases/1705_csub_cache_objects/csub_cache_objects.test
```diff
@@ -0,0 +1,57 @@
...
+Sub cache objects
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0
+
+--SHELL--
+
+#
+# 01. GET /v2/entities with a very complete 'q'
```
fgalan: The test is named csub_cache_objects.test but it is not actually related with csub cache... Maybe the file names at .test and `--NAME--`s inside 1705_csub_cache_objects/ need to be revieweed a bit...

kzangeli: Yeah, this is the test I used just to see that the qString filter parse was working.
I had planned to not include it (or make it grow to actually do what it was meant to do initially) but in the last minute we decided to include it after all ...

test/functionalTest/cases/1705_csub_cache_objects/csub_cache_objects.test
kzangeli: The missing test is to create a subscription with a string filter and then update it (with a different string filter) and make sure it works with the new filter.
I have already seen in the source code that it is taken care of, but a test is always good to 'protect it' against future bugs.

src/lib/apiTypesV2/SubscriptionExpression.h
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/serviceRoutinesV2/patchSubscription.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/serviceRoutinesV2/postBatchQuery.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Fixed in bb0ba50ef684dc35cd098014b96fbe058de70b4a

test/functionalTest/cases/1705_csub_cache_objects/csub_cache_objects.test
fgalan: Spined off this (very looooong :) PR to tech debt issue https://github.com/telefonicaid/fiware-orion/issues/1948

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1269,7 +1266,28 @@ static bool addTriggeredSubscriptions_noCache
...
+        if (q != "")
+        {
+          StringFilter* stringFilterP = new StringFilter();
+
+          if (stringFilterP->parse(q.c_str(), &err) == false)
+          {
+            delete stringFilterP;
+          
+            LM_E(("Runtime Error (%s)", err.c_str()));
+            TIME_STAT_MONGO_READ_WAIT_STOP();
```
fgalan: Maybe this TIME_STAT_blabla() should be removed? It seems it is already been done at https://github.com/telefonicaid/fiware-orion/pull/1900/files#diff-4d88547ce2fb3f34e774c32845272c20R1197

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: NTC

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -201,6 +221,7 @@ HttpStatusCode mongoSubscribeContext
     LM_T(LmtSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
 
     cacheSemTake(__FUNCTION__, "Inserting subscription in cache");
+    LM_W(("Calling subCacheItemInsert with status == %s", status.c_str()));
```
fgalan: This seems to be a "regular" trace... why is this in warning level? Maybe a leftover?

src/lib/rest/rest.cpp
```diff
@@ -988,6 +988,7 @@ static int connectionTreat
 
     LM_T(LmtRequest, (""));
     LM_T(LmtRequest, ("--------------------- Serving request %s %s -----------------", method, url));
+    LM_W(("KZ: --------------------- Serving request %s %s -----------------", method, url));
```
fgalan: Oh oh... did your grep for "KZ" failed? ;)

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: You mean it has been removed and then put back again?
Perhaps via one of the many merges with develop ...
I have no problem in removing it (again?), no reason to measure times in failed requests, I guess ...

src/lib/serviceRoutines/postSubscribeContext.cpp
```diff
@@ -51,6 +52,7 @@
...
+*   - q
```
fgalan: "q" as URL param is only valid in "GET /v2/entities" operation (please check http://telefonicaid.github.io/fiware-orion/api/v2/)

src/lib/serviceRoutinesV2/postBatchQuery.cpp
```diff
@@ -51,6 +53,7 @@
 *   - limit=NUMBER
 *   - offset=NUMBER
 *   - options=count,keyValues
+*   - q
```
fgalan: "q" as URL param is only valid in "GET /v2/entities" operation (please check http://telefonicaid.github.io/fiware-orion/api/v2/)

src/lib/serviceRoutinesV2/postBatchQuery.cpp
```diff
@@ -62,6 +65,7 @@ std::string postBatchQuery
 {
   BatchQuery*           bqP  = &parseDataP->bq.res;
   QueryContextRequest*  qcrP = &parseDataP->qcr.res;
+  std::string           q    = ciP->uriParam[URI_PARAM_Q];
```
fgalan: "q" as URL param is only valid in "GET /v2/entities" operation (please check http://telefonicaid.github.io/fiware-orion/api/v2/)

...and in this case, note that is not a fix in a function preamble but in actual code.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Removed in 38cfe0a

fgalan: The issue is not about measuing times in failed request... is that the "end mark" for that time measure has been already done. In other words, you were calling twice in a row TIME_STAT_MONGO_READ_WAIT_STOP().

However, I may wrong. Please review the whole function to ensure that now is working properly for the different ways the function may `return`.

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Yes, I use LM_W(("KZ; ...")) for my traces, and then I grep on KZ to remove them. Here it seems I forgot the 'KZ' part so I didn't find it in my grep for removal

src/lib/rest/rest.cpp
kzangeli: Strange ... I remember removing that trace ...
I might have put it back and then forgotten it ...

Fixed in c4344c3

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: Fixed in c4344c3

src/lib/serviceRoutinesV2/postBatchQuery.cpp
kzangeli: Fixed in f72def2

src/lib/serviceRoutinesV2/postBatchQuery.cpp
kzangeli: Fixed in f72def2

src/lib/serviceRoutines/postSubscribeContext.cpp
kzangeli: Fixed in f72def2

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, I checked. Only two measurements in this file and both of them correctly started/ended

fgalan:
LGTM!! Finally!

PR #1914, opened 2016-03-17
Some accumulated changes for release/0.28.0

fgalan:
Note that the landing branch is release/0.28.0, not develop (thus **don't delete branch after merging, as we would need to merge into develop**).

This PR correspond to a sum of minor changed developed as hotfixes for orion.lab. Most of them are related with loging, although there are some things to discuss (see specific comments on the PR).

@kzangeli 
@crbrox

### DIFF
--- a/src/lib/alarmMgr/AlarmManager.cpp
+++ b/src/lib/alarmMgr/AlarmManager.cpp
@@ -172,6 +172,7 @@ bool AlarmManager::dbError(const std::string& details)
   {
     if (dbErrorLogAlways)
     {
+      //FIXME P10: to be discussed at PR time. DB alarms are loged at ERROR level, so this should be LM_E... copy-paste error? :)
       LM_W(("Repeated Database Error: %s", details.c_str()));
     }
 
--- a/src/lib/alarmMgr/AlarmManager.cpp
+++ b/src/lib/alarmMgr/AlarmManager.cpp
@@ -172,8 +172,7 @@ bool AlarmManager::dbError(const std::string& details)
   {
     if (dbErrorLogAlways)
     {
-      //FIXME P10: to be discussed at PR time. DB alarms are loged at ERROR level, so this should be LM_E... copy-paste error? :)
-      LM_W(("Repeated Database Error: %s", details.c_str()));
+      LM_E(("Repeated Database Error: %s", details.c_str()));
     }
 
     return false;
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2871,8 +2871,8 @@ static void updateEntity
 
   // The hasField() check is needed as the entity could have been created with very old Orion version not
   // supporting modification/creation dates
-  notifyCerP->contextElement.creDate = r.hasField(ENT_CREATION_DATE)     ? getLongField(r, ENT_CREATION_DATE)     : -1;
-  notifyCerP->contextElement.modDate = r.hasField(ENT_MODIFICATION_DATE) ? getLongField(r, ENT_MODIFICATION_DATE) : -1;
+  notifyCerP->contextElement.creDate = r.hasField(ENT_CREATION_DATE)     ? getIntOrLongFieldAsLong(r, ENT_CREATION_DATE)     : -1;
+  notifyCerP->contextElement.modDate = r.hasField(ENT_MODIFICATION_DATE) ? getIntOrLongFieldAsLong(r, ENT_MODIFICATION_DATE) : -1;
 
   if (!processContextAttributeVector(ceP,
                                      action,
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1428,8 +1428,6 @@ static bool addTriggeredSubscriptions_noCache
   DBClientBase* connection = getMongoConnection();
   if (collectionQuery(connection, collection, query, &cursor, &errorString) != true)
   {
-    // FIXME P10: collectionQuery already calls dbMgr in the case of problem. Redundant alarmMsg.dbError() ?
-    //alarmMgr.dbError(errorString);
     TIME_STAT_MONGO_READ_WAIT_STOP();
     releaseMongoConnection(connection);
     return false;
@@ -1444,7 +1442,7 @@ static bool addTriggeredSubscriptions_noCache
     std::string err;
     if (!nextSafeOrErrorF(cursor, &sub, &err))
     {
-      LM_E(("Runtime Error (exception in nextSafe(): %s - with query: <%s>)", err.c_str(), query.toString().c_str()));
+      LM_E(("Runtime Error (exception in nextSafe(): %s - query: %s)", err.c_str(), query.toString().c_str()));
       continue;
     }
     BSONElement  idField  = getFieldF(sub, "_id");
@@ -3278,7 +3276,7 @@ void processContextElement
     BSONObj r;
     if (!nextSafeOrErrorF(cursor, &r, &err))
     {
-      LM_E(("Runtime Error (exception in nextSafe(): %s - with query: <%s>)", err.c_str(), query.toString().c_str()));
+      LM_E(("Runtime Error (exception in nextSafe(): %s - query: %s)", err.c_str(), query.toString().c_str()));
       continue;
     }
     docs++;
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -311,7 +311,7 @@ static bool matchMetadata(BSONObj& md1, BSONObj& md2)
 {
 
   // Metadata is identified by name, type is not part of identity any more
-  return  getStringField(md1, ENT_ATTRS_MD_NAME) == getStringField(md2, ENT_ATTRS_MD_NAME);
+  return  getStringField(md1, ENT_ATTRS_MD_NAME, __FUNCTION__) == getStringField(md2, ENT_ATTRS_MD_NAME, __FUNCTION__);
 
 }
 
@@ -336,11 +336,11 @@ static bool equalMetadataValues(BSONObj& md1, BSONObj& md2)
   // If type exists in both metadata elments, check if they are the same
   if (md1TypeExist && md2TypeExist)
   {
-    if (getField(md1, ENT_ATTRS_MD_TYPE).type() != getField(md2, ENT_ATTRS_MD_TYPE).type())
+    if (getField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__).type() != getField(md2, ENT_ATTRS_MD_TYPE, __FUNCTION__).type())
     {
       return false;
     }
-    switch (getField(md1, ENT_ATTRS_MD_TYPE).type())
+    switch (getField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__).type())
     {
       /* FIXME not yet, issue #1068 Support array and object in metadata value
       case Object:
@@ -353,21 +353,21 @@ static bool equalMetadataValues(BSONObj& md1, BSONObj& md2)
       */
 
     case NumberDouble:
-      if (getField(md1, ENT_ATTRS_MD_TYPE).Number() != getField(md2, ENT_ATTRS_MD_TYPE).Number())
+      if (getField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__).Number() != getField(md2, ENT_ATTRS_MD_TYPE, __FUNCTION__).Number())
       {
         return false;
       }
       break;
 
     case Bool:
-      if (getBoolField(md1, ENT_ATTRS_MD_TYPE) != getBoolField(md2, ENT_ATTRS_MD_TYPE))
+      if (getBoolField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__) != getBoolField(md2, ENT_ATTRS_MD_TYPE, __FUNCTION__))
       {
         return false;
       }
       break;
 
     case String:
-      if (getStringField(md1, ENT_ATTRS_MD_TYPE) != getStringField(md2, ENT_ATTRS_MD_TYPE))
+      if (getStringField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__) != getStringField(md2, ENT_ATTRS_MD_TYPE, __FUNCTION__))
       {
         return false;
       }
@@ -381,19 +381,19 @@ static bool equalMetadataValues(BSONObj& md1, BSONObj& md2)
       break;
 
     default:
-      LM_E(("Runtime Error (unknown JSON type for metadata NGSI type: %d)", getField(md1, ENT_ATTRS_MD_TYPE).type()));
+      LM_E(("Runtime Error (unknown JSON type for metadata NGSI type: %d)", getField(md1, ENT_ATTRS_MD_TYPE, __FUNCTION__).type()));
       return false;
       break;
     }
   }
 
   // declared types are equal. Same value ?
-  if (getField(md1, ENT_ATTRS_MD_VALUE).type() != getField(md2, ENT_ATTRS_MD_VALUE).type())
+  if (getField(md1, ENT_ATTRS_MD_VALUE, __FUNCTION__).type() != getField(md2, ENT_ATTRS_MD_VALUE, __FUNCTION__).type())
   {
     return false;
   }
 
-  switch (getField(md1, ENT_ATTRS_MD_VALUE).type())
+  switch (getField(md1, ENT_ATTRS_MD_VALUE, __FUNCTION__).type())
   {
     /* FIXME not yet
     case Object:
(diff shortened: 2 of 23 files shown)


src/lib/alarmMgr/AlarmManager.cpp
```diff
@@ -172,6 +172,7 @@ bool AlarmManager::dbError(const std::string& details)
   {
     if (dbErrorLogAlways)
     {
+      //FIXME P10: to be discussed at PR time. DB alarms are loged at ERROR level, so this should be LM_E... copy-paste error? :)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1428,7 +1428,8 @@ static bool addTriggeredSubscriptions_noCache
   DBClientBase* connection = getMongoConnection();
   if (collectionQuery(connection, collection, query, &cursor, &errorString) != true)
   {
-    alarmMgr.dbError(errorString);
+    // FIXME P10: collectionQuery already calls dbMgr in the case of problem. Redundant alarmMsg.dbError() ?
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
```diff
@@ -74,7 +74,8 @@ static HttpStatusCode processDiscoverContextAvailability
                           &count))
   {
     responseP->errorCode.fill(SccReceiverInternalError, err);
-    alarmMgr.dbError(responseP->errorCode.details);
+    // FIXME P10: overloggin (registrationQuery() false return menas that collectionRangedQuery() retursn false, corresponding to internal alarms in that function)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -356,7 +356,8 @@ HttpStatusCode mongoQueryContext
       else
       {
         /* Different from errors in DB at entitiesQuery(), DB fails at registrationsQuery() are not considered "critical" */
-        alarmMgr.dbError(err);
+        // FIXME P10: overloggin (registrationQuery() false return menas that collectionRangedQuery() retursn false, corresponding to internal alarms in that function)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -375,7 +376,8 @@ HttpStatusCode mongoQueryContext
...
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -394,7 +396,8 @@ HttpStatusCode mongoQueryContext
...
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoQueryContext.cpp
```diff
@@ -414,7 +417,8 @@ HttpStatusCode mongoQueryContext
       else
       {
         /* Different from fails in DB at entitiesQuery(), DB fails at registrationsQuery() are not considered "critical" */
-        alarmMgr.dbError(err);
+        // FIXME P10: overloggin (registrationQuery() false return menas that collectionRangedQuery() retursn false, corresponding to internal alarms in that function)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -409,7 +409,8 @@ void mongoSubCacheRefresh(const std::string& database)
   DBClientBase* connection = getMongoConnection();
   if (collectionQuery(connection, collection, query, &cursor, &errorString) != true)
   {
-    alarmMgr.dbError(errorString);
+    // FIXME P10: overloggin (collectionQuery() false involves to internal alarms in that function)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/mongoUnsubscribeContextAvailability.cpp
```diff
@@ -84,7 +84,8 @@ HttpStatusCode mongoUnsubscribeContextAvailability
   {
     reqSemGive(__FUNCTION__, "ngsi9 unsubscribe request (mongo db exception)", reqSemTaken);
     responseP->statusCode.fill(SccReceiverInternalError, err);
-    alarmMgr.dbError(responseP->statusCode.details);
+    // FIXME P10: overloggin (collectionQuery() false involves to internal alarms in that function)
```
fgalan: @kzangeli what do you think?

src/lib/mongoBackend/safeMongo.cpp
```diff
@@ -234,8 +234,9 @@ bool moreSafe(const std::auto_ptr<DBClientCursor>& cursor)
 * nextSafeOrError -
 *
 */
-bool nextSafeOrError(const std::auto_ptr<DBClientCursor>& cursor, BSONObj* r, std::string* err)
+bool nextSafeOrError(const std::auto_ptr<DBClientCursor>& cursor, BSONObj* r, std::string* err, const std::string& caller, int line)
 {
+  // FIXME P4: use line in the trace (just "+ line" doesn't work ;)
```
fgalan: Not so important for this PR, but if somebody have suggestions, please, go ahead with the code.

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
kzangeli: If the error is already logged (and **always** logged in case it occurs), then this alarm is not necessary.
The question is always, which one should be removed, low level or high level?

I guess the best info is given at lower level in this case, so

'LGTM'

src/lib/alarmMgr/AlarmManager.cpp
kzangeli: Yes, of course. This was probably copy-pasted from Bad Input.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Detailed 'explanation' only in one of these redundant logging cases.

NTC, except removing these temporary lines

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
fgalan: In this case, I think that low level. Along the PR you would see that doing so (i..e removing at high level) makes code simpler.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1485,25 +1486,25 @@ static bool addTriggeredSubscriptions_noCache
 
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
-      long long throttling       = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLong(sub, CSUB_THROTTLING)       : -1;
-      long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLong(sub, CSUB_LASTNOTIFICATION) : -1;
+      long long throttling       = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLongF(sub, CSUB_THROTTLING)       : -1;
```
kzangeli: Where does these 'F'-functions come from?
And, especially, what is the difference between the pairs?

fgalan: F-functions are actually macros that include __FUNCTION__ and __LINE__. Detail in safeMongo.h.

fgalan: By the way, naming by sufixing with 'F' leads to awful names :). If you have a better proposal for these macros names, please go ahead.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2239,14 +2239,14 @@ bool entitiesQuery
...
-      LM_E(("Runtime Error (exception in nextSafe(): %s)", e.what()));
+      LM_E(("Runtime Error (exception in nextSafe() at %s:%d: <%s> - query: <%s>)", __FUNCTION__, __LINE__, e.what(), query.toString().c_str()));
       releaseMongoConnection(connection);
       return false;
     }
     catch (...)
     {
       *err = "generic exception at nextSafe()";
-      LM_E(("Runtime Error (generic exception in nextSafe())"));
+      LM_E(("Runtime Error (generic exception in nextSafe() at %s:%d - query: <%s>)", __FUNCTION__, __LINE__, query.toString().c_str()));
```
kzangeli: LM_E already supplies __LINE__ and __FUNCTION__
I guess you have some reason to have them in two places ...

src/lib/mongoBackend/mongoDiscoverContextAvailability.cpp
kzangeli: NTC, except removing these two lines

src/lib/mongoBackend/safeMongo.h
```diff
@@ -32,47 +32,57 @@
...
+#define getBoolFieldF(b, field)             getBoolField(b, field, __FUNCTION__, __LINE__)
+#define getFieldF(b, field)                 getField(b, field, __FUNCTION__,  __LINE__)
+#define nextSafeOrErrorF(c, r, err)         nextSafeOrError(c, r, err, __FUNCTION__,  __LINE__)
+
 /* ****************************************************************************
 *
 * getObjectField -
 */
-extern mongo::BSONObj getObjectField(const mongo::BSONObj& b, const std::string& field);
+extern mongo::BSONObj getObjectField(const mongo::BSONObj& b, const std::string& field, const std::string& caller = "<none>", int line = 0);
```
fgalan: I don't like too much the default part for the parameters but I needed bo about changing thousand lines in the unit tests.

Maybe a "safer" alterantive could be:

```
#if UNIT_TEST
 extern mongo::BSONObj getObjectField(const mongo::BSONObj& b, const std::string& field);
 extern mongo::BSONObj getStringField(const mongo::BSONObj& b, const std::string& field);
  ...
#else
 extern mongo::BSONObj getObjectField(const mongo::BSONObj& b, const std::string& field, const::string& function, int line);
 extern mongo::BSONObj getStringField(const mongo::BSONObj& b, const std::string& field, const::string& function, int line);
  ...#endif
```

What do you think?

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: NTC, except removing these two lines

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: NTC, except removing these two lines

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: NTC, except removing these two lines

src/lib/mongoBackend/mongoQueryContext.cpp
kzangeli: NTC, except removing these two lines

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: LM_E suplies the line and function of the point of the code in which LM_E appears, i.e. within get*Field(). But I need to know the function and line corresponding to the caller function.

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -236,14 +236,38 @@ HttpStatusCode mongoEntityTypes
...
-    EntityType*               entityType  = new EntityType(getStringField(resultItem, "_id"));
-    std::vector<BSONElement>  attrsArray  = getField(resultItem, "attrs").Array();
+    BSONObj                   resultItem  = resultsArray[ix].embeddedObject();    
+    std::vector<BSONElement>  attrsArray  = getFieldF(resultItem, "attrs").Array();
+
+
+    EntityType*               entityType;
 
-    entityType->count = countEntities(tenant, servicePathV, entityType->type);
+    bool nullId = ((resultItem.hasField("")) && (getFieldF(resultItem, "").isNull())) || (getStringFieldF(resultItem, "_id") == "");
```
kzangeli: Difficult line ...
resultItem.hasField("") I imagine refers to that resultItem has a field with empty name?
Perhaps a comment line or two ...

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -236,14 +236,38 @@ HttpStatusCode mongoEntityTypes
...
+      entityType           = emptyEntityType;
+      emptyEntityTypeUsed  = true;
+    }
+    else
+    {
+      entityType = new EntityType(getStringFieldF(resultItem, "_id"));
+    }
+
+    /* Note we use += due to emptyEntityType accumulates */
+    entityType->count += countEntities(tenant, servicePathV, entityType->type);
```
kzangeli: The += I understand, but ... where is emptyEntityType->count set?

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Yes, a comment is neded. I will include it.

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Mmm. Good point. I guess that by construction count is 0, but I won't rely on this and fix this to be sure that is 0.

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -264,7 +288,20 @@ HttpStatusCode mongoEntityTypes
...
+    if (!nullId)
+    {      
+      responseP->entityTypeVector.push_back(entityType);
+    }
+  }
+
+  if (emptyEntityTypeUsed)
+  {
+    responseP->entityTypeVector.push_back(emptyEntityType);
+  }
```
kzangeli: ok, so this is to add a maximum of ONE 'emptyEntityType' to responseP->entityTypeVector, right?

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: same same,

NTC - remove the two comments

src/lib/mongoBackend/mongoUnsubscribeContextAvailability.cpp
kzangeli: idem

src/lib/mongoBackend/safeMongo.cpp
kzangeli: If you want it included, must convert to a string:

    char lineV[32];  // we have a macro for int-as-string-size
    snprintf(lineV, sizeof(lineV), "%d", line);

OR
those ostream constructions that I don't at all like ... :-)

src/lib/mongoBackend/safeMongo.cpp
```diff
@@ -243,12 +244,12 @@ bool nextSafeOrError(const std::auto_ptr<DBClientCursor>& cursor, BSONObj* r, st
...
   catch (const std::exception &e)
   {
-    *err = e.what();
+    *err = std::string(e.what()) + " at " + caller;
     return false;
   }
   catch (...)
   {
-    *err = "generic exception";
+    *err = "generic exception at " + caller;
```
kzangeli: I still can't understand how this works ...

Perhoaps like this:

1) *err = "generic exception at "  => result is a std::string
2) append 'caller' to the string

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Both results for entities without type (`null` in the results array) and type "" ("" in the results array) are aggregated into a single entry "" in the response to get types operation. Thus, I cannot push_back inside the for loop (as with a regular type) because that could lead to duplicated entries for "" (probably that would generate ilegal JSON). Thus, pushing back y deferred after the for loop.

kzangeli: As I thought.
You could do it inside also, with a `if (!emptyEntityTypeUsedPushed) push()`, I imagine.
However, not interesting.

NTC

src/lib/mongoBackend/mongoQueryTypes.cpp
kzangeli: But not always 0, right?
If empty type is found it should be set to 1, I imagine ...
If not, the += doesn't make much sense

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Nah, it's OK.
I just couldn't see them in the PR. Guess they already existed.
Too long names is bad too. 'xxxF' might be a little ugly but it is very clear.

NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I still don't understand. You should get the exact same func and line in two places.
Are you sure this does exactly what you want?
`__LINE__` will be 2249  and `__FUNCTION__` whatever this very function is called.
And the exact same info you already have in the same log line ...

I'm just trying to see whether we have some misunderstanding here.

src/lib/mongoBackend/safeMongo.h
kzangeli: Nah, why bother.
Leave it the way it is.

NTC.

Only 'interesting' thing to fix this very slight problem would be creating a macro with the same name as the function. A macro accepting two params and then calling the function with four. Something like:

    #define getObjectField(a, b)   getObjectField(a, b, __FUNCTION__, __LINE)

However, this is pretty cryptic code.
Better too keep it with default params.

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: This count is for the number of entities in a given type. However, note that no-type and type "" "share" the same type, so we need a sum.

For example, if we have in a given moment 5 entities without type and 10 entities with type "", the entry corresponding to no-type + type "" will count for 15.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: Mmmm... you are right. In the of LM_E the information is already there. I'm afraid that I extend the same solution that I used in get*Fiel() at safeMongo module to a place where it is not actually needed.

I'll fix all them.

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: After a second look I think is safe to go without explicit initalization. Looking to the constructors, 0 is set for count:

```
EntityType::EntityType(): count(0)
{

}



*
* EntityType::EntityType -
*/
EntityType::EntityType(std::string _type): type(_type), count(0)
{

}
```

fgalan: NTC

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Yes, this is exactly what I assumed :-)

src/lib/mongoBackend/mongoQueryTypes.cpp
kzangeli: So, somehow/somewhere emptyEntityType->count is set to 1, right?

kzangeli: Clarified over skype

NTC

src/lib/alarmMgr/AlarmManager.cpp
fgalan: Fixed in 843b4e8

src/lib/mongoBackend/safeMongo.cpp
fgalan: Fixed in 843b4e8

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Fixed in 843b4e8

kzangeli:
LGTM

PR #1928, opened 2016-03-21
Fix CMake to be more generic wrt Ubuntu builds

jmcanterafonseca:


src/app/contextBroker/CMakeLists.txt
```diff
@@ -75,14 +75,8 @@ link_directories("/opt/lib/")
...
-  MESSAGE("contextBroker: Ubuntu DISTRO: '${DISTRO}'")
-  TARGET_LINK_LIBRARIES(contextBroker ${STATIC_LIBS} -lmongoclient -lboost_thread-mt -lboost_filesystem-mt -lboost_system-mt -lboost_regex-mt -lpthread -lssl -lcrypto -lgnutls -lgcrypt)
-ELSEIF((${DISTRO} STREQUAL "Ubuntu_13.10") OR (${DISTRO} STREQUAL "Ubuntu_14.04.1_LTS") OR (${DISTRO} STREQUAL "Ubuntu_14.10"))
-  MESSAGE("contextBroker: Ubuntu DISTRO: '${DISTRO}'")
-  TARGET_LINK_LIBRARIES(contextBroker ${STATIC_LIBS} -lmongoclient -lboost_thread -lboost_filesystem -lboost_system -lpthread -lssl -lcrypto -lgnutls -lgcrypt)
-ELSEIF((${DISTRO} STREQUAL "Ubuntu_14.04.2_LTS") OR (${DISTRO} STREQUAL "Ubuntu_14.04.3_LTS") OR (${DISTRO} MATCHES "Linux_Mint_.*"))
-  MESSAGE("contextBroker: Ubuntu DISTRO: '${DISTRO}'")
+IF ((${DISTRO} MATCHES "Ubuntu_.*") OR (${DISTRO} MATCHES "Linux_Mint_.*"))
+  MESSAGE("contextBroker: DISTRO: '${DISTRO}'")
   TARGET_LINK_LIBRARIES(contextBroker ${STATIC_LIBS} -lmongoclient -lboost_thread -lboost_filesystem -lboost_system -lboost_regex -lpthread -lssl -lcrypto -lgnutls -lgcrypt)
```
kzangeli: There might be an unwanted change here, treating all Ubuntu versions the same ...
The older ubuntu versions use the -mt versions of the boost libs.

With this change they will use 'non -mt version'.

Should be tests, however, perhaps those old ubuntus are so old so we don't really care ...

jmcanterafonseca:
can we fix it by treating separately Ubuntu 12, 13, and 14 but not dealing with particular sub-versions?

fgalan:
However, it seems that within the same major version (e.g. 14) different minor versions use different configuration (e.g. in the current CMakeList.txt 14.10 and 14.02 are in different branches of the "if" clause)... Maybe the current CMakeList.txt is buggy?

If we want to be "strict" with this, the final CMakeList.txt should be checked for at least all Ubuntu versions included in the current CMakeList.txt. Doing it manually would be an overhelming task, but maybe docker could ease it a lot.

fgalan:
This PR should be rebased after merging PR https://github.com/telefonicaid/fiware-orion/pull/2062 into develop.

jmcanterafonseca:
With latest cmake compiles perfectly in my Ubuntu box

fgalan:
> With latest cmake compiles perfectly in my Ubuntu box 

In that case, I'd suggest to close this PR.

jmcanterafonseca:
closed as per latest comments

PR #1932, opened 2016-03-21
FIX GIT_REV_ORION for release/0.26.1

### LINKED ISSUE #1922: [docker] Dockerfile always points to 'Develop' branch
I've found that, Orion's Dockerfile points to `Develop` in every release.This means that, all the tags available in Dockerhub build the image from `Develop` at the moment it's triggered. This can cause 2 issues:

- The first one is, if the image is somehow 'triggered' after the release, it will always build an image from `Develop` branch
- There's no way to know that the image built is exactly the one pointing to the right `tag`. The only way to make sure of that is building the image exactly right after the release (but the first issue still remains there).

```dockerfile
FROM centos:centos6.6

MAINTAINER FIWARE Orion Context Broker Team. Telefónica I+D

ENV ORION_USER orion
ENV GIT_REV_ORION develop
ENV CLEAN_DEV_TOOLS 1
...
```

fgalan:
Fix #1922 in release/0.26.1 branch

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,7 +3,7 @@ FROM centos:centos6.6
 MAINTAINER FIWARE Orion Context Broker Team. Telefónica I+D
 
 ENV ORION_USER orion
-ENV GIT_REV_ORION develop
+ENV GIT_REV_ORION release/0.26.1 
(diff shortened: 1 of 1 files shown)


kzangeli:
LGTM

PR #1934, opened 2016-03-21
Hardening/functest failing

kzangeli:
typeGiven was not initialized in one of the constructors of Metadata.

I wonder if perhaps a better implementation would be like this:

    if (mdB.hasField(ENT_ATTRS_MD_TYPE)
    {
      type      = getStringFieldF(mdB, ENT_ATTRS_MD_TYPE);
      typeGiven = true:
    }
    else
    {
      type      = "";
      typeGiven = false;
    }

The 'else' part could be initialized in the initialization list of the constructor.

This change would cover the case   "type": ""    (type given but empty)
Opinions?


fgalan:

crbrox:
LGTM

PR #1935, opened 2016-03-21
FIX overloging false Runtime Error

fgalan:
This fixes loging the following ERROR traces (which is not actually an ERROR):

```
Runtime Error (field '_id' was supposed to be a string but type=10 in BSONObj
```

**DO NOT DELETE BRANCH AFTER MERGE**. We would need it to merge this also in develop


crbrox:
LGTM

PR #1951, opened 2016-03-24
Fix problems after update branch

fortizc:
Fix some problems found after update  feature/1181_websockets from develop


src/lib/orion_websocket/ws.cpp
```diff
@@ -176,6 +182,21 @@ void *runWS(void *ptr)
     pthread_mutex_lock(&mtx);
     lws_service(ws->ctx, 50);
     pthread_mutex_unlock(&mtx);
+/*
```
fgalan: Leftover comment to be removed

fortizc: Oh! Sorry my mistake. This file should be out of this commit. It was fixed as soon as I can put my hands in the laptop.

fgalan: (Pls, remember the "Fixed in X" message. Otherwise, I don't have any mean to know that this has been solved, apart from the costly process or reviewing the whole PR again -easy in this case, as the PR is short, but not in a general case, as the PR could include hundred of lines changed. Thanks!)

src/lib/rest/ConnectionInfo.cpp
```diff
@@ -211,7 +210,28 @@ ConnectionInfo::~ConnectionInfo()
   servicePathV.clear();
 }
 
-void ConnectionInfo::modify(const std::string &_url, const std::string &_verb, const std::string &_payload, const HttpHeaders &head)
+void ConnectionInfo::reset()
```
fgalan: I understand that both modify() and reset() methods has been included in the ConnectionInfo for the feature/1181_websockets and are used from the ws library. It would be great if you could provide a function preamble for each of them, in order to provide some documentation about the purpose of the method, how it is used, etc. I doesn't need to be too much detailed, but explain at least a little about the method.

E.g (in this same file):

```
`/* ****************************************************************************
*
* uriParamTypesParse - parse the URI param 'type' into uriParamTypes vector
*/
```

src/lib/rest/ConnectionInfo.cpp
```diff
@@ -211,7 +210,28 @@ ConnectionInfo::~ConnectionInfo()
...
+  servicePath = "/";
+
+  httpHeaders.gotHeaders = true;
+  httpHeaders.userAgent = "orionWS/0.1";
+  httpHeaders.accept = "*/*";
+  httpHeaders.contentLength = 0;
+  httpHeaders.servicePath = "/";
+  httpHeaders.tenant.clear();
+
+  uriParam["details"] = "off";
```
fgalan: I think we have constant for "details", "limit", etc. in `#define` in some .h file. Not sure, but it is worth to have a look and, in that case, use them.

src/lib/rest/ConnectionInfo.cpp
fortizc: Fixed in d1db7baa4198ef01a24c34bfd55bae696753352e

src/lib/rest/ConnectionInfo.cpp
fortizc: I found constants for all in src/lib/rest/uriParamNames.h except for notifyFormat

fgalan: notifyFormat is going to be deprecated in Orion 1.0.0 (to be released tomorrow, if everything went ok with soak test).

So, I'd suggest to remove `uriParam["notifyFormat"] = "JSON";` line.

fortizc: Fixed in 9f32e2061a521738bf6645a484f0fae2dff90cdd

fgalan:
LGTM

Issue #1669, opened 2016-01-14
Websocket module: notifications through websocket


fgalan:
(Parent issue: #1181)

We could consider a new kind of subscriptions that instead of sending notifications through HTTP REST use the websocket used to create the subscription in order to send notifications back. A new URL schema (e.g. ws://) could be used so Orion can clearly identify them.

Probably this kind of subscriptions doesn’t need to be persisted in DB (it would live only in CB memory) given that once the websocket is closed by the client, then the subscription is immediately removed.

[ See issue #1943 for details about notifications in Active-Active environment ]

fgalan:
After today discussions on skype, more info on how implement this. 

Let's consider that ws module keeps a table relating subIds (as string) and ws_hdls (websocket handles, in the structure used by the module, @fortizc I understand that your control that part). Table implementation details are out of the scope of this description (although probably a `std::map` could be used). Let's name this tables subsWs for convenience in this description (please, use a better name at implementation stage if you find so).

At CB startup time, the table starts empty. Let's consider the different flows:

* Subscription done through websocket. The ws module analyzes if the subscription is of ws nature looking to the reference for the `ws://` string. In that case, it processes the message as usual but looks the subscription response before sending it back to the client to get the subId associated to the subscription. A new entry is added to subsWs linking the subscriptionId to the ws handler (which is kept opened).

* Update context triggering the subscription. The CB processes this update as usual (no matter if it arrived on the REST API or in the WS API). In some point of mongoBackend the notification is triggered and Notifier::sendNotifyContextRequest() is invoked. This function will be modified to look at the reference string:
 * If reference starts with `http://` or `https://` then perform regular process.
 * If reference starts with `ws://` then a callback to ws module is done, including the subId and the notification content. Let's name this function sendNotifyContextRequestWs() by the moment (please, use a better name at implementation stage if you find so). This functions looks for subId in the subsWs table to get the corresponding ws_hdls. The ws module sends the notification using ws_hdl. More detail on sendNotifyContextRequestWs() signature in a separate comment.

* If the client closes the ws_hdl, the the ws module does a reverve lookup in the table to get the associated subId (if any) and in that case does a unsubscribe for subId on the lower layer and remove the entry from the table.

fgalan:
Draft for sendNotifyContextRequestWs() signature:

```
bool sendNotifyContextRequestWs(std::string subId, std::map<string, string> headers, std::string data);
```

[...]

PR #1955, opened 2016-03-28
ws:// processing at Notifier library

fgalan:
Part of #1669 

Unfortunatelly, make ft is broken in feature/1181_websocket due to current WS implementation in that branch, I cannot assure this modification doesn't break anything...

@fortizc

### DIFF
--- a/src/lib/ngsiNotify/Notifier.cpp
+++ b/src/lib/ngsiNotify/Notifier.cpp
@@ -36,14 +36,9 @@
 #include "ngsiNotify/Notifier.h"
 
 
-
-/* ****************************************************************************
-*
-* Select 'method' to send notifications - only one can be uncommented
-*/
-//#define SEND_BLOCKING
-#define SEND_IN_NEW_THREAD
-
+// FIME P11 #1669: move this to WS library. @fortizc please take care of this
+#define WS_SCHEME         "ws://"
+#define WS_SCHEME_LENGTH  5
 
 
 /* ****************************************************************************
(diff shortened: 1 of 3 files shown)


src/lib/ngsiNotify/senderThread.cpp
```diff
@@ -30,6 +30,18 @@
 #include "ngsiNotify/senderThread.h"
 
 
+// FIXME P11 #1669: remove this stub when the actual sendNotifyContextRequestWs() gets developed a the end. @fortizc please take care of this
+// in the ws library
+int sendNotifyContextRequestWs(const std::string& subId, const std::map<std::string, std::string>& headers, const std::string& data)
```
fgalan: At the begining I proposed (at #1669) to use `bool` as return value for this function. However, in order to align with httpRequestSent() signature used at startSenderThread() function, I think it should be better to use `int` the the usual semantics: 0 -> everythin ok, not 0 -> error.

(I have edited signature at https://github.com/telefonicaid/fiware-orion/issues/1669#issuecomment-199907859 has been changed `void` -> `int`)

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -104,53 +99,26 @@ void Notifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::st
...
-    if (!parseUrl(url, host, port, uriPath, protocol))
-    {
-      LM_E(("Runtime Error (not sending NotifyContextRequest: malformed URL: '%s')", url.c_str()));
-      return;
-    }
-
-    /* Set Content-Type */
-    std::string content_type = "application/json";
-
-#ifdef SEND_BLOCKING
```
fgalan: SEND_BLOCKING has not been in use sinde months (or years?). In addition, keeping this dead code aligned with the changes implemented in this PR for WS support will complicate things, so let's definetively remove it.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -36,14 +36,9 @@
...
-/* ****************************************************************************
-*
-* Select 'method' to send notifications - only one can be uncommented
-*/
-//#define SEND_BLOCKING
-#define SEND_IN_NEW_THREAD
-
+// FIME P11 #1669: move this to WS library. @fortizc please take care of this
+#define WS_SCHEME         "ws://"
+#define WS_SCHEME_LENGTH  5
```
fortizc: This is only a suggest, but may be is better use a static const std::string and to get its length you can use size() method, moreover when you assign a char * to std::string internally make something like that: std::string(my_char_pointer) in this way is better too a const std::string

fgalan: Ok. However, after skype talk, this is going to be done by @fortizc in a next PR, so NTC with regards to this PR.

src/lib/ngsiNotify/Notifier.cpp

src/lib/ngsiNotify/senderThread.cpp
fgalan: NTC

fortizc:
LGTM !

PR #1959, opened 2016-03-28
UPDATE NGSIv2 cookbook - geo query

jmcanterafonseca:


doc/apiary/v2/fiware-ngsiv2-cookbook.apib
```diff
@@ -397,16 +397,18 @@ This resource allows you to obtain a list of subresources published by the API:
...
 + Parameters
     + q: temperature==22.5;humidity==60 (optional, string) - General query parameter that allows
     to filter by attribute content
-    + coords: 41.3763726, 2.1864475 (optional, string) - Coordinates. Comma separated values.
-    The number of values will depend on the 'geometry' attribute.
-    + geometry: circle;radius=10000 (optional, string) - Geospatial area for the query. 
-    The 'exterior' modifier allows to indicate that the referred area is the outside
-    of the region, for instance, 'polygon;exterior'
+    + coords: 41.3763726, 2.1864475 (optional, string) - Coordinates of a reference geometry used for the query.
+    Comma separated values. The number of values and its interpreation depends on the `geometry` attribute.
```
fgalan: nit: interpreation -> interpretation

In addition, note that we are using `;` to separate each coordinate pairs (e.g. https://github.com/telefonicaid/fiware-orion/blob/develop/test/functionalTest/cases/1038_geo_point_support/geoquery_polygon_geopoint.test#L115). Maybe that should be also clarified here?

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
```diff
@@ -397,16 +397,18 @@ This resource allows you to obtain a list of subresources published by the API:
 
 ## Query Entities [/v2/entities{?q,limit,offset,coords,geometry,attrs,options}]
 
-### Value filter and Geoquery [GET /v2/entities{?q,limit,offset,coords,geometry,attrs,options}]
+### Value filter and Geoquery [GET /v2/entities{?q,limit,offset,coords,georel,geometry,attrs,options}]
```
fgalan: Note L427 below:

```
### No results [GET /v2/entities{?q,limit,coords,geometry,attrs}]
```

To be coherent, `georel` should be added also there.

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
```diff
@@ -397,16 +397,18 @@ This resource allows you to obtain a list of subresources published by the API:
...
+    + coords: 41.3763726, 2.1864475 (optional, string) - Coordinates of a reference geometry used for the query.
+    Comma separated values. The number of values and its interpreation depends on the `geometry` attribute.
+    + geometry: point (optional, string) - Geometry (point, line, polygon) represented by the `coords` parameter. 
+    + georel: near;maxDistance:1000 (optional, string) - Geospatial relationship that must be met between
+    the reference geometry (specified by the `coords` and `geometry` parameters) and the matching entities.
+    This example refers to those entities located at a distance of 1000 meters at a maximum
+    (from the point specified as reference geometry).  
     + limit: 20 - Limit the number of results (optional, number) - Maximum number of results to return
     + attrs: seatsNumber (optional, string) - Attributes to be returned
     + options: count,keyValues (optional, string) - query options (count total, keyValues mode)
```
fgalan: You are asking for `count`. However, is `X-Total-Count` being rendered in the Apiary document?

(Not sure about this... looking only to .apib is difficult to say, but if we want to be fully consistent it should be added or, alternatively, the `count` flag should be removed from `options`).

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
jmcanterafonseca: I was not aware of the ';' is that specified? I'm pretty sure it wasn't when we wrote the spec. Why is it needed?, to my mind it is redundant ... and it only makes it more difficult to generate the location string

jmcanterafonseca: fixed https://github.com/telefonicaid/fiware-orion/pull/1959/commits/8194b68e37784e7029be14690716a4fb2c47204c 

(not the ';' issue as per the comment above)

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
jmcanterafonseca: fixed https://github.com/telefonicaid/fiware-orion/pull/1959/commits/f3fd76c3817fd1a5b8f5d1059e37ebd4be19f371 

the example has been modified appropriately

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
jmcanterafonseca: you are right. dropped at https://github.com/telefonicaid/fiware-orion/pull/1959/commits/f3fd76c3817fd1a5b8f5d1059e37ebd4be19f371 

and considering adding a new example for count in an upcoming PR

doc/apiary/v2/fiware-ngsiv2-cookbook.apib
fgalan: See https://github.com/telefonicaid/fiware-orion/blob/develop/doc/apiary/v2/fiware-ngsiv2-reference.apib#L775

(Not sure when this was introduced, but I guess that using "Blame" buttom and doing some git-archeology based on commit hash you could get the particular branch/PR).

I don't think is redundant. On the contrary, it helps to introduce some structure in the coords field, which makes them easier to read. For example comparing

```
coord=12,11,13,4,5,1,16,17,7,15,10,8,19,20,9,6,18,3,14,2
```

with

```
coord=12,11;13,4;5,1;16,17;7,15;10,8;19,20;9,6;18,3;14,2
```

The former is more difficult to read and debug. In fact, it is difficult to see looking to the first one that you have an even number of numbers.

From a generation difficulty point of view, I think both approaches are equal, i.e. a programmatic client wouldn't have any problem to generate coords in one format or in the other.

jmcanterafonseca: well, GeoRSS which actually inspired our spec does not add any separator

http://www.georss.org/simple.html

on the other hand, the semicolon has been used for separating query components 

I think this discussion deserves an issue

fgalan: For future reference, this is the issue: https://github.com/telefonicaid/fiware-orion/issues/1969

fgalan:
LGTM

Good to see progress on the cookbook document! :) While specification document has evolved a lot, not the same for cookbook (practically untouched since November 2015, see https://github.com/telefonicaid/fiware-orion/commits/develop/doc/apiary/v2/fiware-ngsiv2-cookbook.apib).

Issue #1920, opened 2016-03-17
GeoJSON support


fgalan:
As described in NSGIv2 specification: http://telefonicaid.github.io/fiware-orion/api/v2/

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1982

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/1982. Moved to QATestPending while the acceptance test to cover this functionality gets implemented.

iariasleon:
LGTM
issue associated: #2426

PR #1982, opened 2016-04-01
Complete geo:xxxx support and add GeoJSON support

fgalan:
Completes #1038 and implements #1920 

The location logic at mongoBackend has been moved to an independent module (location.cpp/h). In addition, some parts has been refactored (e.g. the valueBson() logic into the ContextAttribute class, and a errorAdapation.cpp/h to hide the ugly stuff needed for ngsiv1 -> ngsiv2 conversion at serviceRoutines). I hope that these changes hope to make the code more clear.

_Some hints to review the PR:_

Apart from the usual review of the code changes, don't be afraid about the large numer of added .test under cases 1038 and 1920 :) as many of them are variations of others. In particular:

* All geoquery_box_*.test follow the same pattern, but with different geo:xxxx shapes and in some cases (the ones with after_update in the name) adding an update entity step at the beginning. Please have a look to the .png included in this PR to get a graphical representation of the test.
* All errors_*.test follow the same pattern. The only different are the steps are the begigning, that check different error conditions depending on the shape.
* There are two entity_location_shapes_change files (one with_geojson) which is based on an entity changing it shape step after step and tested with 4 geo-queries after each changed. Please have a look to the .png to get a graphical representation of the test.

@kzangeli 
@crbrox


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -486,7 +485,7 @@ can be useful in urderstanding the format.
       "type": "Point",
       "coordinates": [2.186447514, 41.3763726]
     },
-    "type": "application/vnd.geo+json"
+    "type": "geo:json"
```
fgalan: FYI @jmcanterafonseca 

I have changed this due to two reasons:

* The main one is that `/` is a forbidden char in attribute type (see https://github.com/telefonicaid/fiware-orion/blob/develop/doc/apiary/v2/fiware-ngsiv2-reference.apib#L309)
* In addition, it align the GeoJSON support with the other entity location shapes in the Simple Location Format, as all them start with `geo:`. Thus, using `geo:json` for GeoJSON seems to be a smart solution.

src/lib/rest/errorAdaptation.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+
+/* ****************************************************************************
+*
+* errorStringForV2 -
+*
+* NGSIv2 uses "compacted" error descriptions, e.g. "BadRequest" instead of
+* "Bad Request". This function makes the conversion.
+*/
+std::string errorStringForV2(const std::string& _reasonPhrase)
```
fgalan: Originally this function was a member of OrionError class. However, it wasnt' actually used as a member function (i.e. it doesn't uses any of the field of the object) so I have moved to this errorAdaptation module.

test/functionalTest/cases/1038_ngsiv2_geo_location_entities/entity_location_shape_changes.test
```diff
@@ -0,0 +1,539 @@
...
+
+03. Entity changes to point
+===========================
+HTTP/1.1 204 No Content
+Content-Length: 0
+Date: REGEX(.*)
+
+
+
+04. B1: ok, B2: nok, B3: nok, B4: nok
```
fgalan: It would be great to have someting more "compact" for this step output, instead the HTTP response for each case. Something like:

```
04. B1: ok, B2: nok, B3: nok, B4: nok
=====================================
B1: ok
B2: nok
B3: nok
B4: nok
```

After some time (one hour or so) trying to get that with grep, awk and similar bash black magic, I gave up... If you want to improve, please go ahead :)

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fine with me. don't forget to change all the occurrences of the apib document

fgalan:
```
~/src/fiware-orion$ grep -R 'application/vnd.geo+json' doc/ | wc 
      0       0       0
```

Thus, it seems that no more occurrrences exists.

NTC

doc/manuals/admin/database_model.md
```diff
@@ -71,17 +71,49 @@ Fields:
...
+* Representing a polygon (the one used by NGSIv2 geo:box and NGSIv2 geo:polygon):
+
+```
+{
+  "type": "Polygon",
+  "coordinates": [ [ [ 10, 0], [0, 10], [0, 0], [10, 0] ] ]
+}
+```
+
+* Finally, `location.coords` could held an arbitrary JSON object, representing a location
```
kzangeli: typo: held -> hold

src/lib/mongoBackend/location.cpp
```diff
@@ -0,0 +1,467 @@
...
+    *errDetail = "geo:line, geo:box and geo:polygon needs array of strings as value";
+    return false;
+  }
+
+  for (unsigned int ix = 0; ix < caP->compoundValueP->childV.size() ; ix++)
+  {
+     CompoundValueNode* item = caP->compoundValueP->childV[ix];
+     if (!item->isString())
+     {
+       *errDetail = "geo:line, geo:box and geo:polygon needs array of strings but some element is not";
```
kzangeli: is not -> is not a string

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -41,8 +41,203 @@
...
+    else if (child->valueType == ValueTypeObject)
+    {
+      BSONObjBuilder bo;
+
+      compoundValueBson(child->childV, bo);
+      b.append(bo.obj());
+    }
+    else
+    {
+      LM_E(("Runtime Error (Unknown type in compound value)"));
```
kzangeli: Not sure if this error can ever occur, but if it does, it is silently ignored, except for a log-line ...
My question is, do we want something 'more' here ... ?

fgalan: It shouldn't, but I'd like to keep the LM_E in the else just in case some bug (e.g. some change in the if/else if chain) causes the program flow end in that part. Just loging a Runtime Error (so we can detect in running instances) would be enough.

NTC

test/functionalTest/cases/1038_ngsiv2_geo_location_entities/entity_location_shape_changes.test
fgalan: NTC (I guess)

doc/manuals/admin/database_model.md
fgalan: Fixed in 8a3b4dbf43c3bc7ec253d3321f264d80655d7d0f

src/lib/mongoBackend/location.cpp
fgalan: Fixed in 8a3b4dbf43c3bc7ec253d3321f264d80655d7d0f

test/functionalTest/cases/1038_ngsiv2_geo_location_entities/entity_location_shape_changes.test
kzangeli: Yes, too much work to make it 'perfect'

src/lib/ngsi/ContextAttribute.cpp
kzangeli: Sure,

NTC

kzangeli:
LGTM

crbrox:
LGTM

Issue #1918, opened 2016-03-17
Add "description" field to subscriptions


fgalan:
Subscriptions should have an optional description field so clients can use it to annotate things about subscription (e.g. subscription purpose). This description could be added at subscription creation time, modified afterwards (updating it with "" will delete the description).

Details to be clarified at NGSIv2 .apib (the first PR for this issue).

fgalan:
.apib changes in PR #1926

fgalan:
Max length: 1024 characters (not in the .apib, as it is an implementation-specific aspect).

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/2001#event-615592163. Moving to QATestPending while the acceptance test to cover this functionality get implemented.

anabelengp:
related issue: #2042

iariasleon:
LGTM
implemented in https://github.com/telefonicaid/fiware-orion/blob/develop/test/acceptance/behave/components/ngsiv2/subscriptions/create_a_new_subscription/create_a_new_subscription_description.feature

PR #2001, opened 2016-04-05
ADD description field for NGSIv2 subscriptions

fgalan:
Implements #1918 

@crbrox 
@kzangeli


crbrox:
LGTM

doc/manuals/admin/database_model.md
```diff
@@ -252,6 +252,7 @@ Fields:
     the subscription.
 -   **format**: the format to use to send notification, either "XML" or "JSON".
     However, note that XML has been deprecated in Orion 0.23.0 and that this field eventually will be removed.
+-   **description** (optional field): a free string describing the subscription. Maximum length is 1024.
```
kzangeli: free string -> free text   or free text string

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -100,6 +100,25 @@ std::string parseSubscription(ConnectionInfo* ciP, ParseData* parseDataP, JsonDe
...
   }
 
+  // Description field
+  destination->descriptionProvided = false;
+  if (document.HasMember("description"))
+  {
+    const Value& description      = document["description"];
+    std::string descriptionString = description.GetString();
+
+    if(descriptionString.length() > MAX_DESCRIPTION_LENGTH)
```
kzangeli: if (

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -100,6 +100,25 @@ std::string parseSubscription(ConnectionInfo* ciP, ParseData* parseDataP, JsonDe
     return oe.render(ciP, "");
   }
 
+  // Description field
+  destination->descriptionProvided = false;
+  if (document.HasMember("description"))
+  {
+    const Value& description      = document["description"];
+    std::string descriptionString = description.GetString();
```
kzangeli: description doesn't seem to be used after this.
These two lines could be one (and not too long) line.

It's a matter of taste.

NTC

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -231,6 +231,24 @@ HttpStatusCode mongoUpdateContextSubscription
...
 
+  /* Description */
+  if (requestP->descriptionProvided)
+  {
+    // Note that in the case of description "" the field is deleted
+    if (requestP->description != "")
+    {
+      newSub.append(CSUB_DESCRIPTION, requestP->description);
+    }
+  }
```
kzangeli: I might be wrong, but it seems like the 'descriptionProvided' isn't changing anything at all.
Only reason to have it would be to allow an empty description, but that is not the case ...

It seems to be you could simply compare requestP->description with "" and forget about descriptionProvided ... ?

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: Check it is a string before calling GetString or "something bad" (TM) can happen
 (.IsString())

fgalan: Good point. I'll for string.

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
fgalan: First version used requestP->description comparation to "". However, the problem is that, from a parsing layer point of view, description equal "" correspond to two situations:

1. description field was included but with value ""
2. description field was not included

while the bevahiour is not the same, as it is respectively

1. delete the subscription field
2. don't touch description (i.e. keep the one in the original subscription, if any)

Please, have a look to this fragment after taking that into account, and let's see if now it makes sense...

doc/manuals/admin/database_model.md
fgalan: Fixed in 8ccac25

src/lib/jsonParseV2/parseSubscription.cpp
fgalan: Fixed in 8ccac25

src/lib/jsonParseV2/parseSubscription.cpp
fgalan: Fixed in 8ccac25

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Ah, ok, there are two different situations then.

NTC

kzangeli:
LGTM

Issue #1671, opened 2016-01-14
Websocket module: unhardwire configurable aspects


fgalan:
(Parent issue: #1181)

Some tokens in the module could be unhardwired so they could be set through CLI parameters. For example:

* Port
* Sleep time in the main loop
* Maximum message size

PR #2004, opened 2016-04-05
Add notifications through websockets

fortizc:
Implements https://github.com/telefonicaid/fiware-orion/issues/1669


src/lib/orion_websocket/constants.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+
+#include "constants.h"
+
+const int WSConstants::Port = 9010;
+const int WSConstants::Pooling = 50;
+const size_t WSConstants::DataSize = 128;
+// Ok, this is can be a macro, but I prefer
+// to be consistent
+const std::string WSConstants::ProtocolName = "orion-ws";
```
fgalan: Maybe we have talked about this in the past but I don't rembember... what is the purpose of this "orion-ws"?

src/lib/orion_websocket/constants.cpp
```diff
@@ -0,0 +1,35 @@
...
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Felipe Ortiz
+*/
+
+
+#include "constants.h"
+
+const int WSConstants::Port = 9010;
```
fgalan: At the end, did we make this configurable by CLI? Or that is part of another issue (not yet addressed)?

src/lib/orion_websocket/constants.cpp
fortizc: Is the protocol name used for the clients to connect with our service

src/lib/orion_websocket/constants.cpp
fortizc: MMmm.. maybe we must create a new issue for that

fgalan: Actually, it seems we have it. This is: https://github.com/telefonicaid/fiware-orion/issues/1671

src/lib/orion_websocket/constants.cpp
fgalan: While we decide the right name ( @jmcanterafonseca , any suggestion?) maybe we should something that clearly specifies that the name is temporal, e.g. `orion-ws-experimental`.

src/lib/common/string.cpp
```diff
@@ -242,6 +242,10 @@ bool parseUrl(const std::string& url, std::string& host, int& port, std::string&
     return false;
   }
 
+ if (url == "ws://")
```
kzangeli: WSConstants::Scheme instead of "ws://" ?

src/lib/orion_websocket/wsNotify.cpp
```diff
@@ -0,0 +1,51 @@
...
+  client[subId] = ws;
+}
+
+void removeSenders(const std::vector<std::string>& subIds)
+{
+  for (unsigned i = 0; i < subIds.size(); ++i)
+  {
+    client.erase(subIds[i]);
+  }
+}
```
kzangeli: C++ lesson (for me :-)):
Wouldn't client.clear() do exactly this?

src/lib/orion_websocket/wsNotify.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+int sendNotifyContextRequestWs
```
kzangeli: This function should perhaps return a bool, and it would be good to have the different conditions to return true/false (right now it returns 0/1) described in a function header

src/lib/orion_websocket/wsNotify.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+  unsigned char *buff = (unsigned char *) malloc(size);
+  unsigned char *p = &buff[LWS_SEND_BUFFER_PRE_PADDING];
+  sprintf((char *)p, "%s", msg);
+  int result = lws_write(it->second, p, strlen(msg), LWS_WRITE_TEXT);
+
+  free((char *)msg);
+  free(buff);
+
+  return result < 0;
```
kzangeli: What if a buffer with 50 enters, but only 25 bytes are written?
Does  **lws_write** take care of this?
Because if not, we'd have here a possible error without treatment ...

src/lib/orion_websocket/wsNotify.cpp
fortizc: No, because I need remove senders in the subIds vector, not all senders in the clients map

src/lib/orion_websocket/wsNotify.cpp
fortizc: addSender or removeSender? If you speak about addSender your are right. Let me add this change

src/lib/orion_websocket/wsNotify.cpp
fortizc: lws_write use a const char * created with std::strings in ws_parser library. Sorry I can see the problem. Can you explain me better?

src/lib/common/string.cpp
fortizc: Fixed in ea76cbb1ccda285b22ae9c2a79f1e60fb121eab6

src/lib/orion_websocket/wsNotify.cpp
kzangeli: Ah, ok, it's a partial removal, not a complete one.
Understood.

NTC

src/lib/orion_websocket/wsNotify.cpp
kzangeli: I was thinking about sendNotifyContextRequestWs

src/lib/orion_websocket/wsNotify.cpp
kzangeli: I can.

Are we guaranteed that lws_write always sends the entire buffer?

Look at the 'normal' write function.
It returns the number of bytes written. It can perfectly be called with a buffer of 1024 bytes and return an 'ok', saying that "I have sent 259 bytes". What you do then is to call *write* again, with the remaining 765 bytes, and so on, in a loop until the entire buffer has been sent.

So, my question is:  do we know for sure that lws_write sends the entire buffer?

src/lib/orion_websocket/ws.cpp
```diff
@@ -53,12 +55,41 @@ struct _orion_websocket
...
+
+    size_t pos = headValue[i].find("subscriptions");
+    if (pos == std::string::npos)
+      return false;
+
+    pos = headValue[i].find_last_of('/');
+    if (pos == std::string::npos)
+      return false;
+
+    char buff[25];
```
fgalan: I'd suggest to use (in globals/limits.h)

#define MAX_LENGTH_SUBID 24

and size the buff variable with MAX_LENGTH_SUBID + 1. Otherwise, the raw 24 and 25 may be a bit obscure.

(const is also ok if you prefer that to #deffine)

src/lib/orion_websocket/wsNotify.cpp
fortizc: Ok, at this moment return 0 or 1 because https://github.com/telefonicaid/fiware-orion/issues/1669#issuecomment-199907859

kzangeli: Sure, 0/1 or true/false is all the same.
However, an explanation on when 0/1 is returned is always good. Preferribly a comment in the function header

src/lib/orion_websocket/wsNotify.cpp
fortizc: Fixed 05fb052c8d58fe1f0e656d5fe3108b5b888c8669

src/lib/orion_websocket/constants.cpp
jmcanterafonseca: well, the sub-protocol, application level protocol, is an optional part in the Javascript API. 

Having said that, wouldn't it be 'ngsiv2-json' a proper name for the sub-protocol? Maybe in the future we implement 'ngsiv2-text' or 'ngsiv2-binary' ... 

By the way, there is a registry  of subprotocols 

https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name

if we say 'orion-xxx' it can be more difficult if others decide to implement this as we have done for Orion.

fgalan: ngsiv2-json +1

src/lib/orion_websocket/ws.cpp
fortizc: Fixed in 661796d5f3dc2c1c04532ddaba0316dc2d2acf58

src/lib/orion_websocket/wsNotify.cpp
fortizc: Fixed in ec0276246dcbb7fb3bbcc7b535acba72fb7e68be

src/lib/orion_websocket/constants.cpp
fortizc: Ok, changet it in c824a6ced8e63ee53dbd7ec2ae20bc5afab7b4a0

fgalan:

src/lib/orion_websocket/wsNotify.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  int written = 0;
+
+  while (written < msg_size)
+  {
+    written += lws_write(it->second, (p + written), strlen(msg), LWS_WRITE_TEXT);
+    if (written == -1)
+        return 1;
+  }
+
```
kzangeli: A few things about this loop ...
To detect -1 from lws_write you need a separate variable, the accumulating 'written' cannot be used:

    int bytes = lws_write(it->second, (p + written), strlen(msg), LWS_WRITE_TEXT);
    if (bytes == -1)
      return 1;
    written += bytes

AND, the third parameter to lws_write, the size of the buffer to be sent must adapt to the number of bytes already written, and strlen should not be calculated in each loop, so:

    int bytes = lws_write(it->second, (p + written),  msg_size - written, LWS_WRITE_TEXT);

Also (this is just my opinion), isn't **&p[written]** easier to understand than **p + written** ?

fortizc: Fixed in 51652e9822cbcb7e3db5dfdaba91e86c0edbb932
and about to use &p[written] or p + writter, I prefer the second option, because is really a pointer not an array.

kzangeli: As I said, just my opinion :-)

kzangeli:
LGTM

PR #2011, opened 2016-04-08
XML removal (part 8)

fgalan:
Part of #1862 

Regarding tests, maybe it is better to review the "final status" of cases/0000_ngsi10_compounds in order to check if it covers all cases than review .test changes themselves in this PR.

@kzangeli


kzangeli:
LGTM

PR #2016, opened 2016-04-08
FIX attributes not shown in get types operation when attributes use metadata ID

fgalan:
Bug recently introduced in a past PR.


kzangeli:
LGTM

PR #2020, opened 2016-04-11
FIX csub dup script: status and description fields out of csub signature

fgalan:
@iariasleon

### DIFF
--- a/scripts/managedb/csubs_control/csub_dups_checker.py
+++ b/scripts/managedb/csubs_control/csub_dups_checker.py
@@ -100,12 +100,22 @@ for doc in db[COL].find().sort('expiration', DESCENDING):
     n += 1
     csub_id = str(doc['_id'])
 
-    # The following fields are removed from signature calculation: _id, lastNotification, count and expiration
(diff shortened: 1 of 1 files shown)


scripts/managedb/csubs_control/csub_dups_checker.py
```diff
@@ -100,12 +100,22 @@ def log_status_restart():
...
+    # - expiration
+    # - description
+    # 
+    # (Not sure about throttling... as I have the doubt, I'm leaving it by the moment)
     doc.pop('_id', None)
     doc.pop('lastNotification', None)
     doc.pop('count', None)
+    doc.pop('status', None)
     doc.pop('expiration', None)
+    doc.pop('description', None)
```
iariasleon: I would use a try-except in all pop because if same field does not exist an error is returned,ex:
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
KeyError: 'description'
```

iariasleon: Forget the previous comment, I don´t see the "None" value

iariasleon:
LGTM

Issue #1916, opened 2016-03-17
Propagate or generate transaction ID


fgalan:
In the case of receiving a transaction ID (in the incoming HTTP request in a given header), it has to be used (and propagated to any outgoing HTTP request derived from such incoming HTTP request) instead of auto-generating one. 

If no transaction ID is provided in the HTTP request then CB has to auto-generate one and propagate it to any outgoing HTTP request derived from that transaction (forwarding and notifications).

Details about which particular header to use for transaction ID are under discussion in the moment of writing this.

kzangeli:
I imagine that we will use the same transaction ID for forwarding messages also.

What about notifications?
Will they carry the transaction ID of the update that provoked the notification?

fgalan:
> Will they carry the transaction ID of the update that provoked the notification?

That's correct, in the same way the for example carry the same X-Auth-Token header that the one in the updated provoking the notification.

fgalan:
At the end, the name of the header is going to be:

```
Fiware-Transaction
```

fgalan:
Currently we are using 

```
<CB start timestamp (seconds)>-<CB start timestamp (ms)>-<transaction number from startup>
```

eg: 1459515461-844-00000000001

However, maybe at the end we will use random UUIDv4 (https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29), which use the following format:

We should use existing libraries for generation, such libuuid (http://linux.die.net/man/3/libuuid) which seems to be available at CentOS after `yum search libuuid`.

See also: http://stackoverflow.com/questions/16884288/which-uuid-library-to-use

fgalan:
However, note that the internal transaction counter (either auto-generated or re-transmited) will be still in use in order to fill the information in `SUMMARY` traces.

fgalan:
Additional considerations:

* The field to map Fiware-Transaction is corr=
* The old transactionID= (or trans=... I don't remember exact name right now) would be "N/A" from now on. In the future it could be use for internal ID corresponding to the (global) transaction, but that is another history...

fgalan:
Final name for the header: `Fiware-Correlator`.

fgalan:
After discussion, the correlator (either auto-generated or received in the Fiware-Correlator header in the request) has to be included in the response.

fgalan:
Implemented in PR #2025. Asigned to QATestPending issue until corresponding functional test gets implemented.

fgalan:
[...]

PR #2025, opened 2016-04-11
Feature/1916 fiware correlator

kzangeli:
Fixed issue #1916 

Fiware-Correlator in log file and in HTTP header.
The correlator is present in all outgoing messages, i.e. responses, forwards and notifications.

Almost ALL functests have been modified, 2 have been added.
The two new functests are in test/func*/cases/1916*/*.test:

test/functionalTest/cases/1916_fiware_correlator/fiware_correlator.test
test/functionalTest/cases/1916_fiware_correlator/fiware_correlator_forward_and_notifications.test


doc/manuals/admin/logs.md
```diff
@@ -77,13 +77,22 @@ The different fields in each line are as follows:
...
         section on summary traces](#summary-traces) for details.
+-   **corr (correlator id)**. Can be either "N/A" (for log messages
+    "out of transaction", e.g. log lines corresponding to Orion Context
+    Broker startup), or it is a string in the UUID format.
+    An example:
+      "550e8400-e29b-41d4-a716-446655440000".
+    This 'correlator id' is either transferred from an incoming request,
+    or, if the incoming request doesn't carry any HTTP header "Fiware-Correlator",
+    the correlator is generated by the Orion context broker and then used in
+    the log file (as well as sent as HTTP header in forwarding messages and responses).
```
fgalan: Maybe is better "forwarded messages, *notifications* and responses" to make explicit that notifications are a outstanding case in which the correlator is propagated.

doc/manuals/admin/logs.md
```diff
@@ -77,13 +77,22 @@ The different fields in each line are as follows:
...
+    The correlator id is a common identifier among all services for one specific request.
```
fgalan: Not sure what do you mean by "service" in this sentence. Service in the sense of URL route? Service in the sense of Fiware-Service?

doc/manuals/admin/logs.md
```diff
@@ -77,13 +77,22 @@ The different fields in each line are as follows:
...
     "out of transaction", as the ones corresponding to Orion Context
-    Broker startup) or a string in the
-    format "1405598120-337-00000000001". The transaction id generation
-    logic ensures that every transaction id is unique, also for Orion
-    instances running in different VMs (which is useful in the case you
-    are aggregating logs from different sources), except if they are
+    Broker startup) or a string in the format "1405598120-337-00000000001".
+    The transaction id generation logic ensures that every transaction id is unique,
+    also for Orion instances running in different VMs (which is useful in the case
+    you are aggregating logs from different sources), except if they are
```
fgalan: Maybe we can add the following to remark the differences between transaction ID and correlation ID:

"Note that transaction ID is independent of correlator ID. Transaction ID has a local nature while correlator ID is meaningfull end-to-end, involving other software componentes apart from Context Broker itself"

Or some similar text that captures the same idea.

src/lib/rest/rest.cpp
```diff
@@ -82,6 +83,20 @@ static unsigned int              threadPoolSize;
 
 /* ****************************************************************************
 *
+* correlatorGenerate - 
+*/
+static void correlatorGenerate(char* buffer)
```
fgalan: Is the correlator generation logic thread-safe?

src/lib/rest/rest.cpp
```diff
@@ -1017,6 +1036,24 @@ static int connectionTreat
       ciP->httpHeaders.servicePath = defaultServicePath(url, method);
     }
 
+    char correlator[37];
```
fgalan: I'd suggest not using a 37 literal here. I'd suggest to put the CORRELATOR_ID_SIZE at limits.h or global.h (don't remberbe which one) and size arrays based on that.

doc/manuals/admin/logs.md
kzangeli: As this part explains the log file I didn't put much emphasis on the how corr works, only that it's 'there' (only put a short comment within parenthesis). 
I believe the HTTP header and it's functionality should be described elsewhere ...

That said, I have no problem in including more info here also

doc/manuals/admin/logs.md
kzangeli: Yeah, I struggled a little to find a good word.

"The set of TID/FIWARE applications involved in this 'message chain' of a request".
As I said before, this should be explained in detail somewhere ...

doc/manuals/admin/logs.md
kzangeli: And here we go again ... :-)
I don't think the description of the log file contents is the right place to describe the Fiware Correlator.
However, I will put in all these proposals in logs.md.
Perhaps some day we replace them with a link to some other part of the documentation.

BTW, the HTTP header Fiware-Correlator should also be explained somewhere, right?

src/lib/rest/rest.cpp
kzangeli: It is a thread variable, so that should not be a problem.
But you are probably thinking of the functions from libuuid.
Good question, needs some research.

src/lib/rest/rest.cpp
kzangeli: Fixed in e987104

doc/manuals/admin/logs.md
kzangeli: Fixed in e987104

doc/manuals/admin/logs.md
kzangeli: Fixed in e987104

doc/manuals/admin/logs.md
kzangeli: Fixed in e987104

fgalan: I don't know a better place to put that information. If you find another document/section in whihc i could fit, please go ahead (and add a link from this one). Otherwise, include the documentation here.

doc/manuals/admin/logs.md
fgalan: Ok, excetp mentionding TID/FIWARE, mabe:

"The set of components involved in this 'message chain' of a request".

kzangeli: Fixed in cd68d9d

src/lib/rest/rest.cpp
kzangeli: I've been searching for thread-safeness of the functions used from libuuid and nothing has come up.
From what the functions do, there is no reason to believer they aren't thread safe. Also, no complaints found on the web.
I think it's safe to assume we don't have a problem here. 

NTC

src/lib/common/limits.h
```diff
@@ -65,6 +65,8 @@
 #define IP_LENGTH_MAX           15    // Based on xxx.xxx.xxx.xxx
 #define STRING_SIZE_FOR_INT     16    // Room enough for an integer
 #define STRING_SIZE_FOR_DOUBLE  64    // Room enough for a double
+#define CORRELATOR_ID_SIZE      36
```
fgalan: I'd suggest to add a little comment about where that 36 comes, as in the cases above, e.g:

"Max size of a UUIDv4 string"

or something like that.

kzangeli: I thought about that, but decided that CORRELATOR_ID_SIZE explains pretty well what it is all about.  STRING_SIZE_FOR_DOUBLE also. That comment (Room enough for a double) is pretty useless ... :-)

However, for symmetry, I'll add a comment for it

kzangeli: Fixed in f3eb1d6

crbrox:
LGTM

src/lib/common/limits.h
fgalan: CORRELATOR_ID_SIZE explains pretty well what it is all about. However, I'm still thinkg that it is valuable to explain from where that number comes (in the same sense that for example it is important to explain in the corresponding place that 24 size limits in subsId comes from MongoDB OID length).

fgalan:
LGTM

PR #2034, opened 2016-04-13
[CHANGES] to jenkins

iariasleon:
@crbrox @kzangeli

### DIFF
--- a/test/acceptance/behave/components/ngsiv2/api_entry_point/retrieve_api_resource.feature
+++ b/test/acceptance/behave/components/ngsiv2/api_entry_point/retrieve_api_resource.feature
@@ -45,7 +45,7 @@ Feature: general operations in Context Broker using NGSI API v1 and v2
   Actions After the Feature:
   Setup: stop ContextBroker
 
-  @version
+  @version @skip
   Scenario: launch Context broker version request
     When send a version request
     Then verify that receive an "OK" http code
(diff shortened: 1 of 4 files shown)


test/acceptance/behave/components/ngsiv2/api_entry_point/retrieve_api_resource.feature
```diff
@@ -45,7 +45,7 @@ Feature: general operations in Context Broker using NGSI API v1 and v2
   Actions After the Feature:
   Setup: stop ContextBroker
 
-  @version
+  @version @skip
```
fgalan: Maybe XML removal has affected this test case, now /version works always in JSON. However, I understand that the right approach would be fixing the test instead of skipping. I'd recomment fixing in this very PR.

crbrox:
LGTM

test/acceptance/behave/components/ngsiv2/api_entry_point/retrieve_api_resource.feature
iariasleon: This test is skipped because verifies the version installed, and then it is necessary to modify the version value in the configuration manually with each modification. I will append another test that verify the version structure (x.y.z[-next]) using Regular expression.

fgalan: Maybe it would be a good practice to include a comment about every test with `@skip` in order to explain why that test is skipped (in this case, a copy paste or your above response could be used to fill that comment).

iariasleon: I will remove the skip tag, and I will skip this scenario in execution time with `-t=-version`. This way not generate confusion .

fgalan: You mean to modify the job script run by jenkins to use that `-t=-version` ?

iariasleon: yes, and the scenario with `@version` tag is omitted, only in jenkins

Issue #1945, opened 2016-03-23
Arithmetic and range operators in Simple Query Language also for strings


fgalan:
Current NGSIv2 specification http://telefonicaid.github.io/fiware-orion/api/v2/ allows to use `<`, `>`, `<=`, `>=` and ranges for dates an numbers.

This issue is about extensing the usage of these opertators to strings, so we can express queries like these ones:

```
GET /v2/entities?q=colour<c (returning entities with colour "black" and "blue")
GET /v2/entities?q=colour=o..s (returning entities with colour "pink" and "red")
```

Probably, the current code implementation already works with strings (as the `$lte` and similar operators used at mongoBackend are supposed to work with strings) but we would need at least to modify documentation and cover the functionality with test harness.

fgalan:
Implemented in PR #2036 (and a little bugfix in PR #2039). Moved to QATestPending while functional tests get implemented.

iariasleon:
E2E covered in features/pepflows/019_pep2cbv2_filteronstructuredcontent.feature

fgalan:
Issues in milestone QATestPendingE2ECovered are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeled with QATestPending label in the case things change and we need to recover it.

PR #2036, opened 2016-04-13
ADD string support in <, >, <=, >= and range filters

fgalan:
Implements #1945

Note that some existing .test has reduced their scope, as some cases that previously result in error after this PR are no longer errors.

@kzangeli
@crbrox


kzangeli:
LGTM

crbrox:
LGTM

PR #2041, opened 2016-04-14
Remove libwebsockets warnings and fix JSON responses

fortizc:


src/lib/orion_websocket/parser.cpp
```diff
@@ -160,7 +160,7 @@ const char *ws_parser_message
    * and add status code (3) and the final null character
    */
   char *json = (char *) malloc(strlen(tmpl) - 6 + strlen(headers) + msg.size() + 3 + 1);
-  sprintf(json, tmpl, headers, msg.c_str(), statusCode);
+  sprintf(json, tmpl, headers, msg.empty() ? "\"\"" : msg.c_str(), statusCode);
```
kzangeli: Just for curiosity, a std::string that is empty, doesn't c_str() return "" ?

src/lib/orion_websocket/ws.cpp
```diff
@@ -259,6 +259,7 @@ orion_websocket *orion_websocket_new(RestService *serv)
     0,0,0,0,0,0,0,0,0,0
   };
 
+  lws_set_log_level(0, NULL);
```
fgalan: Just to know... after this fix, does the `make ft` test work?

src/lib/orion_websocket/parser.cpp
fortizc: Yes, you're right, the standard C++98 (21.3.6 basic_string string operations) define that, but in the C++11 standard (21.4.7.1 basic_string accessors) does not show any restriction about it. Now if you want use C++11 in the future (which would be great), this is may be is the better solution. But I do not complete sure.

kzangeli: ok

NTC

kzangeli:
LGTM

src/lib/orion_websocket/ws.cpp
fortizc: lws_set_log_level() remove all LWS output, now I have other problems with make ft, but it must allow to make ft work without problems

fgalan: NTC

fgalan:
LGTM

PR #2048, opened 2016-04-18
Build on Ubuntu 14.04.3

markonikolic:
Added support for make and build on Ubuntu 14.04.4.


fgalan:
Contribution has been done in a way that doesn't break existing Ubuntu cases in the CMakeList.txt so... LGTM.

Thanks for the contribution!

PR #2051, opened 2016-04-19
FIX JSON broken in NGSIv2 csub renders

fgalan:
Recently introduced in a PR

### DIFF
--- a/src/lib/apiTypesV2/Subscription.cpp
+++ b/src/lib/apiTypesV2/Subscription.cpp
@@ -125,7 +125,9 @@ std::string Condition::toJson()
   if (this->expression.coords != "")    jhe.addString("coords", this->expression.coords);
   if (this->expression.georel != "")    jhe.addString("georel", this->expression.georel);
 
-  if (jhe.str() != "{}")                jh.addRaw("expression", jhe.str());
+  std::string expressionString = jhe.str();
+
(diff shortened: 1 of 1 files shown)


src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -125,7 +125,9 @@ std::string Condition::toJson()
   if (this->expression.coords != "")    jhe.addString("coords", this->expression.coords);
   if (this->expression.georel != "")    jhe.addString("georel", this->expression.georel);
 
-  if (jhe.str() != "{}")                jh.addRaw("expression", jhe.str());
+  std::string expressionString = jhe.str();
+
+  if (expressionString != "{}")         jh.addRaw("expression", expressionString);
```
kzangeli: Is jhe.str()  destroyed after usage ?

fgalan: Not exactly, but each invocation adds a '}' to the output stream. Thus, str() is a method that shouldn't be invoked more than one time.

In this sense, uses the same pattern that BSONObjBuilder, which obj() method should be called one time.

kzangeli: NTC

kzangeli:
LGTM

Issue #1875, opened 2016-03-03
Implement NGSIv2 notification formats


fgalan:
After merging PR https://github.com/telefonicaid/fiware-orion/pull/1761, the format for NGSIv2 notification has been defined. This issue is about supporting it, considering the four different flavors for `attrFormat`: normalized, keyValues, values and legacy (legacy is for using NGSIv1 format).

_Implementation hints_

I'd suggest to use the `format` field in csub documents (see https://fiware-orion.readthedocs.org/en/develop/admin/database_model/index.html#csubs-collection) in order to specify the format associated to the subscription. The string "JSON" will remain for NGSIv1 legacy, other formats could use: "NGSIv2-NORMALIZED", "NGSIV2-KEYVALUES" and "NGSIV2-VALUES". The file **database_model.md** document has to be edited to include the new values (either these suggested names or others).

The `attrsFormat` parameter has to be processed at subscription creation/modification to set correctly the format in DB. The notification logic needs to be modified to take this parameter into account and render the notification the right way. The key functions are processOnChangeConditionForUpdateContext() and processOnChangeConditionForSubscription(), both take format into account yet as parameter.

Note that in NGSIv2, the order in which attributes appear in the `attrs` array in csub document is important (especially in NGSIV2-VALUES) as it defines the order in which the attrs will appear.

(I'm talking all the time about "csub document" but the same applies for the cache entries, which should have the same structure as csub documents in DB).

fgalan:
The following answer at SOF needs to be edited once this issue gets solved: http://stackoverflow.com/questions/36504230/doubts-about-notifications-format-in-orion-usin-apiv2

kzangeli:
Stuff left of this issue after PR #2054

* Unify the two parameters 'notifyFormat' and 'renderMode' for ContextAttribute::toJson
* Order of attributes in notifications, especially if notifyFormat is "values"
* sendNotifyContextAvailabilityRequest to accept NotificationFormat instead of Format and similar stuff to align availlability notifications with context notifications
* **Add more action items here**

fgalan:
What is pending for step III (I hope, the last one in this PR):

* sendNotifyContextAvailabilityRequest to accept NotificationFormat instead of Format and similar stuff to align availlability notifications with context notifications
 * [FGM] In particular, using NGSI_V1_JSON in the places where the NGSI9 notification code now uses JSON.

[...]

Issue #2015, opened 2016-04-08
NGSIv2 notification templates


fgalan:
CC: @mrutid , @jmcanterafonseca 

Apart from the different notification formats defined in NGSIv2, we should implement the capability of customizing notifications.

Related with the changes in #2014, the `http` field will be expanded in the following way, to support additional sub-fields to `url`:

      "http": {
        "url": "http://localhost:1234",
        "headers": {
          "X-MyHeader": "foo"
        },
        "method": "PUT",
        "qs": {
          "authToken": "bar"
        }
        "payload": <...>
      }

Fields description (all them optionals):

* `headers`: a key-map to build the HTTP headers to include in the subscription
* `method`: method to use (by default POST is used, but this field would allow to change)
* `qs`: a key-map used to build the query parameter that are appended to the URL. Note that query parameters cannot be used directly in the `url` field, as `=` is a forbidden char.
* `payload`: the payload to use. If omitted, the regular payload according to NGSIv2 is used.

All them should implement a template language that enables to use of variable values. The details of this template language are still pending (suggestions are welcome!).

jmcanterafonseca:
maybe we can use the templating mechanism as well for other notification channels, namely web socket.

so the template would be part of the notification object and not the http object. For instance

```json
{ 
   "notification": {
      "attrs": ["temperature", "humidity"]
      "template": [{
         "id": "$id",
         "t":  "$(temperature)",
         "h": "$(humidity) * 100",
         "custom": "my own key"
      }]
   }
}
````

another nice to have for the templating mechanism is the capability of specifying the container for the elements, it could be an array or a keyValue object, ex.

```json
{ 
   "notification": {
      "attrs": ["temperature", "humidity"]
      "template": {
         "$id": {
            "t":  "$(temperature)",
            "h": "$(humidity) * 100"
         }
      }
   }
}
````

**Comment by mrutid** [comment]
I may want to templatize a header content for example. I don't see how your proposal cover this aspect.

**Comment by fgalan** [comment]
Another interesting case is to templatize URL, e.g:

```
"url": "http://example.com:8080/processEntity/${id}"
```

jmcanterafonseca:
oh yes. we can have template at payload level, or at HTTP request level

2016-04-12 13:31 GMT+02:00 Fermín Galán Márquez <notifications@github.com>:

fgalan:
> oh yes. we can have template at payload level, or at HTTP request level

[...]

PR #2054, opened 2016-04-19
Feature/1875 ngsi v2 notification formats

kzangeli:
Notifications for NGSIv2.

Fixes issue #1875, or a big part of it at least.

All old functests work and a pretty simple one added to test all three formats (cases/1875\*/\*.test).
Another more complete functest to be added in a later PR.
Also, ordering of attributes is not considered in this PR, that is postponed to the next one.
This PR is big enough already ...


scripts/accumulator-server.py
```diff
@@ -44,36 +44,44 @@
...
 
 
 # This function is registered to be called upon termination
 def all_done():
     os.unlink(pidfile)
 
 # Default arguments
-port = 1028
-host='0.0.0.0'
+port       = 1028
```
fgalan: Python style is automatically checked by the pep8 tool. Not sure if this "cross-application" of C/C++ based style would make happy to pep8. However, we can skip this by the moment (probably the accumulator-server.py in its version before this PR is not ok...).

NTC

src/lib/apiTypesV2/Attribute.cpp
```diff
@@ -52,7 +52,8 @@ std::string Attribute::render(ConnectionInfo* ciP, RequestType requestType, bool
     else
     {
       out = "{";
-      out += pcontextAttribute->toJson(true, renderMode, requestType);  // param 1 'true' as it is the last and only element
+      // FIXME PR: renderMode + attrsFormat ... a bit confusing. Should we join these two somehow?
```
fgalan: Which values could take renderMode?

src/lib/common/NotificationFormat.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "common/NotificationFormat.h"
+#include "common/wsStrip.h"
+
+
+
+/* ****************************************************************************
+*
+* notificationFormatToString - 
+*/
+const char* notificationFormatToString(NotificationFormat format, bool noDefault)
```
fgalan: What about returning `cosnt std::string` ?

src/lib/common/NotificationFormat.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+  switch (format)
+  {
+  case NGSI_V1_JSON:                return "JSON";
```
fgalan: I understand that the description of `format` field in https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/admin/database_model.md#csubs-collection should be modified to include the new alternatives.

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -82,7 +83,7 @@ static bool processSubscriptions
                                          trigs->attrL,
                                          mapSubId,
                                          trigs->reference,
-                                         trigs->format,
+                                         JSON,
```
fgalan: This means that format is "hardwired" to "legacy JSON" in NGSI9 notifications? Given that no alternative format has been defined in NGSIv2 for this kind of notifications :) I think is the right way of doing

However, maybe a comment like "// FIXME: this will be unhardwirer once ctx availability notification formats get defined for NGSIv2"

fgalan: However, I've just realized... is "JSON" correct? Or it should be "NGSI_V1_JSON"?

Actually, this lead to a new question... Is "JSON" stills working as "symbol" in the code? After the refactoring from `Format` enumeration to `NotificationFormat` I have said that "JSON" would no be longer valid...

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -273,7 +274,7 @@ static bool addTriggeredSubscriptions
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
       TriggeredSubscription* trigs = new TriggeredSubscription(
-        sub.hasField(CASUB_FORMAT) ? stringToFormat(getStringFieldF(sub, CASUB_FORMAT)) : JSON,
+        sub.hasField(CASUB_FORMAT)? stringToNotificationFormat(getStringFieldF(sub, CASUB_FORMAT)) : NGSI_V1_JSON,  // FIXME PR: hard to decide a default NotificationFormat ... JSON or NGSI_V2_NORMALIZED ?
```
fgalan: This is related with a previous comment. I'd sugges to use NGSI_V1_JSON plus a comment like:

// FIXME: this will be refactored once ctx availability notification formats get defined for NGSIv2"

src/lib/common/NotificationFormat.cpp
kzangeli: That is exactly what we need to try to avoid, returning an entire object on the stack.
We probably have it in a thousand places, but no need to add more ...

What we should do is to transform all functions returning std::string into returning either **const std::string&** or **const char***

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1246,14 +1246,16 @@ static bool addTriggeredSubscriptions_noCache
 
       LM_T(LmtMongo, ("adding subscription: '%s'", sub.toString().c_str()));
 
-      long long throttling       = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLongF(sub, CSUB_THROTTLING)       : -1;
-      long long lastNotification = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLongF(sub, CSUB_LASTNOTIFICATION) : -1;
+      long long           throttling         = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLongF(sub, CSUB_THROTTLING)       : -1;
+      long long           lastNotification   = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLongF(sub, CSUB_LASTNOTIFICATION) : -1;
+      std::string         notifyFormatString = sub.hasField(CSUB_FORMAT)? getStringFieldF(sub, CSUB_FORMAT) : "";
+      NotificationFormat  notifyFormat       = stringToNotificationFormat(notifyFormatString);
```
fgalan: Alternative:

```
NotificationFormat  notifyFormat       = stringToNotificationFormat(sub.hasField(CSUB_FORMAT)? getStringFieldF(sub, CSUB_FORMAT) : "");
```

Pros: you save a variable. More compact.
Const: longer and maybe more obscure.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1982,7 +1983,7 @@ bool processAvailabilitySubscription
     /* Complete the fields in NotifyContextRequest */
     ncar.subscriptionId.set(subId);
 
-    getNotifier()->sendNotifyContextAvailabilityRequest(&ncar, notifyUrl, tenant, fiwareCorrelator, format);
+    getNotifier()->sendNotifyContextAvailabilityRequest(&ncar, notifyUrl, tenant, fiwareCorrelator, JSON);
```
fgalan: JSON -> NGSI_V1_JSON ? + fixme about unhardwiring in the future.

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: In a few places, the parameter *format* has disappeared and *notifyFormat* appeared in its place, with a different type.
I didn't want to touch ngsi9 stuff during the PR (it would never end!) so I believe hardcoding JSON is the best choice (which will be fixed once ngsi9 is handled the same way).

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -424,8 +425,26 @@ HttpStatusCode mongoUpdateContextSubscription
...
+
+  if (requestP->attrsFormat != NGSI_NO_NOTIFICATION_FORMAT)
+  {
+    notifyFormat       = requestP->attrsFormat;
+    notifyFormatString = notificationFormatToString(notifyFormat);
+    newSub.append(CSUB_FORMAT, notifyFormatString.c_str());
+  }
+  else
+  {
+    notifyFormatString = sub.hasField(CSUB_FORMAT)? getStringFieldF(sub, CSUB_FORMAT) : "JSON";  // If not in database, "JSON" is the default value
```
fgalan: "JSON" -> notificationFormatToString(NGSI_V1_FORMAT)

(Maybe "safer" and more manteinable code in the case we change the keyword in the future).

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yeah, I had it like that in my first version.
The line got very long and hard to read so I added a help variable

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -756,13 +756,36 @@ std::string ContextAttribute::render
...
-std::string ContextAttribute::toJson(bool isLastElement, const std::string& renderMode, RequestType requestType)
+std::string ContextAttribute::toJson(bool isLastElement, const std::string& renderMode, NotificationFormat notifyFormat, RequestType requestType)
 {
   std::string  out;
+  bool         uniqueValues = false;
+  std::string  rMode        = renderMode;  // renderMode is 'const' and cannot be modified
+
+  //
+  // FIXME PR: To be discussed during PR-review
+  //   'notifyFormat' overrides 'renderMode' if NGSI_V2_KEYVALUES or NGSI_V2_VALUES but 'uniqueValues' must be saved
```
fgalan: Unifying sound good... However, I don't have the full context of this and not sure if this would be a good or bad idea from the point of view of complexity.

(Sorry for providing such poor feedback :( )

src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -202,7 +203,7 @@ HttpStatusCode mongoSubscribeContext
     }
 
     /* Adding format to use in notifications */
-    sub.append(CSUB_FORMAT, "JSON");
+    sub.append(CSUB_FORMAT, notificationFormatToString(requestP->attrsFormat));
```
fgalan: Could you elaborate a little bit on how requestP->attrsFormat is filled, please? (Or include the links to the parts of the code where that filling takes places, both for NGSIv1 and NGSIv2 cases).

src/lib/apiTypesV2/Attribute.cpp
kzangeli: I'm not an expert ...
I believe they (renderMode and attrsFormat) are pretty similat, however, renderMode has the possible value "unique"

src/lib/ngsi/ContextElement.cpp
```diff
@@ -111,6 +111,32 @@ std::string ContextElement::render(ConnectionInfo* ciP, RequestType requestType,
 
 /* ****************************************************************************
 *
+* ContextElement::toJson - 
+*/
+std::string ContextElement::toJson(ConnectionInfo* ciP, NotificationFormat notifyFormat)
```
fgalan: Is ciP actually used in the body of the method? It doesn't seem so...

(Note that most of the toJson() method doesn't use ciP, differenctly to render() familiy)

src/lib/ngsi/ContextElementResponseVector.cpp
```diff
@@ -73,6 +73,31 @@ std::string ContextElementResponseVector::render
 
 /* ****************************************************************************
 *
+* ContextElementResponseVector::toJson - 
+*/
+std::string ContextElementResponseVector::toJson(ConnectionInfo* ciP, NotificationFormat notifyFormat)
```
fgalan: Same here. ciP is not used, except for passing down to CER toJson() (and, in CER toJson(), ciP is not used, as mentionced in a different comment).

src/lib/ngsi/EntityId.cpp
```diff
@@ -136,6 +136,28 @@ std::string EntityId::render
...
+{
+  std::string  out;
+  std::string  typeEscaped  = htmlEscape(type.c_str());
+  std::string  idEscaped    = htmlEscape(id.c_str());
+
+  out += JSON_VALUE("id", idEscaped);
+  out += ",";
+  out += JSON_VALUE("type", typeEscaped);
+
+  // out += "\"id\":\""   + idEscaped   + "\",";
```
fgalan: Leftovers?

src/lib/ngsi10/NotifyContextRequest.cpp
```diff
@@ -61,6 +64,46 @@ std::string NotifyContextRequest::render(ConnectionInfo* ciP, RequestType reques
 
 /* ****************************************************************************
 *
+* NotifyContextRequest::toJson -
+*/
+std::string NotifyContextRequest::toJson(ConnectionInfo* ciP, NotificationFormat notifyFormat)
```
fgalan: "Cascade comment" :)

removing ciP in CER toJson() -> removing ciP in CER vector toJson() -> removing ciP in NotifyContextRquest toJson()

src/lib/common/NotificationFormat.cpp
kzangeli: Fixed in 51eb54e

src/lib/ngsi10/NotifyContextRequest.cpp
fgalan: ...or maybe not in this case, as it seems that we have an old render() method that may need ciP (just after "First, v1 rendering of notification?").

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: requestP->attrsFormat are filled in during the parse step, and set to default values before the parse step

src/lib/common/NotificationFormat.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+/* ****************************************************************************
+*
+* stringToNotificationFormat -
+*/
+NotificationFormat stringToNotificationFormat(const std::string& s, bool noDefault)
+{
+  if (s == "JSON")       { return NGSI_V1_JSON;       }
```
fgalan: Doubt: this names are aligned with the strings used in the API for attrsName? The string used in DB for `format` field in csub collection? Both?

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -248,6 +254,39 @@ std::string parseSubscription(ConnectionInfo* ciP, ParseData* parseDataP, JsonDe
...
+      alarmMgr.badInput(clientIp, "attrsFormat is not a string");
+      return oe.render(ciP, "");
+    }
+
+    std::string         attrsFormatString = attrsFormat.GetString();
+    NotificationFormat  nFormat           = stringToNotificationFormat(attrsFormatString, true);
+
+    if (nFormat == NGSI_NO_NOTIFICATION_FORMAT)
+    {
+      const char*  details  = "invalid attrsFormat (accepted values: normalized, keyvalues, values)";
```
fgalan: "legacy" is also a valid keyword for attrsFormat (see https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/user/v1_v2_coexistence.md#ngsiv1-notification-with-ngsiv2-subscriptions).

src/lib/ngsiNotify/Notifier.h
```diff
@@ -25,33 +25,37 @@
...
-                                                      const std::string&                fiwareCorrelator,
-                                                      Format                            format           = JSON);
+  virtual ~Notifier(void);
+
+  virtual void sendNotifyContextRequest(NotifyContextRequest* ncr,
+                                        const std::string&    url,
+                                        const std::string&    tenant,
+                                        const std::string&    xauthToken,
+                                        const std::string&    fiwareCorrelator,
+                                        NotificationFormat    notifyFormat = NGSI_V2_NORMALIZED);
```
fgalan: Probably it doesn't matter if sendNotifyContextRequest() is always invoked with all the paramters but, while NGSIv1 is supported and from the point of view of backward comptibility, maybe is "safer" to use NGSI_V1_JSON as default.

src/lib/ngsi/ContextElement.cpp
kzangeli: Yes, true. I added the ConnectionInfo only for symmetry.

If it is true what you say that most toJson don't accepr ConnectionInfo as a parameter then perhaps it was not a good idea.

However, if we truly want symmetry (which is a very good thing) and only a few toJson() needs ciP, then all toJson() must accept it ...

15/8   without/with => 35% of toJsons accept ciP as a parameter.

So, if we want symmetry, I'll leave ciP and if not, then I'll remove it.
Votation! :-)

src/lib/ngsiNotify/Notifier.h
```diff
@@ -25,33 +25,37 @@
...
+
+  virtual void sendNotifyContextAvailabilityRequest(NotifyContextAvailabilityRequest* ncr,
+                                                    const std::string&                url,
+                                                    const std::string&                tenant,
+                                                    const std::string&                fiwareCorrelator,
+                                                    Format                            format           = JSON);
```
fgalan: Probably we should take this opportunity to use also NotificationFormat in this place. The code will be easier to mantain with just one enum, not Format and NotifyFormat at the same time.

src/lib/ngsiNotify/QueueNotifier.cpp
```diff
@@ -100,8 +101,8 @@ void QueueNotifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const st
     spathList = "";
   }
 
-  ci.outFormat = format;
-  std::string payload = ncr->render(&ci, NotifyContext, "");
+  ci.outFormat = JSON;
```
fgalan: JSON -> NGSI_V1_FORMAT?

src/lib/rest/httpRequestSend.cpp
```diff
@@ -168,6 +168,7 @@ int httpRequestSendWithCurl
    const std::string&     orig_content_type,
    const std::string&     content,
    const std::string&     fiwareCorrelation,
+   const std::string&     notifyFormat,
```
fgalan: Maybe this should be NotificationFormat instead of std::string ?

src/lib/serviceRoutinesV2/postSubscriptions.cpp
```diff
@@ -62,6 +62,10 @@ extern std::string postSubscriptions
     return answer;
   }
 
+  if (parseDataP->scr.res.attrsFormat == NGSI_NO_NOTIFICATION_FORMAT)
+  {
+    parseDataP->scr.res.attrsFormat = NGSI_V2_NORMALIZED; // FIXME PR: default notification format ?
```
fgalan: Not sure.. what the .apib spec says about default attrsFormat ?

test/functionalTest/cases/0000_ngsi10_convenience/subscriptions_onchange.test
```diff
@@ -662,4 +662,4 @@ Fiware-Correlator: REGEX([0-9a-f\-]{36})
 --TEARDOWN--
 brokerStop CB
 accumulatorStop
-dbDrop CB
+#dbDrop CB
```
fgalan: Leftover

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
```diff
@@ -26,7 +26,7 @@ Filter with dateCreated in subscriptions
 --SHELL-INIT--
 dbInit CB
 brokerStart CB
-accumulatorStart
+accumulatorStart --pretty-print
```
fgalan: Other commnads use a different style, eg.:

dbInit CB
brokerStart CB 0-255
etc.

So probably we should align and use something like:

accumulatorStart prettyPrint

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
```diff
@@ -0,0 +1,509 @@
...
+
+--SHELL--
+
+#
+# 01. Create E1/A1
+# 02. Create NGSIv1 subscription for E1/A1 (A1, A2)
+# 03. Create NGSIv2 subscription for E1/A1 (A1, A3), without attrsFormat (normalized is default)
+# 04. Create NGSIv2 subscription for E1/A1 (A1, A4), with attrsFormat set to normalized
+# 05. Create NGSIv2 subscription for E1/A1 (A1, A5), with attrsFormat set to keyvalues
+# 06. Create NGSIv2 subscription for E1/A1 (A1, A6), with attrsFormat set to values
```
fgalan: Add step:

...., with attrsFormat set to legacy

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
```diff
@@ -0,0 +1,509 @@
...
+Content-Length: 60
+User-Agent: orion/REGEX(\d+\.\d+\.\d+.*)
+Host: localhost:REGEX(\d+)
+Accept: application/json
+X-Ngsiv2-Attrsformat: values
+Content-Type: application/json; charset=utf-8
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+
+{
+    "data": [
```
fgalan: This test bring me something to memory... From NGSIv2 spec (section describing subscription payload):

> attrs [at notification]: List of attributes to be included in the notification message. If not specified (or empty), all attributes are included in the notification. It also defines the order in which attributes will appear in notifications when attrsFormat value is used (see "Notification messages" section).

An additional .test covering this case should be included (maybe modifications in the code are needed also), both the three NGSIv1 modes: normalized, keyValues and values (it is specially important in the third case).

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
```diff
@@ -0,0 +1,509 @@
...
+
+
+11. Dump accumulator to see notification of (A1, A3) with keyvalues
+===================================================================
+POST http://localhost:REGEX(\d+)/notify
+Content-Length: 95
+User-Agent: orion/REGEX(\d+\.\d+\.\d+.*)
+Host: localhost:REGEX(\d+)
+Accept: application/json
+X-Ngsiv2-Attrsformat: keyvalues
```
fgalan: I think the righ one according to NGSIv2 spec is `keyValues`, not `keyvalues`. As far as I understand, HTTP header names are case-insensitive, but I think values of the headers are case-sensitive.

The same `keyValues` should be used at DB format csub level.

src/lib/common/NotificationFormat.cpp
```diff
@@ -0,0 +1,76 @@
...
+*
+* notificationFormatToString - 
+*/
+const char* notificationFormatToString(NotificationFormat format, bool noDefault)
+{
+  switch (format)
+  {
+  case NGSI_V1_JSON:                return "JSON";
+  case NGSI_V2_NORMALIZED:          return "normalized";
+  case NGSI_V2_KEYVALUES:           return "keyvalues";
```
fgalan: Also mentioned in another place, this should probably be "keyValues".

src/lib/common/NotificationFormat.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+/* ****************************************************************************
+*
+* stringToNotificationFormat -
+*/
+NotificationFormat stringToNotificationFormat(const std::string& s, bool noDefault)
+{
+  if (s == "JSON")       { return NGSI_V1_JSON;       }
+  if (s == "normalized") { return NGSI_V2_NORMALIZED; }
+  if (s == "keyvalues")  { return NGSI_V2_KEYVALUES;  }
```
fgalan: keyvalues -> keyValues

test/unittests/commonMocks.h
```diff
@@ -160,13 +161,14 @@ class NotifierMock : public Notifier {
...
+    MOCK_METHOD6(sendNotifyContextRequest, void(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, NotificationFormat notifyFormat));
     MOCK_METHOD5(sendNotifyContextAvailabilityRequest, void(NotifyContextAvailabilityRequest* ncar, const std::string& url, const std::string& tenant, const std::string& fiwareCorrelator, Format f));
 
     /* Wrappers for parent methods (used in ON_CALL() defaults set in the constructor) */
-    void parent_sendNotifyContextRequest(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, Format format)
+    void parent_sendNotifyContextRequest(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, NotificationFormat notifyFormat)
     {
-      return Notifier::sendNotifyContextRequest(ncr, url, tenant, xauthToken, fiwareCorrelator, format);
+      LM_W(("KZ: In parent_sendNotifyContextRequest"));
```
fgalan: Leftover

(A grep for "KZ" in the repository would help to detect other cases).

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
fgalan: I have just realized about:

>  Also, ordering of attributes is not considered in this PR, that is postponed to the next one.

Thus, I understand that it will be deferred to the very next PR.

NTC

src/lib/apiTypesV2/Attribute.cpp
fgalan: Before unifiying them we should precisely now which values could take renderMode in order to see how to deail with the cartesian product bewteeen renderMode values and attrsFormat values. Makes sense?

src/lib/common/NotificationFormat.cpp
fgalan: Does the returning of `const std::string&` avoids the problem you mention, compared with `const std::string`? In that case, let's use it.

doc/manuals/admin/database_model.md
```diff
@@ -284,8 +284,8 @@ Fields:
     fields: q, georel, geometry and/or coords (optional)
 -   **count**: the number of notifications sent associated to
     the subscription.   
--   **format**: the format to use to send notification, currently "JSON"
-    meaning JSON notifications in NGSIv1 format.
+-   **format**: the format to use to send notification, possible values are **JSON**
+    (meaning JSON notifications in NGSIv1 format), **normalized**, **keyvalues** and **values**.
```
fgalan: I'd suggest to add at the end of the sentence "... (used in NGSIv2 format)" (to be more comprehensive and have moer "symetry" with the NGSIv1 format description).

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: I agree in not making this PR more complex. However, I think that for the sake of "cleaness" we should avoid having NotifyFormat and Format in the code base, unifying them under NotifyFormat. That means that part of the NGSI9 notification code needs a little bit of refactor to use NotifyFormat with hardwired NGSI_V1_JSON, but I think is not too much effort and worth the time (not for this PR and maybe not for the very next PR, but maybe for the next one after the very next PR :).

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: However, I think I have see code for setting attrsFormat *after* the parsing stage (in the serviceRoutines). I have lo take a closer look.

src/lib/ngsi/ContextElement.cpp
fgalan: Don't miss "symmetry" with "exception". I mean, if some toJson() needs ciP, let's add it. But for the most of them that doesn't need, no need to make things more complex. 

I love the `toJson(void)` we have in some places and I'd prefer if all toJson() tend to work as closer as possible to that pattern. In my opinion, this is a HUGE improvement compared with the old render() family used mainly in NGSIv1, using very complex signatures that makes the code very difficult to understand and mantain (I had nightmares in the past with all that stuff about isVector, commaBlabla, isLastElement and the like :S)

src/lib/ngsi/EntityId.cpp
kzangeli: Yes, of course :-)

src/lib/ngsiNotify/Notifier.h
kzangeli: The default value is tricky in this issue.
Clear is that if a v2 request comes in without attrsFormat, the default is 'normalized', and if v1, the default is 'JSON'.
In other places, like extracting a sub from mongo without 'format' it's a little more difficult.

That said, this part is your territory, if you think NGSI_V1_JSON is better here, who am I to argue? :-)
[ Unless the functests start to protest ... ]

src/lib/ngsiNotify/Notifier.h
kzangeli: I would prefer to refer things like this to the next PR, creating an issue ...
As said, this PR is already a lot bigger than desired ...

A change in the signature of a function sometimes carries **a lot** of work extra

src/lib/ngsiNotify/Notifier.h
fgalan: If default value is tricky, then let's remove default values at all.

These method are called from a couple or places or so, so it shouldn't be difficult ot adjust... at least in the "actual" code (not sure about unit test, that may involve a number between 0 or thousand changes, based in previous experiences :). If unit test are a problem, then, let's pospone the un-defaulting to a next PR.

src/lib/ngsiNotify/Notifier.h
fgalan: Instead of creating a new issue, please take down of it in #2015 as "Pending things" in a comment. Not only for this case, but for all the other things we have mention to defer to another PR.

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
kzangeli: Yeah, but what if we look at orionCurl instead?
The good way is to use 'flags', like orionCurl  (in my opinion).
I would suggest to (apart from implementing a real argv handler in the accumulator), try to use the flags in the orionCurl style. So much easier to use.  That's kind of why I decided to use --pretty-print and not prettyPrint

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
kzangeli: Yeah, these are exactly the type of things I wanted to delay to a second PR.
All existing functests work with the current implementation, but that is not a guarantee that the new notifications work 100%. More extensive (and probably bug fixes) functests to be provided in a later (starting tomorrow) PR.

test/unittests/commonMocks.h
kzangeli: Yeah, I did one, as I always do.
But of course, I look in *.cpp, not in *.h ...
This is the only one

src/lib/common/NotificationFormat.cpp
kzangeli: Actually, I don't really know if `const std::string&` works in all cases.
I have seen more than just a little problems in the past with std::string&`, especially when sending a string from function to function.
What I do know for certain is that `const char*` works perfectly in all cases and that it gives 0% overhead.

Is there anything to gain here by using std::string?

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: I am happy to include everything missing in the next PR.
We just need to keep track of the stuff to postpone/include.
Maybe a new issue ?

src/lib/apiTypesV2/Attribute.cpp
kzangeli: Yes, it makes sense.
The render() method of ContextAttribute is pretty complex and we need to look at this carefully

src/lib/ngsiNotify/Notifier.h
kzangeli: Only reason I used a default value (I hate them) here is that unittests use this function in a huge number of places.
It is a pity that the unittests have power over how the source code of the broker is implemented ...

A few strategic macros in unit tests would remedy this problem.

kzangeli: And, if we decided to vote on whether to forbid default values in the broker, if would definitely vote in favor.

src/lib/ngsiNotify/Notifier.h
kzangeli: ok, let's do that.
But, you mean, #1875, not #2015. right?  (2015 is about templates)

src/lib/ngsi/ContextAttribute.cpp
kzangeli: No worries, we still have time to discuss this.
Postponed for next PR and added to 'postponed list' in issue #1875

src/lib/apiTypesV2/Attribute.cpp
kzangeli: Added to Postponed List in #1875

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: True.
Fixed in abaf8ef

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 760db57

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: Fixed in 760db57

src/lib/ngsi/EntityId.cpp
kzangeli: Fixed in 15cd784

src/lib/ngsiNotify/Notifier.h
kzangeli: On the Postponed List in #1875

src/lib/ngsiNotify/QueueNotifier.cpp
kzangeli: No, that is not the same type.
ci.outFormat can be either text/plain or application/json.
This is about the 'language' to use for the response, not about the internal formatting 'inside a language', like 'notifyFormat'.

Of course, we could join all of these variables to use the same type, somehow include 'text/plain' also and have ci.inFormat and ci.outFormat use that same type ...  I don't think it's a bad idea.
[ The three different notification 'formats' for json really complicated things ... :-D ]

src/lib/rest/httpRequestSend.cpp
kzangeli: I was aboput to do that but decided not to change this function to have it more generic.
The way it works right now is that whatever text is found in 'notifyFormat' is sent as X-Ngsiv2-AttrsFormat (exceptions: if empty or JSON, the header is omitted).

To change this would be to complicate the function and at the same time make it less generic.

I'm not saying NO, but right now it doesn't seem like a good idea.

src/lib/ngsi/ContextElement.cpp
kzangeli: Well, these functions, toJson() and render() have (with the exception of 'symmetry') the input they need to make their job.   Another way of doing this is to create N function without parameters and have the job done in N places (without ifs) instead.   This give faster and more readable code and I am almost always in favor of that.

Much less maintainable though ...

You are the one that puts maintainability on top of your list, not me ... ;-)

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: Added to Postponed List in #1875

src/lib/ngsiNotify/QueueNotifier.cpp
kzangeli: Added (with a question mark) in Postponed List in #1875

src/lib/serviceRoutinesV2/postSubscriptions.cpp
kzangeli: Here you have the modification of attrsFormat in a service routine.
Note that it is only changed if it not previously set, like a lifeline ...

And yes, the default here has to be NGSI_V2_NORMALIZED.
After all, we are in serviceRoutinesV2.

FIXME to be removed.

kzangeli: FIXME removed in fdb2412

test/functionalTest/cases/0000_ngsi10_convenience/subscriptions_onchange.test
kzangeli: Fixed in 25b4fc8

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
kzangeli: legacy already in Postponed List in #1875

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats.test
kzangeli: Fixed in 760db57

src/lib/common/NotificationFormat.cpp
kzangeli: Fixed in 760db57

src/lib/common/NotificationFormat.cpp
kzangeli: Fixed in 760db57

test/unittests/commonMocks.h
kzangeli: Fixed in 760db57

doc/manuals/admin/database_model.md
kzangeli: Fixed in adcc34c

src/lib/mongoBackend/MongoCommonRegister.cpp
fgalan: We don't need a new issue. You can use comments on #2015 to keep track of pending things related with notifications NGSIv2 functionality.

src/lib/ngsiNotify/Notifier.h
fgalan: > It is a pity that the unittests have power over how the source code of the broker is implemented ...

It isn't an actual power. It is a matter of chaning unit test to align with the new function signature. Some times an heavy work (as it may involve to change many code lines) but usually very simple a straighforward.

> And, if we decided to vote on whether to forbid default values in the broker, if would definitely vote in favor.

Default values are not fobidden in the code, as it may be cases where they are useful. We should consider case by case and, in this particular case, after discussion (above comments) it seems it is better not having default values.

src/lib/common/NotificationFormat.cpp
fgalan: We have

`NotificationFormat stringToNotificationFormat(const std::string& s, bool noDefault = false)`

so it makes sense that the dual function have a dual parameters:

```
const std::string& notificationFormatToString(NotificationFormat format, bool noDefault)
```
Of course, this is not a must and we can use `const char*` at the end if that is the only option. However, I think that it worth the time to research a little bit in how to return a std::string y C++ without the "stack overhead situation" you describe. In the worst case, at leat we gain some new knowledgue.

fgalan: Some thread to start reseraching after a quick google search: http://stackoverflow.com/questions/3976935/best-way-to-return-an-stdstring-that-local-to-a-function

fgalan: More references to this topic (I haven looked in detail yet)

* http://stackoverflow.com/questions/1032243/should-i-return-stdstrings
* http://stackoverflow.com/questions/13507698/return-stdstring-as-const-reference

src/lib/ngsiNotify/Notifier.h
fgalan: You are right, correct number is #1875.

src/lib/ngsiNotify/QueueNotifier.cpp
fgalan: Maybe the abuse of "format" is making things complicated...

Thus, at the end probably we need two enumaration and my original idea of joining them is not good. One named NotificationFormat (the one about which this PR is mainly about). The other maybe should be MimeType (or something like that) as their values are JSON and Text vay the time being. I think is a bad idea mixing into the same type, as that would mix different kind of objects.

src/lib/rest/httpRequestSend.cpp
fgalan: If that is the interpretation, then probably a better name is `ngisv2AttrFormat` or something closer to the hadear string nature. When somebody reads `notifyFormat` tends to thing in the `NotificationFormat` enum and it may be confusing to have a std::string there (that confussion motivates my first comment).

src/lib/ngsi/ContextElement.cpp
fgalan: I'm in favour of each function having *no more and no less* of the input it need to do if work. This is the basic principle of modularity in programming lanaugages. If we provide ciP in a function that doesn't need, we are breaking the *no more* part (specially when ciP is a complex object that explodes in a lot of sub-field, making more complex).

Regarding manteinability, it depends. If there is a overall cohesion in all the methods using the same signature, it may have sense. But, in this case, I think is more manteinable (as it is easier to understand, specially for newcomers to the Orion code) `toJson(void)` than `toJson(ciP)`. If I were a developer looking the code for the first time and I see `toJson(ciP)` I will start wondering why that funciton have an argument that it doesn't use and I could end getting lost in the details of ConnectionInfo.h in order to figure out what it is all that about.

So, at least for the toJson() family, I will try to adhere to the *no more and no less* principle. Let's see how manteinable is the code and the feedback of new developers after a while of applying it.

src/lib/ngsiNotify/QueueNotifier.cpp
kzangeli: Yeah, I was thinking about that same thing. Actually, mimeType we used ... a long time ago.

kzangeli: Included in Postponed List in #1875

src/lib/common/NotificationFormat.cpp
kzangeli: I will just make a quick change in the code and compile. 
Sometimes empirical methods are faster :-)

That said, personally I don't agree in using std::string as default type, instead of char*.
Unless 'needed' (to avoid using strcmp instead of '=='), it only adds overhead. We gain nothing and the broker gets less efficient. No, instead **char*** should be the default type to use and std::string only in the exceptions where we want to do the comparisons.

That is only my personal opinion, of course.

src/lib/ngsiNotify/QueueNotifier.cpp
kzangeli: If mixed, the values would be prefixed, like jsonValues, jsonKeyvalues, etc.
It's ok and it removes the use of an 'extra' parameter.
Using two parameters is cleaner, no doubt.
To me, both ways are ok

src/lib/rest/httpRequestSend.cpp
kzangeli: ok, seems like a good idea to make the parameter name more descriptive

src/lib/ngsiNotify/QueueNotifier.cpp
fgalan: They are conceptually different things... I tend to prefer not mixing churras with merinas ;)

src/lib/common/NotificationFormat.cpp
fgalan: As a general rule, I think that people (thinking to the kind of potential "audience" of this code) would prefer `std::string` as it provides a better abstraction level with a very little overhead penalty (let's be a bit humble and think that the implementers of std::string did a good job, ok? ;). In this case I think that "big savings in simplicity with little penalty in performance" principle (https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/contribution_guidelines.md#general-principles) applies.

A similar opinion appear in one of the cited referenced (I have filtered out some parts that I think there are unpolite):

> I think the better abstraction is worth it. It took many years to get a good string abstraction into C++. I don't believe that Bjarne Stroustroup, so famous for his conservative "only pay for what you use" dictum, would have permitted an obvious performance killer into the language. Higher abstraction is good.

As with the case of the default values, that doesn't mean that `char*` is forbidden, of course. There are places in which it could make fully sense using it, e.g. string-processing functions.

kzangeli: So, I did the compilation test and got the following errors:

    common/NotificationFormat.cpp: In function 'const std::string& notificationFormatToString(NotificationFormat, bool)':
    common/NotificationFormat.cpp:44: error: returning reference to temporary
    common/NotificationFormat.cpp:45: error: returning reference to temporary
    common/NotificationFormat.cpp:46: error: returning reference to temporary
    common/NotificationFormat.cpp:47: error: returning reference to temporary

Lines 44-47:

    case NGSI_V1_JSON:                return "JSON";
    case NGSI_V2_NORMALIZED:          return "normalized";
    case NGSI_V2_KEYVALUES:           return "keyValues";
    case NGSI_V2_VALUES:              return "values";

Lesson learned, you can't return a 'read-only' string using `const std::string&` as return type.

This could of course be avoided, by creating std::strings for each constant string that you want to return, like this:

    const std::string json       = "JSON";
    const std::string normalized = "normalized";
    const std::string keyValues  = "keyValues";
    const std::string values     = "values";
    
    switch (format)
    {
    case NGSI_V1_JSON:          return json;
    case NGSI_V2_NORMALIZED:    return normalized;
    case NGSI_V2_KEYVALUES:     return keyValues;
    case NGSI_V2_VALUES:        return values;
    ...

However, there must be a limit as to how much we have to sacrifice only to use std::string instead of char*.
If what we want here is *symmetry* between stringToNotificationFormat and notificationFormatToString I would propose to change the `const std::string` in stringToNotificationFormat for a `const char*`.
However, that would go against our 'general' rule to use `const std::string&` for input strings and `char*` for output strings ...

Perhaps it is best to just leave this **as is**.

src/lib/mongoBackend/MongoCommonRegister.cpp
kzangeli: Added to Postponed List in #1875

src/lib/mongoBackend/MongoCommonUpdate.cpp

src/lib/mongoBackend/mongoSubscribeContext.cpp
kzangeli: I looked them up:

1. serviceRoutines/postSubscribeContext.cpp:
```
    if (ciP->version != "v2")
    {
      parseDataP->scr.res.attrsFormat = NGSI_V1_JSON;
    }
```
2. serviceRoutines/postUpdateContextSubscription.cpp:
```
    if (ciP->version != "v2")
    {
      parseDataP->ucsr.res.attrsFormat = NGSI_V1_JSON;
    }
```
3. serviceRoutinesV2/postSubscriptions.cpp:
```
    if (parseDataP->scr.res.attrsFormat == NGSI_NO_NOTIFICATION_FORMAT)
    {
      parseDataP->scr.res.attrsFormat = NGSI_V2_NORMALIZED;
    }
```
1: Only set if v1.
2: Only set if v1.
3: Only set if previously unset (and if v2)

Nothing to worry about

src/lib/ngsi/ContextElementResponseVector.cpp
kzangeli: Fixed in c689381

src/lib/ngsi/ContextElement.cpp
kzangeli: Fixed in c689381

src/lib/ngsi10/NotifyContextRequest.cpp
kzangeli: NTC

src/lib/common/NotificationFormat.cpp
kzangeli: attrsName: Yes, all spec searched
JSON entity representation format:  its possible values aren't mentioned in the spec. JSON is a good candidate though. We've used it in the broker since we first implemented JSON ... 

NTC

src/lib/ngsiNotify/Notifier.h
kzangeli: Removed the default value in 3049d4e.

Actually, the default values wasn't used anywhere, I removed it and recompiled without any compilation errors.
I did find an incoherent thing, which I fixed: QueueNotifier, that inherits Notifier had a different signature of sendNotifyContextRequest. So, the virtual method wasn't used (most probably my fault, but now fixed - in 3049d4e)

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in d41bab6

crbrox:
LGTM

src/lib/common/NotificationFormat.cpp
fgalan: Pre-defining the string that the function may return add complexity, so in this case it seems better to use `const  char*` at the end.

What I wonder if we should consolidate a cosing style rule based in this experience (for not passing again the same discussion in a couple of PRs), maybe this:

> However, that would go against our 'general' rule to use const std::string& for input strings and char* for output strings ...

src/lib/mongoBackend/mongoSubscribeContext.cpp
fgalan: Great exahustive explanation. Thus, all under control. NTC.

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
fgalan: The "ugly" thing is having in a line one style and the other in the next one, eg:

brokerStart CB 0-255)
accumulatorStart --pretty-print

That uglyness doesn't happen with orionCurl due to, in the SHELL part, we are not mixing style.

Another advantage of having a different style is that is easy to make clear it is an arguments for the wrapper (e.g. `0-255`) that are not exactly the same than the ones that the wrapper will use to run the actual command (e.g. `-t 0-255`). That difference could help to debug the test harness suite in the cases of problems.

However, not so important at the present moment. It could be deferred from this PR and fixed (in a way or other) when that ugliness becomes unbearable. As you prefer.

doc/manuals/admin/database_model.md
```diff
@@ -284,8 +284,8 @@ Fields:
     fields: q, georel, geometry and/or coords (optional)
 -   **count**: the number of notifications sent associated to
     the subscription.   
--   **format**: the format to use to send notification, currently "JSON"
-    meaning JSON notifications in NGSIv1 format.
+-   **format**: the format to use to send notification, possible values are **JSON**
+    (meaning JSON notifications in NGSIv1 format), **normalized**, **keyvalues** and **values** (the last three used in NGSIv2 format).
```
fgalan: keyvalues -> keyValues

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
kzangeli: I would propose the following:

New issue about fixing brokerStart (you know I've been complaining about that function quite some time), where we 'migrate' brokerStart and probably a few more function to behave like orionCurl.  accumulator.py included.

doc/manuals/admin/database_model.md
kzangeli: Fixed in d2e6dad

test/functionalTest/cases/0876_entity_dates/filter_datecreated_sub.test
fgalan: Ok. Please, include a reference to this thread comment in the issue, for reference.

kzangeli: Issue #2055

fgalan:
LGTM

PR #2077, opened 2016-04-22
XML removal (part 11)

fgalan:
Part of #1862


src/lib/rest/orionReply.cpp
```diff
@@ -38,11 +38,9 @@ std::string orionReply(ConnectionInfo* ciP, const std::string& what, const std::
...
    std::string out = "";
 
-   out += startTag1("",     "orion");
-   out += startTag1("  ",   "log");
-   out += valueTag1("    ",  what, value);
-   out += endTag("  ");
-   out += endTag("");
+   out += '{';
+   out += valueTag1(" ", what, value);
+   out += '}';
```
kzangeli: While I like this change, it is not very backwards compatible ...
This needs to be clearly explained in the release notes for the next release.
However ... is it only used for `log` REST messages?

fgalan: My theory (a double check would be need although not indispensable) is that the JSON rendering in the response to log operations has never worked. You can check that reverting my change and running the 0000_log_something .test.

Thus, there isn't any break in backwards compatibility, as the functionality never worked, so there is nothing that breaks :)

However, maybe this could be add to CNR:

"Fix: JSON rsponse in log management operations"

Finally, it seems that the orionReply() function (at least this instance, in the case there are  many functions sharing the same name) is only called by the log response logic (as the `log` key is hardwired in the body) so probably it should be renamed to orionLogReply() or something like that.

kzangeli: We didn't have any functest covering this?
orionLogReply seems like an ok name for the function.

fgalan: Fixed in 796fa7c

kzangeli:
LGTM

PR #2079, opened 2016-04-25
Feature/1875 step ii

kzangeli:
Step II of #1875  (Step III still missing).

This PR includes:
* legacy to be accepted as a NotificationFormat
* New functest for notification **value** format (to check the order)
* Unify the two parameters 'notifyFormat' and 'renderMode'


src/lib/common/RenderFormat.cpp
```diff
@@ -28,49 +28,52 @@
...
+* stringToRenderFormat -
 */
-NotificationFormat stringToNotificationFormat(const std::string& s, bool noDefault)
+RenderFormat stringToRenderFormat(const std::string& s, bool noDefault)
 {
-  if (s == "JSON")       { return NGSI_V1_JSON;       }
-  if (s == "normalized") { return NGSI_V2_NORMALIZED; }
-  if (s == "keyValues")  { return NGSI_V2_KEYVALUES;  }
-  if (s == "values")     { return NGSI_V2_VALUES;     }
+  if (s == "JSON")         { return NGSI_V1_LEGACY;        }  // Older content in DB
```
fgalan: "Older content in DB" -> "DB content for NGSIv1 rendering due to legacy reasons"

(In the other way, it could be interpreted that before this PR the content in DB for NGSIv1 was "JSON" and now the content in DB for NGSIv1 is "legacy"... which is not corret. In DB "JSON" is still being used for NGSIv1 format).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -635,7 +635,7 @@ BSONObj fillQueryServicePath(const std::vector<std::string>& servicePath)
   // Note that by construction servicePath vector must have at least one element. Note that the
   // case in which first element is "" is special, it means that the SP were not provided and
   // we have to apply the default
-  if (servicePath[0] == "")
+  if ((servicePath.size() == 0) || (servicePath[0] == ""))
```
fgalan: You shouldn't need that `(servicePath.size() == 0)`... The statement above ("Note that by construction servicePath vector must have at least one element") should be correct... and if itsn't I'd like to know in which place we are constructing a servicePath breaking this rule.

test/functionalTest/cases/1316_create_subscription/post_subscriptions.test
```diff
@@ -228,8 +228,8 @@ Date: REGEX(.*)
     "id": "REGEX([0-9a-f]{24})",
     "notification": {
         "attrs": [
-            "humidity",
-            "temperature"
+            "temperature",
```
fgalan: I understand that you have needed to change attrs order in order to "align" with the orther in which they appear in notifications due to that in this case accumulator doesn't run in pretty-print mode?

(Just to check if I'm understanding correctly the change)

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats_order_in_values.test
```diff
@@ -0,0 +1,142 @@
...
+    }
+  },
+  "notification": {
+    "http": {
+      "url": "http://localhost:'$LISTENER_PORT'/notify"
+    },
+    "attrs": [ "A2", "A4", "A6", "A5", "A3", "A1" ]
+  },
+  "attrsFormat": "values",
+  "expires": "2050-04-05T14:00:00.00Z"
```
fgalan: Note that permanent subscriptions are now supported. Thus, in order to keep .test simpler, you can omit this field (except if the purpose of the test is to test something related with expiration, of course).

Not worth the time to modify existing .test, just a remark for the future. NTC.

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
```diff
@@ -797,7 +797,7 @@ TEST(mongoSubscribeContext, Ent1_AttrN_CNbis)
     EXPECT_FALSE(sub.hasField("lastNotification"));
     EXPECT_FALSE(sub.hasField("throttling"));
     EXPECT_STREQ("http://notify.me", C_STR_FIELD(sub, "reference"));
-    EXPECT_STREQ("JSON", C_STR_FIELD(sub, "format"));
+    EXPECT_STREQ("legacy", C_STR_FIELD(sub, "format"));
```
fgalan: Note that "legacy" is the word used in the API, but the keyword used at DB level doesn't change (due to backward compability) and is remains as "JSON".

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I believe I added this extra check to avoid a crash ...
Will try to reproduce

test/functionalTest/cases/1316_create_subscription/post_subscriptions.test
kzangeli: Before the fix, the attributes always came in alphabetic order. This made ordering for **values** impossible and was fixed. 

About pretty-print ... I thought the order in vectors weren't touched by the pretty-print ...

test/functionalTest/cases/1875_NGSIv2_notification_formats/NGSIv2_notification_formats_order_in_values.test
kzangeli: True.
This was just a copy-paste from an older test.

test/functionalTest/cases/1316_create_subscription/post_subscriptions.test
fgalan: > About pretty-print ... I thought the order in vectors weren't touched by the pretty-print ...

That's right... but I don't refere to the pretty print of the GET subscriptions operation, but the pretty print of the accumulator, in which the attributes came in a key-map where the pretty printer will act. For this particular.test (the one in which this commen has been inserted) that comment doesn't really applies (as no accumulator server is started) but maybe for others .test in which attrs order has been also changed it applies (I haven't reviewed in detail...).

kzangeli: Not sure about the possible problem here.
The attributes come out in the same order as they were sent in the request (and it's not alphabetical order)

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
kzangeli: Fixed in cd8a9b3

test/functionalTest/cases/1316_create_subscription/post_subscriptions.test
fgalan: Not really a problem. Let's discuss by skype.

src/lib/common/RenderFormat.cpp
fgalan: Seems it is also fixed by cd8a9b365496fa90d2374b7b826f43942616034b

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I removed the extra check and ran a full test round.
All good.
Extra check removed in commit 939f656

src/lib/common/RenderFormat.cpp
kzangeli: Yes, that is true ...

test/functionalTest/cases/1316_create_subscription/post_subscriptions.test
fgalan: Discussed. NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1457,14 +1469,36 @@ static bool processSubscriptions
...
-                                                 trigs->reference,
-                                                 trigs->notifyFormat,
-                                                 tenant,
-                                                 xauthToken,
-                                                 fiwareCorrelator))
+
+    CachedSubscription*  cacheP = subCacheItemLookup(tenant.c_str(), trigs->cacheSubId.c_str());
+    bool                 notificationSent;
+
+    if (cacheP)
```
fgalan: In order to avoid duplication, I'd suggest:

```
std::vector<std::string> attrV;
if (cacheP)
{
  attrV = cacheP->attributes;
}
else
{
  attrV = trigs->attrL.attributeV;
}

notificationSent = processOnChangeConditionForUpdateContext(notifyCerP,
                                                                  trigs->attrL,
                                                                  mapSubId,
                                                                  trigs->reference,
                                                                  trigs->renderFormat,
                                                                  tenant,
                                                                  xauthToken,
                                                                  fiwareCorrelator,
                                                                  attrV);
```

(Above fragment probably don't work due to typing, but I guess you get the idea...)

test/unittests/commonMocks.h
```diff
@@ -161,19 +161,18 @@ class NotifierMock : public Notifier {
...
+    MOCK_METHOD7(sendNotifyContextRequest, void(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, RenderFormat renderFormat, const std::vector<std::string>&  attrsFilter));
     MOCK_METHOD5(sendNotifyContextAvailabilityRequest, void(NotifyContextAvailabilityRequest* ncar, const std::string& url, const std::string& tenant, const std::string& fiwareCorrelator, Format f));
 
     /* Wrappers for parent methods (used in ON_CALL() defaults set in the constructor) */
-    void parent_sendNotifyContextRequest(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, NotificationFormat notifyFormat)
+    void parent_sendNotifyContextRequest(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, RenderFormat renderFormat, const std::vector<std::string>&  attrsFilter)
     {
-      return Notifier::sendNotifyContextRequest(ncr, url, tenant, xauthToken, fiwareCorrelator, notifyFormat);
```
fgalan: Just curiosity... why did you remove `return`? Is not working in your system with it?

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -204,7 +204,7 @@ static void prepareDatabase(void) {
                         "lastNotification" << 15000000 <<
                         "throttling" << 60 <<
                         "reference" << "http://notify1.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -217,7 +217,7 @@ static void prepareDatabase(void) {
                         "expiration" << 20000000 <<
                         "lastNotification" << 25000000 <<
                         "reference" << "http://notify2.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -561,7 +561,7 @@ TEST(mongoUpdateContextSubscription, updateDuration)
 
     EXPECT_EQ(60, sub.getIntField("throttling"));
     EXPECT_STREQ("http://notify1.me", C_STR_FIELD(sub, "reference"));
-    EXPECT_STREQ("JSON", C_STR_FIELD(sub, "format"));
+    EXPECT_STREQ("legacy", C_STR_FIELD(sub, "format"));
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -272,7 +272,7 @@ static void prepareDatabase(void) {
                         "lastNotification" << 55000000 <<
                         "throttling" << 10 <<
                         "reference" << "http://notify5.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -258,7 +258,7 @@ static void prepareDatabase(void) {
                         "expiration" << 40000000 <<
                         "lastNotification" << 45000000 <<
                         "reference" << "http://notify4.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -244,7 +244,7 @@ static void prepareDatabase(void) {
                         "expiration" << 30000000 <<
                         "lastNotification" << 35000000 <<
                         "reference" << "http://notify3.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
```diff
@@ -231,7 +231,7 @@ static void prepareDatabase(void) {
                         "expiration" << 20000000 <<
                         "lastNotification" << 25000000 <<
                         "reference" << "http://notify2.me" <<
-                        "format" << "JSON" <<
+                        "format" << "legacy" <<
```
fgalan: It seems this change wasn't correctly "rollbacked".

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: In order to avoid dupliction, you suggest to duplicate the attribute vector every time?
No, we can't do that.
What we can do, I am almost 100% sure of it, is to always send it trigs->attrL.attributeV.

I left it like this to 'prepare' for different treatment between cache and non-cache.
I really think we should implement that, merge TriggeredSubscription and CachedSubscription and stop allocating freeing like we are doing now.
However, the issue already exists and no need to prepare for it in the code.
I will try to unify the two calls, using trigs->attrL.attributeV for both of them. That should work.

[ I actually planned to do that before sending the PR, but I forgot about it in the end ]

test/unittests/commonMocks.h
kzangeli: The methid is void but has a return ...
Does that work in any system?
It sure shouldn't !!!

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Hmmm, I did a grep on "legacy" and after the changes, all unit tests worked.
I will look into it (in all places, no need to add more comment about the same in the PR)

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: This line here is the reason everything works.
I guess my grep was in combination with something else, like a function call or something ...

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: What I'm suggesting is to avoid two calls that are almost equal in the code. The code would be more maintenable having just one call (with two calls if we change something in one of them due to a refactor, e.g. `tenant` by `tenantX`, and we forgot the other... it is error prone).

> I am almost 100% sure of it, is to always send it trigs->attrL.attributeV.

That would also solve the problem.

test/unittests/commonMocks.h
fgalan: But it is a fact that it worked with `return` until this PR, isn't it?

Assuming that, it is a *clever* idea to modify a code that we are not fully sure how it works (I think that I took that "parent wrapping" pattern from some place when I wrote the original code in this part... and in that place I guess that they were using `return`... for unknown reasons but maybe powerful enough)?

Not really sure what is better, anyway :)

kzangeli: It is very strange ...
I did a short test just to see how it 'should' work:

XXX.cpp:

    void a(void)
    {
      return 10;
    }

    int main(int argC, char* argV[])
    {
      a();
      return 0;
    } 

This is the output from the compilation:

    kz@hoppe:fiware-orion> make XXX
    g++     XXX.cpp   -o XXX
    XXX.cpp: In function 'void a()':
    XXX.cpp:4: error: return-statement with a value, in function returning 'void'
    make: *** [XXX] Error 1

Which makes a lot of sense.
I would **not** leave it the way it was, knowing that it is incorrect.
Why doesn't the compiler complain? I have no idea. Strange things with the mocks, perhaps ...

However, I believe I found this 'error' as my compiler complained ...

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/mongoBackend/mongoUpdateContextSubscription_test.cpp
kzangeli: Fixed in 4f26501

test/unittests/commonMocks.h
fgalan: As you prefer... fixit it or NTCice it :) to close this thread.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1343,10 +1354,11 @@ static bool processOnChangeConditionForUpdateContext
   const AttributeList&             attrL,
   std::string                      subId,
   std::string                      notifyUrl,
-  NotificationFormat               notifyFormat,
+  RenderFormat                     renderFormat,
   std::string                      tenant,
   const std::string&               xauthToken,
-  const std::string&               fiwareCorrelator
+  const std::string&               fiwareCorrelator,
+  std::vector<std::string>         attributeV
```
fgalan: The type doesn't conform the in/out criteria. I mean, it should be either

const std::vector<std::string> -> for "read only" parameters

or

std::vector<std::string>*   -> otherwise.

In addition, it this attribute is intended for define the attribute order, maybe `attrsOrderV` (or something like that) would be a better name.

src/lib/ngsiNotify/Notifier.h
```diff
@@ -44,12 +44,13 @@ class Notifier
...
-                                        const std::string&    xauthToken,
-                                        const std::string&    fiwareCorrelator,
-                                        NotificationFormat    notifyFormat);
+  virtual void sendNotifyContextRequest(NotifyContextRequest*            ncr,
+                                        const std::string&               url,
+                                        const std::string&               tenant,
+                                        const std::string&               xauthToken,
+                                        const std::string&               fiwareCorrelator,
+                                        RenderFormat                     renderFormat,
+                                        const std::vector<std::string>&  attrsFilter);
```
fgalan: Just a filter or it also acts as a orderding criteria? In that case attrsOrder would be better...

test/unittests/commonMocks.h
kzangeli: OK, if it's up to me I prefer to leave it the way I left it.

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: True, the whole vector is copied to the stack like this ... That is not good

src/lib/ngsiNotify/Notifier.h
kzangeli: Sure, attrsOrder might be a better name

kzangeli: Fixed in a5e94e2

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in a5e94e2

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in a5e94e2

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1394,7 +1406,7 @@ static bool processOnChangeConditionForUpdateContext
   // FIXME: we use a proper origin name
   ncr.originator.set("localhost");
 
-  getNotifier()->sendNotifyContextRequest(&ncr, notifyUrl, tenant, xauthToken, fiwareCorrelator, notifyFormat);
+  getNotifier()->sendNotifyContextRequest(&ncr, notifyUrl, tenant, xauthToken, fiwareCorrelator, renderFormat, attributeV);
```
fgalan: attrsOrderV ?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1343,10 +1354,11 @@ static bool processOnChangeConditionForUpdateContext
   const AttributeList&             attrL,
   std::string                      subId,
   std::string                      notifyUrl,
-  NotificationFormat               notifyFormat,
+  RenderFormat                     renderFormat,
   std::string                      tenant,
   const std::string&               xauthToken,
-  const std::string&               fiwareCorrelator
+  const std::string&               fiwareCorrelator,
+  std::vector<std::string>&        attributeV
```
fgalan: `const` missing?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -1768,19 +1768,20 @@ AttributeList subToAttributeList(const BSONObj& sub)
...
   const std::string&               notifyUrl,
-  NotificationFormat               notifyFormat,
+  RenderFormat                     renderFormat,
   const std::string&               tenant,
   const std::string&               xauthToken,
   const std::vector<std::string>&  servicePathV,
   Restriction*                     resP,
-  const std::string&               fiwareCorrelator
+  const std::string&               fiwareCorrelator,
+  const std::vector<std::string>&  attrsFilter
```
fgalan: attrsFilter -> attrsOrder ?

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: In addition, maybe attributeV -> attrsOrderV ?

src/lib/ngsiNotify/QueueNotifier.cpp
```diff
@@ -60,7 +60,16 @@ int QueueNotifier::start()
...
-void QueueNotifier::sendNotifyContextRequest(NotifyContextRequest* ncr, const std::string& url, const std::string& tenant, const std::string& xauthToken, const std::string& fiwareCorrelator, NotificationFormat notifyFormat)
+void QueueNotifier::sendNotifyContextRequest
+(
+  NotifyContextRequest*            ncr,
+  const std::string&               url,
+  const std::string&               tenant,
+  const std::string&               xauthToken,
+  const std::string&               fiwareCorrelator,
+  RenderFormat                     renderFormat,
+  const std::vector<std::string>&  attrsFilter
```
fgalan: attrsOrder? (in this case it seem clear, as this method is like the one in the parent Notifeir class, which uses attrsOrder)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Fixed in 16149b1

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 16149b1

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 16149b1

src/lib/ngsiNotify/QueueNotifier.cpp
kzangeli: True.
Fixed in 3e0513e

src/lib/ngsi/ContextAttributeVector.cpp
```diff
@@ -204,7 +202,7 @@ std::string ContextAttributeVector::toJson(bool isLastElement, const std::string
 
   uniqueMap.clear();
 
-  if (attrsFilter == "")
+  if (attrsFilter.size() == 0)
```
crbrox: Perhaps `attrsFilter.empty()` ?

crbrox:
LGTM

src/lib/ngsi/ContextAttributeVector.cpp
kzangeli: I always use size(), it works for more stuff.
But sure, empty() is good here.

crbrox: NTC

fgalan:
LGTM

Issue #978, opened 2015-06-11
Extend the pattern concept to attribute values


fgalan:
OMA NGSI only considers patterns in entity ID. To that end, the `?idPattern` parameter has been defined (see https://github.com/telefonicaid/fiware-orion/issues/969).

However it would be useful to implement also pattern for entity type and attribute values. A way of fitting it in OMA NGSI has to be found, but, by the moment, the following has been suggested:
- `?typePattern` for entity type pattern
- New operator for `?q` filter (see https://github.com/telefonicaid/fiware-orion/issues/967), e.g. `^=` so `?q=color^=b.*` means "all the colors that start with letter 'b'".

fgalan:
As alterantive, `=~` could be used as operator for attribute pattern matching.

fgalan:
Another possiblity is to use `~=` which seems to be more aligned with Perl or Ruby. The final decision, either `~=` or `=~` (or other :) is deferred until the time to implement this issue comes.

fgalan:
Originally this issue covers also patterns for entity types. However, that part has been "moved" to https://github.com/telefonicaid/fiware-orion/issues/1853

fgalan:
> The final decision, either ~= or =~ (or other :) is deferred until the time to implement this issue comes.

Time has come :)

I'll finally use ~= which is aligned with

!=
>=
<=
~=

fgalan:
Implementation in PR #2081. Documentation changes in PR #2088.

fgalan:
Moving to QATestPending milestone while the functional test covering this are implemented.

iariasleon:
still it has not been developed `?typePattern` for entity type pattern? Only New operator for `?q` filter

fgalan:
That's correct: the present issue is about `~=` for attribute values pattern. Note that `?typePattern` has its own issue (#1853).

fgalan:
Issues in milestone QATestPending are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeles with QATestPending label in the case thinks change and we need to recover it.

PR #2081, opened 2016-04-25
Feature/978 attribute patterns

fgalan:
Implements #978 (except .apib changes)

@crbrox 
@kzangeli 
@fgalan


src/lib/cache/subCache.cpp
```diff
@@ -77,6 +77,8 @@ EntityInfo::EntityInfo(const std::string& _entityId, const std::string& _entityT
 
   if (isPattern)
   {
+    // FIXME P5: recomp error should be captured? have a look to other usages of regcomp()
+    // in order to see how it works
```
kzangeli: Doesn't sound like a bad idea.
Capture error and return error response ...

http://linux.die.net/man/3/regcomp

We should create an issue about it

src/lib/rest/StringFilter.cpp
```diff
@@ -497,7 +522,7 @@ bool StringFilterItem::parse(char* qItem, std::string* errorStringP)
   }
   if (forbiddenChars(lhs, ""))
   {
-    *errorStringP = "invalid character found in URI param /q/";
+    *errorStringP = "invalid character found in URI param /q/";  // fix this (and others)
```
kzangeli: Is this a FIXME ?

src/lib/cache/subCache.cpp
fgalan: I would create it but with a broader scope. I'd like to include an anaylsis of std::regex in order to see if using it will make regex management simpler (we have had many problems with them in this PR, suposed to be simple from a functionaly point of view).

src/lib/rest/StringFilter.cpp
fgalan: It seems I missed something... looking to it...

src/lib/cache/subCache.cpp
fgalan: Issue created: https://github.com/telefonicaid/fiware-orion/issues/2082

src/lib/rest/StringFilter.cpp
fgalan: I don't remember where this comment come from... Maybe in a early stage of the work, when I wrongly thought that check could be a problem.

However, make test ensures that all works fine, so I have removed the comment.

Fixed in 5ac5ef1

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1285,8 +1285,7 @@ static bool addTriggeredSubscriptions_noCache
           }
           else
           {
-            trigs->stringFilterSet(stringFilterP);  // Clones the string filter
-            delete stringFilterP;
+            trigs->stringFilterSet(stringFilterP);
```
fgalan: Maybe the comment should be kept?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1447,7 +1446,7 @@ static bool processSubscriptions
     }
 
     /* Check 2: String Filter */
-    if (!trigs->stringFilter.match(notifyCerP))
+    if (trigs->stringFilterP && !trigs->stringFilterP->match(notifyCerP))
```
fgalan: use `if (() && ())`

(Who should do the change? @kzangeli is this your part or mine?)

src/lib/rest/StringFilter.cpp
```diff
@@ -846,6 +892,7 @@ bool StringFilter::parse(const char* q, std::string* errorStringP)
     }
 
     str = NULL;  // So that strtok_r continues eating the initial string
+    item.patternValueToBeFreed = false;
```
fgalan: Why this explicit setting is needed? The StringFilterItem constructor at line L882 sets `patternValueToBeFreed` to `false`. What could happend between L882 and this one that "reset" that setting an needs to do it again?

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Looking to the stringFilterSet() code it seems that now it is not cloning, so removing the comment is ok.

I have some doubt, but I'll use skype.

NTC

src/lib/rest/StringFilter.cpp
fgalan: For reference, the constructor code is here: https://github.com/telefonicaid/fiware-orion/pull/2081/files#diff-ef659d8f6e392bb5d95fdfbcd0997815R46

src/lib/rest/StringFilter.h
```diff
@@ -61,6 +63,7 @@ typedef enum StringFilterOp
 typedef enum StringFilterValueType
 {
   SfvtString,
+  SfvtPattern,   // really needed?
```
fgalan: Leftover. I'll fix it...

src/lib/mongoBackend/TriggeredSubscription.cpp
```diff
@@ -132,12 +134,7 @@ std::string TriggeredSubscription::toString(const std::string& delimiter)
 *
 * TriggeredSubscription::stringFilterSet - 
 */
-void TriggeredSubscription::stringFilterSet(StringFilter* stringFilterP)
+void TriggeredSubscription::stringFilterSet(StringFilter* _stringFilterP)
```
fgalan: At the end (dicussed at skype) we need to get back to clone semantics (cache refresh logic could remove the StringFilter while TriggeredSubscription is using it... and we already discarded the usage of sems for this while we worked on PR #1900).

Thus, explicit copy each field would be a more clear solution than the "obscure" `stringFilter = *stringFilterP`. However, it is not clear yet how to clone a `regex_t` (open question at SOF: http://stackoverflow.com/questions/36846426/best-way-of-cloning-compiled-regex-t-struct-in-c). If we don't find a better solution, then we need to re-compile it using regcomp() from the original string.

src/lib/rest/StringFilter.h
fgalan: Fixed in 250f6abafd47c99430b6db50ccca49908f41b9b1

src/lib/rest/StringFilter.cpp
fgalan: Auto-answer: the parse() processing make change the field between L882 and this point in code.

src/lib/cache/subCache.cpp
```diff
@@ -735,7 +737,21 @@ void subCacheItemInsert
 
   if (stringFilterP != NULL)
   {
-    cSubP->expression.stringFilter = *stringFilterP;  // Object copy
+    std::string  errorString;
+
+    if (!cSubP->expression.stringFilter.fill(stringFilterP, &errorString))
+    {
+      LM_E(("Runtime Error (%s)", errorString.c_str()));
```
fgalan: Looking to fill() code, it seems that LM_E is generated inside that function. Thus, this LM_E should be removed, to avoid over-logging.

src/lib/cache/subCache.cpp
```diff
@@ -735,7 +737,21 @@ void subCacheItemInsert
...
+
+      //
+      // NOTE
+      //   Here, the cached subscription should have a String Filter but as clone failed, it doesn't
+      //   The subscription is already in mongo and hopefully this erroneous situation is fixed 
+      //   once the sub-cache is refreshed.
+      //
+      //   This 'but' should be minimized once the issue 2082 gets implemented.
+      //   [ Only reason for clone to fail (apart from out-of-memory) seems to be an invalid regex ]
```
fgalan: Appart of the LM_E (which I suggest to remove above) I understand that this `{ }` block is left empty just to put the comment (ok with that, I'm only checking if I'm understand correctly).

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: In addition, what about 

trigs->stringFilterP != NULL

I know... NULL is 0 and it works as it is :) However, doing that will make this code more readable.

src/lib/cache/subCache.cpp
```diff
@@ -735,7 +737,21 @@ void subCacheItemInsert
...
-    cSubP->expression.stringFilter = *stringFilterP;  // Object copy
+    std::string  errorString;
+
+    if (!cSubP->expression.stringFilter.fill(stringFilterP, &errorString))
+    {
+      LM_E(("Runtime Error (%s)", errorString.c_str()));
+
+      //
+      // NOTE
+      //   Here, the cached subscription should have a String Filter but as clone failed, it doesn't
```
kzangeli: clone -> fill()

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -329,7 +329,20 @@ int mongoSubCacheItemInsert
 
   if (stringFilterP != NULL)
   {
-    cSubP->expression.stringFilter = *stringFilterP;
+    std::string errorString;
+
+    if (!cSubP->expression.stringFilter.fill(stringFilterP, &errorString))
+    {
```
fgalan: Similar comment than above regarding empty `{ }` block.

src/lib/cache/subCache.cpp
kzangeli: Fixed in 1d38359

src/lib/rest/StringFilter.cpp
```diff
@@ -48,6 +96,13 @@ StringFilterItem::~StringFilterItem()
 {
   stringList.clear();
   numberList.clear();
+
+  LM_W(("FGM this is %s, destroying", stringValue.c_str()));
```
fgalan: Ooooh. This time was my fault!! :D

@kzangeli could you manage to grep for "FGM" and remove potential traces, pls? (you are now in the branch and it is easier for you).

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 8833f26

src/lib/rest/StringFilter.cpp
```diff
@@ -1187,8 +1294,76 @@ bool StringFilter::match(ContextElementResponse* cerP)
...
+
+    if (!sfi.fill(&filters[ix], errorStringP))
+    {
+      delete sfP;
+      return NULL;
+    }
+
+    sfP->filters.push_back(sfi);
+
+    // Next line is to avoid double-free at StringFilterItem destructor
```
fgalan: I like the wording you use for an analogous situation at https://github.com/telefonicaid/fiware-orion/pull/2081/files#diff-ef659d8f6e392bb5d95fdfbcd0997815R938. Please, replace my original comment in that part wiith yours.

src/lib/rest/StringFilter.cpp
```diff
@@ -1187,8 +1294,76 @@ bool StringFilter::match(ContextElementResponse* cerP)
...
+
+    if (!sfi.fill(&sfP->filters[ix], errorStringP))
+    {
+      LM_E(("Runtime Error (error filling StringFilterItem: %s)", errorStringP->c_str()));
+      return false;
+    }
+    
+    filters.push_back(sfi);
+
+    // Next line is to avoid double-free at StringFilterItem destructor
```
fgalan: Same here, pls...

src/lib/cache/subCache.cpp
kzangeli: I considered that when adding the LM_E ... 
And yes, you're right, might provoke over-logging.
Perhaps we should never log if we 'send errorStringP' to calling function

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: So, you propose to leave the if {}   with just a comment inside?
I'm not aganst it, it's a bit strange, but as it gives more info to the comment inside somehow I find it OK

src/lib/rest/StringFilter.cpp
kzangeli: Sure

kzangeli: Just one line.
Fixed in 35ff8c1

src/lib/cache/subCache.cpp
kzangeli: Fixed in 4d9b038 (LM_E removed)

src/lib/cache/subCache.cpp
kzangeli: Fixed in 4d9b038

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in 8ed5c80

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in de9256a

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: != NULL fixed in 9b471cc

src/lib/mongoBackend/TriggeredSubscription.cpp
kzangeli: Fixed in a few recent commits ...

kzangeli:
LGTM

src/lib/mongoBackend/mongoSubCache.cpp
fgalan: Maybe is better to remove the empty { } block and put the comment above the if. Makes sense?

kzangeli: Fixed in af88915

fgalan:
LGTM

crbrox:
LGTM

PR #2090, opened 2016-04-26
ADD S13 rule and object passing and strings return design patterns

fgalan:
@crbrox 
@kzangeli 

Please implement directly in the PR any English language fix.

### DIFF
--- a/doc/manuals/contribution_guidelines.md
+++ b/doc/manuals/contribution_guidelines.md
@@ -685,14 +685,14 @@ If you plan to use some of them, please consult before with the core developers.
   * The flavor of this project is biased to C clearly. Many plain C libraries are used and mixing exceptions and
   traditional if-checking method should be avoided as much as possible.
 * Object passing. In order to avoid inefficient object copies when calling functions, the following criteria apply:
-  * If the function doesn't modify the object (i.e. "read only"), then use const references, e.g. `const BigFish& bf`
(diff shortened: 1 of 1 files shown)


doc/manuals/contribution_guidelines.md
```diff
@@ -672,3 +684,18 @@ If you plan to use some of them, please consult before with the core developers.
...
   traditional if-checking method should be avoided as much as possible.
+* Object passing. In order to avoid inefficient object copies when calling functions, the following criteria apply:
+  * If the function doesn't modify the object (i.e. "read only"), then use const references, e.g. `const BigFish& bf`
+  * If the function modifies the object (i.e. "read/write" or "write"), then use a pointer type, e.g: `BigFish* bf`. Note that in this case we could also use a reference, but we prefer pointers to make more explicit that the object is to be modified by the function.
+
+```
+void myFunction(const& BigFish in, BigFish* out)
+```
+
+* Strings return. In order to avoid inefficient object copies when returning strings, the  `const std::string&` return type is preferred. In the case that pattern cannot be used (e.g. when literal string such as "black", "red", etc. are used in the function) then `const char*` should be used as return type.
```
fgalan: If you find a better way of explaining in which cases `const std::string&` could be a problem and `const char*` has to be used, please go ahead.

kzangeli: No, this is the only case I know of thus far. Probably it is the only one where const std::string& cannot be used, as "xxx" is in a readonly segment.

kzangeli: NTC

kzangeli:
LGTM

crbrox:
LGTM

PR #2097, opened 2016-04-27
FIX problems running some tests under coverage target

fgalan:
@kzangeli 
@crbrox 

Note this PR slightly changes the code. I understand that is safe to merge during the de-leak period, as the changes are very minor and with controlled impact from the point of view of leaks.

### DIFF
--- a/test/functionalTest/testHarness.sh
+++ b/test/functionalTest/testHarness.sh
@@ -134,7 +134,7 @@ function exitFunction()
   errorFile=$5
   forced=$6
 
-  echo -n "(ERROR $exitCode - $errorText) "
+  echo -n "(FAIL $exitCode - $errorText) "
 
   if [ "$stopOnError" == "on" ] || [ "$forced" == "DIE" ]
   then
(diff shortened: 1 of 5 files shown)


test/functionalTest/testHarness.sh
```diff
@@ -134,7 +134,7 @@ function exitFunction()
   errorFile=$5
   forced=$6
 
-  echo -n "(ERROR $exitCode - $errorText) "
+  echo -n "(FAIL $exitCode - $errorText) "
```
fgalan: Why this is needed? Before this PR we can have outputs like:

ERROR (first trial)
ERROR (second trial)
OK (last trial)

that corresponds to a righ execution, but when running test under coverage target the FAIL_REGULAR_EXPRESSION clause at https://github.com/telefonicaid/fiware-orion/pull/2097/files#diff-5961c9f02197252964031a47b2e1a240L30 makes this to be interpreted as a failed execution.

Which this change (which uses FAIL in the last trial to mark definitive failure) and FAIL_REGULAR_EXPRESSION changes it seems to work now.

kzangeli:
LGTM

crbrox:
LGTM

PR #2105, opened 2016-04-29
FIX adapt unittest and ft run to remote mongodb servers

fgalan:
@kzangeli 
@crbrox


src/lib/common/globals.cpp
```diff
@@ -188,6 +189,69 @@ void orionInit
 
 /* ****************************************************************************
 *
+* mongoInit -
+*/
+void mongoInit
```
kzangeli: Why in common/globals.cpp?
Would it not be better to have the function in mongoBackend?
mongoBackend/mongoInit.h/cpp ...

test/functionalTest/harnessFunctions.sh
```diff
@@ -428,6 +428,16 @@ function localBrokerStart()
...
   ps=$(ps aux | grep contextBroker)
   dMsg $ps
+
+  # Some times (specially when using remote DB) CB needs some time to connect DB and
+  # the test execution needs to a guard time. The righ value for CB_WAIT_AFTER_START
+  # needs to be find empirically (it seems that when DB is localhost this time is
+  # not needed at all)
+  #
+  # (A better solution will be to detect when CB is ready to start the test, but
+  # it can be complex. Note that and check above ensures that the port is ready, but
```
kzangeli: Note that **the** check above ?

test/unittests/main_UnitTest.cpp
```diff
@@ -110,9 +129,26 @@ int main(int argC, char** argV)
...
+  //
+  // However, if we introduce gtest arguments, it breaks, e.g:
+  //
+  // BUILD_UNITTEST/test/unittests/unitTest -t 0-255 -dbhost localhost --gtest_filter=*.*
+  // ...
+  // option '--' is a boolean option - cannot have a value (gtest_filter=*.*)
+  //
+  // I don't know how to fix this... paArgs wisdom is required :)
+
+  paParse(paArgs, argC, (char**) argV, 1, false);
```
kzangeli: ok, interesting problem.
I will have to look into this ...

kzangeli: Fixed in e5391f7    (However, I didn't remove the FIXME and I added an outdeffed loop in the beginning of main(), just in case you want to see how exactly the unit test is called ...
This must still be cleaned before merging.

test/functionalTest/harnessFunctions.sh
```diff
@@ -428,6 +428,16 @@ function localBrokerStart()
...
+  # it doesn't guarantee that DB connection is ok).
```
kzangeli: Wouldn't a similar 'nc check' with the port 27017 do the trick ?

Or even something like this:

    echo 'version();' | mongo --host $CB_DATABASE_HOST $CB_DB_NAME

In a loop, capturing the output

src/lib/common/globals.cpp
fgalan: I chose globals.cpp/h due to orionInit() was already there. However, you are right that probably makes more sense as part of mongoBackend. To be moved.

test/unittests/main_UnitTest.cpp
fgalan: Just be sure, with the fix implemented in https://github.com/telefonicaid/fiware-orion/commit/e5391f7e613ffe653e7ddb7e69b4c88cad62e4c7, the `--gtest*` arguments... are accepted and nullified (i.e. they haven't any effect)? Or are they accepted and used "internally" by the unit test library in the way they are expected (e.g. `--gtest_filter` to filter test to execute, etc.)?

I understand is the second one, but looking to the commit is not so clear to me... It would be great to unserstand how to "configure" `paArgs[]` if I need this in some other place.

test/functionalTest/harnessFunctions.sh
fgalan: nc on 27017 doesn't work, as we are not waiting for the database service to start (mongod will be typically be running for days or months), but waiting for CB to connect to DB.

Although restInit() cames after mongoInit() is seems that it doesn't suffice... Probably some is happening asynchronously. Maybe we should implement some mechanim in mongoInit() so the function doesn't return if the connection has not succesuffy stablished (but not in this PR but as a separate issue, by the moment CB_WAIT_AFTER_START would suffice).

test/unittests/main_UnitTest.cpp
kzangeli: With this change, paLib simply accepts the arguments and does nothing with them.
Later they are passed to InitGoogleMock and used there by gtest.
If there is ever any need for more args, just add them to the paArgs array.

Before, as paLib complained about unknown params and doing an exit(), we never got that far.

fgalan:
We are making some progress...

Unit test works perfectly:

```
make unit_test MONGO_HOST=localhost
make unit_test MONGO_HOST=qa-orion-cep-01
make unit_test MONGO_HOST=qa-orion-cep-01 TEST_FILTER=*.*
make unit_test MONGO_HOST=qa-orion-cep-01 TEST_FILTER=mongoQueryContextRequest.*
```

Regarding, functional test, some test are failing (running with `CB_WAIT_AFTER_START=2s CB_DATABASE_HOST=qa-orion-cep-01 ./testHarness.sh`):

```
  o  1323_put_replacing_dot_in_attrnames/put_replacing_dot_in_attrnames.test
  o  1661_bad_input_on_notification_error/bad_input_on_notification_error.test
  o  1661_bad_input_on_notification_error/bad_input_on_notification_error_api_v1.test
  o  1661_bad_input_on_notification_error/bad_input_on_notification_error_with_relogged_alarms.test
  o  1661_bad_input_on_notification_error/bad_input_on_notification_error_with_relogged_alarms_api_v1.test
```

Cause of the problem is yet under investigation.

test/functionalTest/harnessFunctions.sh
fgalan: Issue created https://github.com/telefonicaid/fiware-orion/issues/2123

src/lib/common/globals.cpp
fgalan: Fixed in 1de055328d99e5e762d9320837c8842447858325

test/functionalTest/harnessFunctions.sh
fgalan: Fixed in 1de055328d99e5e762d9320837c8842447858325. In fact, the whole paragraph has been re-written to refer to the issue just created about this.

test/unittests/main_UnitTest.cpp
fgalan: Cleaned in 37e3d486fafb2fc6e5c98da2319ffe85a5e45488

fgalan:
Failing test fixed in 063c65ad55feddf36a9b40ea91b62db0280baeb6.

The PR is ready for final review and (hopefully) merging.

kzangeli:
LGTM

crbrox:
LGTM

PR #2113, opened 2016-05-03
FIX crash when updating entities which corresponding document at DB is somehow large

### LINKED ISSUE #2112: Crash when updating entities which corresponding document at DB is somehow large
As described at http://stackoverflow.com/questions/36917731/context-broker-crashing-with-certain-update-queries

fgalan:
Fixes #2112 

Unit test pass. Test harness passed. Valgrind result in 2 test failing:

```
Test 001/2: 1328_pause_subscriptions/pause_subscriptions_notif_cache ................................................................................ FAILED (lost: 16).
Test 002/2: 1661_bad_input_on_notification_error/bad_input_on_notification_error_with_relogged_alarms_api_v1 ........................................ FAILED (lost: 16)
```
Double-check needed in order to discard glithes and specific local system potential problems (my test has been done in Debian 8.x).

I'm still trying to understand better why this fix works. Please have a look http://stackoverflow.com/questions/36998896/storing-bsonobj-object-in-stdvector-crashes-with-big-documents before/after reviewing this PR.

@crbrox 
@kzangeli


fgalan:
PR is ready from a code contributions point of view (LGTM pending, anyway). We have dicussed internally that leaks could be due to another cause (maybe related with csub sem) and be even false leaks.

A RPM has been generated and provide to users which report this problem. Ok from their side is also required before merging.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2923,7 +2924,9 @@ void processContextElement
       alarmMgr.dbError(details);
       continue;
     }
-    results.push_back(r);
+    // we need to getOwned() here, otherwise we have empirically found that bad things may happend with long BSONObjs
```
kzangeli: we need to **use** getOwned() here ... bad things may **happen**

fgalan:
User report was ok.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Fixe in 7436f936c3af2e2a4363dfd3df135a351882caae

kzangeli:
LGTM

Issue #2114, opened 2016-05-03
Throttling set to -1 at creating v2 subscriptions


kzangeli:
It seems we have a bug in the sub-cache with newly created subscriptions in v2.

What seems to be a method meant only for v1 subscriptions is called in subCacheItemInsert:

    cSubP->throttling = scrP->throttling.parse();

This method parses the string of the Throttling of the subscription, and as in v1 the throttling comes in as a string, this makes sense. In v2 however, throttling is an integer and the parsing stops making sense.

fgalan:
Just to understand the problem from an user perspective... does this bug mean that thorttling was not correctly stored in cache when the user used NGSIv2 subscription API? 

E.g. user creates a subscription with `throttling: 5`, then the subscriptions is stored in cache as -1. Thus, when a updateCtx arrives, throtlling has no effect, as -1 means "no thottrling" in the mongoBackend/Notifier logic that evaluates if the throttling blocks a notification or not. Is this correct?

fgalan:
Fixed in PR #2116

PR #2116, opened 2016-05-03
Fixed a bug in throttling with v2 subscriptions

kzangeli:
Fixed issue #2114.


CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix: bug setting throttling to -1 in the subscription cache for v2 subscriptions (Issue #2114)
```
fgalan: It would be better to have a description closer to what the bug means from the user perspective (have a look to https://github.com/telefonicaid/fiware-orion/issues/2114#issuecomment-216587273 in order to understand what I mean).

test/functionalTest/cases/0000_large_requests/notification_different_sizes.test
```diff
@@ -241,7 +241,7 @@ Date: REGEX(.*)
 1
 ** Notification sizes:
  Sending message 1 to HTTP server: sending message of REGEX((19669|19670|19671|19672|19673|19674|19675)) bytes to HTTP server
- Sending message 2 to HTTP server: sending message of REGEX((21670|21671|21672|21673|21674|21675|21676)) bytes to HTTP server
+ Sending message 2 to HTTP server: sending message of REGEX((21669|21670|21671|21672|21673|21674|21675|21676)) bytes to HTTP server
```
fgalan: Probably this is not needed... use to be a glitch effect when develop evolves (in this case) from 1.0.0-next -> 1.1.0 -> 1.1.0-next.

Let's rollback this .test to its original status after merging develop now (note that this PR includes a develop merge, but not at the moment in which develop was prepared for the next cycle) and re-run the .test in order to check if my theory is correct.

src/lib/cache/subCache.cpp
```diff
@@ -757,7 +757,20 @@ void subCacheItemInsert
...
+  if (scrP->throttling.string != "")
+  {
+    //
+    // NOTE: In v1, the throttling of a subscription enters as a string and parsing that string makes sense (right?).
+    //       In v2 however, the throttling is an integer and as the integer value is passed as the sixth parameter of
+    //       this function, the call to scrP->throttling.parse() (the string is empty and parse returns -1) destroys
+    //       the throttling, setting it to -1 (inactive).
+    //       
+    //       For v2, not calling scrP->throttling.parse() is pretty clear.
+    //       What is still a little strange is that v1 has the throttling come is as the sixth parameter but then
```
fgalan: Looking to the code (as in develop, i.e. before this PR):

```
  //
  // Now allocate the subscription
  //
  CachedSubscription* cSubP = new CachedSubscription();
  ...
  cSubP->throttling            = throttling;   // An int64_t, comming from function arguments 
  ...
  cSubP->throttling = scrP->throttling.parse();
```

So it seems that always (i.e. no matter if subscripion came from NGSIv1 and NGSIv2) `scrP->throttling.parse()` is overriding throttling. Thus, probably we could be more "aggrestive" and just remove `cSubP->throttling = scrP->throttling.parse()` (of course, checking that all .test work after that).

Even more aggresive: remove `SubscribeContextRequest*  scrP` from function parameter, replacing it with the actual info that the insert function needs from it (reference string, notifyConditionVector, attributeList and entityIdVector). This is a case in which *too much information* (a throttling-as-a-string, inside de scrP) is causing problems (in particular, overriding another throttling parameter, as number, with the right information).

(Or maybe my reasoning is wrong in some point :))

test/functionalTest/cases/2114_throttling_in_v2_subs/throttling_in_v2_subs.test
```diff
@@ -0,0 +1,193 @@
...
+
+--NAME--
+Throttling in v2 subscriptions
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0
+accumulatorStart --pretty-print
+
+--SHELL--
```
fgalan: In this case (which is tricky) it would be useful to run this test with the code before changes and see it fails, to check that the .test is capturing the bug.

(Maybe you already did it, just want to check).

kzangeli: Yes, I did that, of course :-)

NTC

src/lib/cache/subCache.cpp
kzangeli: I decided not to risk any older code stopping to work, so I added that 'if'.
I can try to remove it altogether of course, just to see

CHANGES_NEXT_RELEASE
kzangeli: Sure, this text is very implementation oriented

kzangeli: Fixed in b9a3d45

src/lib/cache/subCache.cpp
kzangeli: Fixed in 6b4ac24

test/functionalTest/cases/0000_large_requests/notification_different_sizes.test
kzangeli: Yes, it works. Fixed in 63007a7.

Not that the modification did hurt much, instead it would prevent this problem from occurring again ...

fgalan:

crbrox:
LGTM

PR #2125, opened 2016-05-06
ADD functional description of templating mechanism in documentation

fgalan:
Doc part of #2015 

@kzangeli 
@jmcanterafonseca 
@mrutid

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1486,7 +1486,7 @@ An `http` object contains the following subfields:
 + `url` : URL referencing the service which is invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas could also be supported.
 
-An `httpExtende` object conatins the following subfields. If `httpExtened` is used, then
+An `httpExtended` object conatins the following subfields. If `httpExtened` is used, then
 "Notification templates" will be used (see specific section).
 
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -714,7 +714,8 @@ so notification receiver can know the format used without needing processing not
 ## Custom notifications
 
 NGSIv2 clients can customize HTTP notification messages using a simple template mechanism. The
-`notification.httpCustom` field allows to specify the following fields subject to be templatized:
+`notification.httpCustom` property of a subscription allows to specify the following fields subject
+to be templatized:
 
 * `url`
 * `headers` (both header names and values)
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
+At notification time, NGSIv2 implementations will send a notification based on the `notification.http`
+field of the associated subscription. This fields can be interpreted litearly or use a
+templating mechanism, based in macros substitution, that is described in the present subsection.
+
+The following fields inside `notification.http` in subscriptions are intepreted as templates:
+
+* `url`
+* `headers` (both header names and values)
+* `qs` (both in parameters names and values)
+* `payload`
+
+It is worth mentioning that `method` is not a template and it is always used literarly.
+
+Macro substitution is based in the syntax `${..}`. In particular:
+
+* `${id}` is replaced by entity id
+* `${type}` is replaced by entity type
+* Any other `${token}` is replaced by the value of attribute which name is `token` or with
+  an empty string if the attribute is not included in the data to notify. If the value is
+  not a string (i.e. it is a number, a bool, a JSON object or array) it is casted to a string.
+
+Example:
+
+Let's consider the following `notification.http` object in a given subscription.
+
+```
+"http": {
+  "url": "http://foo.com/entity/${id}",
+  "headers": {
+    "Content-Type": "text/plain"
+  },
+  "method": "PUT",
+  "qs": {
+    "type": "${type}"
+  },
+  "payload": "The temperature is ${temperature} degrees"
+}
+```
+
+Now, let's consider that NGSIv2 implementation triggers a notification associated to this subscription.
+Notification data is for entity with id "DC_S1-D41" and type "Room", including an attribute named
+"temperature" with value 23.4. The resulting notification after applying the template would be:
+
+```
+PUT http://foo.com/entity/DC_S1-D41?type=Room
+Content-Type: text/plain
+Content-Length: 31
+
+The temperature is 23.4 degrees
+```
+
+Some considerations to take into account:
+
+* It is user's responsability to ensure that after substitution the notification is a correct
+  HTTP message. Otherwise the NGSIv2 implementation could skip the notification (e.g. if URL results in a
+  malformed URL) or the receiver could get something formally wrong (e.g. the Conent-Type header is
+  application/xml but the payload doesn't correspond to a well-formed XML document).
+* In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
+  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage)
+  if at least one of the following is true:
+  * `url` contains one or more substitutions.
+  * `qs` contains one or more substitutions (either in parameters names or values).
+  * `headers` contains one or more substitutions (either header names or values).
+  * `payload` contains one or more substitutions.
 
 ## Group API Entry Point
 
@@ -1429,7 +1491,8 @@ An `http` object contains the following subfields:
   compliant server must support `http` URL schema, other schemas could also be supported.
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
 + `qs` (optional): a key-map of URL query parameters that are included in the notification message.
-+ `method` (optional): the method to use for sending the notification (default is POST)
++ `method` (optional): the method to use for sending the notification (default is POST). Only valid
+  HTTP methods are allowed.
 + `payload` (optional): the payload to be used in the notification. If omitted, the default payload (see
   "Notification messages" sections) is used.
 
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -711,35 +711,34 @@ If `attrsFormat` is `values` then values entity partial representation mode is u
 Notifications must include a `X-Ngsiv2-AttrsFormat` header with the value of `attrsFormat` of the associated subscription,
 so notification receiver can know the format used without needing processing notification payload.
 
-### Notification templates
+## Custom notifications
 
-At notification time, NGSIv2 implementations will send a notification based on the `notification.httpExtended`
-field of the associated subscription. These fields can be interpreted literally or use a
-templating mechanism, based in macros substitution, that is described in the present subsection.
-
-Only the following fields inside `notification.httpExtended` can be templatized:
+NGSIv2 clients can customize HTTP notification messages using a simple template mechanism. The
+`notification.httpCustom` field allows to specify the following fields subject to be templatized:
 
 * `url`
 * `headers` (both header names and values)
-* `qs` (both in parameters names and values)
+* `qs` (both in parameter names and values)
 * `payload`
 
-It is worth mentioning that `method` is not a template and it is always used literally.
+In addition the HTTP method to be used for delivering the notification can be customized as well using the `method` field,
+but it cannot be a template.
 
-Macro substitution is based on the syntax `${..}`. In particular:
+Macro substitution for templates is based on the syntax `${..}`. In particular:
 
 * `${id}` is replaced by entity id
 * `${type}` is replaced by entity type
 * Any other `${token}` is replaced by the value of attribute which name is `token` or with
-  an empty string if the attribute is not included in the notification. If the value is
-  not a string (i.e. it is a number, a bool, a JSON object or array) it is casted to a string.
+  an empty string if the attribute is not included in the notification. If the value is a number,
+  a bool or null then its string representation is used. If the value is a JSON array or object
+  then its JSON representation as string is used.
 
 Example:
 
-Let's consider the following `notification.httpExtended` object in a given subscription.
+Let's consider the following `notification.httpCustom` object in a given subscription.
 
 ```
-"httpExtended": {
+"httpCustom": {
   "url": "http://foo.com/entity/${id}",
   "headers": {
     "Content-Type": "text/plain"
@@ -766,12 +765,12 @@ The temperature is 23.4 degrees
 
 Some considerations to take into account:
 
-* It is user's responsability to ensure that after substitution the notification is a correct
+* It is the NGSIv2 client's responsability to ensure that after substitution the notification is a correct
   HTTP message (e.g. the Content-Type header is application/xml but the payload doesn't correspond to a
-  well-formed XML document). Specifically, if the resulting URL aftering appliying template is malformed,
-  then notification is skipped.
+  well-formed XML document). Specifically, if the resulting URL after appliying template is malformed,
+  then the notification is skipped.
 * In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
-  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage).
+  for each of the entities (contrary to default behaviour, which is to send all entities in the same HTTP message).
 
 ## Group API Entry Point
 
@@ -1473,7 +1472,7 @@ A `notification` object contains the following subfields:
 + `attrs`: List of attributes to be included in the notification message. If not specified (or empty), all attributes are included
   in the notification. It also defines the order in which attributes will appear in notifications when `attrsFormat` `value` is used
   (see "Notification messages" section).
-+ `http` or `httpExtended` (one of both has to be used, but not both at the same time): It is used to convey parameters
++ `http` or `httpCustom` (one of them has to be used, but not both at the same time): It is used to convey parameters
   for notifications delivered through the HTTP protocol.
 + `attrsFormat` (optional): specifies how the entities will be represented in notifications. Accepted values are
   `normalized` (the one used if the field is omitted), `keyValues` or `values`. If `attrsFormat` takes any value different
@@ -1486,15 +1485,17 @@ An `http` object contains the following subfields:
 + `url` : URL referencing the service which is invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas could also be supported.
 
-An `httpExtended` object conatins the following subfields. If `httpExtened` is used, then
-"Notification templates" will be used (see specific section).
+An `httpCustom` object contains the following subfields.
 
++ `url`: same as in `http` above.
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
 + `qs` (optional): a key-map of URL query parameters that are included in the notification message.
-+ `method` (optional): the method to use for sending the notification (default is POST). Only valid
-  HTTP methods are allowed, otherwise a 400 Bad Request error will be returned.
++ `method` (optional): the method to use for sending the notification (default is POST). Only valid HTTP methods are allowed.
+   On specifying an invalid HTTP method, a 400 Bad Request error is returned.
 + `payload` (optional): the payload to be used in the notification. If omitted, the default payload (see
   "Notification messages" sections) is used.
+
+If `httpCustom` is used, then the considerations described in "Custom notifications" section apply.
 
 Notification rules are as follow:
 
@@ -1544,7 +1545,7 @@ Response:
                     }
                 },
                 "notification": {
-                    "httpExtended": {
+                    "httpCustom": {
                       "url": "http://localhost:1234",
                       "headers": {
                         "X-MyHeader": "foo"
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -768,7 +768,7 @@ Some considerations to take into account:
 
 * It is the NGSIv2 client's responsability to ensure that after substitution the notification is a correct
   HTTP message (e.g. the Content-Type header is application/xml but the payload doesn't correspond to a
-  well-formed XML document). Specifically, if the resulting URL after appliying template is malformed,
+  well-formed XML document). Specifically, if the resulting URL after appliyng the template is malformed,
   then the notification is skipped.
 * In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
   for each of the entities (contrary to default behaviour, which is to send all entities in the same HTTP message).
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -713,33 +713,33 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
 
-At notification time, NGSIv2 implementations will send a notification based on the `notification.http`
-field of the associated subscription. This fields can be interpreted litearly or use a
+At notification time, NGSIv2 implementations will send a notification based on the `notification.httpExtended`
+field of the associated subscription. These fields can be interpreted literally or use a
 templating mechanism, based in macros substitution, that is described in the present subsection.
 
-The following fields inside `notification.http` in subscriptions are intepreted as templates:
+Only the following fields inside `notification.httpExtended` can be templatized:
 
 * `url`
 * `headers` (both header names and values)
 * `qs` (both in parameters names and values)
 * `payload`
 
-It is worth mentioning that `method` is not a template and it is always used literarly.
+It is worth mentioning that `method` is not a template and it is always used literally.
 
-Macro substitution is based in the syntax `${..}`. In particular:
+Macro substitution is based on the syntax `${..}`. In particular:
 
 * `${id}` is replaced by entity id
 * `${type}` is replaced by entity type
 * Any other `${token}` is replaced by the value of attribute which name is `token` or with
-  an empty string if the attribute is not included in the data to notify. If the value is
+  an empty string if the attribute is not included in the notification. If the value is
   not a string (i.e. it is a number, a bool, a JSON object or array) it is casted to a string.
 
 Example:
 
-Let's consider the following `notification.http` object in a given subscription.
+Let's consider the following `notification.httpExtended` object in a given subscription.
 
 ```
-"http": {
+"httpExtended": {
   "url": "http://foo.com/entity/${id}",
   "headers": {
     "Content-Type": "text/plain"
@@ -767,16 +767,11 @@ The temperature is 23.4 degrees
 Some considerations to take into account:
 
 * It is user's responsability to ensure that after substitution the notification is a correct
-  HTTP message. Otherwise the NGSIv2 implementation could skip the notification (e.g. if URL results in a
-  malformed URL) or the receiver could get something formally wrong (e.g. the Conent-Type header is
-  application/xml but the payload doesn't correspond to a well-formed XML document).
+  HTTP message (e.g. the Content-Type header is application/xml but the payload doesn't correspond to a
+  well-formed XML document). Specifically, if the resulting URL aftering appliying template is malformed,
+  then notification is skipped.
 * In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
-  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage)
-  if at least one of the following is true:
-  * `url` contains one or more substitutions.
-  * `qs` contains one or more substitutions (either in parameters names or values).
-  * `headers` contains one or more substitutions (either header names or values).
-  * `payload` contains one or more substitutions.
+  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage).
 
 ## Group API Entry Point
 
@@ -1478,7 +1473,8 @@ A `notification` object contains the following subfields:
 + `attrs`: List of attributes to be included in the notification message. If not specified (or empty), all attributes are included
   in the notification. It also defines the order in which attributes will appear in notifications when `attrsFormat` `value` is used
   (see "Notification messages" section).
-+ `http`: It is used to convey parameters for notifications delivered through the HTTP protocol.
++ `http` or `httpExtended` (one of both has to be used, but not both at the same time): It is used to convey parameters
+  for notifications delivered through the HTTP protocol.
 + `attrsFormat` (optional): specifies how the entities will be represented in notifications. Accepted values are
   `normalized` (the one used if the field is omitted), `keyValues` or `values`. If `attrsFormat` takes any value different
   than those an error will be raised. See detail in "Notification messages" section.
@@ -1489,10 +1485,14 @@ An `http` object contains the following subfields:
 
 + `url` : URL referencing the service which is invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas could also be supported.
+
+An `httpExtende` object conatins the following subfields. If `httpExtened` is used, then
+"Notification templates" will be used (see specific section).
+
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
 + `qs` (optional): a key-map of URL query parameters that are included in the notification message.
 + `method` (optional): the method to use for sending the notification (default is POST). Only valid
-  HTTP methods are allowed.
+  HTTP methods are allowed, otherwise a 400 Bad Request error will be returned.
 + `payload` (optional): the payload to be used in the notification. If omitted, the default payload (see
   "Notification messages" sections) is used.
 
@@ -1544,7 +1544,7 @@ Response:
                     }
                 },
                 "notification": {
-                    "http": {
+                    "httpExtended": {
                       "url": "http://localhost:1234",
                       "headers": {
                         "X-MyHeader": "foo"
--- a/doc/manuals/user/forbidden_characters.md
+++ b/doc/manuals/user/forbidden_characters.md
@@ -64,17 +64,17 @@ enable them for NGSIv1, as described in [this section of the documentation](v1_v
 
 NGSIv2 allows a templating mechanism (see "Notification templates" section in
 the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/)). Forbidden
-characters restrictions apply to the `http.payload` field in NGSIv2 API operations, such as
+characters restrictions apply to the `httpExtended.payload` field in NGSIv2 API operations, such as
 POST /v2/subscription or GET /v2/subscriptions.
 
-However, at notification time, any URL encoded characters is `http.payload` is decoded.
+However, at notification time, any URL encoded characters in `httpExtended.payload` is decoded.
 
 Example:
 
-Let's consider the following `notification.http` object in a given subscription.
+Let's consider the following `notification.httpExtended` object in a given subscription.
 
 ```
-"http": {
+"httpExtended": {
   "url": "http://foo.com/entity/${id}",
   "headers": {
     "Content-Type": "aplication/json"
@@ -86,6 +86,9 @@ Let's consider the following `notification.http` object in a given subscription.
   "payload": "{ %22temperature%22: ${temperature}, %22asString%22: %22${temperature}%22 }"
 }
 ```
+
+Note that the above payload value is the URL encoded version of this string:
+`{ "temperature": ${temperature}, "asString": "${temperature}" }`.
 
 Now, let's consider that NGSIv2 implementation triggers a notification associated to this subscription.
 Notification data is for entity with id "DC_S1-D41" and type "Room", including an attribute named
--- a/doc/manuals/user/forbidden_characters.md
+++ b/doc/manuals/user/forbidden_characters.md
@@ -62,22 +62,22 @@ enable them for NGSIv1, as described in [this section of the documentation](v1_v
 
 ### Notification payload template
 
-NGSIv2 allows a templating mechanism (see "Notification templates" section in
+NGSIv2 allows a templating mechanism (see "Custom notifications" section in
 the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/)). Forbidden
-characters restrictions apply to the `httpExtended.payload` field in NGSIv2 API operations, such as
+characters restrictions apply to the `httpCustom.payload` field in NGSIv2 API operations, such as
 POST /v2/subscription or GET /v2/subscriptions.
 
-However, at notification time, any URL encoded characters in `httpExtended.payload` is decoded.
+However, at notification time, any URL encoded characters in `httpCustom.payload` is decoded.
 
 Example:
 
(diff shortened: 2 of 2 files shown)


doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,42 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
 or metadata name/type) which are described in the "Field syntax restrictions" section in the
 [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/). You can also
 enable them for NGSIv1, as described in [this section of the documentation](v1_v2_coexistence.md#checking-id-fields).
+
+### Notification payload template
+
+NGSIv2 allows a templating mechanism (see "Notification templates" section in
```
fgalan: Note that the special case of URL decoding in payload template is Orion implementation specific (a consecuence of forbidden chars in Orion) so it is not described in the .apib but in the .md. If you think that there is a better place than as sub-section in forbidden_characters.md, please suggest it.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
+At notification time, NGSIv2 implementations will send a notification based on the `notification.http`
+field of the associated subscription. This fields can be interpreted litearly or use a
```
jmcanterafonseca: These

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+
+The following fields inside `notification.http` in subscriptions are intepreted as templates:
+
+* `url`
+* `headers` (both header names and values)
+* `qs` (both in parameters names and values)
+* `payload`
+
+It is worth mentioning that `method` is not a template and it is always used literarly.
+
```
jmcanterafonseca: literally

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+Macro substitution is based in the syntax `${..}`. In particular:
+
```
jmcanterafonseca: based on

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+* `${id}` is replaced by entity id
+* `${type}` is replaced by entity type
+* Any other `${token}` is replaced by the value of attribute which name is `token` or with
+  an empty string if the attribute is not included in the data to notify. If the value is
```
jmcanterafonseca: in the notification?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+  not a string (i.e. it is a number, a bool, a JSON object or array) it is casted to a string.
```
jmcanterafonseca: that's controversial 

if I have { a: ${val} } I expect to keep the type and not convert to a string

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+Content-Length: 31
+
+The temperature is 23.4 degrees
+```
+
+Some considerations to take into account:
+
+* It is user's responsability to ensure that after substitution the notification is a correct
+  HTTP message. Otherwise the NGSIv2 implementation could skip the notification (e.g. if URL results in a
+  malformed URL) or the receiver could get something formally wrong (e.g. the Conent-Type header is
```
jmcanterafonseca: Content-Type

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+  application/xml but the payload doesn't correspond to a well-formed XML document).
+* In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
+  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage)
+  if at least one of the following is true:
+  * `url` contains one or more substitutions.
+  * `qs` contains one or more substitutions (either in parameters names or values).
+  * `headers` contains one or more substitutions (either header names or values).
+  * `payload` contains one or more substitutions.
```
jmcanterafonseca: That is risky, and we are altering the behavior of the API, depending on whether there are templates or not, sounds not very consistent

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Actually, if understanding you correctly, it is able to work in the way you say. Have a look to this examples: https://github.com/telefonicaid/fiware-orion/pull/2125/files/dd6e4f439c64890312ba21fbafcc0b07be86a5b2#diff-7ed462fcb6ea86180aabc505a3c08a41R72

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1429,7 +1491,8 @@ An `http` object contains the following subfields:
   compliant server must support `http` URL schema, other schemas could also be supported.
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
 + `qs` (optional): a key-map of URL query parameters that are included in the notification message.
-+ `method` (optional): the method to use for sending the notification (default is POST)
++ `method` (optional): the method to use for sending the notification (default is POST). Only valid
+  HTTP methods are allowed.
```
jmcanterafonseca: what happens if an invalid method is used?

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,42 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
...
+the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/)). Forbidden
+characters restrictions apply to the `http.payload` field in NGSIv2 API operations, such as
+POST /v2/subscription or GET /v2/subscriptions.
+
+However, at notification time, any URL encoded characters is `http.payload` is decoded.
+
```
jmcanterafonseca: in

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,42 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
...
+"http": {
+  "url": "http://foo.com/entity/${id}",
+  "headers": {
+    "Content-Type": "aplication/json"
+  },
+  "method": "PUT",
+  "qs": {
+    "type": "${type}"
+  },
+  "payload": "{ %22temperature%22: ${temperature}, %22asString%22: %22${temperature}%22 }"
```
jmcanterafonseca: This is really weird and confusing. Why don't you allow any character in the payload template, it is a special attribute

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
+At notification time, NGSIv2 implementations will send a notification based on the `notification.http`
+field of the associated subscription. This fields can be interpreted litearly or use a
+templating mechanism, based in macros substitution, that is described in the present subsection.
+
+The following fields inside `notification.http` in subscriptions are intepreted as templates:
```
jmcanterafonseca: can be templatized

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+  "method": "PUT",
+  "qs": {
+    "type": "${type}"
+  },
+  "payload": "The temperature is ${temperature} degrees"
+}
+```
+
+Now, let's consider that NGSIv2 implementation triggers a notification associated to this subscription.
+Notification data is for entity with id "DC_S1-D41" and type "Room", including an attribute named
```
jmcanterafonseca: nit: `DC_xx` `Room`

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,70 @@ so notification receiver can know the format used without needing processing not
...
+"temperature" with value 23.4. The resulting notification after applying the template would be:
+
```
jmcanterafonseca: nit: `` needed

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: this is controversial as well. I expect the implementation to properly encode parameters in the URL

and I don't expect the implementation to check whether XML or any other fancy format used is syntactically correct

fgalan: > I expect the implementation to properly encode parameters in the URL

Parameters (I understand you mean to processing of the `qs` field) are not the problem. The problem is, e.g:

```
"http": {
  "url": "${sendTo}"
}

If the value of "sendTo" attribute is "http://foo.bar/myApp", everything is ok. However, if its value is eg "blablabla", then URL is not wellformed. We need to protect agains this, and the best protection is just not sending anything. I.e. "the NGSIv2 implementation could skip the notification"

> I don't expect the implementation to check whether XML or any other fancy format used is syntactically correct

That exactly what the spec tries to say, that "is user's responsability to ensure that after substitution the notification is a correct HTTP message". If you find it is not clear enough, please propose an alternative wording.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 8 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I tend to agree... but we haven't found a better solution yet. What do you propose?

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 9 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I think the best option would be to report an error at csub creation/update time (easy to detect by the parsing layer).

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 8 above)
Comment (kzangeli) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I agree to a certain point.
On the one hand, for notifications with a vector of entities, we have to split the notification into N notifications, if we want to be sure to support substitutions correctly. This is a big overhead and not desired if templates are not in use.

On the other hand, if the user has requested to use templates, but doesn't want any substitutions (perhaps the user just wants to avoid receiving payload and get a 'ping' whenever something has changed, and perhaps react on it every 5 'pings'), then perhaps the broker should do what the user asked for and not trying to do anything 'smarter'.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 9 above)
Comment (kzangeli) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: The subscription-creation would fail.
Was this a question or do you mean that this information should be a part of the spec?

File: doc/manuals/user/forbidden_characters.md (hunk 11 above)
Comment (kzangeli) [inline doc/manuals/user/forbidden_characters.md]: Perhaps we could write out the quotes

    "payload": "{ "temperature": ${temperature}, "asString": "${temperature}" }"

And under the example explain that the value of "payload" must be encoded?:

*Note that the value of the field "payload" must be URL-encoded, as it contains forbidden characters and otherwise will not be accepted*

Comment (fgalan) [inline doc/manuals/user/forbidden_characters.md]: I think is better to be exact, I mean, to show in the example the request as it would be recieved by Orion "on the wire". So, we can do it somehow in the other way around, i.e. leave the example as it is now and include a note below like:

Note that the above `payload` value is the URL encoded version of this string: `{ "temperature": ${temperature}, "asString": "${temperature}" }`.

Comment (kzangeli) [inline doc/manuals/user/forbidden_characters.md]: Sure, that's another way of getting the same advantage

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 8 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: here we have two options:

A/ the templating language allows to deals with collections. That can be done in the future. 

B/ for the time being, given that collections are not supported, can we do the following which is, map the first element in the collection as if it were the only one and add the remaining content of the notification as an extra attribute in the payload requested by the origin service i.e.

```json

payload: "{ "attr1": ${prop1}, "attr2": ${prop2} }" 

{
  "attr1": "value11",
  "attr2": "value21",
  "extra": [{
    "id": "entity2",
    "type": "xx",
    "prop1": 34,
    "prop2" 45
  }]
}
````

File: doc/manuals/user/forbidden_characters.md (hunk 11 above)
Comment (jmcanterafonseca) [inline doc/manuals/user/forbidden_characters.md]: that's an option yes

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 9 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: we need to say clearly that an error will be raised, indicating error name

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 7 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I think we need to make it clear that if the URL is not a valid URL the implementation must skip the notification. I understand that a message will be logged for tracing purposes.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 6 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I think there is no need to convert to a string if the user does not want to. 

payload: { "val": ${myVal} } 

then myVal will be kept as it is conserving its data type, number, object, array, whatever. 

if I say "payload: "val": "${myVal}" then in that case a conversion to a string could be needed, for those values which are not a string. in that case we can agree on serializing an object or array as a JSON serialization. but for me that's a feature for a next iteration.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 13 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Not sure what do you mean here. Could you clarify, please?

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 14 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: To which word/fragment?

File: doc/manuals/user/forbidden_characters.md (hunk 1 above)
Comment (fgalan) [inline doc/manuals/user/forbidden_characters.md]: NTC (it was just informative)

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 2 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in ffb27c6

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 3 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Typo fixed in ffb27c6.

It think it is better to leave the sentences. Additional clarification doesn't hurt. However, I'd like to heard more opinions on this.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 4 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in ffb27c6

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 5 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in ffb27c6

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 7 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: I have re-writter the paragram in ffb27c6, to make it clear that malformed URLs are skipped.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 8 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: After our last discussions, multi-request is done when `httpExtended` is used, to matter which substitutions take place. Thus, all the sub-bullets here has been removed in ffb27c6. 

Please, have a look to the new wording, specially @kzangeli , which is implementing this part right now.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 9 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in ffb27c6

File: doc/manuals/user/forbidden_characters.md (hunk 10 above)
Comment (fgalan) [inline doc/manuals/user/forbidden_characters.md]: Fixed in ffb27c6

File: doc/manuals/user/forbidden_characters.md (hunk 11 above)
Comment (fgalan) [inline doc/manuals/user/forbidden_characters.md]: Fixed in ffb27c6

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 12 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in ffb27c6

**Comment by fgalan** [comment]
PR changes in order to align with the `httpExtended` concept we were discussing at skype. Please, have a look in order to see if it is aligned with the dicussions (I could have make missundesrtand something) and @kzangeli check that the implementation of the notification part is aligned with what you read.

At the present point, I think only two issues remain:

* The one about "string-fy" attribute values upon substitution or do a more inteligent type-based substitution.
* The one about URL encode the payload in create/get subscription but using it in decoded form 

Let's try to close them as soon as possible early next week.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 13 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: you need to enclose DC_Room between "`" chars

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 14 above)
Comment (jmcanterafonseca) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: for temperature

File: doc/apiary/v2/fiware-ngsiv2-reference.apib [hunk 15]
```diff
@@ -1416,7 +1473,8 @@ A `notification` object contains the following subfields:
 + `attrs`: List of attributes to be included in the notification message. If not specified (or empty), all attributes are included
   in the notification. It also defines the order in which attributes will appear in notifications when `attrsFormat` `value` is used
   (see "Notification messages" section).
-+ `http`: It is used to convey parameters for notifications delivered through the HTTP protocol.
++ `http` or `httpExtended` (one of both has to be used, but not both at the same time): It is used to convey parameters
```
jmcanterafonseca: nit: one of them

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1427,9 +1485,14 @@ An `http` object contains the following subfields:
 
 + `url` : URL referencing the service which is invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas could also be supported.
+
+An `httpExtended` object conatins the following subfields. If `httpExtened` is used, then
```
jmcanterafonseca: nit: httpExtended

kzangeli: And 'conatins' :-)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1427,9 +1485,14 @@ An `http` object contains the following subfields:
...
+"Notification templates" will be used (see specific section).
+
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
```
fgalan: `url` should be also included in `httpExtended` section (just a copy-paste from the `http` section).

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+PUT http://foo.com/entity/DC_S1-D41?type=Room
+Content-Type: text/plain
+Content-Length: 31
+
+The temperature is 23.4 degrees
+```
+
+Some considerations to take into account:
+
+* It is user's responsability to ensure that after substitution the notification is a correct
```
kzangeli: **the** user's   (or perhaps client instead of user?)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1427,9 +1485,14 @@ An `http` object contains the following subfields:
 
 + `url` : URL referencing the service which is invoked when a notification is generated. A NGSIv2
   compliant server must support `http` URL schema, other schemas could also be supported.
+
+An `httpExtended` object conatins the following subfields. If `httpExtened` is used, then
+"Notification templates" will be used (see specific section).
```
jmcanterafonseca: can

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1481,7 +1544,7 @@ Response:
                     }
                 },
                 "notification": {
-                    "http": {
+                    "httpExtended": {
```
fgalan: Overall replace: httpExtended -> httpCustom

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
```
jmcanterafonseca: Maybe this can defined at ## level or call it simply Templates

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
```
jmcanterafonseca: I would say here:

NGSIv2 clients can customize HTTP notification messages using a template language. The notification.httpCustom field allows to specify the following fields subject to be templatized: .... 

In addition the HTTP method to be used for delivering the notification can be customized as well, but it cannot be a template.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+Only the following fields inside `notification.httpExtended` can be templatized:
+
+* `url`
+* `headers` (both header names and values)
+* `qs` (both in parameters names and values)
+* `payload`
+
+It is worth mentioning that `method` is not a template and it is always used literally.
+
+Macro substitution is based on the syntax `${..}`. In particular:
```
jmcanterafonseca: Macro substitution for templates

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+
+* `${id}` is replaced by entity id
+* `${type}` is replaced by entity type
+* Any other `${token}` is replaced by the value of attribute which name is `token` or with
+  an empty string if the attribute is not included in the notification. If the value is
+  not a string (i.e. it is a number, a bool, a JSON object or array) it is casted to a string.
```
jmcanterafonseca: if the payload is actually JSON, then it is still not clear to me how this would work ... I think it needs to be clarified

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: client

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
 
 ### Notification templates
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
+At notification time, NGSIv2 implementations will send a notification based on the `notification.httpExtended`
```
kzangeli: Better not to use 'will' (and plural):

    At notification time, NGSIv2 implementations send notifications based on

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+field of the associated subscription. These fields can be interpreted literally or use a
```
kzangeli: or **using** a templating mechanism, based **on** **macro** substitution

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+templating mechanism, based in macros substitution, that is described in the present subsection.
+
+Only the following fields inside `notification.httpExtended` can be templatized:
+
+* `url`
+* `headers` (both header names and values)
+* `qs` (both in parameters names and values)
```
kzangeli: in parameters => parameter

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: casted ... perhaps 'treated' is ab better word ?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+Content-Length: 31
+
+The temperature is 23.4 degrees
+```
+
+Some considerations to take into account:
+
+* It is user's responsability to ensure that after substitution the notification is a correct
+  HTTP message (e.g. the Content-Type header is application/xml but the payload doesn't correspond to a
+  well-formed XML document). Specifically, if the resulting URL aftering appliying template is malformed,
```
kzangeli: aftering appliying ... :-)  

after applying

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+  then notification is skipped.
```
kzangeli: then **the** notification ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: These two phrases need to be switched. We can't end a sentence with 'contains the following subfields', without showing the list right after.

I propose:

    If the field `httpExtended` is used then Notification templates are supported (see specific section).
    An `httpExtended` object contains the following subfields:
    * 
    * ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1427,9 +1485,14 @@ An `http` object contains the following subfields:
...
   compliant server must support `http` URL schema, other schemas could also be supported.
+
+An `httpExtended` object conatins the following subfields. If `httpExtened` is used, then
+"Notification templates" will be used (see specific section).
+
 + `headers` (optional): a key-map of HTTP headers that are included in the notification messages.
 + `qs` (optional): a key-map of URL query parameters that are included in the notification message.
-+ `method` (optional): the method to use for sending the notification (default is POST)
++ `method` (optional): the method to use for sending the notification (default is POST). Only valid
+  HTTP methods are allowed, otherwise a 400 Bad Request error will be returned.
```
kzangeli: otherwise ...

Only valid HTTP methods are allowed. On receiving an invalid HTTP method, a 400 Bad Request error is returned.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -713,8 +713,65 @@ so notification receiver can know the format used without needing processing not
...
+```
+
+Some considerations to take into account:
+
+* It is user's responsability to ensure that after substitution the notification is a correct
+  HTTP message (e.g. the Content-Type header is application/xml but the payload doesn't correspond to a
+  well-formed XML document). Specifically, if the resulting URL aftering appliying template is malformed,
+  then notification is skipped.
+* In case the data to notify contains more than one entity, a separate notification (HTTP message) is sent
+  for each of the entities (contrary to default behaviour, which is to send all that in the same HTTP mesage).
```
kzangeli: all that => all notifications
mesage => message

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,45 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
...
+
+Example:
+
+Let's consider the following `notification.httpExtended` object in a given subscription.
+
+```
+"httpExtended": {
+  "url": "http://foo.com/entity/${id}",
+  "headers": {
+    "Content-Type": "aplication/json"
```
kzangeli: application

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After skype discusion with @jmcanterafonseca we have found that the following:

"If the value is a number, a bool or null then its string representation is used. If the value is a JSON array or object then its JSON representation as string is used."

captures the meaning more precisely so we will use it at the end.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

I'm using a new name for the section: "Custom notifications"

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0 (rewritten)

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0 (rewritten)

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 9585da0

doc/manuals/user/forbidden_characters.md
fgalan: Fixed in 9585da0

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -711,10 +711,66 @@ If `attrsFormat` is `values` then values entity partial representation mode is u
 Notifications must include a `X-Ngsiv2-AttrsFormat` header with the value of `attrsFormat` of the associated subscription,
 so notification receiver can know the format used without needing processing notification payload.
 
-### Notification templates
+## Custom notifications
 
-Based on the "template" field defined as part of the subscription and in a macro replacement language
-that could be used in "payload", "url", "headers", "method" and "qs". Details TBD.
+NGSIv2 clients can customize HTTP notification messages using a simple template mechanism. The
+`notification.httpCustom` field allows to specify the following fields subject to be templatized:
```
jmcanterafonseca: notification.httpCustom`  property  of a subscription allows to specify allows to specify the following parameters

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -711,10 +711,66 @@ If `attrsFormat` is `values` then values entity partial representation mode is u
...
+
+* `url`
+* `headers` (both header names and values)
```
jmcanterafonseca: headers .- HTTP headers

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -711,10 +711,66 @@ If `attrsFormat` is `values` then values entity partial representation mode is u
...
+* `qs` (both in parameter names and values)
```
jmcanterafonseca: qs .- query string

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,45 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
 or metadata name/type) which are described in the "Field syntax restrictions" section in the
 [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/). You can also
 enable them for NGSIv1, as described in [this section of the documentation](v1_v2_coexistence.md#checking-id-fields).
+
+### Notification payload template
+
+NGSIv2 allows a templating mechanism (see "Custom notifications" section in
```
jmcanterafonseca: provides a templating mechanism

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,45 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
...
+the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/)). Forbidden
+characters restrictions apply to the `httpCustom.payload` field in NGSIv2 API operations, such as
+POST /v2/subscription or GET /v2/subscriptions.
+
+However, at notification time, any URL encoded characters in `httpCustom.payload` is decoded.
+
```
jmcanterafonseca: are decoded

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: I think you used " instead of `

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,45 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
 or metadata name/type) which are described in the "Field syntax restrictions" section in the
 [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/). You can also
 enable them for NGSIv1, as described in [this section of the documentation](v1_v2_coexistence.md#checking-id-fields).
+
+### Notification payload template
```
jmcanterafonseca: Custom notifications , better as title, right?

doc/manuals/user/forbidden_characters.md
```diff
@@ -59,3 +59,45 @@ NGSIv2 introduces syntax restrictions for ID fields (such as entity id/type, att
...
+
+NGSIv2 allows a templating mechanism (see "Custom notifications" section in
+the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/)). Forbidden
+characters restrictions apply to the `httpCustom.payload` field in NGSIv2 API operations, such as
+POST /v2/subscription or GET /v2/subscriptions.
+
+However, at notification time, any URL encoded characters in `httpCustom.payload` is decoded.
```
jmcanterafonseca: will be decoded

jmcanterafonseca:
with my latest comments addressed, LGTM

thanks for the work

doc/manuals/user/forbidden_characters.md
kzangeli: Or 'are decoded'.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The field has been already described in the subscription section.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The field has been already described in the subscription section.

NTC

doc/manuals/user/forbidden_characters.md
fgalan: At the end I'll use "Custom payload special treatment"

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 972943c

doc/manuals/user/forbidden_characters.md
fgalan: Fixed in 972943c

doc/manuals/user/forbidden_characters.md
fgalan: Fixed in 972943c

doc/manuals/user/forbidden_characters.md
fgalan: Fixed in 972943c

kzangeli:
LGTM

PR #2139, opened 2016-05-11
Task/2015 part5 notifications without substitutions

kzangeli:
Synch PR, just to check that we are all on the track,
Lots of garbage to remove before a 'real' PR, like #if 0/#endif, LM_W(("KZ ...    etc
Please ignore those things, will be fixed eventually.


src/lib/cache/subCache.cpp
```diff
@@ -734,6 +737,8 @@ void subCacheItemInsert
   cSubP->expression.geometry   = geometry;
   cSubP->expression.coords     = coords;
   cSubP->expression.georel     = georel;
+  cSubP->httpInfo.url          = url.c_str();
+  cSubP->httpInfo.verb         = POST;  // FIXME PR: Unhardwire
```
fgalan: I understand that at the end the method should come as a new argument in the subCacheItemInsert() function. Is that the idea?

fgalan: Or maybe using `httpInfo*` as argument. If that parameter is not NULL, then the cache inserting was called from the NGSIv2 logic and it is a matter of copying into cSubP->httpInfo. If it is NULL, you can use defaults for cSubP->httpInfo.

src/lib/ngsi/ContextElement.cpp
```diff
@@ -113,14 +113,17 @@ std::string ContextElement::render(ConnectionInfo* ciP, RequestType requestType,
...
-std::string ContextElement::toJson(RenderFormat renderFormat, const std::vector<std::string>& attrsFilter)
+std::string ContextElement::toJson(RenderFormat renderFormat, const std::vector<std::string>& attrsFilter) const
 {
   std::string out;
 
   if (renderFormat != NGSI_V2_VALUES)
   {
     out += entityId.toJson();
-    out += ",";
+    if (contextAttributeVector.size() != 0)
```
fgalan: A bug fix not related with the notificaitons functionality?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
   LM_T(LmtNotImplemented, ("Notifier destructor is not implemented"));
 }
 
+
+
+/* ****************************************************************************
```
fgalan: Notifier.cpp is focused in Notifier class method implementation. Maybe new methods substitude(), etc. should be in a separate .cpp/.h.

Not really sure if this is the best alternative... just proposing in case you haven't realize.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+)
+{
+  Verb                                verb = httpInfo.verb;
+  std::string                         method;
+  std::string                         url;
+  std::string                         payload;
+  std::map<std::string, std::string>  qs;
+  std::map<std::string, std::string>  headers;
+
+  LM_W(("KZ: In templateNotify"));
```
fgalan: feature/xxx would need a grep KZ check to catch them all.

src/lib/ngsi/ContextElement.cpp
kzangeli: Yeah, I got this unconditional comma always when using only ContextElement in the payload.
Had to change this.
Funny that NO OTHER toJson suffers from this until now ...

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+{
+  *sP = in;
+}
+
+
+
+/* ****************************************************************************
+*
+* templateNotify - 
+*/
```
fgalan: A little explanation in the function preamble would be greate... what this method does? build the notification payload after applyting template? Maybe I haven't reach the point in the PR where this gets clear, but a little explanation would help anyway.

src/lib/cache/subCache.cpp
kzangeli: Yes, I believe `httpInfo*` might be good as argument to this function.
I haven't really thought this over though, been concentrating on other stuff

src/lib/ngsiNotify/Notifier.cpp

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+                      method,
+                      tenant,
+                      ce.entityId.servicePath,
+                      xauthToken,
+                      uri,
+                      "application/json",
+                      payload,
+                      fiwareCorrelator,
+                      "",                  // ngisv2AttrFormat
+                      false,               // don't use rush ... ?
```
fgalan: Actually, this should depend in the CLI, isn't it?

src/lib/ngsiNotify/Notifier.cpp
kzangeli: I am always a friend to creating new files and modules.
And it is true, the substitution in itself has nothing to do with notifications.
Perhaps it should be a module in common lib ...

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+                      protocol,
+                      method,
+                      tenant,
+                      ce.entityId.servicePath,
+                      xauthToken,
+                      uri,
+                      "application/json",
+                      payload,
+                      fiwareCorrelator,
+                      "",                  // ngisv2AttrFormat
```
fgalan: If the payload hasn't been touched, then attrsFormat should be used.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+                      false,               // don't use rush ... ?
+                      true,                // wait for response
+                      &out,
+                      headers,
+                      "application/json",  // Accept Format
+                      5000);               // Timeout in milliseconds: FIXME PR: unhardwire
```
fgalan: Yes... this comes from httpTimeout from the CLI.

src/lib/ngsiNotify/Notifier.cpp
kzangeli: oh yes, that will come, but later.
I'm still using my traces ...

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+*
+* sendNotifyContextRequestAsPerTemplate -
+*/
+void* sendNotifyContextRequestAsPerTemplate(void* p)
+{
+  NotificationAsTemplateParams* paramP = (NotificationAsTemplateParams*) p;
+
+  LM_W(("KZ: In sendNotifyContextRequestAsPerTemplate (%d contextElementResponses)", paramP->ncrP->contextElementResponseVector.size()));
+
+  for (unsigned int ix = 0; ix < paramP->ncrP->contextElementResponseVector.size(); ++ix)
```
fgalan: This is the part in which you split into N HTTP requests in the case of templates?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+    //
+    // If any of the 'template parameters' is used by the subscripion, then it is not a question of an ordinary notification but one using templates.
+    // Ordinary notifications are simply sent, all of it as one message.
+    //
+    // Notifications for subscriptions using templates are more complex:
+    //   - if there is more than one entity for the notification, split into N notifications, one per entity
+    //   - if 'url' contains substitution keys, substitute the keys for their current values
+    //   - if 'qs' contains query strings, add this to 'url', with the proper substitutions done
+    //   - if 'headers' is non-empty, perform eventual substitutions and make sure the information is added as HTTP headers for the notification
+    //   - if 'payload' is given, use that string as template instead of the default payload string, substituting all fields that are to be substituted
```
fgalan: If you want to be complete, I'd add the next:

```
- if 'method' is given, then a custom HTTP method is used (instead of POST default)
```

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+    // Redirect to the method sendNotifyContextRequestAsPerTemplate() when 'httpInfo.extended' is TRUE.
+    // 'httpInfo.extended' is FALSE by default and set to TRUE by the json parser.
+    //
+    if (httpInfo.extended)
+    {
+      LM_W(("KZ: Calling sendNotifyContextRequestAsPerTemplate"));
+
+      NotificationAsTemplateParams* paramP = new NotificationAsTemplateParams();
+
+      paramP->ncrP             = ncrP->clone();
```
fgalan: I have seen the new clone() method... Is the memmory from clone() controlled (freed) in all the places where it is use?

src/lib/ngsiNotify/Notifier.cpp
kzangeli: I really didn't know what to put here and I didn't spend any time looking into it.
Let's do that now ...

From httpRequestSend:

    //
    // Rush
    // Every call to httpRequestSend specifies whether RUSH should be used or not.
    // But, this depends also on how the broker was started, so here the 'useRush'
    // parameter is cancelled in case the broker was started without rush.
    //
    // If rush is to be used, the IP/port is stored in rushHeaderIP/rushHeaderPort and
    // after that, the host and port of rush is set as ip/port for the message, that is
    // now sent to rush instead of to its final destination.
    // Also, a few HTTP headers for rush must be setup.
    //
    if ((rushPort == 0) || (rushHost == ""))
    {
      useRush = false;
    }
    
    if (useRush)
    {
      ...

So, if we put a `true` for useRush, then Rush will be used only if the CLI arg tells the broker to use Rush.

Using `useRush = false` would provoke rush to **never** be used for the notification.
A value of `true` seems like the better alternative, right?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+      paramP->xauthToken       = xauthToken;
+      paramP->fiwareCorrelator = fiwareCorrelator;
+      paramP->renderFormat     = renderFormat;
+      paramP->attrsOrder       = attrsOrder;
+
+#if 0
+      sendNotifyContextRequestAsPerTemplate(paramP);
+#else
+      pthread_t  tid;
+      int        r = pthread_create(&tid, NULL, sendNotifyContextRequestAsPerTemplate, (void*) paramP);
```
fgalan: What is the purpose of creating a new thread at this point?

I mean, it could make sense to create one thread per HTTP request in the for loop inside sendNotifyContextRequestAsPerTemplate(), so each notification could be send in a paralel thread. However, I don't see the point of doing it at this point...

src/lib/rest/Verb.cpp
```diff
@@ -48,3 +48,24 @@ const char* verbName(Verb verb)
 
   return "Unknown verb";
 }
+
+
+
+/* ****************************************************************************
+*
+* stringToVerb -
```
fgalan: The list comes from the HTTP RFC?

test/unittests/commonMocks.h
```diff
@@ -220,5 +224,28 @@ MATCHER_P(MatchNcar, expected, "") {
...
+
+/* ****************************************************************************
+*
+* MatchHttpInfo -
+*
+* NOTE
+*   HttpInfo is a class, but after fighting with operator== quite some time I (kz) gave up
+*   and tried this way instead.
+*/
+MATCHER_P(MatchHttpInfo, expected, "")
```
fgalan: Maybe this needs to be extended to compare all the possible fields in httpInfo class?

test/unittests/mongoBackend/mongoNotifyContext_test.cpp
```diff
@@ -898,3 +900,77 @@ TEST(mongoNotifyContextRequest, createEntity)
     utExit();
 }
 
+
+
+/* ****************************************************************************
+*
+* templateNotificationVerb -
```
fgalan: This (and the next) are the "stimulation unit test" you are using for manual testing? I'd suggest to include a "FIXME: to be removed" as a reminder.

src/lib/ngsi/ContextElement.cpp
fgalan: Need to mention in CNR? Probably no, as it is not actually impacting any use case...

src/lib/ngsiNotify/Notifier.cpp
kzangeli: I looked into this also. If we send `-1` as Timeout, then the **default timeout** is used:

    static long defaultTimeout = DEFAULT_TIMEOUT;

src/lib/common/limits.h:

    #define DEFAULT_TIMEOUT     5000

However, we also have a function, httpRequestInit, that overwrite this value with what came from the CLI.

So, if  `-1` is used, whatever (default is 5000) comes in via CLI will be used.

I think that is exactly what we want.

src/lib/ngsiNotify/Notifier.cpp
fgalan: What are using other callers of httpRequestSend() ?

src/lib/ngsiNotify/Notifier.cpp
fgalan: What are using other callers of httpRequestSend() ?

src/lib/ngsi/ContextElement.cpp
kzangeli: This was probably not a bug until ContextElement was being used (in this PR) as sole content in the payload.  I would not mention it in the CNR

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Notifications in senderThread.cpp use `true` while forwards in postQueryContext.cpp and postUpdateContext.cpp use `false`.

Makes sense to use `true` here.

src/lib/ngsi/ContextElement.cpp
fgalan: Makes sense. NTC.

src/lib/ngsiNotify/Notifier.cpp
kzangeli: All use -1  (using the default value of -1 for the function)

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Yes - comment in header to explain this is still missing

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Nope, no valgrind stuff still.  I have it in mind though.

However, right now I am thinking of trying to 'steal' the pointer instead of copying it.
If the pointer is no longer used after this (except by the templates, of course) then we  would save some time by stealing the pointer (something like: paramP->ncrP = ncrP; ncrP = NULL).
This to avoid ncrP to be released before it is used by the notification thread. Something similar to 'smart pointers', I think.

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Well, we said that the notifications when split into N notifications must be sent serialized, awaiting an ACK before continuing with the next one.
That's why I implemented it this way, just following the instructions ...

src/lib/rest/Verb.cpp
kzangeli: Yeah, I included all of them (and added NOVERB)

test/unittests/commonMocks.h
kzangeli: Yes, of course.
For now this was enough to pass the unit tests, but of course, the comparison should be complete. A FIXME PR is missing here.

src/lib/ngsiNotify/Notifier.cpp
fgalan: I mean if you have controlled freed *before* valgrind test. We should ensure that, to the best of our knowledgue and ability, the memory is correctly managed in every PR. I mean, not adopt a "bah... I don't need to pay attention to this, as valgrind we save the day after all... in a couple of weeks" way of thinking.

src/lib/ngsiNotify/Notifier.cpp
fgalan: But in that case... why do you need to split a new thread here? I think to remember that the thread in which sendNotifyContextRequest() lives is already independent...

kzangeli: Ah, if so ... then I can remove the whole thing!
We need to make sure ...

kzangeli: And now I remember ...
If you look at Notifier::sendNotifyContextRequest, you will see that the thread you are talking about is created there, inside the method, to start 'startSenderThread'.
So no, that thread cannot be used.

Another thing is that we could refactor the entire thing and move lots of contents from `sendNotifyContextRequest` into `startSenderThread`.

Which such a refactoring, we would gain in lines of code but nothing more.
I'm not saying it's not interesting,

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +50,302 @@ Notifier::~Notifier (void)
...
+        return;
+      }
+      pthread_detach(tid);
+#endif
+      return;
+    }
+
 
     //
     // Creating the value of the Fiware-ServicePath HTTP header.
```
fgalan: Fiware-Service and Fiware-ServicePath headers has to be included in the templatized notifications, but this part if after the `if (httpInfo.extended)` branch in the logic. Maybe `if (httpInfo.extended)` should be move down? In fact, it should be probably "delayed" until the final HTTP request (I mean, just before or near this point: https://github.com/telefonicaid/fiware-orion/pull/2139/files/b54c884946adbab4955b825b9c2afe65d24e214b#diff-4203db90bc90ad52ab2e6fbeccfdac1eL126)

kzangeli: Not following really ... in templateNotify, `tenant` and `ce.entityId.servicePath` are sent to httpRequestSend, so that is already taken care of.  Also, the service path 'down here' is a possible concatenation of various service paths, so it cannot be used for individual notifications ... right?

fgalan: Makes sense.

NTC

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Moved to separate module **substitute.h/cpp** in common lib - in 197bb27

fgalan: Maybe 'substitude' is very generic for a module name... what about macroSubscritions ?

kzangeli: not that I am very fond of the name 'substitute', but, it doesn't necessarily have to do with subscriptions.
The only thing the function does (well, will do in the future) is to substitute ${X} for something else ...

It is not **that** generic, as **ContextElement** is input to the function. If that had been a map instead, with key->value, then it would be 100% generic.

However, I would not use the word 'subscription' is the module name. Subscriptions just happens to be the only user to this more generic function. [ not that I can find a better name ... ]

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Short explanation in dc500c7

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 197bb27

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 197bb27

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 197bb27

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 197bb27

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 197bb27  (relese() and delete of ncrP - to reuse would require a modification of the signature of the function)

test/unittests/commonMocks.h
kzangeli: Fixed in 197bb27   (not a FIXME, but the rest of the comparisons)

test/unittests/mongoBackend/mongoNotifyContext_test.cpp
kzangeli: Fixed in 197bb27

src/lib/ngsiNotify/Notifier.cpp
fgalan: I mean 'macro**Substitutions**'... sorry for the typo ;)

src/lib/rest/Verb.cpp
fgalan: Maybe a reference to the HTTP RFC from which this has been taken could be a good idea.

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Sending in `renderFormatToString(renderFormat)`, in 197bb27

src/lib/ngsiNotify/Notifier.cpp
kzangeli: function and file names changed to macroSubstitute in 82da3a2

src/lib/rest/Verb.cpp
kzangeli: fixed in 824e399

src/lib/cache/subCache.cpp
kzangeli: Fixed in another branch, merged to develop then merged into this branch, and into its 'parent'

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
+  macroSubstitute(&url, httpInfo.url, ce);
+
+
+  //
+  // 3. Payload
+  //
+  if (httpInfo.payload == "")
+  {
+    payload = "{" + ce.toJson(renderFormat, attrsOrder) + "}";
+    LM_W(("KZ: payload: %s", payload.c_str()));
```
crbrox: Is this trace ok as it is?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
+
+  //
+  // 8. Send the request
+  //
+  //    NOTE: the HTTP headers are sent to httpRequestSend via parameter 'extraHeaders'
+  //
+  std::string  out;
+  int          r;
+
+  LM_W(("KZ: Sending notification: %s [%s]", uri.c_str(), payload.c_str()));
```
crbrox: is the trace right that way ?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
+*/
+void* sendNotifyContextRequestAsPerTemplate(void* p)
+{
+  NotificationAsTemplateParams* paramP = (NotificationAsTemplateParams*) p;
+
+  LM_W(("KZ: In sendNotifyContextRequestAsPerTemplate (%d contextElementResponses)", paramP->ncrP->contextElementResponseVector.size()));
+
+  for (unsigned int ix = 0; ix < paramP->ncrP->contextElementResponseVector.size(); ++ix)
+  {
+    LM_W(("KZ: calling templateNotify (%d)", ix));
```
crbrox: trace ok?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
   const std::string&               fiwareCorrelator,
   RenderFormat                     renderFormat,
   const std::vector<std::string>&  attrsOrder
 )
 {
-    ConnectionInfo ci;
+    ConnectionInfo  ci;
+    Verb            verb = httpInfo.verb;
+
+    LM_W(("KZ: In sendNotifyContextRequest"));
```
crbrox: check all right ...

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
+    //   - if 'headers' is non-empty, perform eventual substitutions and make sure the information is added as HTTP headers for the notification
+    //   - if 'payload' is given, use that string as template instead of the default payload string, substituting all fields that are to be substituted
+    //   - if 'method' is given, then a custom HTTP method is used (instead of POST, which is default)
+    //
+    // Redirect to the method sendNotifyContextRequestAsPerTemplate() when 'httpInfo.extended' is TRUE.
+    // 'httpInfo.extended' is FALSE by default and set to TRUE by the json parser.
+    //
+    if (httpInfo.extended)
+    {
+      LM_W(("KZ: Calling sendNotifyContextRequestAsPerTemplate"));
```
crbrox: trace to check

src/lib/ngsiNotify/Notifier.cpp
kzangeli: This PR is 'temporal', not going to develop.
I am still using my traces, they will go away once 'definitive'

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Same here, and in the next few comments (see my PR description)

src/lib/ngsiNotify/Notifier.cpp
fgalan: Not fully sure about the overall way of dealing with threads in Notifier... but let this as it is by the moment. NTC.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -50,32 +51,312 @@ Notifier::~Notifier (void)
...
+                      "application/json",
+                      payload,
+                      fiwareCorrelator,
+                      renderFormatToString(renderFormat),
+                      true,                // Use Rush if CLI '--rush' allows it
+                      true,                // wait for response
+                      &out,
+                      headers,
+                      "application/json",  // Accept Format
+                      -1);                 // Timeout in milliseconds, depends on CLI '--httpTimeout'
```
fgalan: Actually, the field is named `-httpTimeout`. Note that double '-' is not used in the CLI (except for `--version`).

2 occurrences (this and other in another place).

crbrox:
LGTM

src/lib/ngsiNotify/Notifier.cpp
crbrox: NTC

src/lib/ngsiNotify/Notifier.cpp
crbrox: NTC

src/lib/ngsiNotify/Notifier.cpp
crbrox: NTC

src/lib/ngsiNotify/Notifier.cpp
crbrox: NTC

fgalan:
if (httpTimeut comments are fixed) then LGTM

:)

PR #2143, opened 2016-05-12
New signature for subCacheItemInsert

kzangeli:
HttpInfo as parameter to subCacheItemInsert, instead of just the url string

[ make test running, ready by 16.35, if all goes smoothly ]

### DIFF
--- a/src/lib/mongoBackend/mongoSubscribeContext.cpp
+++ b/src/lib/mongoBackend/mongoSubscribeContext.cpp
@@ -223,7 +223,6 @@ HttpStatusCode mongoSubscribeContext
 
     LM_T(LmtSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
 
-    // FIXME Px: In the near future, the subscription carries the HttpInfo
     ngsiv2::HttpInfo httpInfo;
 
     httpInfo.url = requestP->reference.get();
--- a/src/lib/mongoBackend/mongoSubscribeContext.cpp
+++ b/src/lib/mongoBackend/mongoSubscribeContext.cpp
@@ -29,6 +29,7 @@
 #include "common/globals.h"
 #include "common/defaultValues.h"
 #include "common/sem.h"
+#include "apiTypesV2/HttpInfo.h"
 #include "alarmMgr/alarmMgr.h"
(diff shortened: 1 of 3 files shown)


src/lib/mongoBackend/mongoSubscribeContext.cpp
```diff
@@ -222,10 +223,15 @@ HttpStatusCode mongoSubscribeContext
 
     LM_T(LmtSubCache, ("inserting a new sub in cache (%s)", oidString.c_str()));
 
+    // FIXME Px: In the near future, the subscription carries the HttpInfo
```
fgalan: NGSIv1 wouldn't support any of the fields in httpInfo, except url. Thus, I guess that with the wrapping in line 229 is enought.

I mean, all ok, except this comment :) that should be removed.

kzangeli: ok, you mean that with:

    httpInfo.url = requestP->reference.get();
    
    if (apiVersion == "v2")
    {
        httpInfo.xxx = yyy;
        etc
    }

... everything is clear enough ...

Howver, what I meant with this comment is that right now, the HttpInfo is 'simulated', instead of being input to this function.
Are we on the same page?

fgalan: Not exactly... what I mean is that after the refactoring we are doing righ now to create independent mongoCreateSubscription() and mongoUpdateSubscription() methods for NGSIv2 all the `if (apiVersion == "v2")` is no longer needed.

Thus, mongoSubscribreContext() will be used only in the NGSIv1 case and, under that circustances, you only need 

```
httpInfo.url = requestP->reference.get();
```

now and forever :)

fgalan: Bottom line: just removing the comment would make everything ok.

kzangeli: ah, ok ... no probs :-)

kzangeli: Fixed in b64068b

fgalan:
LGTM

crbrox:
LGTM

PR #2144, opened 2016-05-13
Refactor subscription logic in NGSIv2

fgalan:
This PR is a deep refactor create/update subscription logic in NGSIv2. Parsing and mongoBackend logic has been rewritten (in the later case, new function mongoCreateSubscription() and mongoUpdateSubscription() exists, independent from the corresponding ones in NGSIv1).

make unit_test and make ft pass ok.

This is a combinated work by @crbrox and @fgalan. Thus, triple-LGTM is required:

@kzangeli 
@crbrox 
@fgalan

A important point to remark is that .test weren't modified (except some error message texts). That provides a good guarantee that the refactor hasn't modified the functional behavior of the NGSIv2 subscription management logic.


src/lib/apiTypesV2/ngsiWrappers.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "apiTypesV2/Subscription.h"
+#include "ngsi/EntityIdVector.h"
+#include "ngsi/AttributeList.h"
+#include "ngsi/NotifyConditionVector.h"
+#include "ngsi/NotifyCondition.h"
+
+using namespace ngsiv2;
+
+// FIXME: fill with a cut-paste from #if 1 in the MongoGlobal.cpp file
```
fgalan: The three functions in this `#if 1` should be moved here https://github.com/telefonicaid/fiware-orion/pull/2144/files#diff-4dffbb65c790f76c51c29cf0ca65c978R806 (removing `static` of course). Once moved, the `#if 1`blocks in subCache.cpp and MongoGlobal.cpp has to be removed and the two lines in CMakeList.txt uncommented.

This was my first attempt and everything seems to work find... until the linker step.

kzangeli: We had another problem really similar to this one a few days ago ...

In the top-level CMakeLists.txt, the library apiTypesV2 comes before mongoBackend in the list of libraries (a variable called ORION_LIBS).

mongoBackend is the only user of these three functions and they are in a separate module (ngsiWrappers.h/cpp), so when the linker gets to apiTypesV2, nobody is using any function/variable in ngsiWrappers and the module is thrown away, considered garbage. Later, the linker gets to mongoBackend that actually needs the module (ngsiWrappers) but it is too late, it has already been discarded.

Fix:  put apiTypesV2 after mongoBackend in ORION_LIBS (in top-level CMakeLists.txt).

kzangeli: Fixed in 67769ba

kzangeli: Forgot to remove the copis from subCache.cpp ...
Also, the function `attrsStdVector2AttributeList` is not needed, the already existing method `AttributeList::fill()` does the same thing, so `attrsStdVector2AttributeList` has been removed and now `AttributeList::fill()` is used instead.

Fixed in 2b07248 (and `make test` OK)

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -382,6 +382,7 @@ HttpStatusCode mongoUpdateContextSubscription
 
     //
     // Update sub-cache
+    // FIXME PR: this is safe without sem?
```
fgalan: @kzangeli pay special attention to this. Some lines below the cache is accesses without sem take.

Maybe we have talked about this in the past (I have a kind of deja-vu... :). Maybe it is safe to to in this way, but al least a comment should be added about it.

(Similar code in the new mongoUpdateSubscription() function)

CMakeLists.txt
```diff
@@ -133,14 +133,15 @@ SET (ORION_LIBS
     rest
     jsonParse
     jsonParseV2
-    apiTypesV2
+    rest
```
fgalan: `rest` appears twice... here and in line 133 above. Is that correct?

src/lib/apiTypesV2/Subscription.h
```diff
@@ -41,28 +44,51 @@ struct EntID
   std::string idPattern;
   std::string type;
   std::string toJson();
+
+  EntID(std::string idA, std::string idPatternA, std::string typeA):
```
kzangeli: const std::string&   ?

src/lib/apiTypesV2/ngsiWrappers.cpp
```diff
@@ -0,0 +1,95 @@
...
+  ncVP->push_back(nc);
+}
+
+
+/* ****************************************************************************
+*
+* attrsStdVector2AttributeList -
+*
+*/
+void attrsStdVector2AttributeList(const std::vector<std::string>& attrs, AttributeList* attrLP)
```
fgalan: @kzangeli if I'm remembering correctly, one of these 3 wrappers was not used at the end. I'm not sure, but I'd say it was `attrsStdVector2AttributeList()`. Could you confirm?

src/lib/apiTypesV2/Subscription.h
```diff
@@ -41,28 +44,51 @@ struct EntID
...
+    id(),
+    idPattern(),
+    type()
+  {}
+
 };
 
+inline bool operator==(const EntID& lhs, const EntID& rhs)
+{
+  return lhs.id == rhs.id && lhs.idPattern == rhs.idPattern && lhs.type == rhs.type;
```
kzangeli: I would add some parenthesis here:

    return () && () && ();

src/lib/cache/subCache.cpp
```diff
@@ -30,12 +30,19 @@
...
 #include "common/string.h"
 #include "apiTypesV2/HttpInfo.h"
+#include "apiTypesV2/ngsiWrappers.h"
 #include "mongoBackend/MongoGlobal.h"
 #include "mongoBackend/mongoSubCache.h"
 #include "ngsi10/SubscribeContextRequest.h"
 #include "cache/subCache.h"
 #include "alarmMgr/alarmMgr.h"
 
+#include "apiTypesV2/Subscription.h" // FIXME PR: remove after moving things to ngsiWrapper
```
fgalan: Old this stuff should be needed after "fixing" ngsiWrapper.cpp/h. I'll remove this.

src/lib/apiTypesV2/SubscriptionUpdate.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "apiTypesV2/Subscription.h"
+
+namespace ngsiv2
+{
+
+  class SubscriptionUpdate : public Subscription
+  {
+  public:
+
```
kzangeli: Unnecessary empty line

src/lib/apiTypesV2/SubscriptionUpdate.h
```diff
@@ -0,0 +1,35 @@
...
+    SubscriptionUpdate():
+      descriptionProvided  (false),
+      subjectProvided      (false),
+      expiresProvided      (false),
+      statusProvided       (false),
+      notificationProvided (false),
+      attrsFormatProvided  (false),
+      throttlingProvided   (false)
+      {}
+
```
kzangeli: Unnecessary empty line

src/lib/apiTypesV2/ngsiWrappers.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+  for (unsigned int ix = 0; ix < attrs.size(); ix++)
+  {
+    attrLP->push_back(attrs[ix]);
+  }
+}
+
```
kzangeli: I'm pretty sure I removed attrsStdVector2AttributeList (one if the three wrapper functions).
Did it come back?

src/lib/jsonParseV2/jsonRequestTreat.cpp
```diff
@@ -122,25 +123,30 @@ std::string jsonRequestTreat
...
     {
-      alarmMgr.badInput(clientIp, "invalid subscription");
-      return answer;
+
+      answer = parseSubscription(ciP, &parseDataP->subsV2);
+      if (answer != "OK")
+      {
+        return answer;
+      }
+      /*
```
fgalan: Leftover?

src/lib/cache/subCache.cpp
```diff
@@ -799,6 +806,71 @@ void subCacheItemInsert
...
+)
+{
+
+  NotifyConditionVector ncV;
+  EntityIdVector        enV;
+  AttributeList         attrL;
+
+  attrsStdVector2NotifyConditionVector(condAttributes, &ncV);
+  entIdStdVector2EntityIdVector(entities, &enV);
+  attrL.fill(notifAttributes);
```
kzangeli: Here is where attrsStdVector2AttributeList is no longer used, as the functionality is covered by `AttributeList::fill()`

src/lib/cache/subCache.cpp
```diff
@@ -799,6 +806,71 @@ void subCacheItemInsert
...
+                     stringFilterP,
+                     status,
+                     q,
+                     geometry,
+                     coords,
+                     georel);
+
+  enV.release();
+  ncV.release();
+
```
kzangeli: Unnecessary empty line

src/lib/cache/subCache.cpp
```diff
@@ -799,6 +806,71 @@ void subCacheItemInsert
...
+}
+
+
+
+
```
kzangeli: Three empty lines between functions, not four

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -30,50 +30,44 @@
...
+#include "common/string.h"
 #include "rest/ConnectionInfo.h"
 #include "rest/OrionError.h"
-#include "ngsi/ParseData.h"
 #include "ngsi/Request.h"
 #include "jsonParseV2/jsonParseTypeNames.h"
 #include "jsonParseV2/jsonRequestTreat.h"
+#include "jsonParseV2/utilsParse.h"
+#include "rest/Verb.h"
+
```
kzangeli: Unnecessary empty lines

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -30,50 +30,44 @@
...
-static std::string parseNotification(ConnectionInfo* ciP, SubscribeContextRequest* scrP, const Value& notification);
-static std::string parseSubject(ConnectionInfo* ciP, SubscribeContextRequest* scrP, const Value& subject);
-static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP, const Value& entities);
-static std::string parseNotifyConditionVector(ConnectionInfo* ciP, SubscribeContextRequest* scrP, const Value& condition);
-
+static std::string parseNotification(ConnectionInfo* ciP, SubscriptionUpdate* subsP, const Value& notification);
+static std::string parseSubject(ConnectionInfo* ciP, SubscriptionUpdate* subsP, const Value& subject);
+static std::string parseEntitiesVector(ConnectionInfo* ciP, std::vector<EntID>* eivP, const Value& entities);
+static std::string parseNotifyConditionVector(ConnectionInfo* ciP, SubscriptionUpdate* subsP, const Value& condition);
+static std::string error(ConnectionInfo* ciP, const std::string& msg);
```
kzangeli: error ... perhaps too a generic name?

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -462,73 +380,172 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
...
+    }
+
+
+    // qs
+    if (httpExt.HasMember("qs"))
+    {
+      const Value& qs = notification["qs"];
+      if (!qs.IsObject())
+      {
+        return error(ciP, "notificaction httpExt qs is not an object");
```
fgalan: If httpExt refers to "the extended alternative to the `http` field in `notificationsp" the right name is `httpExtended` (see https://github.com/telefonicaid/fiware-orion/pull/2125).

Several ocurrences of `httpExt`. Overall search should be done.

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -164,128 +144,102 @@ std::string parseSubscription(ConnectionInfo* ciP, ParseData* parseDataP, JsonDe
...
+  Opt<std::string> expiresOpt = getStringOpt(document, "expires");
 
-    if (!expires.IsString())
-    {
-      OrionError oe(SccBadRequest, "expires is not a string");
-
-      alarmMgr.badInput(clientIp, "expires is not a string");
-      return oe.render(ciP, "");
-    }
+  if (!expiresOpt.ok())
```
kzangeli: We have many methods already for checking validity of parts of the payload. All with the name `check`. I would propose to keep using that name.

src/lib/apiTypesV2/ngsiWrappers.cpp
fgalan: I'll have a look.

src/lib/jsonParseV2/parseSubscription.cpp
fgalan: On the contrary, using "new fresh" checking methods would allow to have a more clear separation between NGSIv1 and NGSIv2 that could be benefitial in the long run.

To be discussed at skype.

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -355,101 +299,75 @@ static std::string parseSubject(ConnectionInfo* ciP, SubscribeContextRequest* sc
...
-      {
-        alarmMgr.badInput(clientIp, "subject entities element id is not a string");
-        OrionError oe(SccBadRequest, "subject entities element id is not a string");
+      id = idOpt.value;
 
-        return oe.render(ciP, "");
+      if (id.empty())
+      {
+        idPattern = ".*";
       }
```
kzangeli: So, if `id` is given as empty string, we change it to **all** entities?
I didn't know, but it makes sense, analogous to empty `type`.
Is this something new?

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -355,101 +299,75 @@ static std::string parseSubject(ConnectionInfo* ciP, SubscribeContextRequest* sc
...
-
     std::string  id;
+    std::string  idPattern;
     std::string  type;
-    std::string  isPattern  = "false";
 
-    if (iter->HasMember("id"))
     {
-      if ((*iter)["id"].IsString())
+      Opt<std::string> idOpt = getStringOpt(*iter,"id", "subject entities element id");
```
kzangeli: space after comma

src/lib/jsonParseV2/utilsParse.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "rapidjson/document.h"
+
+#include "utilsParse.h"
+
+using namespace rapidjson;
+
+using std::string;
+
+static Opt<string> getStringAux(const Value& parent, const char* field, const string& description, bool optional) {
```
fgalan: Functions in this .cpp should include the comment preamble, e.g:

```
*
* error -
*
*/
```

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -355,101 +299,75 @@ static std::string parseSubject(ConnectionInfo* ciP, SubscribeContextRequest* sc
...
+      if (id.empty())
+      {
+        idPattern = ".*";
       }
     }
-    if (iter->HasMember("idPattern"))
+
     {
-      if ((*iter)["idPattern"].IsString())
+      Opt<std::string> idPatOpt = getStringOpt(*iter,"idPattern", "subject entities element idPattern");
```
kzangeli: space after comma

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -355,101 +299,75 @@ static std::string parseSubject(ConnectionInfo* ciP, SubscribeContextRequest* sc
...
       }
-      else
+      else if (idPatOpt.given)
       {
-        alarmMgr.badInput(clientIp, "subject entities element idPattern is not a string");
-        OrionError oe(SccBadRequest, "subject entities element idPattern is not a string");
-
-        return oe.render(ciP, "");
+        idPattern = idPatOpt.value;
       }
```
kzangeli: Is the case "idPattern": "" taken care of?   (perhaps inside ok()).
It should be an error, right?  (either that or transformed to ".*" ...)

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -462,73 +380,172 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
...
+      Opt<std::string> urlOpt = getStringMust(http, "url", "url http notification");
 
-      alarmMgr.badInput(clientIp, "http notification is not an object");
-      return oe.render(ciP, "");
+      if (!urlOpt.ok())
+      {
+        return error(ciP, urlOpt.error);
+      }
 
+      {
```
kzangeli: Any reason the create a new block here ?   (and later)

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -462,73 +380,172 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, EntityIdVector* eivP
...
-      return oe.render(ciP, "");
+      Opt<std::string> methodOpt = getStringMust(httpExt, "method", "method httpExtended notification");
+
+      if (!methodOpt.ok())
+      {
+        return error(ciP, methodOpt.error);
+      }
+      Verb  verb         = str2Verb(methodOpt.value);
+      /*
+       *  CHECK IT IS VALID
```
kzangeli: This check should probably be implemented before the PR is merged

src/lib/jsonParseV2/utilsParse.cpp
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+#include "rapidjson/document.h"
+
+#include "utilsParse.h"
+
+using namespace rapidjson;
+
+using std::string;
+
```
kzangeli: Function headers, 3 blank lines between functions

src/lib/apiTypesV2/SubscriptionUpdate.h
crbrox: Fixed fb156d0

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: Fixed fb156d0

CMakeLists.txt
kzangeli: Not a problem.
Could be better, as we seem to have a cross dependency between libraries.
However, it's no problem having the linker go over a library twice.

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -0,0 +1,358 @@
...
+* along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+*
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
+* Author: Fermín Galán
+*
+*/
+
+#include "mongoBackend/MongoCommonSubscription.h"
```
kzangeli: It's true that this is a 'special' include, but I would still put it with its 'friends' from mongoBackend.
Also, the extra blank line after the func header is not necessary.

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  if (httpInfo.payload != "")
+  {
+    b->append(CSUB_PAYLOAD, httpInfo.payload);
+    LM_T(LmtMongo, ("Subscription payload: %s", httpInfo.payload.c_str()));
+  }
+}
+
+/* ****************************************************************************
```
kzangeli: 3 blank lines ...

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -0,0 +1,358 @@
...
+  for (unsigned int ix = 0; ix < sub.subject.entities.size(); ++ix)
+  {
+    EntID en = sub.subject.entities[ix];
+
+    if (en.type == "")
+    {
+      if (en.id != "")
+      {
+        entities.append(BSON(CSUB_ENTITY_ID << en.id << CSUB_ENTITY_ISPATTERN << "false"));
+
```
kzangeli: empty line

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -0,0 +1,358 @@
...
+*
+* setEntities -
+*
+*/
+void setEntities(const Subscription& sub, BSONObjBuilder* b)
+{
+  BSONArrayBuilder entities;
+  for (unsigned int ix = 0; ix < sub.subject.entities.size(); ++ix)
+  {
+    EntID en = sub.subject.entities[ix];
```
kzangeli: This here is an object copy ... not possible to use EntID* ?

src/lib/mongoBackend/MongoCommonSubscription.h
```diff
@@ -0,0 +1,172 @@
...
+);
+
+
+
+/* ****************************************************************************
+*
+* setLastNotification -
+*
+*/
+void setLastNotification(long long lastNotification, BSONObjBuilder* b);
```
kzangeli: extern

src/lib/mongoBackend/MongoCommonSubscription.h
```diff
@@ -0,0 +1,172 @@
...
+/* ****************************************************************************
+*
+* setFormat -
+*
+*/
+extern void setFormat(const ngsiv2::Subscription& sub, mongo::BSONObjBuilder* b);
+
+
+#endif // SRC_LIB_MONGO_BACKEND_MONGOCOMMONSUBSCRIPTION_H
+
```
kzangeli: empty line

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: It's a internal (static) function, only visible inside this module. It would be more related to parsing than to rendering. But we can change its name whatever we think it's clearer ...

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Yes, it seems calls to semCacheTake/Give are missing here

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -0,0 +1,608 @@
...
+#include "mongoBackend/MongoGlobal.h"
+#include "mongoBackend/MongoCommonSubscription.h"
+#include "mongoBackend/dbConstants.h"
+#include "mongoBackend/safeMongo.h"
+#include "mongoBackend/mongoSubCache.h"
+#include "common/defaultValues.h"
+
+#include "cache/subCache.h"
+
+#include "mongo/client/dbclient.h"
```
kzangeli: This include should go right after C++ headers, before logMsg, as 3rd party

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -0,0 +1,608 @@
...
+*/
+static void setThrottling(const SubscriptionUpdate& subUp, const BSONObj& subOrig, BSONObjBuilder* b)
+{
+  if (subUp.throttlingProvided)
+  {
+    setThrottling(subUp, b);
+  }
+  else
+  {
+    if (subOrig.hasField(CSUB_THROTTLING))
```
kzangeli: Why not ...

    else if (subOrig.hasField(CSUB_THROTTLING))

 ... and we save two lines ... ?

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -0,0 +1,608 @@
...
+  const BSONObj&                   subOrig,
+  const std::string&               tenant,
+  const std::vector<std::string>&  servicePathV,
+  const std::string&               xauthToken,
+  const std::string&               fiwareCorrelator,
+  BSONObjBuilder*                  b,
+  bool*                            notificationDone
+)
+{
+  // notification can be changed by true by setCondsAndInitialNotify()
```
kzangeli: changed **to** true ?

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -0,0 +1,608 @@
...
+    // If the cached value of lastNotificationTime is higher, then use it.
+    //
+    if (subCacheP != NULL && (subCacheP->lastNotificationTime > lastNotification))
+    {
+      lastNotification = subCacheP->lastNotificationTime;
+    }
+
+    setLastNotification(lastNotification, b);
+  }
+
```
kzangeli: Unnecessary empty line

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -0,0 +1,608 @@
...
+* setLastNotification -
+*
+*/
+static void setLastNotification(const BSONObj& subOrig, CachedSubscription* subCacheP, BSONObjBuilder* b)
+{
+  // FIXME P1: if CSUB_LASTNOTIFICATION is not in the original doc, it will also not be in the new doc.
+  //           Is this necessary?
+  //           The implementation would get a lot simpler if we ALWAYS add CSUB_LASTNOTIFICATION and CSUB_COUNT
+  //           to 'newSub'
+  if (subOrig.hasField(CSUB_LASTNOTIFICATION))
```
kzangeli: Here you could do:

    if (!subOrig.hasField(CSUB_LASTNOTIFICATION))
    {
      return
    }

And you save the extra intentation for the rest

src/lib/apiTypesV2/Subscription.h
crbrox: fixed aa861cf

src/lib/apiTypesV2/Subscription.h
crbrox: Fixed aa861cf

src/lib/rest/Verb.cpp
```diff
@@ -34,17 +34,58 @@ const char* verbName(Verb verb)
...
+  case PUT:        return "PUT";
+  case POST:       return "POST";
+  case DELETE:     return "DELETE";
+  case PATCH:      return "PATCH";
+  case HEAD:       return "HEAD";
+  case OPTIONS:    return "OPTIONS";
+  case TRACE:      return "TRACE";
+  case CONNECT:    return "CONNECT";
+  case NOVERB:     return "NOVERB";
+  case UNKOWNVERB: return "UNKONWVERB";
```
kzangeli: Typo: UNKOWNVERB => UNK**N**OWNVERB

src/lib/rest/Verb.cpp
```diff
@@ -34,17 +34,58 @@ const char* verbName(Verb verb)
...
   }
 
   return "Unknown verb";
 }
+
+/* ****************************************************************************
+*
+* str2Verb -
+*/
+Verb str2Verb(const std::string& str) {
```
kzangeli: { on its own line

src/lib/rest/Verb.cpp
```diff
@@ -34,17 +34,58 @@ const char* verbName(Verb verb)
...
+  }
+  else if (str == "CONNECT") {
+    return CONNECT;
+  }
+  else if (str == "NOVERB") {
+    return NOVERB;
+  }
+
+  return UNKOWNVERB;
+
```
kzangeli: Unwanted empty line

src/lib/apiTypesV2/SubscriptionUpdate.h
crbrox: Fixed 47d2d6e

src/lib/apiTypesV2/SubscriptionUpdate.h
crbrox: fixed 47d2d6e

test/functionalTest/cases/1316_create_subscription/post_subs_errors.test
```diff
@@ -243,13 +243,13 @@ echo
...
 HTTP/1.1 400 Bad Request
-Content-Length: 84
+Content-Length: 93
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
-    "description": "subject entities element has not id/idPattern",
+    "description": "subject entities element has not id/idPattern nor type",
```
kzangeli: This is bad English (not from this PR, but from before). Why not change it now that were here ... ?

    subject entities element has no id/idPattern nor type

src/lib/apiTypesV2/ngsiWrappers.cpp
kzangeli: Yes, I sent a comment about this.
I was surprised to find it here, I'm pretty sure I made a commit removing that function ...

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: Sure, it's static and its name is not very important as it is not visible outside of the source file.
However, I saw it and I was far from sure what it did.

Good idea to create the function though, a lot of repeated code has gotten removed thanks to it

CMakeLists.txt
fgalan: However, repetition is a bit uncommon in our CMakeList.txt (maybe this is the first time we do it?). 

Thus, I'd suggest to add a short comment line to one of the rest occurrences to explain why it is duplicated. Otherwise, we have the risk to forget this, wrongly removing one of them in a future refactor, thus causing problems.

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: Ah, and careful with notificaction ...
A little better that nofitication, but still :-)
It's a very difficult word ... :-)

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: It is the include corresponding to the .h associated to this .cpp file. I think it should be the first one in the file. Separation from the other includes seems to be also a good idea to "remark" that it is special.

Not sure if we should have something in the style guide about this...

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: EntID is a "package" of three strings. Thus, I see it as a "light" copy. Not important enough to change the logic of this function (although if we want to change something, that should be probably the `entities` vector in the `Subject` class, using EntID* instead of EntID).

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: Ok. Is that mentioned in the style guide that 3rd party goes after STL-and-the-like includes?

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: I prefer current way, to have the structure of all set*() functions in this file homogeneous. See for example a case the suggested strategy cannot be used: setHttpInfo().

src/lib/apiTypesV2/ngsiWrappers.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/apiTypesV2/ngsiWrappers.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/cache/subCache.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/cache/subCache.cpp
fgalan: Ok. NTC.

src/lib/cache/subCache.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/cache/subCache.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/rest/Verb.cpp
crbrox: fixed e6a1f7e

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/mongoBackend/MongoCommonSubscription.h
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/mongoBackend/MongoCommonSubscription.h
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/rest/Verb.cpp
crbrox: fixed e6a1f7e

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261. At the end this include was not actually needed, so it has been removed.

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: Fixed in a83018c07aaddda04d94a8f6cff35b3187ba5261

src/lib/rest/Verb.cpp
crbrox: N/A changes in e6a1f7e

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
fgalan: Issue created (and FIXME PR properly changed in the code): https://github.com/telefonicaid/fiware-orion/issues/2146

test/functionalTest/cases/1316_create_subscription/post_subs_errors.test
crbrox: Fixed 4a5440da3ee4605c5f111b61d83fc864d3fd4f6d

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: fixed 4a5440da3ee4605c5f111b61d83fc864d3fd4f6d

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: fixed 4a5440da3ee4605c5f111b61d83fc864d3fd4f6d

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: fixed 4a5440da3ee4605c5f111b61d83fc864d3fd4f6d

src/lib/mongoBackend/MongoCommonSubscription.cpp
kzangeli: I always put the 'module include' as the **last** include.
The include files from its own library should always be the last ones, and that is the 'group' it belongs to.
And yes, we should probably mention this is the style guide.

src/lib/mongoBackend/MongoCommonSubscription.cpp
kzangeli: change the logic of this function ...
I only propose the use EntID* and -> insetad of EntID and `.`.
It hardly changes the logic of the function.
Do we **really** want to just throw away time like this?

And yes, having the vector use EntID* instead of EntID is a very good idea. We have seen that no less that three copies are made of the objects when a vector is made of objects instead of pointers. This we should always avoid (IMO).

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: The style guide mentions to start with std-c headers, then std-c++  headers. It might also say something about 3rd party headers. It definitely should.
It should also say that low level includes (from low level libs inside the repo) should come before high level and that the 'same library includes' should be the last ones

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: Not important

NTC

src/lib/jsonParseV2/utilsParse.cpp
crbrox: fixed d04a478

src/lib/jsonParseV2/utilsParse.cpp
crbrox: fixed d04a478f3c85261d4d4bc236f1fd0a773eca6213

CMakeLists.txt
kzangeli: Also, if the name of the function that is causing the problem is mentioned, then it will be easier to fix the problem later (assuming we have a case of cross dependencies in libraries)

fgalan: Good idea to include the name of the function... in the case we already know it. If knowning it requires research that could delay merging this PR, I'd suggest jus to include a "neutral" comment without mentioning it.

src/lib/jsonParseV2/jsonRequestTreat.cpp
crbrox: fixed 3d458d8

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: discussed at Skype
NTC

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: discussed at Skype (some validations are missing until a future PR)
NTC

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: This needs a more in-deep discussion, involving how Subscription is defined. Thus, I'd leave as it is now in the PR. If in the future Subscription changes, this would automatically change ("automatically" in the sense that changes in Subscription class would require changes in this part in order to being able to build).

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: Fixed in ac30916

src/lib/mongoBackend/MongoCommonSubscription.cpp
kzangeli: OK, I see you really want that copy there ...
Not extremely important

NTC

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: fixed d0a6c4f

src/lib/mongoBackend/MongoCommonSubscription.cpp
fgalan: That's correct, but, just to be sure that the reason behind it is clear: I don't want the copy due to the copy itself, I want the copy as it is aligned with the way in which Subscription class is (currently) defined.

CMakeLists.txt
crbrox: fixed 8129bc0

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: Fixed 73d284c

src/lib/mongoBackend/MongoCommonSubscription.cpp
kzangeli: I think I have seen a line like this somewhere.

    EntID& en = sub.subject.entities[ix];

I wasn't sure what was accomplished with that '&' (it is pretty cryptic), but ... perhaps this avoids the copy for us and exactly everything else remains the same ... ?

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: So. let's use `ok()`

NTC

fgalan:
LGTM

kzangeli:
LGTM

crbrox:
LGTM

PR #2150, opened 2016-05-18
FIX accumulator-server.py able to accumulate query params

fgalan:
I haven't tested extensively but after a quick chec it seems to work as expected:

```
POST http://localhost:1028/accumulate?x=1&b=2
Content-Length: 
User-Agent: curl/7.38.0
Host: localhost:1028
Accept: */*
Content-Type: 

::1 - - [18/May/2016 12:50:20] "POST /accumulate?x=1&b=2 HTTP/1.1" 200 -
```

### DIFF
--- a/scripts/accumulator-server.py
+++ b/scripts/accumulator-server.py
@@ -167,7 +167,20 @@ def record():
     #
     #  request.url = request.scheme + '://' + request.host + request.path
     #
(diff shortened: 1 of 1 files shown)


kzangeli:
Idea stolen and used in another branch (feature 2015 branch)

Issue #1851, opened 2016-02-25
[Pending: csub cache simplification] Database model in post-ONTIMEINTERVAL times


fgalan:
Currently we use the following data model for the csubs collection:

       "conditions" : [
                {
                        "type" : "ONCHANGE",
                        "value" : [
                                "temperature "
                        ]
                }
        ],

However, given that after removing ONTIMEINTERVAL we don't need to distinguish between different types, it could be simplified to just:

       "conditions" : [ "temperature " ],

This of course involves a data migration process (i.e. migration scripts need to be created and tested as part of this issue implementation). Thus, this issue should be scheduled along with other issue requiring database migration to "pack" them all as much as possible in the same Orion version.

**Effort: 2 man day**

fgalan:
In order to align with APIv2, we could use `condition` instead of `conditions` in the DB model.

fgalan:
Code and modifications to database_model.md done in PR #2356.

fgalan:
Migration script (+doc) done at PR https://github.com/telefonicaid/fiware-orion/pull/2363

fgalan:
Most of the work done, although some adjustment at cache structures is required. Look for "FIXME #1851" at mongoSubCache.cpp.

fgalan:
Almost completed by PR https://github.com/telefonicaid/fiware-orion/pull/2483. Some function cleanup pending.

fgalan:
Cleanup completed in PR https://github.com/telefonicaid/fiware-orion/pull/2487

PR #2152, opened 2016-05-19
Refactor of csub create/update NGSIv1 logic to use fresh new NGSIv2 csub create/update backend

fgalan:
Lot of old code removed. In addition, NGSIv1 parts isolation has improved (thus, its future removal will be easier).

It is worth mentioning that **no .test** what touched in this PR, which provides enough guarantee regarding the refactoring.

Unit test however have some modifications that should be reviewed. I'll add some comment about them.

@kzangeli 
@crbrox


test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
```diff
@@ -331,7 +330,7 @@ TEST(mongoSubscribeContext, Ent1_Attr0_C1)
     EXPECT_EQ(id, sub.getField("_id").OID().toString());
     EXPECT_EQ(1360236300, sub.getIntField("expiration"));
     EXPECT_FALSE(sub.hasField("lastNotification"));
-    EXPECT_FALSE(sub.hasField("throttling"));
+    EXPECT_EQ(-1, sub.getIntField("throttling"));
```
fgalan: Slight change in how throttling is persisted, but I doesn't involve a functional e2e change at the end. What do you think?

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
```diff
@@ -56,487 +46,43 @@ HttpStatusCode mongoUpdateContextSubscription
...
   {
-    //
-    // The hasField check is needed as Throttling might not be present in the original doc
-    // FIXME P1: However, we could include Throttling in the new doc anyway ...
-    //
-    if (sub.hasField(CSUB_THROTTLING))
+    // Check OrionError. Depending the error kind, details are included or not in order
+    // to have a better backward compatiblity
+    // FIXME: should we? or it is better modify .test and provide more accurate errors with 'details'?
+    if (oe.code == SccContextElementNotFound)
```
fgalan: Maybe is better not doing exceptions and modifying the .test adding the extra "detail" field. In some way, it would be a kind of "fixing", as now we can provide more info about the error.

What do you think?

src/lib/apiTypesV2/SubscriptionUpdate.h
```diff
@@ -32,23 +32,26 @@ namespace ngsiv2
...
       statusProvided       (false),
       notificationProvided (false),
       attrsFormatProvided  (false),
-      throttlingProvided   (false)
-      {}
+      throttlingProvided   (false),
+      fromNgsiv1           (false)
+      {
+        descriptionProvided = false;
+      }
```
kzangeli: So, descriptionProvided has been removed from the class but a variable `descriptionProvided` is set in the constructor ...  ?

src/lib/apiTypesV2/SubscriptionUpdate.h
```diff
@@ -32,23 +32,26 @@ namespace ngsiv2
...
   public:
-    bool  descriptionProvided;
     bool  subjectProvided;
     bool  expiresProvided;
     bool  statusProvided;
     bool  notificationProvided;
     bool  attrsFormatProvided;
     bool  throttlingProvided;
 
+    bool  fromNgsiv1;          // to support an special case when the SubscriptionUpdate comes from NGSIv1
```
kzangeli: typo: an => a

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
```diff
@@ -1100,17 +1083,13 @@ TEST(mongoSubscribeContext, EntN_Attr0_CN)
 
     std::vector<BSONElement> conds = sub.getField("conditions").Array();
     std::vector<BSONElement> condValues;
-    ASSERT_EQ(2, conds.size());
+    ASSERT_EQ(1, conds.size());
```
fgalan: Another interesting consequence of this refactoring is that now it "collapse" all attributes in no matter how much condition vectors element into just one. I mean, previous to this PR if we have in the API:

conditions: [
{ type: ONCHANGE, values: [ A, B ] },
{ type: ONCHANGE, values: [ C, D ] },
]

that ended in DB in a similar estructure.

Now, with this PR, that will be stored in DB as:

conditions: [ 
{type: ONCHANGE, value: [ A, B, C, D ]
]

I think this is benefitial, in some way a step in the road to https://github.com/telefonicaid/fiware-orion/issues/1851

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -238,6 +240,89 @@ static void setAttrs(const SubscriptionUpdate& subUp, const BSONObj& subOrig, BS
...
+
+  std::vector<std::string> attributes;
+  std::vector<BSONElement> attrs = getFieldF(subOrig, CSUB_ATTRS).Array();
+  for (unsigned int ix = 0; ix < attrs.size(); ++ix)
+  {
+    attributes.push_back(attrs[ix].String());
+  }
+
+
+  /* Conds vector (and maybe and initial notification) */
```
kzangeli: and => any ?

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
```diff
@@ -2887,7 +2838,7 @@ TEST(mongoSubscribeContext, matchEnt1_AttrN_CN_disjoint)
     attrsFilter.push_back("A2");
 
     EXPECT_CALL(*notifierMock, sendNotifyContextRequest(MatchNcr(&expectedNcr),"http://notify.me", "", "", "no correlator", NGSI_V1_LEGACY, attrsFilter))
-            .Times(2);
+            .Times(1);
```
fgalan: This is another interesting side-effect... Several TEST has changed with regards the number of Times the sendNotifyContextRequest() was hit.

Not sure why 2 was the value before this PR but, as far as I understand, the right thing is 1 time. I mean, as a consequence of a single csub create or update as much as one call should hit sendNotifyContextRequest().

What do you think?

Surprisingly, no .test was needed modification regarding this. Maybe is "internally hiden". 

Not sure if this should be mentioned in CNR... actually, I'm not sure about what to mention exactly :)

src/lib/apiTypesV2/SubscriptionUpdate.h
fgalan: It has been removed from the child class (so I cannot use initialization list), but stills in the parent class.

I mean, descriptionProvided was (wrongly) being used in both chil and parent. I have removed in child, but I cannot use initialization list now (or maybe is possible but I don't know how :)

src/lib/ngsi10/UpdateContextSubscriptionRequest.cpp
```diff
@@ -129,6 +102,42 @@ void UpdateContextSubscriptionRequest::present(const std::string& indent)
...
+  // that 'restriction.release()' would cause double-free problems
+
   notifyConditionVector.release();
 }
+
+
+
+/* ****************************************************************************
+*
+* UpdateContextSubscriptionRequest::release -
```
kzangeli: release => toNgsiv2Subscription

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
kzangeli: So, throttling is always present, but a value of 0 or -1 means throttling is 'off'.

Sounds ok.

src/lib/apiTypesV2/SubscriptionUpdate.h
kzangeli: ok, I see.

NTC

src/lib/mongoBackend/mongoUpdateContextSubscription.cpp
kzangeli: Not very important.
I wouldn't mind giving better error descriptions, but, not very important.

NTC

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
kzangeli: As long as we haven't lost any notifications, all should be good.
Perhaps this is a bugfix ...

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
kzangeli: Sounds good

NTC

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
kzangeli: So, probably a bugfix but as we don't know of any bug, almost better to not mention this in the CNR.
Especially as we wouldn't even know how to describe it ... :-)

NTC

crbrox:
LGTM

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
fgalan: NTC

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
fgalan: NTC

src/lib/apiTypesV2/SubscriptionUpdate.h
fgalan: To be fixed in a further PR.

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: To be fixed in a next PR.

test/unittests/mongoBackend/mongoSubscribeContext_test.cpp
fgalan: NTC

src/lib/ngsi10/UpdateContextSubscriptionRequest.cpp
fgalan: To be fixed in a next PR.

PR #2155, opened 2016-05-19
FIX -noCache now avoids populating cache on create/update subscription operations

### LINKED ISSUE #2147: -noCache must avoid populating cache on create/update subscription operations
It seems that currently `-noCache` has no effect in create/update subscription operation, so cache is always populated, althought at the end is not going to be used. This is unefficient: it incresases the CB memory footprint witout any gain.

fgalan:
Fixes #2147 

@kzangeli

### DIFF
--- a/src/lib/mongoBackend/mongoCreateSubscription.h
+++ b/src/lib/mongoBackend/mongoCreateSubscription.h
@@ -48,7 +48,6 @@ extern std::string mongoCreateSubscription
 (
   const ngsiv2::Subscription&          sub,
   OrionError*                          oe,
-  std::map<std::string, std::string>&  uriParams,
   const std::string&                   tenant,
   const std::vector<std::string>&      servicePathV,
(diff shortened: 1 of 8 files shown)


src/lib/mongoBackend/mongoCreateSubscription.h
```diff
@@ -48,7 +48,6 @@ extern std::string mongoCreateSubscription
 (
   const ngsiv2::Subscription&          sub,
   OrionError*                          oe,
-  std::map<std::string, std::string>&  uriParams,
```
fgalan: In the past we opted for a "have the method signature homogeneous, even when some parameters could not be used in some cases". However, after the experience, I now think that is better to adopt a "each function uses only what it needs" to improve function/method encapsulation. Thus, this change.

NTC (just informative).

kzangeli:
LGTM

PR #2172, opened 2016-05-23
Task/2015 forbidden chars

kzangeli:
Added check for forbidden chars in http info.
Included http info in the rendering of subscriptions (for GET /v2/subscriptions/[<subId]>)
Http info 'payload' field is now decoded on rendering (e.g. %22 is converted to ")

@fgalan,
@crbrox (please look closely at the template function **objectToJson**)


src/lib/mongoBackend/mongoGetSubscriptions.cpp
```diff
@@ -130,46 +130,94 @@ static void setSubject(Subscription* s, const BSONObj& r)
...
+  subP->notification.timesSent        = r.hasField(CSUB_COUNT)?            getIntOrLongFieldAsLongF(r, CSUB_COUNT)            : -1;
+
 
   //
   // Check values from subscription cache, update object from cache-values if necessary
   //
-  CachedSubscription* cSubP = subCacheItemLookup(tenant.c_str(), s->id.c_str());
+  // NOTE: only 'lastNotificationTime' and 'count'
+  //
+  cacheSemTake(__FUNCTION__, "get lastNotification and count");
```
fgalan: Is this fix about taking/giving sem a bugfix worth to mention in CNR?

test/functionalTest/cases/2015_notification_templates/notification_templates_with_decoded_chars_in_notifications.test
```diff
@@ -0,0 +1,179 @@
...
+POST http://127.0.0.1:9997/notify?qs2=a1&qs1=1
+Fiware-Servicepath: /
+Content-Length: 33
+User-Agent: orion/REGEX(\d+\.\d+\.\d+.*)
+Hs1: 1
+Hs2: a1
+Ngsiv2-Attrsformat: custom
+Host: 127.0.0.1:9997
+Accept: application/json
+Content-Type: application/json; charset=utf-8
```
fgalan: This is interesting... passed unoticed until now...

Should we use "Conent-Type: text/plain; charset=utf-8" when payload is used (i.e. httpInfo.payload different from "")? It seems a good idea, given that we cannot guarantee that the final payload is JSON (as in this case).

kzangeli: This is so true ...
Ngsiv2-Attrsformat: custom
There is no way the broker can tell its user the Content-Type.
It should probably be a part of the **httpCustom** of the subscription ...

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: Not sure when this potential bug was introduced.
Was it part of last release?

fgalan: I think this code was last modified before Orion 1.1 (current last version).

However, to be sure you can use "Blame" on that file (in develop branch) to know in which commit (and its associated date) that part of the code was last modified.

test/functionalTest/cases/2015_notification_templates/notification_templates_with_decoded_chars_in_notifications.test
fgalan: > It should probably be a part of the httpCustom of the subscription ...

Actually, it is include in httpCustom, somehow... I mean, the user can use `headers: {  Content-Type: ... }` to set the right content type. However, if that custom header is not provided, CB should use the "lowest common denominator", which is text/plain.

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: Fixed in 28887c9 (line added to CNR)

test/functionalTest/cases/2015_notification_templates/notification_templates_with_decoded_chars_in_notifications.test
kzangeli: Changed mime-type to text/plain for templates with payload given, always modifyable using `headers` in the subscription.

Fixed in 28887c9

test/functionalTest/cases/2015_notification_templates/notification_templates.test
```diff
@@ -337,7 +337,7 @@ User-Agent: orion/REGEX(\d+\.\d+\.\d+.*)
 Ngsiv2-Attrsformat: normalized
 Host: 127.0.0.1:REGEX(\d+)
 Accept: application/json
-Content-Type: application/json; charset=utf-8
+Content-Type: text/plain
```
fgalan: Note this notification is `Ngsiv2-Attrsformat: normalized`. Thus `Content-Type: application/json; charset=utf-8` should be used.

In addition, not sure if we should use "text/plain" or "text/plain; charset=utf-8" for the "custom" default... I'd say that the charset used at text level is the same in both cases, so if we are using charset=utf-8 in JSON case, we should do the same in the text plain case.

fgalan: > Note this notification is Ngsiv2-Attrsformat: normalized. Thus Content-Type: application/json; charset=utf-8 should be used.

Looking to the code (https://github.com/telefonicaid/fiware-orion/pull/2172/files#diff-4203db90bc90ad52ab2e6fbeccfdac1eR109) the logic seems to be correct. This is werid... Maybe some "sorting effect" due to SORT_START / SORT_END ?

fgalan: Regarding the utf-8 stuff, looking to the code we have:

```
  // For content-type application/json we add charset=utf-8
  if (orig_content_type == "application/json")
  {
    content_type += "; charset=utf-8";
  }
```

Not sure why we are doing that exception... maybe is obsolete (coming from the XML times) and should be applied to both defaults application/json and text/plain ?

kzangeli: Yeah, I also don't know why we do this.
I added charset=utf-8 to text/plain for custom notifications however, in commit 2ac853c.

We can take a look at removing (moving) this strange thing adding charset=utf-8 if json, or including text/plain, but perhaps not in this very PR ?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -112,13 +115,15 @@ static bool templateNotify
...
+    payload  = ncr.toJson(renderFormat, attrsOrder);
+    mimeType = "application/json";
   }
   else
   {
     macroSubstitute(&payload, httpInfo.payload, ce);
+    payload      = curl_unescape(payload.c_str(), payload.length());
     renderFormat = NGSI_V2_CUSTOM;
-    LM_W(("KZ: substituted payload: '%s'", payload.c_str()));
+    mimeType     = "text/plain; charset:utf-8";  // May be overridden by 'Content-Type' in 'headers'
```
fgalan: To be homogeneous with L119 for the JSON case, I'd suggest to use `mimeType = "texp/plain";` here, then add to the `if` in httpSendRequest() (or something like that, don't remembering the exact name of the function right now) `orig_content_type == "application/text"` to sum the charset part.

By the way, is `charset=utf-8` not `charset:utf-8`...

kzangeli: Fixed in 220fe84

fgalan:
LGTM

PR #2179, opened 2016-05-24
Remove smallfiles from Docker configurations

fgalan:
Given that WiredTiger is default storage in MongoDB 3.2 (according to https://docs.mongodb.com/manual/release-notes/3.2/#wiredtiger-as-default) and that smallfiles applies only to mmapv1 (see https://docs.mongodb.com/manual/reference/configuration-options/#storage-options) all the references to `smallfiles` could be removed from Docker configurations.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -25,7 +25,7 @@ Follow these steps:
 	
 		mongo:
 		  image: mongo:3.2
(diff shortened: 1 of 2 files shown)


kzangeli:
LGTM

PR #2180, opened 2016-05-24
FIX accessing to csub cache in update subscription logic to update count and lastNotification without taking semaphorewq

fgalan:


kzangeli:
LGTM

PR #2185, opened 2016-05-24
subscription templates now work for the noCache case as well

kzangeli:
The title says it all ...

Also fixed a few functests from a previous PR, about the mime-type.

### DIFF
--- a/src/lib/apiTypesV2/HttpInfo.cpp
+++ b/src/lib/apiTypesV2/HttpInfo.cpp
@@ -25,9 +25,12 @@
 #include <string>
 #include <map>
 
+#include "mongo/client/dbclient.h"
 #include "logMsg/logMsg.h"
 
 #include "common/JsonHelper.h"
+#include "mongoBackend/dbConstants.h"
+#include "mongoBackend/safeMongo.h"
 #include "apiTypesV2/HttpInfo.h"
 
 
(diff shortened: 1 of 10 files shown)


src/lib/apiTypesV2/HttpInfo.cpp
```diff
@@ -91,4 +94,52 @@ std::string HttpInfo::toJson()
   return jh.str();
 }
 
+
+
+/* ****************************************************************************
+*
+* HttpInfo::fill -
+*/
```
fgalan: Maybe `using namespace mongo` would make the fill() code cleaner.

We typically include `using namespace` in .cpp, so it wouldn't be a problem.

kzangeli: This is a style-guide issue.
Personally, I don't like the 'using', as information is 'removed' somehow.
When mongo:: (well, any namespace, like 'std::', for example) is present, it is 100% clear where the type came from.
Using `using namespace` it becomes a whole less clearer.
At the same time, it is true that the code gets shorter and 'easier' to read.
But, easier to read only if you know where each type comes from.
So, not easier to read, but more cryptic, IMHO.

Let's discuss this, but my opinion is definitely that namespaces should always be written out.

fgalan: In my opiniong the decission should be done by the author. In some case, having all the time `blabla::` makes the code harder to read and follow. In some cases, it doesn't. So I'm not sure that could be defined as a general rule for .cpp files.

Let's be pragmatic: if you read `BSONObj` in our code there is no doubt where that `BSONObj` came from. There isn't any risk of confussion between the `mongo` namespace or any other.

However, what should be a general rule (in MUST category) is that `using namespaces` cannot be used in .h files. That could cause conflicts (we have had some of those in the pasts). In fact, maybe we already have this one?

kzangeli: It is true that `BSONObj` is not like `string`, as BSON already is some kind of prefix that clearly ties it to mongo.
So, if you really want to, I'll do it 'your way' in this file.
My personal vote is still for leaving it the way it is.
When namepaces become really long, like over 15 chars, then I will probably have another opinion :-)
Right here it is not so important (due to the prefix in BSONObj - had it been mongobson::Obj instead, it would be very different).

fgalan: Citing myself :)

> In my opinion the decission should be done by the author [in .cpp files]

You are the author of this part of the code. So, it's up to you.

kzangeli: `using namespace mongo` in commit f810bbf.

Why?
Well, while 'in general' I am still against removing the explicit namespace names, the BSONxxx types have their prefixes and especially, they are used in a million places already in our code, without the explicit namespace.
So ...

fgalan:
LGTM

PR #2186, opened 2016-05-24
Hardening/using namespace in headers for style guide

kzangeli:
See title

### DIFF
--- a/doc/manuals/contribution_guidelines.md
+++ b/doc/manuals/contribution_guidelines.md
@@ -395,7 +395,21 @@ xyzInit(user,   pwd  host);
 
 *How to check*: checked using the internal script ```scripts/style_check.sh```
 
+#### M12 (`using namespace` in header files)
+
+*Rule*: the construct ```using namespace XXX``` must not be used in header files.
+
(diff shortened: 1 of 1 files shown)


doc/manuals/contribution_guidelines.md
```diff
@@ -395,6 +395,16 @@ xyzInit(user,   pwd  host);
...
 
+#### M12 (`using namespace` in header files)
+
+*Rule*: the construct ```using namespace XXX``` must not be used in header files.
+
+*Rationale*: a header file should not decide for the source files that include the header file.
+If a header file contains this construct then all source files that include the header file,
+directly or indirectly are forced to have that construct as well.
+
+*How to check*: easily added to the internal script ```scripts/style_check.sh```, using an intelligent ```grep```.
```
fgalan: What about just:

"using an intelligent grep```

?

kzangeli: Well, the style_check script is mentioned in other 'how to checks'

kzangeli: Fixed in dac25cc

fgalan:
LGTM

PR #2196, opened 2016-05-26
Align fiware-ngsiv2-cookbook with Tour Guide Application

FiwareULPGC:
This PR aligns the fiware-ngsiv2-cookbook content with the Tour Guide Application (<https://github.com/Fiware/tutorials.TourGuide-App>)

A compiled version can be consulted at <http://docs.oriontourguide.apiary.io/#>


fgalan:
Given that this PR doesn't allow inline comments (as the diff of changes is not rendered at github) please find my feedback as general comments:

* You use `idPattern=0115206c51f60b48b77e4*` which in regex terms means "any string with '0115206c51f60b48b77e' followed by any number of occurrences of the digit 4". Maybe is a typo and the right one is `idPattern=0115206c51f60b48b77e4.*` (which means "any string '0115206c51f60b48b77e4' followed by any character).

* If I have undestood correctly, the "One entity attribute. Not present" case shows the case of a given entity (existing) with a non-existing attribute (I think so given the attribute name example is "mistake"). In that case, the error message shown ("The requested entity has not been found. Check type and id") is wrong and the right one should be: "The entity does not have such an attribute".

* I have seent: `&offset=offset`. While that style is correct for the NGSIv2 spec document, I understand that the idea of the cookbook is using always actual example, eg. `&offset=2`.

* Note that `PUT http://fiware-ngsi2.apiblueprint.org/v2/entities/entityId/attrs/attrName/value` with payload `{ "value": 25 }` you set the value of attrName to the JSON `{ "value": 25 }` not to 25 (which probably is what you are going to achieve). The example could lead to confussion, af after that operation you will get something like this when you get that entity:
```
    ...
    "A": {
        ...
        "value": {
            "value": 28
        }
    }
...
```


* Remove entity by id. It should use DELETE (and without payload)

* List all subscritpion in the system. Maybe another receipt showing pagination (as shown in the case of entities) would be a good addition.

* Partial update of subscription. The example is wrong: "duration" field is not used in NGSIv2 at the end. You can use "expires".

fgalan:
More stuff:

* The `Link` header in the pagination operation is not yet part of NGSIv2 spec. It could be added in the future but, while that moment comes and in order to have fully alingment cookbook-spec, I'd suggest remove Link from cookbook.

* All your examples (or the most) use `Fiware-Service` header, which is an specific addition of Orion Context Broker implementaton, but not part of the NGSIv2 by the time being. Thus, one of the following should be done: 1) remove the header from the cookbook, 2) keep it but explain in the introduction the reason to keep it (e.g. to use the cookbook examples with the Tour Guide Application if that application needed it).

iariasleon:
In the case of implementing option 2 in last bullet in previous @fgalan, then `Fiware-ServicePath` headers  could be also ilustrated in all examples

FiwareULPGC:
> Given that this PR doesn't allow inline comments (as the diff of changes is not rendered at github) please find my feedback as general comments:

> You use idPattern=0115206c51f60b48b77e4* which in regex terms means "any string with '0115206c51f60b48b77e' followed by any number of occurrences of the digit 4". Maybe is a typo and the right one is idPattern=0115206c51f60b48b77e4.* (which means "any string '0115206c51f60b48b77e4' followed by any character).

Yes, it was a typo. But sing this, then we found an unexpected behaviour. If we perform the request as described we get the entity.

 ```
    curl -v "orion:1026/v2/entities?idPattern=0115206c51f60b48b77e4*&attrs=name,id" --header 'Fiware-Service: tourguide'
    * Hostname was NOT found in DNS cache
    *   Trying 172.17.0.3...
    * Connected to orion (172.17.0.3) port 1026 (#0)
    > GET /v2/entities?idPattern=0115206c51f60b48b77e4*&attrs=name,id HTTP/1.1
    > User-Agent: curl/7.35.0
    > Host: orion:1026
    > Accept: */*
    > Fiware-Service: tourguide
    > 
    < HTTP/1.1 200 OK
    < Connection: Keep-Alive
    < Content-Length: 127
    < Content-Type: application/json
    < Fiware-Correlator: 371e8fe2-2355-11e6-bea0-0242ac110003
    < Date: Thu, 26 May 2016 15:19:18 GMT
    <
    * Connection #0 to host orion left intact
    [{"id":"0115206c51f60b48b77e4c937835795c33bb953f","type":"Restaurant","name":{"type":"none","value":"Elizalde","metadata":{}}}]
 ```
So, the restaurant is returned, and it should not be. 

> If I have undestood correctly, the "One entity attribute. Not present" case shows the case of a given entity (existing) with a non-existing attribute (I think so given the attribute name example is "mistake"). In that case, the error message shown ("The requested entity has not been found. Check type and id") is wrong and the right one should be: "The entity does not have such an attribute".

If we perform the request we get that message. We did not notify it because we thought that this message is shown because is a entity level operation and maybe someone could find an entity with the specified attribute (taking into account that is possible to have two entities with same id and different types).

Note that here we are querying by an entity and not by an attribute like in `v2/{entityId}/attrs/{attrName}`

These are the request and the response that we get with a wrong attribute name:

```
curl -v "orion:1026/v2/entities/0115206c51f60b48b77e4c937835795c33bb953f?attrs=name2" --header 'Fiware-Service: tourguide'
* Hostname was NOT found in DNS cache
*   Trying 172.17.0.3...
* Connected to orion (172.17.0.3) port 1026 (#0)
> GET /v2/entities/0115206c51f60b48b77e4c937835795c33bb953f?attrs=name2 HTTP/1.1
> User-Agent: curl/7.35.0
> Host: orion:1026
> Accept: */*
> Fiware-Service: tourguide
> 
< HTTP/1.1 404 Not Found
< Connection: Keep-Alive
< Content-Length: 95
< Content-Type: application/json
< Fiware-Correlator: a4210140-2357-11e6-aa25-0242ac110003
< Date: Thu, 26 May 2016 15:36:40 GMT
<
* Connection #0 to host orion left intact
{"error":"NotFound","description":"The requested entity has not been found. Check type and id"}
```

> I have seent: &offset=offset. While that style is correct for the NGSIv2 spec document, I understand that the idea of the cookbook is using always actual example, eg. &offset=2.

Ok. We will check and update it.

> Note that PUT http://fiware-ngsi2.apiblueprint.org/v2/entities/entityId/attrs/attrName/value with payload { "value": 25 } you set the value of attrName to the JSON { "value": 25 } not to 25 (which probably is what you are going to achieve). The example could lead to confussion, af after that operation you will get something like this when you get that entity:

    ...
    "A": {
        ...
        "value": {
            "value": 28
        }
    }
    ...


Ok. We will check and update it.


> Remove entity by id. It should use DELETE (and without payload)

Ok. We will check and update it.

> List all subscritpion in the system. Maybe another receipt showing pagination (as shown in the case of entities) would be a good addition.

We think that is a good idea. We will add it.

> Partial update of subscription. The example is wrong: "duration" field is not used in NGSIv2 at the end. You can use "expires".

Ok. We will check and update it.

> The Link header in the pagination operation is not yet part of NGSIv2 spec. It could be added in the future but, while that moment comes and in order to have fully alingment cookbook-spec, I'd suggest remove Link from cookbook.

Ok. We will remove it and it will be added once it is implemented.

> All your examples (or the most) use Fiware-Service header, which is an specific addition of Orion Context Broker implementaton, but not part of the NGSIv2 by the time being. Thus, one of the following should be done: 1) remove the header from the cookbook, 2) keep it but explain in the introduction the reason to keep it (e.g. to use the cookbook examples with the Tour Guide Application if that application needed it).

Ok, so, @jmcanterafonseca  what approach should we follow?

fgalan:
> So, the restaurant is returned, and it should not be. 

I don't get your point... I mean, the `0115206c51f60b48b77e4*` *does* match `0115206c51f60b48b77e4c937835795c33bb953f`. You can check for example here http://rubular.com

Maybe I'm missing something... :)

fgalan:
> If we perform the request we get that message. We did not notify it because we thought that this message is shown because is a entity level operation and maybe someone could find an entity with the specified attribute (taking into account that is possible to have two entities with same id and different types).
> Note that here we are querying by an entity and not by an attribute like in v2/{entityId}/attrs/{attrName}

Maybe I'm not interpreting correclly the case...

So you mean that "One entity attribute. Not present" correspond to a case in which the entity doesn't exist? Is that correct

------------------

By the way, I have done a quick test to be sure of the exact message for "entity doesn't exist" and "entity exist but attribute doesn't exist" cases.

fermin@neodeb:~/src/fiware-orion$ curl -s -S localhost:1026/v2/entities/notexist | python -mjson.tool
{
    "description": "The requested entity has not been found. Check type and id",
    "error": "NotFound"
}
fermin@neodeb:~/src/fiware-orion$ curl localhost:1026/v2/entities --header 'Content-Type: application/json' -d @- <<EOF
> {
>   "id": "Room1",
>   "type": "Room",
>   "temperature": {
>     "value": 10
>   }
> }
> EOF
fermin@neodeb:~/src/fiware-orion$ curl -s -S localhost:1026/v2/entities/Room1/attrs/notexistattr | python -mjson.tool
{
    "description": "The entity does not have such an attribute",
    "error": "NotFound"
}
```

Thus, the

**Comment by FiwareULPGC** [comment]
>I don't get your point... I mean, the 0115206c51f60b48b77e4* does match 0115206c51f60b48b77e4c937835795c33bb953f. You can check for example here http://rubular.com
>
>Maybe I'm missing something... :)

We thought that the ID must *fully match* (all the indentifier must be contained in the match) with the pattern because we read in the documentation (<http://fiware-orion.readthedocs.io/en/latest/user/walkthrough_apiv2/index.html>) :


> You can filter using entity id patterns, using the `idPattern` URL parameter (whose value is a regular expression). For example, to get all entities whose id starts with `Room` and is followed by a number in the 2 to 5 range (in this case retrieving Room2) you can use (note the `-g` in curl command line to avoid problems with brackets):

```
curl localhost:1026/v2/entities?idPattern=Room[2-5] -g -s -S  \
    --header 'Accept: application/json' | python -mjson.tool
```

But, if *partial match* is used to filter the entities, the entity which id is `aaaaaRoom2` would be returned when the documentation example is used. Note that it is not consistent with the text **all entities whose id starts with Room**.  We think that the pattern should be something like `^Room[2-5]` instead of `Room[2-5]`.

**Comment by FiwareULPGC** [comment]
>Maybe I'm not interpreting correclly the case...
>
>So you mean that "One entity attribute. Not present" correspond to a case in which the entity doesn't exist? Is that correct

>By the way, I have done a quick test to be sure of the exact message for "entity doesn't exist" and "entity exist but attribute doesn't exist" cases.


We think we are referencing different operations.

If **partial entity retrieval** operation (`/v2/entities/exampleId?attrs=mistakenAttribute`)  is used, we get:

```
 {
   "error": "NotFound",
    "description": "The requested entity has not been found. Check type and id"
}
```

If **get single attribute** operation (`/v2/entities/exampleId/attrs/mistakenAttribute`) is used, we get:

```
{
  "error": "NotFound",
  "description": "The entity does not have such an attribute"
}
```

We are refering to the first operation. We are trying to get the **entity** which has the specified attribute.

An example of this kind of operation could be: 
If we have two entities with the same id (Bcn-Welt) with different types (Room1 and Room) and if  entities of type Room have two attributes (temperature, humidity) and entities of type Room1 only have one (temperature).

We can perform the following requests:

```
curl orion:1026/v2/entities/Bcn-Welt?attrs=humidity

{"error":"TooManyResults","description":"More than one matching entity. Please refine your query"}
```


```
curl orion:1026/v2/entities/Bcn-Welt?attrs=humidity
{"id":"Bcn-Welt","type":"Room","humidity":{"type":"none","value":60,"metadata":{}}}
```

**Comment by fgalan** [comment]
> But, if partial match is used to filter the entities, the entity which id is aaaaaRoom2 would be returned when the documentation example is used. Note that it is not consistent with the text all entities whose id starts with Room. We think that the pattern should be something like ^Room[2-5] instead of Room[2-5].

Could you fix user/walkthrough_apiv2.md in that way (checking first that curl with `idPattern=^Room[2-5]` actually works) in this PR, please?

**Comment by fgalan** [comment]
Regarding to https://github.com/telefonicaid/fiware-orion/pull/2196#issuecomment-222652681 (too long to cite :):

* Regarding "One entity attribute. Not present" if what this operation is showing is the problem that happen when ID is not found, then I'd suggest to use betters name in the example, e.g: 
`GET http://fiware-ngsi2.apiblueprint.org/v2/entities/mistakenEntity?attrs=temperature`. Note that, as a reader, if I see `entityId` (for entity) and `mistake` (for attribute) I tend to think that the problem the operation is showing is related with the thing named "mistaked" and my reasoning would be wrong.

* Which operation (if any) of the ones included in the cookbook is showing the case for "entity exist but not the attribute"? Maybe is there but I haven't seen...

**Comment by jmcanterafonseca** [comment]
@FiwareULPGC  wrt fiware-service we can keep it in the cookbook examples but add a note at some place saying that it is something specific to Orion Context Broker and it is used to allow testing the cookbook examples against the FIWARE tour guide application.

**Comment by FiwareULPGC** [comment]
> Regarding "One entity attribute. Not present" if what this operation is showing is the problem that happen when ID is not found, then I'd suggest to use betters name in the example, e.g:
    GET http://fiware-ngsi2.apiblueprint.org/v2/entities/mistakenEntity?attrs=temperature. Note that, as a reader, if I see entityId (for entity) and mistake (for attribute) I tend to think that the problem the operation is showing is related with the thing named "mistaked" and my reasoning would be wrong.
    
In this case, the entity id is correct but the attribute name is not. The problem that we are trying to explain is with the attribute name.

> Which operation (if any) of the ones included in the cookbook is showing the case for "entity exist but not the attribute"? Maybe is there but I haven't seen...


We are refering to the case when we try to partially retrieve an **entity** whose id exists but specifying a wrong attribute name. (<http://docs.oriontourguide.apiary.io/#reference/context-entity/partial-entity-retrieval/one-entity-attribute.-not-present>).

Using the previous example of two entities with the same id (Bcn-Welt) but different types (Room and Room1) and with different attributes (Room type has temperature and humidity attributes and Room1 only has temperature).

```
curl orion:1026/v2/entities

    [
        {
            "id": "Bcn-Welt",
            "type": "Room",
            "humidity":
            {
                "type": "none",
                "value": 60,
                "metadata":
                {
                }
            },
            "location":
            {
                "type": "geo:point",
                "value": "41.3763726, 2.1864475",
                "metadata":
                {
                    "crs":
                    {
                        "type": "none",
                        "value": "WGS84"
                    }
                }
            },
            "temperature":
            {
                "type": "none",
                "value": 21.7,
                "metadata":
                {
                }
            }
        },
        {
            "id": "Bcn-Welt",
            "type": "Room1",
            "location":
            {
                "type": "geo:point",
                "value": "41.3763726, 2.1864475",
                "metadata":
                {
                    "crs":
                    {
                        "type": "none",
                        "value": "WGS84"
                    }
                }
            },
            "temperature":
            {
                "type": "none",
                "value": 21.7,
                "metadata":
                {
                }
            }
        }
    ]

```


If we try to make a partial **entity** retrieve of `Bcn-Welt` using the attribute `temperature`, it fails because there are two entities which satisfy the condition

```
curl orion:1026/v2/entities/Bcn-Welt?attrs=temperature

{"error":"TooManyResults","description":"More than one matching entity. Please refine your query"}
```

If the request is performed using the `humidity` attribute instead of the `temperature` attribute, only one entity satisfies the condition and it is returned.

```
curl orion:1026/v2/entities/Bcn-Welt?attrs=humidity
{"id":"Bcn-Welt","type":"Room","humidity":{"type":"none","value":60,"metadata":{}}}
```


Finally, if we try to retrieve a **nonexistent attribute** (the case that we are talking about), we get the error :

```
curl orion:1026/v2/entities/Bcn-Welt?attrs=mistakenAttribute
{"error":"NotFound","description":"The requested entity has not been found. Check type and id"}
```

Note that we tried to retrieve an **entity**, not an attribute so, we think that it is ok that Orion returns **The requested entity has not been found**. But maybe the message **Check type and id** could be improved?

Regarding to a previous comment made by you, if you still think that the message is wrong, should an issue be opened?

You can also notice that the entity id is correct, there are two entities which use them.

All these examples were performed using:

```
{
  "orion" : {
  "version" : "1.1.0-next",
  "uptime" : "0 d, 0 h, 8 m, 9 s",
  "git_hash" : "2397fd46be54008409e7b2e4d29e572a2c598f2c",
  "compile_time" : "Tue May 24 10:38:00 UTC 2016",
  "compiled_by" : "root",
  "compiled_in" : "838a42ae8431"
}
}

```


In order to clarify the operation we can replace the title `One entity attribute. Not present` with `One entity attribute. Attribute not present`

**Comment by FiwareULPGC** [comment]
PR updated.

File: doc/apiary/v2/fiware-ngsiv2-cookbook.apib [hunk 1]
```diff
@@ -5,8 +5,8 @@ HOST: http://fiware-ngsi2.apiblueprint.org/
 
 FIWARE-NGSI Simple API (v2) Cookbook.
 
-This is a cookbook with recipes that show how to use FIWARE-NGSI v2 aka FIWARE-NGSI Simple API.
-Use this examples for learning and enjoy!
+This is a cookbook with recipes that show how to use FIWARE-NGSI v2 aka FIWARE-NGSI Simple API. The examples of this guide contain the `Fiware-Service: tourguide` header. This header is specific to Orion Context Broker and it is used to allow testing the cookbook examples against the FIWARE tour guide application.
```
fgalan: Line too long.

fgalan:
In RC spec the following is going to be removed from the `GET /v2` op:

```
"registrations_url": "/v2/registrations"
```

So maybe the cookbook should do the same.

fgalan:
LGTM. Thanks!

PR #2205, opened 2016-05-27
FIX documentation for Orion 1.2 releaseing along with some fixes in the code

fgalan:


test/functionalTest/cases/1080_batch_query_context/batch_query_context_error_conditions.test
```diff
@@ -327,21 +310,7 @@ Date: REGEX(.*)
 }
 
 
-10. POST /v2/op/query, with a scope with an unknown field (not 'value' nor 'type') - see error
```
fgalan: This is a consecuence of the changing in processing strategy at parseSubscription(), based in FindMember():

```
Value::ConstMemberIterator iter = valueP->FindMember("type");
if (iter == valueP->MemberEnd())
   {
     return "scope without a type";
   }
```

The bad thing is that we lose the ability to detect not-expected fields. The good thing is that we use one lookup instead two ones (according to http://rapidjson.org/md_doc_tutorial.html#QueryObject), so I think is better  in this way.

doc/manuals/index.md
```diff
@@ -5,8 +5,8 @@ Orion is a C++ implementation of the NGSI9/10 REST API binding developed as a pa
 Orion Context Broker allows you to manage all the whole lifecycle of context information including updates, queries, registrations and subscriptions.
 Using the Orion Context Broker, you are able to register context elements and manage them through updates and queries.
 In addition, you can subscribe to context information so when some condition occurs
-(e.g. an interval of time has passed or the context elements have changed) you receive a notification.
+(e.g. a context elements have changed) you receive a notification.
```
kzangeli: Remove 'a' OR:

a context element has changed

doc/manuals/user/pagination.md
```diff
@@ -80,11 +135,15 @@ POST <orion_host>:1026/v1/queryContext?offset=1000&limit=100
...
-specify a different order. In particular, the value of `orderBy` is a comma-separated
-list of attributes and the results are ordered by the first attribute. On ties,
-the results are ordered by the second attribute and so on.
-A "!" before the attribute name means that the order is reversed. For example:
+In the case of entities query, the `orderBy` URL parameter can be use to
+specify an order different from creation time. In particular, the value of
+`orderBy` is a comma-separated list of attributes and the results are ordered
+by the first attribute. On ties, the results are ordered by the second attribute and so on.
+A "!" before the attribute name means that the order is reversed. For example (NGSIv2):
+
```
kzangeli: Perhaps we should mention here that if not ordering by creation date, pagination might suffer ... ?

fgalan: You mean MongoDB suffering due to the `sort()` used in the DB query?

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Disable attribute detail in GET types operation
+
+Related with: https://github.com/telefonicaid/fiware-orion/issues/2073
+
+## Metadata vector and object values not implemented yet
+
+## Default type for entities
+
+## Default type for attributes
```
fgalan: Selft-note: include also a bullet about scope filters in POST /v2/op/query

doc/manuals/user/pagination.md
kzangeli: I mean that if a new entity is created between pagination 1 and 2, an entity might be lost as the order has changed due to the new entity.

When the order is on creation-time, any new entity is always placed last in the sort, thus never 'destroying' the list

fgalan: Note that the list can be destroyed anyway in the case of deleting entities already shown in early pages. Thus, pagination may suffer whatever the case :)

kzangeli: Yeah, that is also true.
However, the probability for 'errors' is much higher when other sort criteria than creation-date is used

kzangeli:
LGTM

test/functionalTest/cases/1080_batch_query_context/batch_query_context_error_conditions.test
kzangeli: NTC

doc/manuals/user/ngsiv2_implementation_notes.md
fgalan: Fixed in dc71b68e562de12c3637bc8d66fd46e899e3d944

Issue #1849, opened 2016-02-25
APIv2: attribute values and multiple accept types


marcc-orange:
The current APIv2 specification states that attribute value requests support client-server HTTP negotiation :
> The payload MIME type in responses is based on client-server HTTP negotiation, based on client `Accept` header.

But the given algorithm does not seem to handle `accept` header with multiple values :

> * If response payload MIME type is `application/json`: [...]
>    * If attribute value is a string, number, null or boolean, a HTTP error "406 Not Acceptable: accepted MIME types:
    text/plain" is returned.
> * If response payload MIME type is `text/plain`, [...]

Example: If `Accept: application/json, text/plain`, the implementation should not return a `HTTP error "406 Not Acceptable: accepted MIME types: text/plain"`. It should be able to return `application/json` when value is an object or array and fallback to `text/plain` if the value is a string/number/boolean/null.

I propose that the content negotiation algorithm work like the following:

* if attribute value is an array or object and `Accept` header contains `application/json` or `text/plain`:
     return the value as a JSON with a response type of `application/json` or `text/plain` (whichever is the first in `Accept` header).
* if attribute value is a string, number, null or boolean:
   * if `Accept` header contains `text/plain` return the value as text
   * else return a HTTP error  "406 Not Acceptable: accepted MIME types: text/plain"

Another solution would simply to remove "based on client-server HTTP negotiation" and mention that the `Accept` header only handles a single type.

fgalan:
(Sorry for the delay answering this...)

I think that your approach improves the one currently in the NGSIv2 specifiction, so I think it would be a good idea to use it.

Only a minor detail... you say:

> (whichever is the first in Accept header)

But what to do in the cases of ties, e.g.: `Accept: */*`. In my opinion, in that case `application/json` should win (the most the server can say about the "structure" of the response, the better... and in this sense application/json is more structured data than text/plain).

fgalan:
This question a SOF is related with this issue http://stackoverflow.com/questions/37458863/get-attribute-value-behaviour-with-accept-header/37485098#37485098 and should be edited once this issue gets closed.

fgalan:
> But what to do in the cases of ties, e.g.: Accept: */*. In my opinion, in that case application/json should win (the most the server can say about the "structure" of the response, the better... and in this sense application/json is more structured data than text/plain).

[...]

PR #2215, opened 2016-05-30
FIX NGSIv2 spec get attribute value MIME type chosing algorithm

fgalan:
Algorithm for Get Attribute Value improved as suggested in #1849. 

@marcc-orange it would be great if you could review this PR and provide feedback. I have introduced some modifications on your original text in order to cover the  `Accept: */*` case and to make it more structured (i.e. two levels bullet list). I hope you find modifications fine :)

Deadline for merging: May 31st EoB (time enough to review so samll PR, +9 -13 lines).

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1388,8 +1388,8 @@ Response:
 
 This operation returns the `value` property with the value of the attribute.
 
-* If attribute value is an array or object:
-  * If `Accept` header contains application/json or text/plain return the value as a JSON with a
+* If attribute value is JSON Array or Object:
+  * If `Accept` header contains `application/json` or `text/plain` return the value as a JSON with a
     response type of application/json or text/plain (whichever is the first in `Accept` header or
     application/json in the case of `Accept: */*`).
   * Else return a HTTP error "406 Not Acceptable: accepted MIME types: application/json, text/plain"
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1389,12 +1389,12 @@ Response:
 This operation returns the `value` property with the value of the attribute.
 
 * If attribute value is JSON Array or Object:
-  * If `Accept` header contains `application/json` or `text/plain` return the value as a JSON with a
+  * If `Accept` header can be expanded to `application/json` or `text/plain` return the value as a JSON with a
     response type of application/json or text/plain (whichever is the first in `Accept` header or
-    application/json in the case of `Accept: */*`).
+    `application/json` in the case of `Accept: */*`).
   * Else return a HTTP error "406 Not Acceptable: accepted MIME types: application/json, text/plain"
 * If attribute value is a string, number, null or boolean:
-  * If `Accept` header contains text/plain return the value as text. In case of a string, citation
+  * If `Accept` header can be expanded to text/plain return the value as text. In case of a string, citation
     marks are used at the begining and end.
   * Else return a HTTP error "406 Not Acceptable: accepted MIME types: text/plain"
 
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -195,6 +195,12 @@ is typically done in the REST interface)
 (placeholder to describe roles like context provider, context producer,
 context broker, context registry, context consumer)
 
+## MIME Types
+
+The API response payloads in this specification are based in `application/json` and  (for attribute value 
+type operation) `text/plain` MIME types. Client issuing HTTP requests with accept types different 
+than those will get a 406 Not Acceptable error.
+
 ## JSON Entity Representation
 
 An entity is represented by a JSON object with the following syntax:
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -197,8 +197,8 @@ context broker, context registry, context consumer)
 
 ## MIME Types
 
-The API response payloads in this specification are based in `application/json` and  (for attribute value 
-type operation) `text/plain` MIME types. Client issuing HTTP requests with accept types different 
+The API response payloads in this specification are based on `application/json` and  (for attribute value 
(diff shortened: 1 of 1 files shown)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1388,19 +1388,15 @@ Response:
...
-* If the response payload MIME type is `text/plain`,
-  * If the attribute value is a JSON object or array, the payload in a stringification of that JSON
-    object or array.
-  * If the attribute value is a string, number, null or boolean, the payload is a text
-    representation of that value. In case of a string, citation marks are used at the begining and
-    end.
-
-The payload MIME type in responses is based on client-server HTTP negotiation, based on the client
-`Accept` header.
+* If attribute value is an array or object:
```
jmcanterafonseca: value is a JSON Array or Object:

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1388,19 +1388,15 @@ Response:
...
+  * If `Accept` header contains application/json or text/plain return the value as a JSON with a
```
jmcanterafonseca: nit: `application/json` `text/plain`

jmcanterafonseca:
to the best of my knowledge I thought that the agreement was that if a `string`, `number`, etc. value is requested as JSON, simply the response should be

```json
{ 
   "value": 5
}
````

Take into account that if this PR lands, then the only way to get just the value would be to stick to text/plain, which is really weird. (/attrs/myAttr would return value, type and metadata).

**Comment by marcc-orange** [comment]
Looks good to me, but @jmcanterafonseca's comment makes a lot of sense. It seems logical that all NGSIv2 API requests support JSON.

**Comment by fgalan** [comment]
The problem with @jmcanterafonseca's proposal is that it would introduce ambiguety. I mean, if I do

`GET /v2/entities/E/attrs/A/value` with `Accept: application/json` and I get:

{
  "value": 5
}

I cannot distinguish between the following situations:

1. The value of A is the number '5'
2. The value of A is the JSON objecT: { "value": 5 }

Thus, we need the algorithm as it is described right now in the PR in order to avoid such ambiguety.

**Comment by marcc-orange** [comment]
Ok, thanks for the clarification.

**Comment by jmcanterafonseca** [comment]
fair enough, then I think we need to update the corresponding to PUT
operations to make them consistent with this approach, right?

2016-05-31 9:54 GMT+02:00 Marc Capdevielle <notifications@github.com>:

**Comment by fgalan** [comment]
> fair enough, then I think we need to update the corresponding to PUT operations to make them consistent with this approach, right?

I have reviewed the `PUT /v2/entities/E/attrs/A/value` operation and its description right now in the spec documents, seems to be ok. However, it would be great if @marcc-orange and you could have a look in order to re-check, please.

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 1 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in 01eeac06d761d0ce8dd8e616460ff3d7f43dd1f5

File: doc/apiary/v2/fiware-ngsiv2-reference.apib (hunk 2 above)
Comment (fgalan) [inline doc/apiary/v2/fiware-ngsiv2-reference.apib]: Fixed in 01eeac06d761d0ce8dd8e616460ff3d7f43dd1f5

File: doc/apiary/v2/fiware-ngsiv2-reference.apib [hunk 3]
```diff
@@ -1388,19 +1388,15 @@ Response:
...
-* If the response payload MIME type is `text/plain`,
-  * If the attribute value is a JSON object or array, the payload in a stringification of that JSON
-    object or array.
-  * If the attribute value is a string, number, null or boolean, the payload is a text
-    representation of that value. In case of a string, citation marks are used at the begining and
-    end.
-
-The payload MIME type in responses is based on client-server HTTP negotiation, based on the client
-`Accept` header.
+* If attribute value is JSON Array or Object:
```
jmcanterafonseca: I think this can be made clearer:

If accept header does not include a MIME type which can expand to `text/plain` or `application/json` an error `406` must be raised. 

In the absence of any hint from accept headers, or if accept headers specify a generic mime type (`*/*`), an attribute value will be returned as JSON (Content-Type: `application/json`) if the attribute value is a JSON object or array. Conversely, it will be returned as plain text (Content-Type: `text/plain`) in any other case. In the particular case of a `string`, citation (`"`) marks must be used at the beginning and end.

If accept headers include only `application/json`  and the attribute value has to be represented as text plain (as per the rules above) it will be considered an error `406` Not Acceptable: accepted MIME types: text/plain. In any other case the implementation will honor the requested MIME type provided by the client.

fgalan: Regarding using "expansion" rather than "contain" when referring to Accept, fixed in ae27c94c17e3be70c9ae8a3d60b65cffd793e484.

Regarding changing the orientation in the description, NTC, as I think is more clear to describe from the perspective of the attribute value (first level bullets), then Accept (second level bullets), as #1849 suggest.

fgalan:
Add general section about MIME types in response in commit https://github.com/telefonicaid/fiware-orion/pull/2215/commits/e90395eea1c9428bd385fd61f6895ec0a4f2e7c5

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -195,6 +195,12 @@ is typically done in the REST interface)
 (placeholder to describe roles like context provider, context producer,
 context broker, context registry, context consumer)
 
+## MIME Types
+
+The API response payloads in this specification are based on `application/json` and  (for attribute value 
+type operation) `text/plain` MIME types. Clients issuing HTTP requests with accept types different 
+than those will get a 406 Not Acceptable error.
```
jmcanterafonseca: nit: `406 Not Acceptable` HTTP error.

jmcanterafonseca:
LGTM, though have a look at literal values and enclose them between `` , there are more left

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in fde7fe14bf1889ebc01e5a14e1d72af18d65ef1a

kzangeli:
LGTM

fgalan:
Having reached deadline, PR has been merged. Thanks @marcc-orange and @jmcanterafonseca for your feedback!

PR #2219, opened 2016-05-31
ADD new .apib file for stable release candidate May 2016

fgalan:
Just to create a copy of existing .apib file.

The only interesting review that could be done in this branch is just to compare MD5SUM of both files (new an existing) to check they are exactly equal.


kzangeli:
LGTM

PR #2236, opened 2016-06-02
Hardening/2232 utc in logfile (into release/1.2.0)

fgalan:
Twin PR (into release/1.2.0) of #2233

* UTC in logfile and not the nodes time-zone (which did not correspond to the times)
* gmtime_r seems to solve the problem
* CHANGES_NEXT_RELEASE


kzangeli:
LGTM

PR #2239, opened 2016-06-03
Doc/better test for batch query response

kzangeli:
Branch merged with develop and the diff is OK.
**Six** commits ... no explanation for that ...


fgalan:
LGTM

Don't worry for extra commits. It use to be due to "git branch inheritance". These extra commits use to cancel one each other. As long a diff shows the changes we want in a clean way, PR is safe to merge.

Issue #2237, opened 2016-06-03
CB does not forward compund values to contextProviders


manucarrace:
When sending an updatecontext to CB for a providing application (iot platform commands), CB does not forward the compound values:

The registration:

```
POST /NGSI9/registerContext
{
  "contextRegistrations": [
    {
      "entities": [
        {
          "type": "device",
          "isPattern": "false",
          "id": "devicetest"
        }
      ],
      "attributes": [
        {
          "name": "ping",
          "type": "command",
          "isDomain": "false"
        }
      ],
      "providingApplication": "http://10.0.0.2:4061"
    }
  ],
  "duration": "P1M"
}
```

The entity creation:

```
POST /v1/updateContext
{
  "contextElements": [
    {
      "type": "device",
      "isPattern": "false",
      "id": "devicetest",
      "attributes": [
        {
          "name": "humidity",
          "type": "percet",
          "value": " "
        },
        {
          "type": "type1",
          "name": "name1",
          "value": "value1"
        },
        {
          "name": "ping_status",
          "type": "commandStatus",
          "value": "UNKNOWN"
        },
        {
          "name": "ping_info",
          "type": "commandResult",
          "value": " "
        },
        {
          "name": "TimeInstant",
          "type": "ISO8601",
          "value": " "
        }
      ]
    }
  ],
  "updateAction": "APPEND"
}
```

The update command:

```
PUT /v1/contextEntities/type/device/id/devicetest
{
  "attributes": [
    {
      "type": "command",
      "name": "ping",
      "value": {
        "param3": "33",
        "param2": "22",
        "param1": "11"
      }
    }
  ]
}
```

And finally, this is what the CB forwards:

```
POST //updateContext
{
  "contextElements" : [
    {
      "type" : "device",
      "isPattern" : "false",
      "id" : "devicetest",
      "attributes" : [
        {
          "name" : "ping",
          "type" : "command",
          "value" : ""
        }
      ]
    }
  ],
  "updateAction" : "UPDATE"
}
```

you can check it in integration with the following data:

service: city012
subservice: electricidad
entityid: devicetest
user: admin_city012
password: password

CB version:

```
{
  "orion" : {
  "version" : "1.1.0_20160531020227",
  "uptime" : "1 d, 17 h, 2 m, 38 s",
  "git_hash" : "e8fb906dc7278b2799698eb917edde8b2da3e700",
  "compile_time" : "Tue May 31 02:05:08 CEST 2016",
[...]

PR #2244, opened 2016-06-06
Bug/2237 compound values in forwarded messages (release/1.2.0)

kzangeli:
Issue #2237 fixed

[ This fix includes cloning of compound values so the valgrind test suite should be run ]

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1747,7 +1747,7 @@ static void updateAttrInNotifyCer
 
   if (caP->compoundValueP)
   {
-    // The ContextAttribute constructor steals the compound, but in this case, it must be cloed
+    // The ContextAttribute constructor steals the compound, but in this case, it must be cloned
     targetAttr->compoundValueP = caP->compoundValueP->clone();
   }
 
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1744,6 +1744,13 @@ static void updateAttrInNotifyCer
 
   /* Reached this point, it means that it is a new attribute (APPEND case) */
   ContextAttribute* caP = new ContextAttribute(targetAttr, useDefaultType);
+
+  // KZ: this is the fix for issue 2237 (2 lines)
+  if (caP->compoundValueP)
+  {
+    targetAttr->compoundValueP = caP->compoundValueP->clone();
+  }
+
   notifyCerP->contextElement.contextAttributeVector.push_back(caP);
 }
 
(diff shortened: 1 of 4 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1744,6 +1744,13 @@ static void updateAttrInNotifyCer
 
   /* Reached this point, it means that it is a new attribute (APPEND case) */
   ContextAttribute* caP = new ContextAttribute(targetAttr, useDefaultType);
+
+  if (caP->compoundValueP)
+  {
+    // The ContextAttribute constructor steals the compound, but in this case, it must be cloned
+    targetAttr->compoundValueP = caP->compoundValueP->clone();
```
fgalan: Is this leak-free?

(Every time I see a clone() method, an alarm about leaks triggers in my mind :)

test/functionalTest/cases/2237_compound_values_in_forwarded_messages/compound_values_in_forwarded_messages.test
```diff
@@ -0,0 +1,235 @@
...
+payload='{
+  "contextElements": [
+    {
+      "type": "T1",
+      "id":   "E1",
+      "attributes": [
+        {
+          "name": "A1",
+          "type": "string",
+          "value": { "vec": [ "this", "is", "a", "compound", "value" ], "obj": { "and": "this", "is": "also", "a": "compound", "value": "." } }
```
fgalan: It would be clearer to break this line in multiple one (jsonlint.com may help), as we have done in other test related with compounds.

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Funny you should say that ...
See the description :-)

fgalan: Discussed at skype. Seems to be leak-safe from in principle (valgrind job will confirm) but the following comment in 'updateAttrInNotifyCer' should be adjusted:

```
  // Note that updateAttrInNotifyCer() may "ruin" targetAttr, as compoundValueP is moved
  // (not copied) to the structure in the notifyCerP and null-ified in targetAttr. Thus, it has
  // to be called after the location processing logic (as this logic may need the compoundValueP
```

test/functionalTest/cases/2237_compound_values_in_forwarded_messages/compound_values_in_forwarded_messages.test
kzangeli: No probs

kzangeli: Fixed in 0d16260

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Comment removed in f1c6860

fgalan:
LGTM

Note that branch is not deleted after merging, as it would be reused for develop PR.

PR #2255, opened 2016-06-09
Fix subscription equal filter evaluation on update context logic (release/1.2.0)

fgalan:
Fixes #2222

Issue #2222, opened 2016-06-01
in POST v2/subscriptions if we use q=attr==true (boolean) the notification is not sent


iariasleon:
in POST v2/subscriptions if we use q=attr==true (boolean) the notification is not sent

#### subscription request:
```
url: POST http://192.168.21.64:10026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: qap1
    Fiware-ServicePath: /ssp1
    X-Auth-Token: c2a0744403f0469c8d764d4f27c4c0fc
payload: { "subject": {"entities": [{"idPattern": ".*"}], "condition": { "attrs": ["temperature"], "expression": {  "q": "value_bool==true"} }}, "notification": {"http": { "url": "http://10.95.82.95:10032/notify"}, "attrsFormat": "values", "attrs": ["pressure"]} }
```

#### entity request:
```
url: POST http://192.168.21.64:10026/v2/entities
headers:
    Content-Type: application/json
    Fiware-Service: qap1
    Fiware-ServicePath: /ssp1
    X-Auth-Token: c2a0744403f0469c8d764d4f27c4c0fc
payload: {"type": "cMen", "id": "room_2",  "temperature":{"type": "celsius", "metadata": {"very_hot": {"value": "default", "type": "alarm"}}, "value": 34}, "pressure":{"value": "high"}, "humidity":{"type": "porcent", "metadata": {"very_hot": {"value": "default", "type": "alarm"}}, "value": "air,density"}, "timestamp":{"type": "DateTime", "value": "2017-06-15T07:21:24.00Z"}, "value_str":{"value": "true"}, "value_bool":{"value": true}}
```
But if we invert the order, first, we create the entity and after, the subscription, the notiication is sent correctly

kzangeli:
This is also true for non-bool attributes, right?

iariasleon:
With reverse order, another type (ex:string)  are notified correctly

fgalan:
Given that is is related with e2e tests in IoTP, its priority has been raised.

fgalan:
Fixed in PR #2255 (release/1.2.0) and #2256 (develop).

fgalan:
@iariasleon pls have a look in order to check if this is now fixed.

iariasleon:
LGTM

PR #2256, opened 2016-06-09
Fix subscription equal filter evaluation on update context logic (develop)

fgalan:
Implements #2222. Twin PR for #2255


kzangeli:
LGTM

PR #2261, opened 2016-06-10
clearly stating that condition is optional

kzangeli:
The ngsiv2 spec a little clearer about 'condition' in subscriptions.
Ah, and an error encountered and fixed.

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -561,7 +561,7 @@ might be useful in understanding the format.
 The Simple Query Language provides a simplified syntax to retrieve entities which match a set of
 conditions.
 A query is composed by a list of statements separated by the ';' character.
-Each statement express a matching condition.
+Each statement expresses a matching condition.
 The query returns all the entities that match all the matching conditions (AND logical operator). 
 
 There are two kinds of statements: *unary statements* and *binary statements*.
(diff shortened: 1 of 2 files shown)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -561,7 +561,7 @@ might be useful in understanding the format.
 The Simple Query Language provides a simplified syntax to retrieve entities which match a set of
 conditions.
 A query is composed by a list of statements separated by the ';' character.
-Each statement express a matching condition.
+Each statement expresses a matching condition.
```
fgalan: Maybe this should be rollbacked, waiting for the next proof-read process upon releasing next RC based on latest.

Maybe I bit contra-natural :) but note that if we start fixing proof-read errors now, then a lot of noise will be added in the future when we compare current RC with lastet to look for actual changes...

kzangeli: I see three choices:
1. Maintain this kind of fixes in a separate file - to not forget 
2. Change also in the freezed (frozen? :-) doc
3. Live with the noise

Easiest option: 3 (I don't see that this noise will bother too much - of course, bigger changes should not be made - just simple typos)

I would put a 'weak vote' for option 3

kzangeli: ok, no probs

kzangeli: Fixed in 361df60

fgalan:
LGTM (passing the ball to @jmcanterafonseca just in case he has some feedback)

jmcanterafonseca:
LGTM

Issue #2193, opened 2016-05-25
Assessment of NGSIv2 query/update redirect to CPrs (it was: Problem mixing NGSIv2 queries with faulty CPrs)


fgalan:
If clients uses NGSIv2 query which result to being forwareded due to registrations and the CPr doesn't reponds or does in a faulty way, then CB responds to the client in NGSIv1 format.

This is probably a side-effect of not having pay attention yet to NGSI9 functionality in NGSIv2. CPrs functionality *shouldn't* use NGSIv2 until we define NGSI9 functionality in NGSIv2 (at spec level) and implement it (at Orion code level).

Some information about the query used to test this:

 > Url: http://192.168.21.64:1026/v2/entities
 > Method: get
 > Headers: {'fiware-service': 'city012', 'fiware-servicepath': '/electricidad', 'Accept': 'application/json'}
 > Params sent: {'q': 'dateCreated>2016-04-04T14:00:00.00Z', 'options': 'dateCreated'}

CC: @mcarracedo

fgalan:
We have a large batery of CPr test in cases/0787_cprs_full_functional. It would be a good idea to create a version of such .tests using NGSIv2 query/post operations in order to ensure that the NGSIv2 query/post functionality keeps working for CPrs registed using NGSIv1.

kzangeli:
I was pretty sure that the service routines that issue the forwarding to CPrs (serviceRoutines/postUpdateContext and postQueryContext) looked at the API version and did not forward anything if API v2 ...  Perhaps this is broken?

However, we could of course support forwarding (in V1) even though the petitions are made with V2, just sending the forward in V1 and receiving the response of the forward also in V1 to finally respond to the client using V2 ...

So, two possible fixes to this problem.
First, let's investigate a little and estimate effort for the two options.

fgalan:
Whatever the conclusion be, the doc at https://fiware-orion.readthedocs.io/en/develop/user/v1_v2_coexistence/index.html should include a paragraph regarding this explaining if it is advisable to use NGSIv2 query/udpates with CPRs registred using NGSIv1.

kzangeli:
Some info on this after some testing:

First of all, forwarding on queries in NGSIv2 is **not** capped and it works just fine. The query is sent in NGSIv2 and its response comes in NGSIv2. The forwarding messages are all done in NGSIv1.

After some slight modification of the broker, I managed to make the same thing happen for Updates also, using the following request:

[...]

PR #2271, opened 2016-06-13
Hardening/2193 op tests for forwarding

kzangeli:
Functests covering operations that provoke forwards.
Solves remaining part of issue #2193


test/functionalTest/cases/2193_v1_reply_to_v2_response_on_error_with_cpr/forward_using_GET_v2_entities.test
```diff
@@ -0,0 +1,220 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: Note that previous new files related with #2193 went to cases/0787... directory and used _in_ngsIv2.test suffix.

Not sure what is the best naming/ubication convention.

kzangeli: Those tests in 0787 were pure copies of oldertests in 0787, converted to v2.
These new v2 tests have no v1 counterpart and don't belong in 0787.
Perhaps we should move those v2 tests from 0787 to 2193?

fgalan: What about to move all *_in_ngsiv2.test cases in `cases/0787_cpr_full_functional` to `cases/0787_cpr_full_functional_v2` (new directory), then change "_in_ngsiv2" sufix by "ngsiv2_" prefix in all them?

Regarding the test in this PR, I agree in that being new test (not conversions), cases/2193 is the right place to be.

fgalan: Actually, sufix vs. prefix is probably a matter of taste (at least I don't remember any style rule about it). Thus, sufixes are also ok if you prefer so.

fgalan: NTC (whatever the change, it is better in a next PR, as I'm touching some `_in_ngsiv2.test` files in ongoing work.

kzangeli: I like the idea of grouping all the V2 CPr tests in one directory, `0787_cpr_full_functional_v2` for example.
And removing the `_in_ngsiv2` in those from 0787 plus moving all tests files from 2193 to that new directory.

fgalan: Due to the way `make coverage` works, all .test files needs a different name (although they are located in different directories). Thus, just removing the sufix is not enough (as the name will be equal to the one used by the NGSIv1 variant and it would colide), prefixing would be needed.

kzangeli: ok, perhaps we should start adding the issue number to the test files as well.
I think Carlos is doing that already.

fgalan: I don't think we need to be so "strict". Actually, the probability of name collision is very low, only in cases like this (a .test that is generated from another one, with the idea of keeping the same name) we need a prefix.

fgalan:
LGTM

PR #2277, opened 2016-06-15
hardening/2193_entity_not_found_when_forwarding

kzangeli:
Two tests to ensure 404 Not Found is returned when a forward-updated entity doesn't exist


kzangeli:
Only for develop

PR #2305, opened 2016-06-17
NGSIv2 spec (latests): create/update subscriptions operation in NGSIv2 spec with regards to expires field equal to empty string

fgalan:


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1722,6 +1722,9 @@ Response:
 Creates a new subscription.
 The subscription is represented by a JSON object as described at the beginning of this section.
 
+Using `expires` with value equal to empty string (`""`) equals to omit the field, i.e. creating
```
kzangeli: equal-equals ... I don't like it very much.

I propose (with a few more 'tweaks'):

    Using `expires` with a value of the empty string (`""`) equals omitting the field, i.e. creating ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1816,6 +1819,9 @@ Response:
 
 Only the fields included in the request are updated in the subscription.
 
+Using `expires` with value equal to empty string (`""`) equals to removing the field, i.e.
```
kzangeli: And same here:

    Using `expires` with a value of the empty string (`""`) equals removing the field, i.e. ...

jmcanterafonseca:
and what about null? I think it is the same. and what about " " ?

kzangeli:
null is not a string, so error right away there (I hope).
" " is a string but not a 'valid time', so that should give an error as well.

jmcanterafonseca:
null is equivalent as having no expire date, so it should not raise an error

2016-06-20 10:45 GMT+02:00 Ken Zangelin <notifications@github.com>:

kzangeli:
If we wish to accept null as another way for permanent subscriptions, no problem.
Right now I believe we force the expiration to be a string, in the parsing stage and an error is returned if not a string. Very easy to modify. 

I actually believe `null` is better than the empty string for a 'non-expiring subscription' ...

jmcanterafonseca:
sure

2016-06-20 11:12 GMT+02:00 Ken Zangelin <notifications@github.com>:

fgalan:
Moved to master as base branch: https://github.com/telefonicaid/fiware-orion/pull/2573

(I have just remember that some days ago I reralized that git allows to edit PR to change base branch... too late :( )

Issue #2221, opened 2016-06-01
replacing a non existent attribute results in incorrect error


jmcanterafonseca:
PUT  

http://{{host}}:1026/v2/entities/restaurant-utrecht-abcdef987/attrs/removeMe

payload:

```json
{
  "value": false
}
```

HTTP response code is `472` and body is empty

fgalan:
Fixed by PR #2307

PR #2307, opened 2016-06-20
Bug/2221 invalid error updating non existing attribute

kzangeli:
Fixed issue #2221 
Now, the issue is describing an older behavior of the broker, part of the issue has been fixed by another PR. In this PR, the request in #2221 is made receive a 'Attribute Not Found' instead of the former 'Entity Not Found'  (the entity exists but doesn't have the attribute that issue #2221 tries to replace).

Also, plenty of older 'context element not found' texts replaced for the nicer message 'The requested entity has not been found. Check type and id', and 'Attribute Not Found' is also used in more places than before.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -22,3 +22,4 @@
 - Fix: Error payload returned on encountering non-existing entity/attribute in PUT /v2/entities/<entity-id>/attrs/<attr-name> (Issue #1360)
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
+- Fix: Correct error returned when trying to replace a non-existing attribute (Issue #2221)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -23,4 +23,4 @@
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
-- Fix: Correct error returned when trying to replace a non-existing attribute (Issue #2221)
+- Fix: Correct error returned when trying to replace a non-existing attribute in PUT /v2/entities/<entity-id>/attrs/<attr-name> (Issue #2221)
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -3056,7 +3056,7 @@ void processContextElement
 
         if (apiVersion == "v1")
         {
-          responseP->oe.fill(SccContextElementNotFound, "Entity not found. Check type and id", "NotFound");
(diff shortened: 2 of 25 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -23,3 +23,4 @@
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
+- Fix: Correct error returned when trying to replace a non-existing attribute (Issue #2221)
```
fgalan: Maybe the operation that has been fixed (PUT /v2/entities/id/attrs/attrName) should be mentioned?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3053,14 +3053,22 @@ void processContextElement
       if (forwardsPending(responseP) == false)
       {
         cerP->statusCode.fill(SccContextElementNotFound);
-        responseP->oe.fill(SccContextElementNotFound, "No context element found", "NotFound");
+
+        if (apiVersion == "v1")
+        {
+          responseP->oe.fill(SccContextElementNotFound, "Entity not found. Check type and id", "NotFound");
```
fgalan: nit: third parameter is not needed for v1 (althout it doesn't hurt, anyway...)

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -690,7 +690,14 @@ std::string postUpdateContext
     if (fails == response.contextElementResponseVector.size())
     {
       // If all CER result in error, then it isn't a partial update, but a regular NotFound
-      parseDataP->upcrs.res.oe.fill(SccContextElementNotFound, "No context element found", "NotFound");
+      if (ciP->apiVersion == "v1")
+      {
+        parseDataP->upcrs.res.oe.fill(SccContextElementNotFound, "No context element found", "NotFound");
```
fgalan: nit: third parameter is not needed for v1 (althought it doesn't hurt, anyway...)

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 248de8e

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: ok, but with it the calls looks similar between v1 and v2.

However, I just noticed that I have changed the text also for v1 here ... That's not what I intented to do ...

Rollback of that part?

fgalan: Discussed at skype. Rollback.

kzangeli: Fixed in 4066516

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: NTC

fgalan:
LGTM

PR #2310, opened 2016-06-20
Added 'errata' suffix for check_files_compliance.py

kzangeli:
Added 'errata' suffix for check_files_compliance.py, and a copyright header for the errata file, to avoid problems ...


fgalan:
LGTM

Issue #1037, opened 2015-06-26
MIME types right support


fgalan:
(Originally this was a comment on #947)

Raised in https://github.com/telefonicaid/fiware-orion/pull/958/files#r31917180, right processing of Accept in /v2/entities operations is not part of PR #958 and would be addressed in a next PR (maybe developing an overall solution for all /v2/ operations, which are purely based on JSON).

Request only support Content-Type application/json and (in some operations) text/plain. Other types (e.g. XML) should be rejected (I don't remember right now the HTTP code to use, but I think there is one special code for this).

Payload is returned in application/json, thus Accept must include application/json (note that _/_, the default Accept if client doesn't include the header, includes application/json). The exception are the operations that also can return text/plain when the client uses `options=text`. In that case, payload is returned in text/plain, thus Accept must include text/plain.

fgalan:
Implemented in PR https://github.com/telefonicaid/fiware-orion/pull/2323. Moved to the QATestPending milestone for final validation by QA.

fgalan:
Issues in milestone QATestPending are going to be closed (it is not realistic to think we would have people working on behave testing at the present moment). However, issue is labeles with QATestPending label in the case thinks change and we need to recover it.

Issue #2208, opened 2016-05-27
406 Not Acceptable must be returned is Accept doesn't include right MIME type


fgalan:
From http://stackoverflow.com/questions/37458863/get-attribute-value-behaviour-with-accept-header:

```
curl -v orion:1026/v2/entities/Bcn-Welt/attrs/humidity/value --header "Accept: audio/*"
* Hostname was NOT found in DNS cache
*   Trying 172.17.0.13...
* Connected to orion (172.17.0.13) port 1026 (#0)
> GET /v2/entities/Bcn-Welt/attrs/humidity/value HTTP/1.1
> User-Agent: curl/7.35.0
> Host: orion:1026
> Accept: audio/*
> 
< HTTP/1.1 200 OK
< Connection: Keep-Alive
< Content-Length: 9
< Content-Type: text/plain
< Fiware-Correlator: a216e33c-232b-11e6-83f3-0242ac11000d
< Date: Thu, 26 May 2016 10:21:39 GMT
< 
* Connection #0 to host orion left intact
60.000000
```

Note that text/plain is not part of the accepted MIME types in the client `Accept` header (which is: `Accept: audio/*`). Thus, the response should be a 406 Not Acceptable error.

fgalan:
Question answer to http://stackoverflow.com/questions/37458863/get-attribute-value-behaviour-with-accept-header should be edited when this issue got fixed.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/2323.

Issue #2175, opened 2016-05-23
inconsistent behaviour when asking for a value


jmcanterafonseca:
let entity be

```json
{
  "id": "d7fafe27f0416d9ee89ebd2ab42baaddc746fa47",
  "type": "Restaurant",
  "address": {
    "type": "PostalAddress",
    "value": {
      "streetAddress": "Zarate Bidea",
      "addressRegion": "Gipuzkoa",
      "addressLocality": "Zizurkil",
      "postalCode": "20159"
    },
    "metadata": {}
  },
  "capacity": {
    "type": "PropertyValue",
    "value": 200,
    "metadata": {}
  }
}
````

if I run http://192.168.59.103:1026/v2/entities/d7fafe27f0416d9ee89ebd2ab42baaddc746fa47/attrs/capacity/value

response is 

```json
{
  "error": "NotAcceptable",
  "description": "accepted MIME types: text/plain"
}
````
but if I run 

http://192.168.59.103:1026/v2/entities/d7fafe27f0416d9ee89ebd2ab42baaddc746fa47/attrs/address/value

response is

```json
{
  "streetAddress": "Zarate Bidea",
  "addressRegion": "Gipuzkoa",
  "addressLocality": "Zizurkil",
  "postalCode": "20159"
}
````

in the first case response should be 

```json
{
   "value": 200
}
````

fgalan:
Which `Accept` header is being used in each case?

fgalan:
Which `Accept` header is being used in each case?

jmcanterafonseca:
none

jmcanterafonseca:
I think this is a bug, right?

fgalan:
Not sure... we need to research on this a bit to diagnose.

kzangeli:
This bug is being fixed in a current branch, PR soon to be sent.
However, the spec says only compound values can retuen JSON. In the case of String, Number Bool, and Null, only text/plain is accepted.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/2323.

Issue #1886, opened 2016-03-08
[BUG] in "/v2/entities/<id>/attrs/<attr_name>/value" with Accept header different of application/json it is taken as text/plain


iariasleon:
in `/v2/entities/<id>/attrs/<attr_name>/value` with Accept header different of `application/json` it is  taken as text/plain.

#### Datasets
```
      | Accept                            |
      |-----------------------------------|
      | application/x-www-form-urlencoded |
      | application/xml                   |
      | multipart/form-data               |
      | text/html                         |
      | dsfsdfsdf                         |
      | <sdsd>                            |
      | (eeqweqwe)                        |
```

#### Get Request
```
GET http://localhost:1026/v2/entities/room_1/attrs/temperature_0/value
Content-Type: application/json
Fiware-Service: test_update_happy_path
Fiware-ServicePath: /test
Accept: application/x-www-form-urlencoded
```

#### Get Response
```
http code: 200
date: Tue, 08 Mar 2016 14:36:49 GMT
connection: Keep-Alive
content-type: text/plain
content-length: 5
34.67
```

#### Expected Response
```
http code: 406
{"error":"NotAcceptable","description":"accepted MIME types: text/plain"}
```

fgalan:
Test cases_xml/1049_values_mode/attrs_order.test fail could be related with this issue and not really related with XML removal.

fgalan:
Fixed in PR https://github.com/telefonicaid/fiware-orion/pull/2323. @iariasleon please have a look and close if you find it ok.

iariasleon:
LGTM 
```
http code: 406 - Not Acceptable
headers:
   date: Thu, 21 Jul 2016 08:07:19 GMT
   fiware-correlator: 256bb928-4f1a-11e6-81d7-005056a20feb
   connection: Keep-Alive
   content-type: text/plain
   content-length: 81
payload: {"error":"NotAcceptable","description":"acceptable MIME types: application/json"}
```

PR #2323, opened 2016-06-30
Bug/1849 attribute values and multiple accept types

kzangeli:
Fixes issues #1037, #1849, #1886, #2175, and #2208


CHANGES_NEXT_RELEASE
```diff
@@ -30,3 +30,4 @@
 - Fix: Avoid rendering at the same time id and idPattern fields in NGSIv2 subscriptions
 - Fix: Error returned on GET/DELETE requests specifying Content-Type (Issue #2128)
 - Fix: Error responses with '405 Method Not Allowed' now return payload to describe the error, for v2 requests only (Issue #2075, #2078 and #2083)
+- Fix: Implemented more sofisticated HTTP Accept Header handling (Issue #1849)
```
fgalan: The five issue numbers solved by this PR should be mentioned.

In addition, not sure about "more sofisticated" is the best alternative... It is not about sofistication, but about correctness or accuracy.

src/lib/rest/HttpHeaders.cpp
```diff
@@ -0,0 +1,130 @@
...
+  acceptHeaderV.clear();
+}
+
+
+
+/* ****************************************************************************
+*
+* HttpHeaders::accepted - 
+*/
+bool HttpHeaders::accepted(const std::string& mime)
```
fgalan: What about `*/json` and `*/plain`?

Not sure how to deal with them... the "* thing" should be checked in the MIME spec.

src/lib/rest/HttpHeaders.h
```diff
@@ -26,6 +26,37 @@
 * Author: Ken Zangelin
 */
 #include <string>
+#include <vector>
+
+#include "common/MimeType.h"
+
+
+
+// -----------------------------------------------------------------------------
```
fgalan: We typically use `/* ... */` for preambles. Not really important, but it has grabed my attention...

src/lib/rest/rest.cpp
```diff
@@ -262,18 +262,236 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
...
+    ciP->httpStatusCode    = SccBadRequest;
+    ciP->outMimeType       = mimeTypeSelect(ciP, "Error");
+    delete acceptHeaderP;
+    return false;
+  }
+
+
+  //
+  // And now the accept-extensions  (which is ignored for now)
+  // FIXME P4: Implement treatment of accept-extensions
```
fgalan: What about accept-params? Are we doing something with them? In negative case, it deserves a similar FIXME line.

src/lib/rest/HttpHeaders.cpp
kzangeli: */json and */plain are not valid.
only application/* and text/*

NTC

test/functionalTest/cases/0951_get_v2_param_type/get_v2_param_type.test
```diff
@@ -181,12 +181,12 @@ Date: REGEX(.*)
...
 ==============================================
 HTTP/1.1 200 OK
-Content-Length: 11
+Content-Length: 13
 Content-Type: text/plain
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
-V1 A1 E1/T1
+"V1 A1 E1/T1"
```
fgalan: This is another fix different than the one already described in the CNR about Accept header processing. I sugges to describe it in an additional entry in CNR (in this case, without associated issue number, as far as I remember).

src/lib/rest/HttpHeaders.h
kzangeli: :-)
My mistake

src/lib/rest/rest.cpp
kzangeli: q=  is accept-params. There should be a comment somewhere about the anatomy of an 'accept item'

test/functionalTest/cases/1849_attribute_values_and_multiple_accept_types/attribute_values_and_multiple_accept_types_for_number_attribute.test
```diff
@@ -0,0 +1,220 @@
...
+# 08. GET I with Accept header text/*;q=0.3,application/json;q=0.8 - see plain text (q is overridden)
+# 09. GET I with Accept header application/json - see error
+#
+
+echo "01. Create E1 with an attribute of each type (S: string, I: number, B: bool, N: null, O: json object, A: json array)"
+echo "===================================================================================================================="
+payload='{
+  "id": "E1",
+  "S": "S",
+  "I": 3.14,
```
fgalan: I guess that "I" was intended for Integer, but at the end the number is decimal... Maybe D should be more clear.

Actually not very important (you would need to touch a lot of .test files to implemnt that change, so probably it doesn't worth the effort).

test/functionalTest/cases/0951_get_v2_param_type/get_v2_param_type.test
kzangeli: Ah true. This is a 'bug' that I fixed on the way ...

src/lib/rest/HttpHeaders.h
kzangeli: Wait, this was no mistake.

The text in the comment contains  `*/*` and thus would end the C-comment right there, resulting in a compilation error.
That's why I used C++ comments here.

NTC

src/lib/rest/rest.cpp
kzangeli: The comment header of HttpAcceptHeader in HttpHeaders.h - that's where the whole thing is explained.

NTC

test/functionalTest/cases/1849_attribute_values_and_multiple_accept_types/attribute_values_and_multiple_accept_types_for_number_attribute.test
kzangeli: I wanted to use N, but N I needed for Null.
I took I instead. Should have picked D instead, perhaps.
As you say, not really worth the effort.

NTC

test/functionalTest/cases/0951_get_v2_param_type/get_v2_param_type.test
kzangeli: Fixed in 03d1ed1

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 03d1ed1

Issue #2030, opened 2016-04-12
in PATCH v2/subcription if try to update throttling to 0 (default), it is not changed


iariasleon:
in PATCH v2/subcription if try to update throttling to 0 (default), it is not changed in mongo.

#### update subcription request
```
curl -v -X PATCH http://localhost:1026/v2/subscriptions/pplication/json" -H "Fiware-Service: test_update_happy_path" -H "Fiware-ServicePath: /test"  -d '{ "notification": {  "callback": "http://localhost", "throttling": 0, "attributes": ["pepe"] }}'
```

#### update subcription response
```
HTTP 204 No Content
Connection: Keep-Alive
Content-Length: 0
Date: Tue, 12 Apr 2016 16:02:07 GMT
```

#### mongo doc
```
{ "_id" : ObjectId("570d0dcaeb4996bff55a0330"), "reference" : "http://localhost", "entities" : [ { "id" : "Room", "isPattern" : "false" } ], "attrs" : [ "pepe" ], "expiration" : NumberLong(1459864800), "servicePath" : "/test", "throttling" : NumberLong(5), "description" : "One subscription to rule them all", "status" : "active", "conditions" : [ { "type" : "ONCHANGE", "value" : [ "temperature " ] } ], "expression" : { "q" : "temperature>40", "geometry" : "", "coords" : "", "georel" : "" }, "format" : "JSON" }

```

fgalan:
Functest added in PR https://github.com/telefonicaid/fiware-orion/pull/2331 during the 1.3.0 development cycle. However, we have found that the bug was fixed in 1.2.0. Thus I've changed the milestone.

@iariasleon please have a look (in both 1.2.0 and develop) and close if you find it solved.

iariasleon:
LGTM
Request updated
```
curl -v -X PATCH http://localhost:1026/v2/subscriptions/5790c0572b452ae6b944839d -H "Content-Type: application/json" -H "Fiware-Service: test_update_happy_path" -H "Fiware-ServicePath: /test"  -d '{ "throttling": 0}'
```
mongo doc before:
```
[...]

PR #2331, opened 2016-07-07
Bug/2030 set throttling to zero

kzangeli:
Fixes issue #2030.
Actually, that issue was already fixed. This PR just adds a functest for it.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -31,7 +31,7 @@
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
 - Fix: Distinguish between strings and numbers in q string filter (Issue #1129)
 - Fix: proper response for 'GET /v2/entities/<id>?attrs=<attrName>' and 'GET /v2/entities/<id>/attrs?attrs=<attrName>' when <attrName> doesn't exist (#2241)
-- Fix: throttling set to zero (Issue #2030)
+- Fix: throttling set to zero (Issue #2030) (This issue was actually fixed for the release 1.2.0)
 - Fix: In error responses for ngsi9 registrations, the duration is no longer rendered.
 - Fix: Do not allow forbidden chars in description for subscription (#2308)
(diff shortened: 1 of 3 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -30,6 +30,7 @@
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
 - Fix: Distinguish between strings and numbers in q string filter (Issue #1129)
 - Fix: proper response for 'GET /v2/entities/<id>?attrs=<attrName>' and 'GET /v2/entities/<id>/attrs?attrs=<attrName>' when <attrName> doesn't exist (#2241)
+- Fix: throttling set to zero (Issue #2030)
```
fgalan: This seems to be another case of "it was fixed in the past but we don't know when"... We should use the .test to locate the newest failing Orion version in this order: 1.2.1, 1.2.0, 1.1.0, 1.0.0, etc.

kzangeli: Yes, this is another one of those :-(
Will perform those tests and hopefully I can tell when the bug was fixed ...

kzangeli: Fixed in dbfd21e

CHANGES_NEXT_RELEASE
```diff
@@ -31,6 +31,7 @@
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
 - Fix: Distinguish between strings and numbers in q string filter (Issue #1129)
 - Fix: proper response for 'GET /v2/entities/<id>?attrs=<attrName>' and 'GET /v2/entities/<id>/attrs?attrs=<attrName>' when <attrName> doesn't exist (#2241)
+- Fix: throttling set to zero (Issue #2030) (This issue was actually fixed for the release 1.2.0)
```
fgalan: Given this is a change in an already release version, this entry should be moved to the 1.2.0 section in contextBroker.spec.

kzangeli: Fixed in ec7947d

fgalan:
LGTM

Issue #1156, opened 2015-08-12
Simple Query : Support conditions over attribute structured content


jmcanterafonseca:
Example 

``` js
{ 
   "temperature": {
      "value": 23,
      "accuracy": 0.1
   }
}
```

```
q=temperature.accuracy < 0.1
```

Please note that this should work in the case of metadata or in the case of values which are actually Javascript objects. 

``` js
{
   "category": ["luxury", "spanish"]
}
```

```
q=category|=spanish
```

category field contains spanish

or

```
q=category|=[spanish, luxury]
```

category either spanish or luxury 

```
q=category&=[spanish, luxury]
```

category contains at least both spanish and luxury 

```
q=category==[spanish, luxury] 
```

category is exactly an array with two elements spanish and luxury.

fgalan:
What about referring to a particular element of a list? Do we want to support that kind of use case?

jmcanterafonseca:
I cannot see the use case because depending on an specific position in the array can be problematic

fgalan:
The answer to the following question at SOF have to be edited before closing this issue: http://stackoverflow.com/questions/32735329/search-by-attribute-value-orion-ngsiv2/32742522#32742522

fgalan:
JsonPath (http://goessner.net/articles/JsonPath) could be useful as source of inspiration (thanks to 
marcc-orange for the feedback!)

jmcanterafonseca:
My wishlist here is to allow querying without specifying the whole path
i.e. if I have

{
   address {
        "addressCountry": "ES",
        "postalCode": "39001",
        "addressLocality": "Santander"
    }
}

I could specify my query as

/v2/entities?q=addressCountry:ES

without having to say explicitly address.addressCountry

Of course whenever there is ambiguity, an error should be raised.

2016-02-02 20:16 GMT+01:00 Fermín Galán Márquez <notifications@github.com>:

fgalan:
> My wishlist here is to allow querying without specifying the whole path

Achieve that could be hard, taking into account the DB model for the entities. I'd suggest to go step by step. First, to implement filtering for structure values (and metadata) knowing the path to the element used to filter. In a second phase, we can analyze how to achieve "free path" filtering.

jmcanterafonseca:
oh, yes probably you are going to need to extend the DB Model ... as this
is a new functionality ... not present in previous versions ...

[...]

PR #2337, opened 2016-07-07
NGSIv2 spec: 'q' (and new 'mq' filter) for structured attribute values and metadata

fgalan:
This PR includes (partially) the changes described in #1156 in the WIP .apib file of the NGSIv2 specification.

@kzangeli
@jmcanterafonseca

Feedback is welcome!

Merging deadline: Friday July 15th EoB.

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -959,7 +959,7 @@ Response code:
       See [Simple Query Language specification](#simple_query_language). 
 
     + mq: temperature.accuracy<0.9 (optional, string) - A query expression for attribute metadata,
-      composed of a list of statements separated by `;`, i.e., mq=statement;statements;statement.
+      composed of a list of statements separated by `;`, i.e., mq=statement1;statement2;statement3.
       See [Simple Query Language specification](#simple_query_language).
     
     + georel: near (optional, string) - Spatial relationship between matching entities and a
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -566,8 +566,8 @@ The query returns all the entities that match all the matching conditions (AND l
 
 There are two kinds of statements: *unary statements* and *binary statements*.
 
-Binary statements are composed by an attribute path (e.g. `temperature`), an operator and a value
-(whose format depends on the operator), e.g.:
+Binary statements are composed by an attribute path (e.g. `temperature` or `brand.name`), an operator
+and a value (whose format depends on the operator), e.g.:
 
 ```
 temperature==50
@@ -581,7 +581,7 @@ addresses a JSON property name, in accordance with the following rules:
 * If filtering by attribute value (i.e. the expression is used in a `q` query), the rest of tokens (if present)
   represent the path to a sub-property of the *target NGSI attribute value* (which should be a JSON object).
   Such sub-property is defined as the *target property*.
-* If filtering by metadata (i.e. the expression is used in a `q` query), the second token represents a metadata
+* If filtering by metadata (i.e. the expression is used in a `mq` query), the second token represents a metadata
   name associated to the target NGSI attribute, *target metadata*, and the rest of tokens
   (if present) represent the path to a sub-property of the *target metadata value* (which should be a
   JSON object). Such sub-property is defined as the *target property*.
@@ -589,81 +589,82 @@ addresses a JSON property name, in accordance with the following rules:
 The *target property value* is defined as the value of the JSON property addressed by the list of tokens described
 above i.e. the value of the *target property*.
 
-In case only one token is provided (two in the case of filtering by metadata), then the *target property* will
-be the target NGSI attribute itself (or the target metadata in the case of filtering by metadata) and the
-*target property value* will be the target NGSI attribute value (or the target metadata value in the case
-of filtering by metadata). It is assumed that the value of the target NGSI attribute (or the target metadata
-in the case of filtering by metadata) is not a JSON object in this case.
+In case only one token is provided (two in case of filtering by metadata), then the *target property* will
+be the *target NGSI attribute* itself (or the *target metadata* in case of filtering by metadata) and the
+*target property value* will be the *target NGSI attribute* value (or the *target metadata* value in case
+of filtering by metadata). The value of the *target NGSI attribute* (or the *target metadata*
+in case of filtering by metadata) should not be a JSON object in this case.
 
-In the case some of the tokens include `.`, you can use simple quote (`'`) as separator. For example, the following
+In case some of the tokens include `.`, you can use single quote (`'`) as separator. For example, the following
 attribute path `'a.b'.w.'x.y'` is composed by three tokens: the first token is `a.b`, the second token is `w` and
 the third token is `x.y`.
 
 The list of operators (and the format of the values they use) is as follows:
 
 + **Equal**: `==`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature==40`. For an entity to match, it must contain the target
-      property (temperature) and its value be the query value (40) (or include the value, in the case
-      the target property value is an array).
+    + Single element, e.g. `temperature==40`. For an entity to match, it must contain the *target
+      property* (temperature) and the *target property value* must be the query value (40)
+      (or include the value, in case the *target property value* is an array).
     + A list of comma-separated values, e.g. `color==black,red`. For an entity to match, it must
-      contain the target property and its value must be **any** of the values in the list (OR clause)
-      (or include **any** of the values in the list in the case the target property value is an array).
+      contain the *target property* and the *target property value* must be **any** of the values
+      in the list (OR clause) (or include **any** of the values in the list in case the *target
+      property value* is an array).
       E.g. entities with an attribute named `color`, whose value is `black` are a match, while
       entities with an attribute named `color` but whose value is `white` do not match.
     + A range, specified as a minimum and a maximum, separated by `..`, e.g. `temperature==10..20`.
-      For an entity to match, it must contain the target property (temperature),
-      and its value must be in the interval between the minimum and maximum of the range (both included).
-      Ranges can only be used with target properties that represent dates (in ISO8601 format),
-      numbers or strings.
+      For an entity to match, it must contain the *target property* (temperature),
+      and the *target property value* must be in the interval between the minimum and maximum
+      of the range (both included). Ranges can only be used with *target properties* that represent
+      dates (in ISO8601 format), numbers or strings.
 + **Unequal**: `!=`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature!=41`. For an entity to match, it must contain the target
-      property (temperature) and its value must **not** be the query value (41).
+    + Single element, e.g. `temperature!=41`. For an entity to match, it must contain the *target
+      property* (temperature) and the *target property value* must **not** be the query value (41).
     + A list of comma-separated values, e.g. `color!=black,red`. For an entity to match, it must
-      contain the target property, and its value is **not** any of the values
-      in the list (AND clause) (or not include **any** of the values in the list in the case the target
-      property value is an array).
+      contain the *target property* and the *target property value* must **not** be any of the values
+      in the list (AND clause) (or not include **any** of the values in the list in case the *target
+      property value* is an array).
       Eg. entities whose attribute `color` is set to `black` will not match, while entities whose
       attribute `color` is set to `white` will match.
     + A range, specified as a minimum and maximum separated by `..`, e.g. `temperature!=10..20`.
-      For an entity to match, it must contain target property (temperature) and its value must **not**
-      be in the interval between the minimum and the maximum (both included). Ranges can only be used
-      with elements target properties that represent dates (in ISO8601 format), numbers or strings.
+      For an entity to match, it must contain the *target property* (temperature) and the
+      *target property value* must **not** be in the interval between the minimum and the maximum
+      (both included). Ranges can only be used with elements *target properties* that represent dates
+      (in ISO8601 format), numbers or strings.
 + **Greater than**: `>`. The right-hand side must be a single element, e.g. `temperature>42`.
-    For an entity to match, it must contain the target property (temperature) and
-    its value must be strictly greater than that query value (42).
-    This operation is only valid for target properties of type date, numbers or strings (used with
-    target properties of other types may lead to unpredictable results).
+    For an entity to match, it must contain the *target property* (temperature)
+    and the *target property value* must be strictly greater than the query value (42).
+    This operation is only valid for *target properties* of type date, number or string (used with
+    *target properties* of other types may lead to unpredictable results).
 + **Less than**: `<`. The right-hand side must be a single element, e.g. `temperature<43`.
-    For an entity to match, it must contain the target property (temperature) and
-    its value must be strictly less than that value (43).
-    This operation is only valid for target properties of type date, numbers or strings (used with
-    target properties of other types may lead to unpredictable results).
+    For an entity to match, it must contain the *target property* (temperature)
+    and the *target property value* must be strictly less than the value (43).
+    This operation is only valid for *target properties* of type date, number or string (used with
+    *target properties* of other types may lead to unpredictable results).
 + **Greater or equal than**: `>=`. The right-hand side must be a single element, e.g. `temperature>=44`.
-    For an entity to match, it must contain the target property (temperature) and
-    its value must be greater than or equal to that value (44).
-    This operation is only valid for target properties of type date, numbers or strings (used with
-    target properties of other types may lead to unpredictable results).
+    For an entity to match, it must contain the *target property* (temperature)
+    and the *target property value* must be greater than or equal to that value (44).
+    This operation is only valid for *target properties* of type date, number or string (used with
+    *target properties* of other types may lead to unpredictable results).
 + **Less or equal than**: `<=`. The right-hand side must be a single element, e.g. `temperature<=45`.
-    For an entity to match, it must contain the target property (temperature) and
-    its value must be less than or equal to that value (45).
-    This operation is only valid for target properties of type date, numbers or strings (used with
-    target properties of other types may lead to unpredictable results).
+    For an entity to match, it must contain the *target property* (temperature)
+    and the *target property value* must be less than or equal to that value (45).
+    This operation is only valid for *target properties* of type date, number or string (used with
+    *target properties* of other types may lead to unpredictable results).
 + **Match pattern**: `~=`. The value matches a given pattern, expressed as a regular expression, e.g.
-    `color~=ow`.
-    For an entity to match, it must contain the target property (color), and
-    its value must match the string in the right-hand side, 'ow' in this example (`brown` and `yellow`
-    would match, `black` and `white` would not).
-    This operation is only valid for target properties of type string.
+    `color~=ow`. For an entity to match, it must contain the *target property* (color)
+    and the *target property value* must match the string in the right-hand side,
+    'ow' in this example (`brown` and `yellow` would match, `black` and `white` would not).
+    This operation is only valid for *target properties* of type string.
 
 The symbol `:` can be used instead of `==`.
 
-In the case of equal or unequal, if the string to match includes a `,`, you can use simple quote
+In the case of equal or unequal, if s string to match includes a `,`, you can use simple quote
 (`'`) to disable the special meaning of the comma,  e.g: `color=='light,green','deep,blue'`.
 The first example would match a color with the exact value 'light,green' OR 'deep,blue'.
 
 Unary negatory statements use the unary operator `!`, while affirmative unary statements use no
 operator at all.
-The unary statements are used to check for the existence of the target property.
+The unary statements are used to check for the existence of the *target property*.
 E.g. `temperature` matches entities that have an attribute called 'temperature' (no matter its
 value), while `!temperature` matches entities that do not have a temperature called 'attribute'.
  
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -566,7 +566,7 @@ The query returns all the entities that match all the matching conditions (AND l
 
 There are two kinds of statements: *unary statements* and *binary statements*.
 
-Binary statements are composed by an attribute name (e.g. "temperature"), an operator and a value
+Binary statements are composed by an attribute path (e.g. "temperature"), an operator and a value
 (whose format depends on the operator), e.g.:
 
 ```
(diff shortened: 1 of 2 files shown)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+  the metadata value (assuming that the metadata value is a JSON object). If no navigation path is provided
+  (beyond the first element which identifies the metadata name) then the metadata itself
+  is taken into account.
+
 The list of operators (and the format of the values they use) is as follows:
 
 + **Equal**: `==`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an attribute
-      with that exact name (temperature) and value (40).
+    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an element
```
fgalan: Note that "attribute" doesn't works any more here, as the thing under test can be any of attribute, attribute sub-key, metadata or metadata sub-key. Thus, "element under test" is used. Note that the term has been already defined in line 557.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+The attribute path is used to select the *element under evaluation* for the statement. Its general
+syntax consists of an attribute name optionally followed by a *navigation path* using the `.`
+character as delimiter, e.g. "temperature.accuracy.average" refers to the navigation path "accuracy.average"
+within the "temperature" attribute.
+
+How the navigation path is interpreted depends on where the Simple Query Language statement is used:
+
+* If used in a `q` query, then the navigation path refers to a sub-key of the attribute value
+  (assuming that the attribute value is a JSON object). If no navigation path is provided (i.e.
+  the attribute path doesn't have any `.`) then the attribute itself is taken into account.
```
kzangeli: then the **value** of the attribute itself ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+* If used in a `mq` query, then the first element of the navigation path is mandatory and it means the
```
kzangeli: The first element of the path is always the attribute name, right?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+  is taken into account.
+
 The list of operators (and the format of the values they use) is as follows:
 
 + **Equal**: `==`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an attribute
-      with that exact name (temperature) and value (40).
+    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an element
+      under evaluation with that exact name (temperature) and value (40) (or which includes the
+      value, in the case of elements under evaluation which value is an array).
```
kzangeli: in the case of elements under evaluation which value is an array ...

Shorter:
in case the element under evaluation is an array

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+      values in the list (OR clause) (or which includes **any** of the values in the case of
+      elements under evaluation which value is an array).
       E.g. entities with an attribute named `color`, whose value is `black` are a match, while
       entities with an attribute named `color` but whose value is `white` do not match.
     + A range, specified as a minimum and a maximum, separated by `..`, e.g. `temperature==10..20`.
-      For an entity to match, it must contain an attribute with that name (temperature), and its
-      value must be in the interval between the minimum and maximum of the range (both included).
-      Ranges can only be used with attributes that represent dates (in ISO8601 format), numbers or
-      strings.
+      For an entity to match, it must contain an element under evaluation with that name (temperature),
```
kzangeli: In this phrase 'element under evaluation' sounds really strange ...
Just 'element' or 'matching element' would be a lot better

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
+      and its value must be in the interval between the minimum and maximum of the range (both included).
```
kzangeli: minimum and maximum ...   lower and upper limit ?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Accordint to the definition above (line 577):

> Its general syntax [attribute path] consists of an attribute name optionally **followed by** a *navigation path* using the `.`

Thus, strictly speaking, the first element of the *navigation* path is not the attribute name but the token immediately after.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Not sure... At least my aim is to describe how the "element under test" is specified, i.e. the *container* of the value, not the value itself.

However, the whole paragraph is confusing. In particular, I don't like the "sub-key" term, but I haven't figure out a best way of describing...

Alternative wordings (even the whole paragraph) are really welcomed.

marcc-orange:
This _navigation path_ syntax looks like a subset of JsonPath. 

Does the _navigation path_ allows accessing _metadata_ associated to an attribute or just a "navigate" inside the value of the attribute ?

If so, how do you distinguish between both access (e.g. "temperature.metadata.unit" could be a problem if there is a "metadata" key in the attribute value) ?

Do you plan on allowing navigating inside arrays using indexes like "temperatures[0].value" ?

fgalan:
Yes, the syntax is somehow inspired in JsonPath (we took your feedbaci into account https://github.com/telefonicaid/fiware-orion/issues/1156#issuecomment-178766831 ;)

Note that as long as attribute query and metadata query use different URL params (`q` and `mq` in particular) there isn't any right of ambiguity. E.g. `GET /v2/entities?q=temperature.avg<20&mq=temperature.accuracy<0.8` referes to both `avg` sub-key in the `temperature` attribute value and the `accuracy` metadata of the `temperature` attribute.

Regarding navigation inside array, we haven't found yet an use case for that, but it could be added if we find needed. As far as I understand, adding `[`-indexes would be back-ward compatible with what it has been defined so far.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
 
 There are two kinds of statements: *unary statements* and *binary statements*.
 
-Binary statements are composed by an attribute name (e.g. "temperature"), an operator and a value
+Binary statements are composed by an attribute path (e.g. "temperature"), an operator and a value
```
jmcanterafonseca: better `temperature`

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,96 @@ The query returns all the entities that match all the matching conditions (AND l
...
 (whose format depends on the operator), e.g.:
 
 ```
 temperature==50
 temperature<=20
 ```
 
+The attribute path is used to select the *element under evaluation* for the statement. Its general
```
jmcanterafonseca: The syntax of an attribute path consists of a list of tokens separated by the `.` character. This list of tokens addresses a JSON property namer, according to the following rules:

+ the first token is the name of an NGSI attribute (target  NGSI attribute) of a matching entity. 
+ the rest of tokens (if present) represent the path to a sub-property of the target NGSI attribute value, which should be a JSON object.

jmcanterafonseca: The *value under test* is defined as the value of the JSON property addressed by the list of tokens described above.

fgalan: I like the suggested approach... altthough it has to be slighty modified to take into accoun that it is not always a sub-property of the target NGSI attribute value. In some cases (depending on wether `q` or `mq`) it may refer to a metadata, idenfied by the second token.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: what about my proposal *target property* and *target property value* .

jmcanterafonseca: For an entity to match, it must contain a target property which value is equal to the target property value

jmcanterafonseca:
I have made a proposal, defining the concept of target property and target property value, that can help to refine this PR.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I prefer the longer wording. It is more precise, as the thing which is actually an array is the value, not the element itself. Maybe a subtle detail, but in a specification document I prefer to be precise in this case.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 76530a8

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: The new approach has been adopted in 76530a8

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After 76530a8, this has been deeply rewritten, so please consider the original comment as deprecated and re-evaluate it in the new text.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After 76530a8, this has been deeply rewritten, so please consider the original comment as deprecated and re-evaluate it in the new text.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: New approach adopted in 76530a8.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: After 76530a8, this has been deeply rewritten, so please consider the original comment as deprecated and re-evaluate it in the new text.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: This has not been touched by this PR. I mean, if "minimum and maximum of the range" was ok before this PR it should still be ok after this PR.

Anyway, if you still think that is is important, this should be stuff for the "errata" file.

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: In that case, change 'which' for 'whose', or perhaps (as it is plural):

in the case of elements under evaluation whose values are arrays ...

fgalan: The comment may be deprecate after the new changes... could you review on the new text, pls?

fgalan: NTC (closing thread, a new thread will be opened on the new text if needed)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
 ```
 
+The syntax of an attribute path consists of a list of tokens separated by the `.` character. This list of tokens
+addresses a JSON property name, in accordance with the following rules:
+
+* The first token is the name of an NGSI attribute (*target NGSI attribute*) of an entity.
+* If filtering by attribute value (i.e. the expression is used in a `q` query), the rest of tokens (if present)
+  represent the path to a sub-property of the *target NGSI attribute value* (which should be a JSON object).
+  Such sub-property is defined as the *target property*.
+* If filtering by metadata (i.e. the expression is used in a `q` query), the second token represents a metadata
```
kzangeli: typo: ... used in a `mq` query

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+The *target property value* is defined as the value of the JSON property addressed by the list of tokens described
+above i.e. the value of the *target property*.
+
+In case only one token is provided (two in the case of filtering by metadata), then the *target property* will
+be the target NGSI attribute itself (or the target metadata in the case of filtering by metadata) and the
+*target property value* will be the target NGSI attribute value (or the target metadata value in the case
+of filtering by metadata). It is assumed that the value of the target NGSI attribute (or the target metadata
+in the case of filtering by metadata) is not a JSON object in this case.
+
+In the case some of the tokens include `.`, you can use simple quote (`'`) as separator. For example, the following
```
kzangeli: simple -> single

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+attribute path `'a.b'.w.'x.y'` is composed by three tokens: the first token is `a.b`, the second token is `w` and
+the third token is `x.y`.
+
 The list of operators (and the format of the values they use) is as follows:
 
 + **Equal**: `==`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an attribute
-      with that exact name (temperature) and value (40).
+    + Single element, e.g. `temperature==40`. For an entity to match, it must contain the target
+      property (temperature) and its value be the query value (40) (or include the value, in the case
```
kzangeli: and its value *must* be ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+      the target property value is an array).
     + A list of comma-separated values, e.g. `color==black,red`. For an entity to match, it must
-      contain an attribute with that name and the value of the attribute must be **any** of the
-      values in the list (OR clause).
+      contain the target property and its value must be **any** of the values in the list (OR clause)
+      (or include **any** of the values in the list in the case the target property value is an array).
```
kzangeli: in the case => in case

'the' should not be there. Only in very specific constructions, like "In the case of <NAME>, he/she ..."
This is how I 'feel' it, but, I'm not an English expert ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
       E.g. entities with an attribute named `color`, whose value is `black` are a match, while
       entities with an attribute named `color` but whose value is `white` do not match.
     + A range, specified as a minimum and a maximum, separated by `..`, e.g. `temperature==10..20`.
-      For an entity to match, it must contain an attribute with that name (temperature), and its
-      value must be in the interval between the minimum and maximum of the range (both included).
-      Ranges can only be used with attributes that represent dates (in ISO8601 format), numbers or
-      strings.
+      For an entity to match, it must contain the target property (temperature),
+      and its value must be in the interval between the minimum and maximum of the range (both included).
```
kzangeli: its value must be in the interval between the minimum and maximum of the range =>
its value must be between the upper and lower limits of the range

Except, 'its' should be replaced by "the value of the target property", at 'it' refers to the entity.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+      in the list (AND clause) (or not include **any** of the values in the list in the case the target
+      property value is an array).
       Eg. entities whose attribute `color` is set to `black` will not match, while entities whose
       attribute `color` is set to `white` will match.
     + A range, specified as a minimum and maximum separated by `..`, e.g. `temperature!=10..20`.
-      For an entity to match, it must contain an attribute with that exact name (temperature) and
-      whose value is **not** in the interval between the minimum and the maximum (both included).
-      Ranges can only be used with attributes that represent dates (in ISO8601 format), numbers or
-      strings.
+      For an entity to match, it must contain target property (temperature) and its value must **not**
```
kzangeli: must contain **the** target property

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+      be in the interval between the minimum and the maximum (both included). Ranges can only be used
+      with elements target properties that represent dates (in ISO8601 format), numbers or strings.
 + **Greater than**: `>`. The right-hand side must be a single element, e.g. `temperature>42`.
-    For an entity to match, it must contain an attribute with that exact name (temperature) and
-    whose value is strictly greater than that value (42).
-    This operation is only valid for attributes of type date, numbers or strings (used with
-    attributes of other types may lead to unpredictable results).
+    For an entity to match, it must contain the target property (temperature) and
+    its value must be strictly greater than that query value (42).
```
kzangeli: that => the

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+    This operation is only valid for target properties of type date, numbers or strings (used with
```
kzangeli: of type date, number or string

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+    target properties of other types may lead to unpredictable results).
```
kzangeli: unpredictable results ...
Don't we return an error when invalid types are found?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+    For an entity to match, it must contain the target property (color), and
+    its value must match the string in the right-hand side, 'ow' in this example (`brown` and `yellow`
     would match, `black` and `white` would not).
-    This operation is only valid for attributes of type string.
+    This operation is only valid for target properties of type string.
 
 The symbol `:` can be used instead of `==`.
 
-In the case of equal or unequal, if s string to match includes a `,`, you can use simple quote
+In the case of equal or unequal, if the string to match includes a `,`, you can use simple quote
```
kzangeli: single quote

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: Discussed over skype.
Note added to the issue of the PR.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
 
 There are two kinds of statements: *unary statements* and *binary statements*.
 
-Binary statements are composed by an attribute name (e.g. "temperature"), an operator and a value
+Binary statements are composed by an attribute path (e.g. `temperature`), an operator and a value
```
jmcanterafonseca: perhaps for the sake of a better illustration use something like `brand.name` as an example, to note the structure of a path, or even use both examples `temperature` or `brand.name`

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+  (if present) represent the path to a sub-property of the *target metadata value* (which should be a
+  JSON object). Such sub-property is defined as the *target property*.
+
+The *target property value* is defined as the value of the JSON property addressed by the list of tokens described
+above i.e. the value of the *target property*.
+
+In case only one token is provided (two in the case of filtering by metadata), then the *target property* will
+be the target NGSI attribute itself (or the target metadata in the case of filtering by metadata) and the
+*target property value* will be the target NGSI attribute value (or the target metadata value in the case
+of filtering by metadata). It is assumed that the value of the target NGSI attribute (or the target metadata
```
jmcanterafonseca: well, we cannot assume anything. Perhaps we can say nothing.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+the third token is `x.y`.
+
 The list of operators (and the format of the values they use) is as follows:
 
 + **Equal**: `==`. This operator accepts the following types of right-hand side:
-    + Single element, e.g. `temperature==40`. For an entity to match, it must contain an attribute
-      with that exact name (temperature) and value (40).
+    + Single element, e.g. `temperature==40`. For an entity to match, it must contain the target
+      property (temperature) and its value be the query value (40) (or include the value, in the case
+      the target property value is an array).
```
jmcanterafonseca: in case the target property value is an array

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: now it should be `DateTime`

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -566,78 +566,104 @@ The query returns all the entities that match all the matching conditions (AND l
...
+    This operation is only valid for target properties of type date, numbers or strings (used with
+    target properties of other types may lead to unpredictable results).
 + **Less than**: `<`. The right-hand side must be a single element, e.g. `temperature<43`.
-    For an entity to match, it must contain an attribute with that exact name (temperature) and
-    whose value is strictly less than that value (43).
-    This operation is only valid for attributes of type date, numbers or strings (used with
-    attributes of other types may lead to unpredictable results).
+    For an entity to match, it must contain the target property (temperature) and
+    its value must be strictly less than that value (43).
+    This operation is only valid for target properties of type date, numbers or strings (used with
```
jmcanterafonseca: DateTime, Number, Text ... if we want to be consistent with the new types defined

jmcanterafonseca:
LGTM with the comments addressed and the following global changes:

* aligning with the data types defined globally (DateTime, Text, Number, Boolean, etc.)
* when referring to the target property, use *target property* (italics, to stress the fact that is something particular and well defined above).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: However, in the other case (when we have more than one/two token) we are assuming the the value of the attribute/metadata is a JSON object. Literally: "(which should be a JSON object)". Thus, it makes sense to do the opposite assumption here (maybe using the same style, i.e. "it shouldn't be a JSON object" to make the parallelism more explicit).

Having said that, I don't like this paragraph too much. However, it is the best way I have find to describe the special case of "no navigation". Alternatives are welcomed!

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Actually, that refers to a JSON data type, not to the default NGSI that is set when no explicit type is provided. We cannot refer here to DateTime, Number, Text, etc. Note that a given user could use those literals even if the JSON data type is other, eg:

```
POST /v2/entities
{
  "id": "E",
  "A":  {
    "type": "Boolean",
    "value": 5
   }
}
```

As much, we can change the wording to mean "JSON type":

This operation is only valid for target properties of JSON type ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Line 660 is using "simple quote". As it is not related with this PR but would need a future fix before next RC I'll take note at .errata file.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: An interesting link regarding this: https://jakubmarian.com/in-the-case-vs-in-case-in-english/

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: It's just a better wording.
Not sure it belongs to 'errata'.

fgalan: (Side node: the .errata file not only contains errata. It is a kind of log of wording changes we want to introduce in the next RC but that we cannot do right now as they are not related with the new added things due to they introduce "diff noise" when we want to compare current RC with WIPO. Thus,, both actual errata but also better wording cases fit in the file).

kzangeli: Ah, yes, true. I had completely forgotten about that detail.

I would put it in the errata.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Actually, this line shouldn't have changes, as it is not related with the PR. Rollbacking...

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Considering, e.g. "foo: [1, 2, 3]". The target property is "foo". The value of the target property is [1, 2, 3]. What is an array is the value of the target property. Thus, I understand that the origina wording is ok.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 42e964f

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Adde to .errata file in bd077a2

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -930,6 +957,10 @@ Response code:
     + q: temperature>40 (optional, string) - A query expression, composed of a list of statements
       separated by `;`, i.e., q=statement;statements;statement.
       See [Simple Query Language specification](#simple_query_language). 
+
+    + mq: temperature.accuracy<0.9 (optional, string) - A query expression for attribute metadata,
+      composed of a list of statements separated by `;`, i.e., mq=statement;statements;statement.
```
kzangeli: The second *statement* ends in an 's'. Probably not on purpose.

However, why not use;

    i.e., mq=statement1;statement2;statement3 
    
?

Also for 'q' (that might go to errata)

fgalan: Fixed in 9e923e5

kzangeli:
LGTM

crbrox:
LGTM

PR #2340, opened 2016-07-08
Hardening/2339 json native types in v1

kzangeli:
Functest to make sure how the broker treats v1 rendering of v2 created entities, especially about different  JSON native types for attribute and metadata values.

The documentation has been updated to reflect the findings.


fgalan:
LGTM

PR #2345, opened 2016-07-08
Hardening/1368 cppcheck parse args

kzangeli:
Style guide checks incorporated in make targets `di` and `release`.
For now, only `logMsg` and `parseArgs` libraries are included in the style checks.

### DIFF
--- a/scripts/cpplint.py
+++ b/scripts/cpplint.py
@@ -2685,7 +2685,7 @@ def CheckLanguage(filename, clean_lines, linenum, file_extension, include_state,
   if Search(r'\bsprintf\b', line):
     error(filename, linenum, 'runtime/printf', 5,
           'Never use sprintf.  Use snprintf instead.')
-  match = Search(r'\b(strcpy|strcat)\b', line)
+  match = Search(r'\b(strcat)\b', line)  # KZ removed strcpy from the 'verboten list'
   if match:
     error(filename, linenum, 'runtime/printf', 4,
           'Almost always, snprintf is better than %s' % match.group(1))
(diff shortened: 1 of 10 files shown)


scripts/style_check_in_makefile.sh
```diff
@@ -0,0 +1,48 @@
...
+    cat LINT_ERRORS
+    echo "================================================================="
+    cat LINT | grep -v "Done processing"
+    echo 
+    exit 1
+  fi
+}
+
+style_check src/lib/logMsg
+style_check src/lib/parseArgs
```
fgalan: I'd suggest to add now all the list of libraries, but in commented way. That way, we can have always an idea of the pending work.

scripts/cpplint.py
```diff
@@ -2685,7 +2685,7 @@ def CheckLanguage(filename, clean_lines, linenum, file_extension, include_state,
   if Search(r'\bsprintf\b', line):
     error(filename, linenum, 'runtime/printf', 5,
           'Never use sprintf.  Use snprintf instead.')
-  match = Search(r'\b(strcpy|strcat)\b', line)
+  match = Search(r'\b(strcat)\b', line)  # KZ removed strcpy from the 'verboten list'
```
fgalan: There are some comments about this, that are going to be noted in issue #1368

scripts/style_check_in_makefile.sh
kzangeli: Sure

scripts/cpplint.py
fgalan: Added to issue. NTC.

kzangeli: Fixed in fc167ef

scripts/style_check_in_makefile.sh
kzangeli: Fixed in fc167ef

fgalan:
LGTM

Issue #1068, opened 2015-07-13
Support array and object in metadata value


fgalan:
(In some way, this issue is a spin-off of #970 )

fgalan:
After the limitations discovered in NGSIv1 metadata rendering for metadata created with NGSIv2 in PR https://github.com/telefonicaid/fiware-orion/pull/2340, this issue should take into account the following:

* Modify 2339_json_native_types_in_v1/json_native_types_in_v1.test to include metadata cases for JSON array and object.
* Fix the problem, so md created with NGSIv2 gets correctly rendered in NGSIv1 as attributes do.
* Modify documentation in "Native JSON types" section in v1_v2_coexistence.md to remove the sentence about the limitation in metadata.

fgalan:
The part regarding limitations in rendering non-string medatada has been fixed in PR #2348

fgalan:
After completing this issue documentation at https://github.com/telefonicaid/fiware-orion/blob/develop/doc/manuals/user/ngsiv2_implementation_notes.md has to be modified to remove:

> NGSIv2 specification allows metadata value to be JSON Array or Object. However, such datatypes are yet to be implemented for metadata values in Orion.

fgalan:
The two compund cases used in cases/0000_ngsi10_compound_values/compound_convop_json.test for attrs could be used for metadata test.

fgalan:
Impllemented in #2377.

This should be now checked by QA in order to close definitively. Not sure what to do... leave it in 1.3.0 milestone (if verificaiton is going to happen soon) or move to QATestPending in order to accumulate with others for the future. @iariasleon what do you think?

iariasleon:
LGTM

PR #2348, opened 2016-07-11
Correct rendering of metadata values in v1

kzangeli:
Fixes part of #1068
Correct rendering of metadata values in v1

### DIFF
--- a/src/lib/ngsi/Metadata.cpp
+++ b/src/lib/ngsi/Metadata.cpp
@@ -193,12 +193,33 @@ std::string Metadata::render(const std::string& indent, bool comma)
 {
   std::string out     = "";
   std::string tag     = "contextMetadata";
-  std::string xValue  = stringValue;
+  std::string xValue  = toStringValue();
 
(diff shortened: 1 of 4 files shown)


src/lib/ngsi/Metadata.cpp
```diff
@@ -193,12 +193,33 @@ std::string Metadata::render(const std::string& indent, bool comma)
...
+    out += indent + "  " + JSON_STR("value") + ": " + xValue;
+  }
+  else if (valueType == orion::ValueTypeNone)
+  {
+    out += indent + "  " + JSON_STR("value") + ": " + xValue; 
+  }
+  else
+  {
+    out += indent + "  " + JSON_STR("value") + ": " + JSON_STR("unknown json type");
+  }
```
kzangeli: Soon we will treat json vectors and objects also and this last `else` will never be reached. It's dead code but good to have as a security guard ...

fgalan:
LGTM

PR #2351, opened 2016-07-12
FIX change log level to uppercase in log API doc

fgalan:
It has been recently decided that the standard codification for log levels is in upper case. Thus, documentation should use upper case.

(Anyway, it is ok to supporr also lowercase in the implementation, as a "convenience" alternative flavour of the API).


kzangeli:
LGTM

PR #2358, opened 2016-07-14
Reduce image docker size

cesarjorgemartinez:


AlvaroVega:
LGTM

docker/Dockerfile
```diff
@@ -93,8 +90,11 @@ RUN \
            /usr/local/lib/libgtest* \
            /usr/local/share/aclocal/gtest* \
            /opt/fiware-orion \
+           /usr/share/cracklib \
```
fgalan: It is safe to directly remove directories from /usr/share? Shouldn't they be removed using "yum remove XXX"?

docker/Dockerfile
```diff
@@ -6,18 +6,16 @@ ENV ORION_USER orion
...
 
 RUN \
+    adduser --comment "${ORION_USER}" ${ORION_USER} && \
     # Install dependencies
     yum -y install epel-release && \
     yum -y install \
       boost-devel \
       bzip2 \
       cmake \
-      epel-release \
```
fgalan: I though that some epel-release dependencies were required, but I understand that if you have tested with this removed everthing works ok with `FROM centos:centos6.6`.

cesarjorgemartinez:
Its repeated epel-release, I do removed one and install at first

cesarjorgemartinez:
The remove cracklib is a recommendation to reduce image size.
Cannot be erased by yum because yum uses hard dependencies and try to uninstall essential packages

docker/Dockerfile
AlvaroVega: epel-release is just installed by previous yum command

docker/Dockerfile
fgalan: As discussed at skype, it would be good to add some comment above this line to explain a little bit. Something like the following suggestion:

"# We cannot remove the following using yum as yum uses hard dependencies and doing so will uninstall essential packages. However we have empirically check that rm these libs is fine"

docker/Dockerfile
```diff
@@ -106,14 +106,17 @@ RUN \
         boost-iostreams boost boost-date-time \
         boost-test boost-graph boost-signals \
         boost-program-options boost-math \
-        openssh openssh-clients libedit \
+        openssh openssh-clients libedit hwdata dbus-glib fipscheck* *devel sysvinit-tools \
         && \
+    rpm -qa groff | xargs -r rpm -e --nodeps && \
```
fgalan: As discussed at skype, it would be good to add some comment above this line to explain a little bit. Something like the following suggestion:

"# We cannot use yum to remove groff, as as yum uses hard dependencies and doing so will uninstall essential packages"

docker/Dockerfile
```diff
@@ -106,14 +106,17 @@ RUN \
...
     yum clean all && \
-    bash -c 'find /usr/share/locale -maxdepth 1 -mindepth 1 -type d | grep -v -e "en_US" | xargs rm -rfv' && \
+    rpm -vv --rebuilddb && rm -rf /var/lib/yum/yumdb && rm -rf /var/lib/yum/history && \
+    find /usr/share/locale -mindepth 1 -maxdepth 1 ! -name 'en_US' ! -name 'locale.alias' | xargs -r rm -r && rm -f /var/log/*log && \
     bash -c 'localedef --list-archive | grep -v -e "en_US" | xargs localedef --delete-from-archive' && \
-    mv -f /usr/lib/locale/locale-archive /usr/lib/locale/locale-archive.tmpl && \
```
fgalan: As discussed at skype, it would be good to add some comment above this line to explain a little bit. Something like the following suggestion:

"# We use cp instead of mv due to [reason]. We use /bin/cp instead of cp to avoid alias, we in some cases has been problematic"

docker/Dockerfile
fgalan: NTC

fgalan:
LGTM

PR #2366, opened 2016-07-15
Fix for subscription case in  API Walkthrough (v2)

LeonanCarvalho:
In the example it is said that:
"For example, in this case, when Room1 `pressure` changes, the Room1 temperature value is notified"
But the temperature of the object is being used as a condition attribute.


fgalan:
LGTM

Thanks for the contribution!

Issue #643, opened 2014-11-13
Smart change detection logic for compound attribute values


fgalan:
checkAndUpdate() function includes the following comment:

```
// FIXME P6: in the case of compound value, it's more difficult to know if an attribute
// has really changed its value (many levels have to be traversed). Until we can develop th
// matching logic, we consider actualUpdate always true.
```

This issue is about solving this (some users has asked about this at SOF: http://stackoverflow.com/questions/26862953/contextbroker-vectors-update-with-unchanged-value-causes-onchange-subscription-n)

**Effort: 4 man day**

fgalan:
Recently (after implementing #1068) is is not a just a matter of attribute compound values, but also metadata compound values. The logic in equalMetadataValue() is also affected.

See this PR comment for details: https://github.com/telefonicaid/fiware-orion/pull/2377/files#r71670648

fgalan:
PR https://github.com/telefonicaid/fiware-orion/pull/4444

PR #2377, opened 2016-07-20
Feature/1068 metadata value as compound

kzangeli:
Fixed issue  #1068 - support for compounds in metadata value.
Also removed a part of the metadata parsing code, specifically the part that accepts 'keyValue-style' metadata values, e.g.:

    "metadata": { "A1": 12 }

This is pretty old and should have been removed a long time ago.


src/lib/jsonParseV2/parseMetadata.cpp
```diff
@@ -53,34 +56,44 @@ static std::string parseMetadataObject(const Value& start, Metadata* mP)
...
+        mdP->valueType     = orion::ValueTypeNone;
+      }
+      else if ((type == "Array") || (type == "Object"))
+      {
+        mdP->valueType = orion::ValueTypeObject;  // Used both for Array and Object ...
+        std::string r = parseMetadataCompoundValue(iter, mdP, NULL);
+        if (r != "OK")
+        {
+          alarmMgr.badInput(clientIp, "json error in Metadata compound value");
+          return "json error in ContextAttributeObject::Vector";
```
kzangeli: Hmmm, didn't see this in my pre-pr check ...

CHANGES_NEXT_RELEASE
```diff
@@ -22,6 +22,8 @@
 - Fix: A bug fixed in URI param 'q', avoiding false matches when using string match operator (~=) that matches AND some operator that does not match.
```
fgalan: Just for not forgetting: have a look to FIXME "1068" on MongoCommonUpdate.cpp file.

CHANGES_NEXT_RELEASE
```diff
@@ -22,6 +22,8 @@
...
 - Add: New param 'mq' (both in URI param for list all entities operation and expression sub-field in subscriptions) for matching in metadata (#1156)
 - Add: FIWARE::StringQuery::Metadata scope (#1156)
+- Add: accepting compound values for metadatas (Issue #1068)
+- Fix: no longer accepting 'keyValues-style' values for metadatas (e.g. "metadata": { "m1": 6 } )
 - Fix: GET /v2/subscriptions/subId error for non existing subscriptions (now using 404 Not Found, previously using 400 Bad Request)
 - Fix: 472 Unknown returned as HTTP status code by POST /v2/op/update in some cases (now using 404 Not Found).
 - Fix: Correctly returning error on present but empty attribute types in PATCH v2/entites/{id} operation (Issue #1785)
```
fgalan: I'd suggest to include ih this PR the changes in documentation described at: https://github.com/telefonicaid/fiware-orion/issues/1068#issuecomment-232749454

kzangeli:
There is a comment in the function equalMetadataValues(), in MongoCommonUpdate.cpp:

    // If type exists in both metadata elments, check if they are the same
    if (md1TypeExist && md2TypeExist)
    {
      if (getFieldF(md1, ENT_ATTRS_MD_TYPE).type() != getFieldF(md2, ENT_ATTRS_MD_TYPE).type())
      {
        return false;
      }
    
    switch (getFieldF(md1, ENT_ATTRS_MD_TYPE).type())
    {
      /* FIXME not yet, issue #1068 Support array and object in metadata value
      case Object:
      ...
      break;

      case Array:
      ...
        break;
    */

After merging this PR, 1068 is fixed and what is mentioned in this comment could be implemented.

src/lib/jsonParseV2/parseMetadata.cpp
kzangeli: Fixed in f83152d

CHANGES_NEXT_RELEASE
kzangeli: Fixed in bc99af8

src/lib/jsonParseV2/parseMetadataCompoundValue.cpp
```diff
@@ -0,0 +1,337 @@
...
+      //
+      if ((nodeType == "Object") || (nodeType == "Array"))
+      {
+        parseMetadataCompoundValue(iter, mdP, cvnP);
+      }
+
+      ++counter;
+    }
+  }
+
```
fgalan: Two blank lines? Is this "style compliant"?

kzangeli: We don't really have a strict rule for this.
First of all, in this exact case, there is no reason to have two blank lines, and I will fix it.

I use two blank lines to separate 'logical groups' of code inside a function. A 'code-group' always starting with a comment, of course. Not sure this is mentioned in the style guide

test/functionalTest/cases/1068_metadata_value_as_compound/metadata_value_as_compound_array.test
```diff
@@ -0,0 +1,106 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: Some test notifying compound metadata should be included, to check that they are correctly rendered in notification payloads (current test only consider queries, as fas as I have checked).

test/functionalTest/cases/1367_time_measures/time_measure_stats.test
```diff
@@ -43,13 +43,12 @@ brokerStart CB 101,38 IPv4 -statTiming
...
-# 06. GET /statistics
-# 07. GET /v1/contextTypes
-# 08. GET /statistics
+# 02. Create entity E2/A1-A5 with v1 JSON
+# 03. GET /statistics
+# 04. Create entity E3/A1-A5 (identical) with v2 JSON
+# 05. GET /statistics
+# 06. GET /v1/contextTypes
+# 07. GET /statistics
 #
```
fgalan: Test "script" description has been changed, but no actual changes in the steps... was this due to description was wrong so you have take the opportunity to fix in this PR?

kzangeli: Yes, the description was erroneous so I fixed it (now that I had to touch the file anyways)

NTC

CHANGES_NEXT_RELEASE
fgalan: I have been having a closer look to this. The fixme is located in `equalMetadataValue()` function, which in secuence is called from `equalMetadata()` which is secuence is used in the logic to decide if an attribute values has changed and a notification has be triggered or not:

```
    /* In the case of simple value, we consider there is an actual change if one or more of the following are true:
     *
     * 1) the value of the attribute changed (see attrValueChanges for details)
     * 2) the type of the attribute changed (in this case, !attr.hasField(ENT_ATTRS_TYPE) is needed, as attribute
     *    type is optional according to NGSI and the attribute may not have that field in the BSON)
     * 3) the metadata changed (this is done checking if the size of the original and final metadata vectors is
     *    different and, if they are of the same size, checking if the vectors are not equal)
     */
      actualUpdate = (attrValueChanges(attr, caP, apiVersion) ||
                      ((caP->type != "") && (!attr.hasField(ENT_ATTRS_TYPE) ||
                                             getStringFieldF(attr, ENT_ATTRS_TYPE) != caP->type) ) ||
                      mdNew.nFields() != mdSize || !equalMetadata(md, mdNew));
```

So, not implementing the array/object cases in `equalMetadataValue()` as described in the FIXME will result in that part of the || be always true (note the `!`). In other words, any change in metadata compound is considered a change, although the new JSON could be structurally equal to the old one.

What we are doing in the same case for attributes? The same approach:

```
  if (caP->compoundValueP == NULL)
  {
    <the above code, which is for simple attribute values>
  }
  else
  {
      // FIXME P6: in the case of compound value, it's more difficult to know if an attribute
      // has really changed its value (many levels have to be traversed). Until we can develop the
      // matching logic, we consider actualUpdate always true.
      actualUpdate = true;
  }
```

Actually, we have an issue for that: https://github.com/telefonicaid/fiware-orion/issues/643

Thus, what we should do? I don't think we should solve this in this PR, but we can use this PR to "mark" the code in te appropiate places. In particular:

*  "// FIXME P6" above to "// FIXME #643 P6"
* "FIXME not yet, issue #1068 Support array and object in metadata value" -> "FIXME #643 P6: metadata array/object are now supported, but we haven't implemented yet the logic to compare compounds between them".

In addition, #643 will be expanded, to include a comment about this (I'll take care of that).

src/lib/jsonParseV2/parseMetadataCompoundValue.cpp
fgalan: Just called my attention... you can set NTC if you find this shouldn't change.

kzangeli: No no, I agree. There should be only one blank line in this particular place

test/functionalTest/cases/1068_metadata_value_as_compound/metadata_value_as_compound_array.test
kzangeli: Fixed in b260452

src/lib/jsonParseV2/parseMetadataCompoundValue.cpp
kzangeli: Fixed in b260452

CHANGES_NEXT_RELEASE
kzangeli: Fixed in f2aacfd

fgalan:
LGTM

PR #2385, opened 2016-07-27
Feature/1156 qfilters and compounds

kzangeli:
Fixes part of issue #1156 
Namely, q-string-filter matching in queries over attributes with compound values.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -22,6 +22,7 @@
 - Fix: A bug fixed in URI param 'q', avoiding false matches when using string match operator (~=) that matches AND some operator that does not match.
 - Add: New param 'mq' (both in URI param for list all entities operation and expression sub-field in subscriptions) for matching in metadata (#1156)
 - Add: FIWARE::StringQuery::Metadata scope (#1156)
+- Add: q-string filters for compound values of attributes (Issue #1156)
 - Add: accepting compound values for metadatas (Issue #1068)
 - Fix: no longer accepting 'keyValues-style' values for metadatas (e.g. "metadata": { "m1": 6 } )
 - Fix: GET /v2/subscriptions/subId error for non existing subscriptions (now using 404 Not Found, previously using 400 Bad Request)
@@ -36,7 +37,6 @@
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
-- Add: q-string filters for compound values of attributes (Issue #1156)
 - Fix: Metadata identification based on just name instead of name+type (metadata representation at DB changed from vector to object)  (#1112)
 - Fix: Distinguish between strings and numbers in q string filter (Issue #1129)
 - Fix: proper response for 'GET /v2/entities/<id>?attrs=<attrName>' and 'GET /v2/entities/<id>/attrs?attrs=<attrName>' when <attrName> doesn't exist (#2241)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -36,6 +36,7 @@
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
+- Add: q-string filters for compound values of attributes (Issue #1156)
 - Fix: Metadata identification based on just name instead of name+type (metadata representation at DB changed from vector to object)  (#1112)
 - Fix: Distinguish between strings and numbers in q string filter (Issue #1129)
 - Fix: proper response for 'GET /v2/entities/<id>?attrs=<attrName>' and 'GET /v2/entities/<id>/attrs?attrs=<attrName>' when <attrName> doesn't exist (#2241)
--- a/src/lib/rest/StringFilter.cpp
+++ b/src/lib/rest/StringFilter.cpp
@@ -42,7 +42,7 @@ using namespace mongo;
 
 
 
-// Temporary debug function
+// FIXME PR: Temporary debug function
 const char* opName(StringFilterOp op)
 {
   switch (op)
--- a/src/lib/rest/StringFilter.cpp
+++ b/src/lib/rest/StringFilter.cpp
@@ -42,6 +42,27 @@ using namespace mongo;
 
 
 
+// Temporary debug function
+const char* opName(StringFilterOp op)
+{
+  switch (op)
+  {
+  case SfopExists:              return "Exists";
+  case SfopNotExists:           return "NotExists";
+  case SfopEquals:              return "Equals";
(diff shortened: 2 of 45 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -36,6 +36,7 @@
 - Fix: invalid regex patterns detection (#968)
 - Fix: Better error returned on invalid geoquery (Issue #2174)
 - Fix: Better error returned on attribute not found in GET /v2/entities/<entity-id>/attrs/<attr-name>/value (Issue #2220)
+- Add: q-string filters for compound values of attributes (Issue #1156)
```
fgalan: I'd suggest to group all the lines related #1156 together, i.e. move this one to L25

src/lib/rest/StringFilter.cpp
```diff
@@ -42,6 +42,27 @@ using namespace mongo;
 
 
 
+// Temporary debug function
```
fgalan: To avoid "leaking" this function into final develop branch, please add a "FIXME PR" in this place.

src/lib/rest/StringFilter.cpp
```diff
@@ -1255,7 +1328,20 @@ bool StringFilter::mongoFilterPopulate(std::string* errorStringP)
...
+ 
+    //
+    // Also, if the left hand side contains a compound path, then we are
+    // dealing with matching of a compound item.
+    //
+    if (itemP->compoundPath.size() != 0)
+    {
+      if (itemP->type == SftQ)
+      {
+        left = std::string(ENT_ATTRS) + "." + itemP->attributeName + ".value." + itemP->compoundPath;
```
fgalan: If this is going to be a filter for the DB, then ".value." should be "." + ENT_ATTRS_VALUE + "."

(I'm writting this without checking code... pls, check that ENT_ATTRS_VALUE is the right name).

CHANGES_NEXT_RELEASE
fgalan: What about "metadata"? Is this PR covering compound filters on metadata?

src/lib/rest/StringFilter.cpp
```diff
@@ -1587,13 +1662,29 @@ bool StringFilter::qMatch(ContextElementResponse* cerP)
     {
       ContextAttribute* caP = cerP->contextElement.getAttribute(itemP->attributeName);
 
-      if ((itemP->op == SfopExists) && (caP == NULL))
```
fgalan: This changes on qMatch() are not really needed for the .test incluced in this PR? I understand that this is work "in advance" for the next PR, the one that will complete the functionality for notifications. Is my understanding correct?

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
```diff
@@ -0,0 +1,221 @@
...
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0-255
+
+--SHELL--
+
+#
+# 01. Create E1 with A1 with compound value containing all JSON types
+#
+# FIXME P2, once 'null' values are supported, these three test cases must change
```
fgalan: I'd suggest to include a reference to the issue number about null filter implementation.

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
```diff
@@ -0,0 +1,221 @@
+ Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: I'd suggest to rename this .test as .test.DISABLED

src/lib/rest/StringFilter.cpp
kzangeli: Yes, I was already starting with subscriptions when we decided to cut the PR in the middle to avoid having the PR too big.
Functests for it have been delayed to the next PR but this change in  the middle of a touched file I  prefer to include (don't want to lose it .. :-))

NTC

CHANGES_NEXT_RELEASE
kzangeli: Fixed in a43b9e2

No, metadatas will come after.
There will be a total of 4 PRs:
1. queries on attribute compound values
2. subscriptions with q-filters on attribute compound values
3. queries on metadata compound values
4. subscriptions with mq-filters on metadata compound values

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in a43b9e2

src/lib/rest/StringFilter.cpp
kzangeli: Ah, of course, there is a define in there somewhere for "value".
No probs, I'll look it up.

src/lib/rest/StringFilter.cpp
fgalan: Just thinking if the reviewer will get crazy trying to find changes related with the new functionality in the new PR that are not actually in that PR... 

Although from a pragmatic point of view, I wouldn't try to remove this bits from this PR to be  "moved" to the next. That could be complex and error prone. Thus, I agree in NTC (but try to do the cut in a cleaner way next time ;)

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
kzangeli: Sure, not a bad idea.
I fix it as how it must be after implementing NULL (Issue 2359) and I leave it as DISABLED

kzangeli: Fixed in 5f584b0

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
kzangeli: Fixed in 5f584b0

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in 5f584b0

CHANGES_NEXT_RELEASE
fgalan: A summary of "what is pending after this PR" has been included in #1156. I understand it matches your list above.

fgalan:
LGTM

CHANGES_NEXT_RELEASE
kzangeli: Yes.

What we don't have included is that idea to match 'items in the middle':

E.g.

    q=type==12   would match A1.f.type==12

Not sure I like that feature however

fgalan: I understand you refer to "free search without knowing the path". Finally, that is not part of #1156 scope, see: https://github.com/telefonicaid/fiware-orion/issues/1156#issuecomment-233252369

> (*) For the parts not covered in the current NGSIv2 spec (e.g. free search without knowing the path) a new issue should be created.

kzangeli: Yes, that is exactly what I was referring to.
Good :-)

Issue #2055, opened 2016-04-20
funcTests: option style parameters in brokerStart and more functions


kzangeli:
The function brokerStart accepts three fixed parameters and from param 4 is considered 'extra params' to use when launching the broker. This is very unfortunate and leads to problems.

Instead, we should adapt the technique used in orionCurl (same file: harnessFunctions.sh), which is an option parsing technique, like the one used in most executables.

Other functions to consider for the same modification:

* dbInit
* localBrokerStart (brokerStart is a wrapper for localBrokerStart)
* accumulatorStart

And then we have a few functions that doesn't need this new technique, but we could simulate the same thing for symmetry reasons:

* brokerStop
* dbDrop
* accumulatorStop
* accumulatorDump
* accumulatorCount

The need for these modifications is pretty old, but was 're-discovered' during review of PR #2054

kzangeli:
Not to forget, all options to the functions must have valid default values.
For example, brokerStart should start in "CB mode" by default

PR #2409, opened 2016-07-29
hardening/no_cache_in_test_harness

kzangeli:
New options for harness test script and brokerStart function: --noCache and --cache

### DIFF
--- a/test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
+++ b/test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
@@ -25,7 +25,7 @@ notifyFormat_ngsi10
 
 --SHELL-INIT--
 dbInit CB
-brokerStart CB 38,101 IPv4 -statSemWait
+brokerStart CB 38,101 IPv4 -statSemWait --cache
 accumulatorStart
 
 --SHELL--
--- a/test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
+++ b/test/functionalTest/cases/1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
@@ -26,7 +26,7 @@ Subscription Cache Throttling and expiration after Refresh
 --SHELL-INIT--
 dbInit CB
 dbInit CB t1
-brokerStart CB 38,101 IPv4 -multiservice -subCacheIval 10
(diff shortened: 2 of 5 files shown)


test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
```diff
@@ -25,7 +25,7 @@ notifyFormat_ngsi10
 
 --SHELL-INIT--
 dbInit CB
-brokerStart CB 38,101 IPv4 -statSemWait
+brokerStart CB 38,101 IPv4 -statSemWait --cache
```
fgalan: I'd suggest to use `-cache` to align with the same style used for other parameters in `brokerStart` (e.g. `-statSemWait`).

kzangeli: We have a nice little mess between '--' and '-' options for the long names ...

-statSemWait is actually *not* an option for brokerStart.
brokerStart just copies it down to the broker.
--cache *is* an option for brokerStart, and just like all other script functions (I hope), it uses '--'.

So, do you still want to change '--cache' ?

The best option actually would be to modify the broker to use '--' instead of '-' for the long options. Support both '-' and '--', to finally drop '-'.

And, of course, the entire arg-parsing for brokerStart should be modified. That is in a separate issue, I believe.

crbrox:
LGTM

test/functionalTest/cases/1048_subscription_cache/three_subscriptions.test
fgalan: Ok, by the moment and while we can address a better refactoring in the brokerStart function (btw, which number has that issue?) the criteria of `-` for "options to be passed to CB" and `--` for "options to be consumed by the brokerStart() function seems ok.

Regarding:

> The best option actually would be to modify the broker to use '--' instead of '-' for the long options. Support both '-' and '--', to finally drop '-'.

That involves a very deep change which would break all the tooling which relies in CB parametrization (Puppet, Ansible, etc.) with very little gain. Thus, I think is better to keep as CLI works now.

fgalan: I found the issue. This is: https://github.com/telefonicaid/fiware-orion/issues/2055

fgalan: NTC

fgalan:
LGTM

Issue #2207, opened 2016-05-27
Excessive decimal digits in number values when substituted in custom payload


iariasleon:
Replacing attribute values in a Custom template

Excesive decimal digits in number values when substituded in custom payload template

if we try to replace a integer value (34), This is rounded to 6 decimal (34.000000). 

Whit this payload `The temperature is ${temperature} degrees` using `text/plain`. We get these answers for these values.
```
  The temperature is 34.123457 degrees   - (value is 34.12345678)
  The temperature is 34.000000 degrees   - (value is 34)
  The temperature is 34.400000 degrees   - (values is 34.4)
```

Any correction will penalize the performance.

fgalan:
Very similar to https://github.com/telefonicaid/fiware-orion/issues/2176

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2410

fgalan:
Please, @iariasleon close this if you find it fixed (after PR merge, of course ;)

fgalan:
Until https://github.com/telefonicaid/fiware-orion/issues/2425 can be addressed, the right value for the first case would be:

```` 
The temperature is 34.12345700 degrees
```

iariasleon:
```
    | attr value           |          payload                                  |
    |----------------------|---------------------------------------------------|
    | 34                   | The temperature is 34 degrees celsius             |
    | 33.3                 | The temperature is 33.3 degrees celsius           |
    | 34.54654654654654646 | The temperature is 34.5465 degrees celsius        | 
    | 3434534000000        | The temperature is 3.43453e+12 degrees celsius    |
```

Issue #2176, opened 2016-05-23
when asking for a number value in text/plain spurious decimals are added


jmcanterafonseca:
let entity be the same as described in #2175

If 

http://192.168.59.103:1026/v2/entities/d7fafe27f0416d9ee89ebd2ab42baaddc746fa47/attrs/capacity/value

is run (accept text/plain)

result is 

200.000000

instead of 200.

fgalan:
Related with https://github.com/telefonicaid/fiware-orion/issues/2207

fgalan:
Fixed by PR #2410

PR #2410, opened 2016-08-01
FIX render float values with 9 digits precision in all cases, avoding spurious decimals

### LINKED ISSUE #2383: Entity geo:json type rounds geometry coordinate precision
I'm trying to create (and update) a context with geo:json attribute  type with precision of seven decimals but always it's rounded to 4 decimals somehow.

Operation: `POST http://localhost:1026/v2/entities/Device813/attrs?type=Devices&option=append`

Payload:  
```json
{
    "location": {
        "value": {
            "type": "Point",
            "coordinates": [
                -42.5124478,
                -22.3006726
            ]
        },
        "type": "geo:json"
    }
}
```

- Result:
```json
{
    "id": "Device813",
    "type": "Devices",
    "location": {
        "type": "geo:json",
        "value": {
            "type": "Point",
            "coordinates": [
                -42.5124,
                -22.3007
            ]
        },
        "metadata": {}
    }
}
```

> 1.2.1 (git version: af44fd1fbdbbfd28d79ef4f929e871e515b5452e)

fgalan:
Fixes #2383, #2207, #2176 

@kzangeli 
@crbrox

### DIFF
--- a/src/lib/common/JsonHelper.cpp
+++ b/src/lib/common/JsonHelper.cpp
@@ -162,10 +162,6 @@ std::string objectToJson(std::map<std::string, std::string>& list)
 */
 JsonHelper::JsonHelper(): empty(true)
 {
-  /* Set format  for floats (it doesn't affect integers). Note that std::fixed
-   * is not used (we have found that it would add spurious decimals in some cases */
-  ss << std::setprecision(PRECISION_DIGITS);
-
   ss << '{';
 }
 
@@ -224,12 +220,19 @@ void JsonHelper::addNumber(const std::string& key, long long value)
 * JsonHelper::addFloat -
 */
 void JsonHelper::addFloat(const std::string& key, float  value)
-{
+{  
+  unsigned int oldPrecision = ss.precision();
+  ss << std::fixed << std::setprecision(decimalDigits(value));
+
   if (!empty)
   {
     ss << ',';
   }  
   ss << toJsonString(key) << ':' << value;
+
+  // Reset stream to old parameters (whichever they are...)
+  ss.unsetf(std::ios_base::fixed);
+  ss << std::setprecision(oldPrecision);
 
   empty = false;
 }
--- a/src/lib/common/JsonHelper.cpp
+++ b/src/lib/common/JsonHelper.cpp
@@ -25,6 +25,7 @@
 
 #include "common/JsonHelper.h"
 #include "common/string.h"
+#include "common/limits.h"
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -161,8 +162,9 @@ std::string objectToJson(std::map<std::string, std::string>& list)
 */
 JsonHelper::JsonHelper(): empty(true)
 {
-  /* Set format  for floats (it doesn't affect integers) */
-  ss << std::fixed << std::setprecision(9);
+  /* Set format  for floats (it doesn't affect integers). Note that std::fixed
+   * is not used (we have found that it would add spurious decimals in some cases */
+  ss << std::setprecision(PRECISION_DIGITS);
 
   ss << '{';
 }
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -808,7 +808,7 @@ unsigned int decimalDigits(double d)
   double intPart;
   double decimalPart = fabs(modf(d, &intPart));
 
-  while (decimalPart > 0)
+  while (decimalPart > PRECISSION)
   {
     digits++;
     decimalPart *= 10;
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -27,6 +27,7 @@
 #include <string.h>
 #include <string>
 #include <vector>
+#include <math.h>    // modf
 
 #include "logMsg/logMsg.h"
 #include "logMsg/traceLevels.h"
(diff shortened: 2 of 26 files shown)


crbrox:
LGTM

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_deeper.test
```diff
@@ -653,7 +653,7 @@ Fiware-Correlator: REGEX([0-9a-f\-]{36})
                             "F": false, 
                             "N": null, 
                             "T": true, 
-                            "f": 3.1400000000000001, 
+                            "f": 3.14,
```
kzangeli: What about all those 3.14REGEX([01]*) ?
I imagine those can be changed to simply 3.14 now.

test/functionalTest/cases/2176_not_print_spurious_decimals/not_print_spurious_decimals.test
```diff
@@ -0,0 +1,138 @@
...
+
+02. Get A value: 1.123456789 (9 decimals)
+=========================================
+HTTP/1.1 200 OK
+Content-Length: 10
+Content-Type: text/plain
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+1.12345679
```
kzangeli: It would have been more interesting to see a float with 'rounding issue' in the ninth decimal, e.g.:

    1.123456785551

Would it be simply cut, to 1.123456785?
Or correctly rounded, to 1.123456786?

test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_deeper.test
fgalan: Probably, but I cannot check that, as the problem it seems that only happens in your system...

I propose the following: after merging this PR, try to change `3.14REGEX(...)` to just `3.14` and check if it works in your system.

kzangeli: ok

NTC

test/functionalTest/cases/2176_not_print_spurious_decimals/not_print_spurious_decimals.test
fgalan: Actually, I'm realizing that using PRECISION_DIGITS to 9 we round to 8 decimals. I'll change PRECISSING_DIGITS to 10 and readjust the tests.

fgalan: Fixed in acdfad7ae4e9ea4e39efd55ffc61359e17aad454.

Have a look. I think it works as expected.

kzangeli: Great.
However ... strange about 9/10 ...
Do we understand this?

My doubt is this:
PRECISION_DIGITS is only about the number of decimals, right?
I mean, there is a '1' on the other side of the decimal point.
Have you tried  123456.12345678551.
Hope it doesn't give only 3 decimals ...

fgalan: I'm almost sure that PRECISION_DIGITS is only about the decimal part, but let me modify the test to be sure...

(Probably "precision n" means that numbers will use n-1 decimal digits from a mathematical point of view... but I don't remember my match classess good enough ;)

fgalan: Contrary to intuition, default behaviour is that `std::setprecision()` digits include both decimal and integer parts. This can be changed using, `std::fixed`. With that modifier, `std::setprecision()` sets the decimal part length... but the side-effect of using `std::fixed` is that float without decimal part are printed with the `.0`.

I'll try to fix this specializing the template for float and checking if the float have decimal part or not before applying modifiers.

fgalan: Test adjusted in e9162f5, along with several changes in the code to avoid the aforementioned problems.

src/lib/common/JsonHelper.cpp
```diff
@@ -161,9 +162,6 @@ std::string objectToJson(std::map<std::string, std::string>& list)
 */
 JsonHelper::JsonHelper(): empty(true)
 {
-  /* Set format  for floats (it doesn't affect integers) */
```
fgalan: This has been moved to the particular function dealing with float printing, i.e. `addFloat()`.

src/lib/common/string.cpp
```diff
@@ -791,6 +792,68 @@ bool str2double(const char* s, double* dP)
...
+  double decimalPart = fabs(modf(d, &intPart));
+
+  while (decimalPart > 0)
+  {
+    digits++;
+    decimalPart *= 10;
+    decimalPart = modf(decimalPart, &intPart);
+    if (fabs(1 - decimalPart ) < 0.01)
+    {
+      // In the case of 3.139999..., after step 2 modf() will produce 0.999..., so
```
fgalan: 0.01 is due to each step is using 1 decimal digit (the `*10` above) thus I'm using the double of that step. However, not sure if 0.1 should be better...

Any opinion on this?

test/functionalTest/cases/2176_not_print_spurious_decimals/not_print_spurious_decimals.test
```diff
@@ -0,0 +1,138 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create entity with attributes A: 442.123456785551 and B: 200
+# 02. Get A value: 42.123456789 (9 decimals)
```
kzangeli: typo: 42 => 442

src/lib/common/string.h
```diff
@@ -159,12 +162,25 @@ extern bool str2double(const char* s, double* dP = NULL);
...
+
+/* ****************************************************************************
+*
 * toString -
 *
 * If the generic ostringstream-based implementation would have performance
 * problems in the future, a set of per-type specialized functions could be
 * used without changing the toString() usage interface from existing callers
 *
+* In fact, we currently have an specizlized function for float, although not
```
kzangeli: typo: specialized

test/functionalTest/cases/2176_not_print_spurious_decimals/not_print_spurious_decimals.test
fgalan: Fixed in 21d1890

src/lib/common/JsonHelper.cpp
fgalan: NTC (just informative)

src/lib/common/string.h
fgalan: Fixed in 21d1890

src/lib/common/string.cpp
kzangeli: No opinion, not without empirical tests ....
floats is not an exact science ...

I just find it a little sad having to call two functions in the loop to get the number of decimals.
However, these two functions are probably implemented as macros anyways

NTC

fgalan: Still testing... probably 0.01 and 0.1 are both wrong and this should be PRECISION

kzangeli:
LGTM

src/lib/common/string.cpp
fgalan: Another round of modifications along with a .test in commit 073bfc9

crbrox:
LGTM

Issue #2406, opened 2016-07-29
in POST v2/subscriptions, with "geometry" condition expression and without "coords" field


iariasleon:
#### Request
```
url: POST http://localhost:1026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: test_condition_expression_geometry_error
    Fiware-ServicePath: /test
payload:{"notification": {"http": {"url": "http://localhost:1234"}, "attrs": ["temperature"]}, "expires": "2016-04-05T14:00:00.00Z", "subject": {"entities": [{"idPattern": ".*", "type": "room"}], "condition": {"expression": {"georel": "disjoint"}, "attrs": ["temperature"]}}}
```
#### Response 
```
http code: 201 - Created
headers:
   date: Fri, 29 Jul 2016 12:54:45 GMT
   fiware-correlator: a0058e58-558b-11e6-a249-005056a20feb
   connection: Keep-Alive
   content-length: 0
   location: /v2/subscriptions/579b5215c80eb79a267b4aa5
payload:
```
#### Expected
```
http code: 400 - Bad Request
payload: {"error":"BadRequest","description":"error parsing geo-query fields: error parsing georel: geometry does not exist"}
```
#### Request
```
url: POST http://localhost:1026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: test_condition_expression_geometry_error
    Fiware-ServicePath: /test
payload: {"notification": {"http": {"url": "http://localhost:1234"}, "attrs": ["temperature"]}, "expires": "2016-04-05T14:00:00.00Z", "subject": {"entities": [{"idPattern": ".*", "type": "room"}], "condition": {"expression": {"geometry": "point"}, "attrs": ["temperature"]}}}
```
#### Response 
```
http code: 201 - Created
headers:
   date: Fri, 29 Jul 2016 12:54:45 GMT
   fiware-correlator: a0058e58-558b-11e6-a249-005056a20feb
   connection: Keep-Alive
   content-length: 0
   location: /v2/subscriptions/579b5215c80eb79a267b4aa5
payload:
```
#### Expected
```
http code: 400 - Bad Request
payload: {"error":"BadRequest","description":"error parsing geo-query fields: error parsing geometry: coords does not exist"}
```
#### Request
```
url: POST http://localhost:1026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: test_condition_expression_geometry_error
    Fiware-ServicePath: /test
[...]

PR #2412, opened 2016-08-01
[FIX] issue 2406 and http code

iariasleon:
review tests to issue #2406 and changed a http code reason "Method Not Allowed"


fgalan:
LGTM

PR #2423, opened 2016-08-03
Feature/1156 mq in compound value for queries

kzangeli:
Implemented mq for compound values in queries.
Part of #1156


fgalan:
LGTM

PR #2433, opened 2016-08-04
Temporary: not deleting compounds in Metadata::release - to avoid crashes

kzangeli:
Simply commented the call to `delete compoundValueP`in Metadata::release().
All tests in 1156_* run OK,

This is very temporary, a real fix is needed ASAP.


fgalan:
LGTM

Issue #1853, opened 2016-02-26
patterns in entity types


kzangeli:
Like entity::id, _entity::type_ could have patterns.
Right now, without patterns for entity::type, a request gets either **one** entity or **all of them** (either for a specific _id_, or a patterned id).

Type patterns would be treated just like **id**, using either **type** or **typePattern** in the payload, and adding 'typePattern' as a URI param.

fgalan:
Changes in NGSIv2 WIP .apib implemented in PR #2336

fgalan:
Completed at NGSIv2 spec. Changing milestone to 1.3.0 (version in which this is going to be implemented at Orion).

fgalan:
Some implementation notes follow:

Looking to the changes in the spec regarding this functionality (PR https://github.com/telefonicaid/fiware-orion/pull/2336) three operations are affected:

1. GET /v2/entities, where `typePattern` may appear as URL param.
2. POST /v2/op/query, where `typePattern` may appear in the payload
3. POST /v2/subscriptions or PATCH /v2/subscription/subId (most of the payload is the same), where `typePattern` may apper in the payload.

(The .spec also describes typePattern in the context of register context operation. However, given that the whole NGSI9 is not yet part of NGSIv2, we can discard that and it will not be considered part of the scope of the present issue).

Let's consider implication of the above three cases.

Regarding case 1 and 2, both are variants of query operations that at the end will end in the use of mongoQueryContext(), which uses a QueryContextRequest class as input argument. Thus, we need to extend that QueryContextRequest class (probably not QueryContextRequest class itself, but some of its contained clasess... I don't remember right now the details) to "map" the type pattern nature of a type. I mean, to implement a `isTypePattern` as "parallell" of `isPattern`.

*It important to note that typePattern is a functionality only in the NGSIv2 API. In other words, we are extending NGSIv1-biased classes to reuse mongoBackend, but not for exposing the functionality in the NGSIv1 API, i.e. no modifications to NGSIv1 parser.*

Moreover, internally the mongoBackend has to be modified to take into account that `isTypePattern` and adjust the query to MongoDB accordingly. This means changes in entitiesQuery() function, probably in the fillQueryEntity() which is called as first step.

Regarding case 3, it has two aspects: the basic create/update/retrieve subscription logic (3a) and the interpretation of the typePattern which is part of the subscription during an update context (in order to know if the update will trigger the notification associated to that subscription or not) (3b).

* (3a) Basically:
[...]

PR #2438, opened 2016-08-04
typePattern mongoBackend parts

fgalan:
This PR aims to complete (or at least help to complete :) the mongoBackend parts involved in the typePattern functionality (issue #1853).

Unfortunatelly, the PR has "polluted" with some fixing in unit test. I have to recover a previously "hidden" module to add 2 new test cases and doing so, I have to implement many fixes in that module not actually related with the PR. Basically, you can ignore the most of unit_test code, except the part that I'll mark as comments in the PR.

@crbrox 
@kzangeli


src/lib/cache/subCache.cpp
```diff
@@ -211,6 +211,20 @@ void subCacheInit(bool multitenant)
 
 /* ****************************************************************************
 *
+* subCacheDisable -
+*
+*/
+#ifdef UNIT_TEST
+void subCacheDisable(void)
```
fgalan: I have need to add this function to avoid subCacheActive to be `true` during no-cache unit test. Maybe not very clean, but the UNIT_TEST define will ensure that this doesn't "leak" into contextBroker binary.

NTC (just informative)

src/lib/ngsi/EntityId.cpp
```diff
@@ -38,7 +38,7 @@
 *
 * EntityId::EntityId -
 */
-EntityId::EntityId() : keyName("entityId")
+EntityId::EntityId()
```
fgalan: Not actually related with this PR but I found that `keyName` was no longer used when modifying class fields, so I have take the opportunity to remove it.

NTC (just informative)

test/unittests/CMakeLists.txt
```diff
@@ -185,6 +185,7 @@ SET (SOURCES
     mongoBackend/mongoUpdateContext_test.cpp
     mongoBackend/mongoUpdateContextGeo_test.cpp
     mongoBackend/mongoUpdateContext_withOnchangeSubscriptions_test.cpp
+    mongoBackend/mongoUpdateContext_withOnchangeSubscriptionsNoCache_test.cpp
```
fgalan: This little line is the one that has causes so many noisy in this PR :)

NTC (just informative)

test/unittests/mongoBackend/mongoQueryContext_test.cpp
```diff
@@ -3741,6 +3745,91 @@ TEST(mongoQueryContextRequest, queryCustomMetadataPattern)
...
+
+/* ****************************************************************************
+*
+* queryTypePattern -
+*
+* Query:     id pattern = E.* - type Pattern = T[1-2]$ - no attrs
+* Result:    E1 (all attributes) and E2 (all attributes)
+*
+*/
+TEST(mongoQueryContextRequest, queryTypePattern)
```
fgalan: This test worth the time to be reviewed.

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptionsNoCache_test.cpp
```diff
@@ -473,9 +512,130 @@ TEST(mongoUpdateContext_withOnchangeSubscriptionsNoCache, Cond1_appendMatch)
...
+    utExit();
+}
+
+
+
+/* ****************************************************************************
+*
+* Cond1_appendMatch_typePattern -
+*/
+TEST(mongoUpdateContext_withOnchangeSubscriptionsNoCache, Cond1_appendMatch_typePattern)
```
fgalan: This test and the next one worth the time to be reviewed.

Note that two new subscriptions (sub10 and sub11) has been populated in database to make these test to work (see prepareDatabase() function).

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1313,7 +1396,7 @@ static bool addTriggeredSubscriptions_noCache
 
       long long           throttling         = sub.hasField(CSUB_THROTTLING)?       getIntOrLongFieldAsLongF(sub, CSUB_THROTTLING)       : -1;
       long long           lastNotification   = sub.hasField(CSUB_LASTNOTIFICATION)? getIntOrLongFieldAsLongF(sub, CSUB_LASTNOTIFICATION) : -1;
-      std::string         renderFormatString = sub.hasField(CSUB_FORMAT)? getStringFieldF(sub, CSUB_FORMAT) : "";
+      std::string         renderFormatString = sub.hasField(CSUB_FORMAT)? getFieldF(sub, CSUB_FORMAT).String() : "legacy";  // NGSIv1 JSON is 'default' (for old db-content)
```
fgalan: I discover during the fixing of the *noCache test that the logic to retrieve render string may have if the doc doens't have a `format` field. Now, it works as in the csub cache (in fact, this line has been copyed from it).

Not sure if this should be mentioned in CNR. Probably not, as no-`format` would mean "XML". Probably it is only a problem for the existing unittest but not something quite important.

Opinions are welcome :)

src/lib/cache/subCache.cpp
kzangeli: Why not put this function in unittest.cpp (or some other test file) ?
subCacheActive is a global variable, not static ...

src/lib/ngsi/EntityId.cpp
kzangeli: Yeah, keyName was probably used for rendering of XML.

NTC

src/lib/ngsi/EntityId.cpp
```diff
@@ -312,6 +306,11 @@ bool EntityId::equal(EntityId* eP)
     return true;
   }
 
+  if (eP->isTypePattern == eP->isTypePattern)
+  {
+    return true;
+  }
+
   return false;
```
kzangeli: This I don't like very much ...

Before

    if (eP->isTypePattern == eP->isTypePattern)

there is a 

      if (eP->isPatternIsTrue() == isPatternIsTrue())
      {
        return true;
      }

That was OK, as long as it was *the last if*.
Now it is not, so, if `eP->isPatternIsTrue() != isPatternIsTrue()`, the `true` will not be returned and we continue to `if  eP->isTypePattern == eP->isTypePattern`, which will return `true` in case of equality.
However, isPattern differs and `true` must *not* be returned.
The function must be rewritten like this:

    bool EntityId::equal(EntityId* eP)
    {
      if (eP->id != id)
      {
        return false;
      }

      if (eP->type != type)
      {
        return false;
      }

      if (eP->isPatternIsTrue() != isPatternIsTrue())
      {
        return false;
      }
    
      if (eP->isTypePattern != eP->isTypePattern)
      {
        return false;
      }
    
    ￼  return true;
    }

src/lib/ngsi/EntityId.h
```diff
@@ -39,22 +39,21 @@
...
   std::string  servicePath;  // Not part of payload, just an internal field
-  std::string  keyName;      // Help variable for the 'render' method
 
   EntityId();
   EntityId(EntityId* eP);
   EntityId(const std::string&  _id,
            const std::string&  _type,
            const std::string&  _isPattern = "",
-           const std::string&  _keyName = "entityId");
+           bool                 isTypePatter = false);
```
kzangeli: isTypePatter => isTypePattern

Time to compile ... :-D

src/lib/ngsi/EntityId.cpp
fgalan: Yes, seems to be a clearer implementation.

(As discussed by skype, to be fixed by Ken)

src/lib/cache/subCache.cpp
fgalan: Agree.

(As discussed by skype, to be fixed by Ken)

src/lib/ngsi/EntityId.h
fgalan: Agree.

(As discussed by skype, to be fixed by Ken)

src/lib/ngsi/EntityId.cpp
crbrox: fixed 4283812

src/lib/ngsi/EntityId.h
crbrox: Fixed by ba39f52

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: No opinions in two weeks. NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1198,68 +1379,70 @@ static bool addTriggeredSubscriptions_noCache
...
+   * FIXME: condTypeQ, condValueQ and servicePath part could be "factorized" out of the $or clause
+   */
 
-  BSONObjBuilder  queryPattern;
+  //
+  // Allocating buffer to hold all these BIG variables, necessary for the population of 
+  // the four parts of the final query.
+  // The necessary variables are too big for the stack and thus moved to the head, inside BsonGroup.
+  //
+  BsonGroup* bgP = new BsonGroup();
```
fgalan: Closely review the lifecyle of `bgP` taking into account the different exit point of this function. Maybe this has broken after last merge.

kzangeli: I looked at it, in the entire file. With the diffs it's not possible to see anything.

It looks good to me.

crbrox:
LGTM

src/lib/cache/subCache.cpp
kzangeli: Change not really needed.

NTC

kzangeli:
LGTM

test/unittests/mongoBackend/mongoQueryContext_test.cpp
fgalan: NTC (just informative)

test/unittests/mongoBackend/mongoUpdateContext_withOnchangeSubscriptionsNoCache_test.cpp
fgalan: NTC (just informative)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Ok, let's go.

NTC

fgalan:
LGTM

Issue #2457, opened 2016-08-22
Compounds with same tag


kzangeli:
We have a check in method `CompoundValueNode::check()` (in file `parse/CompoundValueNode.cpp`) that checks that two fields in the same level haven't got the same name, e.g.:

    "A1": {
      "value": {
        "temp": 14,
        "temp": 15
      }

That is a thing we want to avoid.
It seems this check doesn't do exactly what it should. This input:

        ...
        "attributes": [
        {
          "name": "configData",
          "type": "compound",
          "value": {
            "acquisition_time": {
              "acquisition_time": "7"
            },
            "adjust_sensor_curves": {
              "coef_a": "8",
              "coef_b": "13",
              "coef_c": "1085"
            }, ...

Provokes this error:

    {
      "errorCode" : {
        "code" : "400",
        "reasonPhrase" : "Bad Request",
        "details" : "compound value error: duplicated tag-name: /acquisition_time/ in path: /acquisition_time"
      }
    }

Now, in the input payload, `acquisition_time` is a child of `acquisition_time` and that should not cause a problem. This is not what the `check()` method should complain about ...
This seems like a bug.
The current implementation looks like this:

    for (uint64_t ix = 0; ix < childV.size() - 1; ++ix)
    {
      for (uint64_t ix2 = ix + 1; ix2 < childV.size(); ++ix2)
      {
        if (childV[ix]->name == childV[ix2]->name)
        {
          rootP->error = std::string("duplicated tag-name: /") + childV[ix]->name + "/ in path: " + path;
          LM_W(("Bad Input (%s)", rootP->error.c_str()));

          return;
        }
      }
    }

Note the `childV[ix]` vs `childV[ix2]` ...

For a quick dirty fix needed asap, I will simply remove the check of 'siblings with same tag'.
However, this is not good enough, we need to look into this and modify the implementation to do what should be done, compare names on the same level but not between levels.

The fix is only for branch 0.26.1

**Effort: 2 man day**

There is a unit test checking this 'feature', in XXX, that will have to be disabled again as the check has been removed.

fgalan:
In sum, there are two situations:

1. `{ value: { a: ..., a: ... }´
2. `{ value: {a: { a: ... } }`

After PR #2458, 0.26.1 branch leave pass both cases.

In develop, both cases are currently detected as wrong input. This is not correct. What is correct is:

1. To detect as error situation 1 (this is currently ok in develop)
2. To leave pass situation 2

[...]

PR #2458, opened 2016-08-22
Bug/compounds with same tag in different levels

kzangeli:
Emergency fix for Repsol.
See issue #2457.
This is **not a fix**, a check has simply been removed as it seems it doesn't work the way it should.
A **real fix** for this issue is needed later on, asap ...

The check is meant to make sure there are not more than one tag with the same name, in the same block, see #2457, however, as it seems, the check flags errors for tags with the same name but in different levels/blocks. This is not what the check was meant to do.

A few unit tests, checking the feature that has been removed, has been disabled and some functests also.
Especfially two functests, testing subscription creation in NGSIv2 (something that this version of the broker doesn't even implement!) have been completely disabled.

Another detail, an old, deprecated, call to MHD has been replaced for the recommended call, which is what the broker uses in newer versions (it didn't compile, as we have warnings set to produce errors)


crbrox:
LGTM

manucarrace:
LGTM

Issue #2419, opened 2016-08-02
in PUT admin/log with FATAL and WARN levels the value changed is not the expected


iariasleon:
in PUT admin/log with FATAL and WARN levels the value changed is not the expected

**FATAL** is changed to **NONE**
**WARN** is changed to **ERROR**

#### PUT Request
```
 PUT http://localhost:1026/admin/log?level=FATAL
```

#### Response
```
http code: 200 - OK
headers:
   date: Tue, 02 Aug 2016 12:06:53 GMT
   fiware-correlator: 9a071a1e-58a9-11e6-b2f1-005056a20feb
   connection: Keep-Alive
   content-length: 0
payload:
```

#### GET Request
```
 GET http://localhost:1026/admin/log
```

#### Response
```
http code: 200 - OK
headers:
   date: Tue, 02 Aug 2016 12:06:53 GMT
   fiware-correlator: 9a11d972-58a9-11e6-88bd-005056a20feb
   connection: Keep-Alive
   content-type: application/json
   content-length: 16
payload: {"level":"NONE"}
```

Also, Always is logged this line (both at successful and unsuccessful cases):
```
time=2016-08-02T12:00:26.990Z | lvl=DEBUG | corr=b374a6d4-58a8-11e6-8008-005056a20feb | trans=1470139009-514-00000000004 | from=127.0.0.1 | srv=pending | subsrv=pending | comp=Orion | op=rest.cpp[212]:uriArgumentGet | msg=Received unrecognized URI parameter: 'level'

```

iariasleon:
CB version used:
```
  "version" : "1.2.0-next",
  "git_hash" : "2952223adca30f1ff71f15dbc7e44c3c1fb1137f"
```

iariasleon:
in CB version:
```
  "version" : "1.2.0-next",
  "git_hash" : "59c5c77580ad764684d2e650cc28d916dea41122",
```
Now:
* the `FATAL` level is returned as `FATAL`
* the `WARN` level is returned as `WARN`

But:
* the `NONE` level is returned as `FATAL`

#### PUT request
```
url: PUT http://localhost:1026/admin/log?level=NONE
```

#### GET request
```
url: GET http://localhost:1026/admin/log
```
#### GET response
```
http code: 200 - OK
headers:
   date: Tue, 23 Aug 2016 10:05:08 GMT
   fiware-correlator: 127b4694-6919-11e6-ab20-005056a20feb
   connection: Keep-Alive
[...]

PR #2460, opened 2016-08-23
Bug/2419 log level

kzangeli:
Fixed the response for modifying log levels, issue #2419.
Everything worked the way it should internally, but the broker didn't respond the way it should.
Also, removed the annoying trace line about 'URI param not recognized' for logLevel requests.


crbrox:
LGTM

Issue #1887, opened 2016-03-09
[BUG] in GET "/v2/entities/<id>/attrs/<attr_name>/value" with error in URI the error returned is not the expected


iariasleon:
in GET `/v2/entities/<id>/attrs/<attr_name>/value` with error in URI the error description returned is not the expected.

#### Get request
```
GET http://localhost:1026/v2/entities/house_//attrs/temperature_0/value
Fiware-Service: test_id_entity_invalid
Fiware-ServicePath: /test
Accept: text/plain
```

#### Get response
```
http code: 400
date: Wed, 09 Mar 2016 09:26:49 GMT
connection: Keep-Alive
content-type: text/plain
content-length: 59
{"error":"BadRequest","description":"unrecognized request"}
```

#### Expected response
```
http code: 400
{"error":"BadRequest","description":"service not found"}
```

kzangeli:
Not sure this is a bug ...
The // is the URL changes completely the path and the request is simply not recognized.
Let's discuss it, but in my opinion, the broker responds the way it should.

Unless of course we want to remove completely the *unrecognized request* response and replace it for a *service not found* ...

fgalan:
The bug is that some casese description is  "service not found" and some other "unrecognized request". This should be unfied.

Which message use other well-known public REST APIs for the same case? Let's try to align with the community.

iariasleon:
The same behaviour in error description with an attribute name as `house_/`, ex:
```
GET http://localhost:1026/v2/entities/room_1/attrs/house_//value
Fiware-Service: test_update_attr_name
Fiware-ServicePath: /test
Accept: text/plain
```
```
http code: 400
date: Wed, 09 Mar 2016 14:53:31 GMT
connection: Keep-Alive
content-type: text/plain
content-length: 59
{"error":"BadRequest","description":"unrecognized request"}
```

fgalan:
In the last case, maybe there is an "inverleaved" bug... the response payload (due to an error) in JSON shouldn't use text/plain as Content-Type... maybe this already raised in a separated issue? (in that case, please specify number).

iariasleon:
I think you refer to https://github.com/telefonicaid/fiware-orion/issues/1886, but both issues are different themes.
This issue is if we have a wrong URI with entity_id or attribute_name as `house_/` and the #1886 issue is that any value in Accept header different to `application/json` is assumed as `text/plain`.

fgalan:
[...]

PR #2466, opened 2016-08-24
Functest for hash in URI as entityId or attribute name

kzangeli:
Functest for the Issue #1887 (part about house_# as EntityId or attrName).
No source code has been changed (except for a few empty lines for readability). It seems to be working.

Note that for the test cases with text/plain, the broker returns JSON.
This is something I believe we have discussed in the past and we've said "Good enough, after all, JSON is plain text ..."
Could be better, but it might be good enough.

Still unable to reproduce the strange error response that is described in issue #1887.

### DIFF
--- a/test/functionalTest/cases/1887_hash_in_url_gives_strange_results/hash_in_url_gives_strange_results.test
+++ b/test/functionalTest/cases/1887_hash_in_url_gives_strange_results/hash_in_url_gives_strange_results.test
@@ -32,6 +32,8 @@ brokerStart CB
 #
 # 01. GET /v2/entities/E1_#/attrs/A1/value
 # 02. GET /v2/entities/E1/attrs/A1_#/value
+# 03. GET /v2/entities/E1_#/attrs/A1/value, Accept text/plain
+# 04. GET /v2/entities/E1/attrs/A1_#/value, Accept text/plain
 #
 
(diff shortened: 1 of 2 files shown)


test/functionalTest/cases/1887_hash_in_url_gives_strange_results/hash_in_url_gives_strange_results.test
```diff
@@ -0,0 +1,120 @@
...
+
+03. GET /v2/entities/E1_#/attrs/A1/value, Accept text/plain
+===========================================================
+HTTP/1.1 400 Bad Request
+Content-Length: 63
+Content-Type: text/plain
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+{"error":"BadRequest","description":"invalid character in URI"}
```
kzangeli: When orionCurl receives a response for a request sent with Accept: text/plain, orionCurl assumes the response is normal text and does **not** try to beautify the response with any JSON beautifier.

This is the reason why the response is shown 'raw' here and in the next step.

[ About getting JSON back when asking for plain text ... see the description of the PR ]

NTC, just informative

crbrox:
LGTM

PR #2482, opened 2016-08-30
Feature/1853 type pattern subscriptions

crbrox:
@fgalan @kzangeli 

Partially, #1853 

(missing  parts of 1) and 2))


src/lib/cache/subCache.cpp
```diff
@@ -99,6 +100,26 @@ entityId(_entityId), entityType(_entityType)
   {
     entityIdPatternToBeFreed = false;
   }
+
+  if (isTypePattern)
+  {
+    // FIXME P5: recomp error should be captured? have a look to other usages of regcomp()
```
fgalan: Is this FIXME correct? As far as I understand, the if and `alarmMgr.badInput(clientIp, "invalid regular expression for typePattern")` is a way of implementing such regcomp error capture.

crbrox: not sure, copy&pasted for code for entity id

fgalan: Not very important anyway... NTC

fgalan:
LGTM

src/lib/cache/subCache.cpp
```diff
@@ -99,6 +100,26 @@ entityId(_entityId), entityType(_entityType)
...
+    // in order to see how it works
+    if (regcomp(&entityTypePattern, _entityType.c_str(), REG_EXTENDED) != 0)
+    {
+      alarmMgr.badInput(clientIp, "invalid regular expression for typePattern");
+      isTypePattern = false;  // FIXME P6: this entity should not be let into the system. Must be stopped before.
+                          //           Right here, best thing to do is simply to say it is not a regex
```
kzangeli: As entityTypePatternToBeFreed seems to be set everywhere, it should probably be set here too (to FALSE).
However, it would be better to set it to FALSE in the constructor and only set to to TRUE where needed.

src/lib/cache/subCache.cpp
```diff
@@ -113,22 +134,44 @@ bool EntityInfo::match
...
+    matchedId =  (regexec(&entityIdPattern, id.c_str(), 0, NULL, 0) == 0);
+  }
+  else if (id == entityId)
+  {
+    matchedId =  true;
+  }
+  else
+  {
+    matchedId = false;
+  }
```
kzangeli: This else (setting matchedId to FALSE), is it needed?
matchedId is set to FALSE upon creation.

src/lib/cache/subCache.cpp
```diff
@@ -113,22 +134,44 @@ bool EntityInfo::match
...
+  {
+    // Check type
+    if (isTypePattern)
+    {
+      // REGEX-comparison this->entityTypePattern VS type
+      matchedType = (regexec(&entityTypePattern, type.c_str(), 0, NULL, 0) == 0);
+    }
+    else if ((type != "")  && (entityType != "") && (entityType != type))
+    {
+      matchedType = false;
```
kzangeli: Similar situation - is "matchedType = false" needed?

src/lib/cache/subCache.cpp
crbrox: On purpose, for clarity and ease of reading ...

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -360,25 +367,53 @@ static std::string parseEntitiesVector(ConnectionInfo* ciP, std::vector<EntID>*
...
-      if (typeOpt.value.length() > MAX_ID_LEN)
+      else if (typePatOpt.given)
       {
-        return badInput(ciP, "max type length exceeded");
+        if (typePatOpt.value.empty())
+        {
+          return badInput(ciP, "subject entities element typePattern is empty");
+        }
+
+        typePattern =typePatOpt.value;
```
kzangeli: Space missing after `=`

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -202,33 +202,43 @@ void setEntities(const Subscription& sub, BSONObjBuilder* b)
   BSONArrayBuilder entities;
   for (unsigned int ix = 0; ix < sub.subject.entities.size(); ++ix)
   {
-    EntID en = sub.subject.entities[ix];
+    EntID       en        = sub.subject.entities[ix];
+    std::string finalId;
+    std::string finalType;
+    std::string isIdPattern;
+    bool        isTypePattern = false;
```
kzangeli: This here is a little funny ...
isTypePattern is bool, but isIdPattern is a string?
If this is changable, bool is the better choice.

fgalan: @crbrox will tell, but I guess that this is not easy to change... Remember that due to regacy reasons the id pattern is a string in many internal places.

What could probably be done at least is to add a comment explainin this "justified weridness".

crbrox: Very funny, i was caught for it and was the cause of the last tests failing ...
Actually it is in that way in the model description https://fiware-orion.readthedocs.io/en/develop/admin/database_model/index.html#csubs-collection
Maybe for another refactor PR ? (the impact could be big)

crbrox: Comment added 28d4ecc

kzangeli: Yeah, I wasn't considering to change the format of isPattern in mongo or anything like that, just to have the local vars of the same type.

All is better with the comment

NFTM (nothing further to modify :-))

src/lib/cache/subCache.cpp
kzangeli: ok, matter of opinion.

NTC

src/lib/cache/subCache.cpp
kzangeli: matter of opinion.

NTC

src/lib/mongoBackend/MongoCommonSubscription.cpp
kzangeli: I took a look at the entire file (without diff and stuff) and understood.

These local vars are used later, like this:
entities.append(BSON(CSUB_ENTITY_ID << finalId << CSUB_ENTITY_ISPATTERN << isIdPattern));

So no, definitely no easy way to change the type of isIdPattern

src/lib/cache/subCache.cpp
crbrox: firxed 28e74d5

src/lib/jsonParseV2/parseSubscription.cpp
crbrox: fixed 98e395c

kzangeli:
LGTM

PR #2487, opened 2016-08-31
Hardening/1851 cleanup

kzangeli:
Changed the signature of the "real" subCacheItemInsert to be like the wrapper function and removed the wrapper function.
The case here was that we had a wrapper function for a function that was used **only** by the wrapper function. Didn't make much sense ...


fgalan:
LGTM

PR #2488, opened 2016-08-31
Bug/2457 json tags with same name

kzangeli:
Functest for issue #2457 (the possible problem with tags with the same name)
It seems to work exactly the way it should ...


crbrox:
LGTM

kzangeli:
Replaced by #2559, that merges into **master**

Issue #2425, opened 2016-08-03
[Number precision] Improve precision for decimalDigits() function


fgalan:
As decribed in the decimalDigits() preamble in the source code:

```
* FIXME #2425: this function is not perfect and could be improved. For example,
* considering the following
*
*   "A1":  42.9,
*   "A2":  42.99,
*   "A3":  42.999,
*   "A4":  42.9999,
*   "A5":  42.99999,
*   "A6":  42.999999,
*   "A7":  42.9999999,
*   "A8":  42.99999999,
*   "A9":  42.999999999,
*   "A10": 42.9999999999,,
*
*   "A1":  42.1,
*   "A2":  42.01,
*   "A3":  42.001,
*   "A4":  42.0001,
*   "A5":  42.00001,
*   "A6":  42.000001,
*   "A7":  42.0000001,
*   "A8":  42.00000001,
*   "A9":  42.000000001,
*   "A10": 42.0000000001,
*
* what we get is:
*
*   "A1":  42.9,
*   "A2":  42.99,
*   "A3":  42.999,
*   "A4":  42.9999,
*   "A5":  42.99999,
*   "A6":  42.999999000, (fail)
*   "A7":  42.999999900, (fail)
*   "A8":  42.999999990, (fail)
*   "A9":  42.999999999,
*   "A10": 43.000000000, (fail, although probably not due to this function but the caller)
*
*   "A1":  42.1,
*   "A2":  42.01,
*   "A3":  42.001,
*   "A4":  42.0001,
*   "A5":  42.00001,
*   "A6":  42.000001000, (fail)
*   "A7":  42.000000100, (fail)
*   "A8":  42.000000010, (fail)
*   "A9":  42,           (fail)
*   "A10": 42,
```

This issue is about to solve these issue, although it doesn't seem to be an easy task...

The following test case can be very useful to test:

cases/2176_not_print_spurious_decimals/one_to_nine_decimals.test

fgalan:
Test 2207_not_spurious_decimals_in_custom_notifications/not_spurious_decimals_in_custom_notifications.test was found problematic at jenkins.

Disabled in PR https://github.com/telefonicaid/fiware-orion/pull/2491, to be recoved when this issue gets completed.

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/3229 (in 1.15.0) but ftest added later in PR #3294.

PR #2491, opened 2016-09-01
Disabling not_spurious_decimals_in_custom_notifications.test

fgalan:
To avoid jenkins compalining this all the time. It isn't a major issue. It would be fixed as part of #2425


iariasleon:
LGTM

Issue #2445, opened 2016-08-11
in POST v2/subscriptions, if we use  "mq" query param to validate a "DateTime" some operators do not notify


iariasleon:
in POST v2/subscriptions, if we use  "mq" query param to validate a "DateTime" some operators do not notify (`==`, `<` and `>`). Anothers operators (`>=`, `<=` and `!=`) notify correctly.

#### dataset
```
      | mq_expression                                                              |
      |----------------------------------------------------------------------------|
      | timestamp.meta_timestamp==2017-06-15T07:21:24.00Z                          |
      | timestamp.meta_timestamp>2016-06-15T07:21:24.00Z                           |
      | timestamp.meta_timestamp<2026-06-15T07:21:24.00Z                           |
      | timestamp.meta_timestamp==2016-06-15T07:21:24.00Z..2026-06-15T07:21:24.00Z |
```

#### subscription request
```
url: POST http://localhost1026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: s01801pep2cbv2
    Fiware-ServicePath: /ssp1
payload: { "subject": {"entities": [{"idPattern": ".*"}], "condition": { "attrs": ["temperature"], "expression": {  "mq": "timestamp.meta_timestamp==2016-06-15T07:21:24.00Z..2026-06-15T07:21:24.00Z"} }}, "notification": {"http": { "url": "http://localhost:10032/notify"}, "attrs": ["temperature","pressure","humidity","timestamp","value_str","value_bool"]} }
```

#### entity request
```
url: POST http://localhost:10026/v2/entities
headers:
    Content-Type: application/json
    Fiware-Service: s01801pep2cbv2
    Fiware-ServicePath: /ssp1
[...]

Issue #2443, opened 2016-08-11
in GET v2/entities if are using "mq" query param, the date are not processed


iariasleon:
in GET v2/entities if are using "mq" query param, the date are not processed. The "DateTime" type is defined in the metadata.

#### dataset
```
      | mq_expression                                                              |
      |----------------------------------------------------------------------------|
      | timestamp.meta_timestamp==2017-06-15T07:21:24.00Z                          | 
      | timestamp.meta_timestamp>=2017-06-15T07:21:24.00Z                          | 
      | timestamp.meta_timestamp>=2016-06-15T07:21:24.00Z                          | 
      | timestamp.meta_timestamp>2016-06-15T07:21:24.00Z                           | 
      | timestamp.meta_timestamp<=2017-06-15T07:21:24.00Z                          | 
      | timestamp.meta_timestamp<=2026-06-15T07:21:24.00Z                          | 
      | timestamp.meta_timestamp<2026-06-15T07:21:24.00Z                           | 
      | timestamp.meta_timestamp==2016-06-15T07:21:24.00Z..2026-06-15T07:21:24.00Z |
```

#### entity request
```
url: POST http://localhost:1026/v2/entities
headers:
    Content-Type: application/json
    Fiware-Service: s01801pep2cbv2
    Fiware-ServicePath: /ssp1
payload: {"type": "hy6K", "id": "room4",  "temperature":{"type": "celsius", "metadata": {"meta_temperature": {"value": 34, "type": "alarm"}}, "value": 34}, "pressure":{"metadata": {"meta_pressure": {"value": "high"}}, "value": 35}, "humidity":{"type": "porcent", "metadata": {"meta_humidity": {"value": "air,density"}}, "value": 36}, "timestamp":{"type": "nothing", "metadata": {"meta_timestamp": {"value": "2017-06-15T07:21:24.00Z", "type": "DateTime"}}, "value": 37}, "value_str":{"metadata": {"meta_value_str": {"value": "true"}}, "value": 38}, "value_bool":{"metadata": {"meta_value_bool": {"value": true, "type": "alarm"}}, "value": 39}}
```

#### GET request with "mq" query parameter
```
url: GET http://localhost:1026/v2/entities?mq=timestamp.meta_timestamp>=2017-06-15T07:21:24.00Z&options=count
headers:
    Fiware-Service: s01801pep2cbv2
    Fiware-ServicePath: /ssp1
```

#### GET response
```
http code: 200 - OK
headers:
[...]

PR #2502, opened 2016-09-08
Fixes DateTime filters for mq

fgalan:
Implements a fix for #2443 and  #2445

### DIFF
--- a/src/lib/jsonParseV2/parseMetadata.cpp
+++ b/src/lib/jsonParseV2/parseMetadata.cpp
@@ -113,6 +113,21 @@ static std::string parseMetadataObject(const Value& start, Metadata* mdP)
     }
   }
 
+  // Is it a date?
+  if (mdP->type == DATE_TYPE)
+  {
+    mdP->numberValue =  parse8601Time(mdP->stringValue);
+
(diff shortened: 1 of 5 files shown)


iariasleon:
LGTM

src/lib/jsonParseV2/parseMetadata.cpp
```diff
@@ -113,6 +113,21 @@ static std::string parseMetadataObject(const Value& start, Metadata* mdP)
...
+  {
+    mdP->numberValue =  parse8601Time(mdP->stringValue);
+
+    if (mdP->numberValue == -1)
+    {
+      alarmMgr.badInput(clientIp, "date has invalid format");
+      return "date has invalid format";
+    }
+
+    // Probably reseting stringValue is not needed, but let's do it for cleanliness
```
kzangeli: Probably not needed, no.
If stringValue is reset, what about boolValue, etc?

I would remove this line

kzangeli: Sure, but lets try to remember to remove the lines in some future PR
NTC

Issue #2495, opened 2016-09-05
in PATCH v2/subscriptions/<subsc_id> the typePattern is not append


iariasleon:
in PATCH `v2/subscriptions/<subsc_id>` the typePattern is not append

If we have a subscription with this fields:
```
{
  "id": "57cd34b588ba3ba0868f841c",
  "description": "One subscription to rule them all",
  "expires": "2017-04-05T14:00:00.00Z",
  "status": "active",
  "subject": {
    "entities": [
      {
        "id": "Room"
      }
    ],
    "condition": {
      "attrs": [],
      "expression": {
        "q": "temperature>40"
      }
    }
  },
  "notification": {
    "attrs": [  "temperature",  "humidity"  ],
    "attrsFormat": "normalized",
    "http": { "url": "http://localhost:1234"  }
  },
  "throttling": 5
}
```

And try to update a `typePattern` field in this subscription:
```
http://localhost:1026/v2/subscriptions/57cd34b588ba3ba0868f841c
Fiware-Service: test_update_happy_path
Fiware-ServicePath: /test
Content-Type: application/json
{
	"subject": {                                                                        
             "entities": [                                               
                 {        
                     "id": "room2",                                        
                     "typePattern": ".*"
                 }
             ]
	}
}
```

We see that `id` has been updated, but the `typePattern` has not been appended, and Condition fields have been removed:
```
`{
  "id": "57cd34b588ba3ba0868f841c",
  "description": "One subscription to rule them all",
  "expires": "2017-04-05T14:00:00.00Z",
  "status": "active",
  "subject": {
    "entities": [
      {
        "id": "room2"
      }
    ],
    "condition": {
      "attrs": []
    }
  },
  "notification": {
    "attrs": [
      "temperature",
      "humidity"
    ],
    "attrsFormat": "normalized",
    "http": {
      "url": "http://localhost:1234"
    }
  },
  "throttling": 5
}
```

Note: tested with `type` field and it is appended correctly.

fgalan:
Not an actual bug in current develop branch. A .test has been developed to cover the case in PR https://github.com/telefonicaid/fiware-orion/pull/2503

iariasleon:
LGTM
```
[...]

PR #2503, opened 2016-09-09
ADD unit test to cover the typePattern update case in csubs

fgalan:
Funcional test to cover the case described in #2495


iariasleon:
LGTM

PR #2509, opened 2016-09-13
Changed attrName example value  in order to be consistent with the payload.

FiwareULPGC:
We suggest the `attrName` example value should be `address` instead of `temperature` because the payload is a JSON with `PostalAddress` format: 

```
{
  "address": "Ronda de la Comunicacion s/n",
  "zipCode": 28050,
  "city": "Madrid",
  "country": "Spain"
}
```


fgalan:
LGTM

fgalan:
Thanks!

Issue #2073, opened 2016-04-22
Disable attribute type detail in GET /v2/types operation


fgalan:
We have found that the attribute type inspection step done as part of the mongoEntityTypes() operation could make 

```
$ time curl -s -S --header 'Accept: application/json' --header 'Fiware-Service: sc_a' --header 'Fiware-ServicePath: /b' localhost:1026/v2/types > /dev/null

real	0m29.487s
user	0m0.020s
sys	0m0.000s
```

(Subservice has 5873 entities in17 types with many attributes per type. Note that the problem gets worse as the number of attribute in each type increases).

Client should have the posibility of disabling attribute type detail, e.g:

```
GET /v2/types?options=noAttrsType
```

In this case, the `type` array will be `[]` for each attribute (a bit ugly, but the only way of ensuring consistent syntax for the operation, regardless the usage of `noAttrsType`.

We have checked that in the same scenario above just disabling getAttributeTypes() call, we can lower the time from ~30s to 0.5s.

fgalan:
This was hotfixed in release/1.2.0 by PR https://github.com/telefonicaid/fiware-orion/pull/2514. However, the issue remains open while a definitive (and not hardwired) solution comes into develop.

fgalan:
We will use `"types: [ ]"` at the end (versus `"types: [ "" ]"` as in the hotfix for release/1.2.0).

The name of the option will be:

```
opttions=noAttrsDetail
```

fgalan:
The definitive fix has been implemented in the following PRs (all them mirroring each other, except for the CNR file):

- Main branch (master). In to steps: #2555 (develop) and #2568 (sync develop into master)
- release/1.4.0: #2569

PR #2514, opened 2016-09-14
Hotfix: remove attribute detail in get types operations (only NGSIv2 API)

fgalan:
Hardwired solution for #2073 (only in release/1.2.0 branch)


manucarrace:
LGTM

Issue #2507, opened 2016-09-12
Mark changing attributes in notifications


fgalan:
NGISv2 subscriptions must support a new optional field `markChanges` (tentative name) within the `notification` field. It will allow to values: boolean `true` or `false`. For those subscriptions with the field to `true`, CB will insert the following metadata in the attribute changed by the update operation triggering the notification

* name: "hasChanged"
* type: "Boolean"
* value: true

The `hasChanged` metadata name will become a reserved keyword for attribute medatada.

Default `markChanges` value will be `false`, thus preserving backward compatibility.

NGSIv1 example notification (legacy):

```
{
    "subscriptionId": "51c0ac9ed714fb3b37d7d5a8",
    "originator": "localhost",
    "contextResponses": [
        {
            "contextElement": {
                "attributes": [
                    {
                        "name": "temperature",
                        "type": "float",
                        "value": "26.5",
                        "metadata": [
                            {
                                "name": "hasChanged",
                                "type": "Boolean",
                                "value": true
                            }
                        ]
                    }
                ],
                "type": "Room",
                "isPattern": "false",
                "id": "Room1"
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}
```

NGSIv2 notification (normalized format):

```
{
    "data": [
        {
            "id": "Room1",
            "temperature": {
                "metadata": {
                    "hasChanged": {
                       "type": "Boolean",
                       "value": true
                    }
                },
                "type": "Float",
                "value": 28.5
            },
            "type": "Room"
        }
    ],
    "subscriptionId": "57458eb60962ef754e7c0998"
}
```

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/2518

fgalan:
Finished. Passing the ball to @iariasleon for final QA verification.

iariasleon:
LGTM

PR #2512, opened 2016-09-13
FIX metadata filter crashes and compound filtering bug

fgalan:
Fixes #2496 and additional (not reported yet) bug.

Issue #1494, opened 2015-11-11
More granular subscriptions modes (create only, update only, delete only and combinations)


fgalan:
NGSIv1 uses ONCHANGE notifications, which triggers each time a new context element (entrity + attribute) is created and/or updated. Thus, it is not granular enough to difference between create and update and it is not able to trigger notifications upon context element deletion.

NGISIv2 should improve this and include the following modes:
- Notify when the context element is created (only)
- Notify when the context element is updated (only)
- Notify when the context element is deleted (only)
- Any combination of them (e.g. created+updated, created+deleted, created+updated+deleted, etc.)

In addition, notifications should include some way of specifying if the context element was created, updated or deleted.

The `condition` field in NGSIv2 subscriptions should be reviewed in order to see how it can be adapted to these behaviours.

(Mabye there are duplicate issues about ONCREATE or ONDELETION semantics...)

tobiasjacobs:
Not sure if we talk about the same thing, but
- In my understanding a context element contains an entity id (+type) and several attributes with values. 
- Context Elements are also not really created/updated/deleted in NGSI, but they are data containers to transmit one or more attribute values of a context entity.

Anyway, I think what you propose is to define different subscription modes in order to
- get a notification whenever an attribute value becomes known
- get a notification whenever the value of an attribute becomes known to have changed
- get a notification whenever an attribute value becomes unknown

fgalan:
A way of specifying to which case (update, create, delete) a given attribute in the notification corresponds (see `actionType` metadata introduced in https://github.com/telefonicaid/fiware-orion/pull/2518). However, the way of setting to which evetns the subscription has to react has to be defined. 

It is a just a matter on how to "code" that configuration in the subscripton payload.

jmcanterafonseca:
We need to distinguish between actions referring to entities and actions referring to attributes.

I guess that if attrs is not defined or empty we will be referring to the whole entity and if not we will be referring to a particular attribute or attributes. 

```
{
  "subject": {
    "actionType": "update" ("delete", "append", "create"), "change"
  }
}
```

The difference between update and change would be that change will imply a change in the attribute value whereas update will imply an update of the element which could result in a attribute value change or not.

For backwards compatibility, no actionType will mean "change". 

What do you think?

fgalan:
Note from https://github.com/telefonicaid/fiware-orion/pull/2518#discussion_r80059510, to take into account in the context of this issue:

[...]

Issue #1298, opened 2015-09-29
JsonHelper usage in all NGSIv2 renders


fgalan:
JsonHelper class was introduced in PR #1284 as a way of simplifying the JSON response rendering (at that time only used in the JSON response for the "get all entities" operations). In addition, the class allows to rendering logic in a common place, which is interesting as it allows to solve problems such the one described in https://github.com/telefonicaid/fiware-orion/issues/1172 in a single point.

This issue is about extending the usage of JsonHelper to the other NGSIv2 types (and adapt the JsonHelper class in the process, if needed). As a result, we should see a reduction of the length and complexity of the current render() methods, as the start to use `addRaw()`, `addString()`.

**Effort: 5 man day**

kzangeli:
Before making this big change, that definitely has its advantages, we need to make sure the broker doesn't loose too much in terms of throughput.

fgalan:
Actually, what we have to check is not only the change in throughput, but the trade off between throughput and code simplification (i.e. small lost in throughput is acceptable if the gain in code simplification is big).

kzangeli:
Yes, that is exactly what I said :-)

fgalan:
After PR #2518, it has been that this could be also applied to NGSIv1 render. However, we should evaluate if it worth the time/effort.

fgalan:
With regards to payload rendering two actions needs to be done.

A first action is related with _unification and simplification_. Due to the organic evolution of the code during years, we have ended with several ways of rendering payloads. We should have ONLY ONE pattern along all code. Having only one pattern will be a big gain in simplification and homogeneity and will allow to focus optimizations (related with the second action described in this comment, see below) in just one point.

The pattern to use should be the one based in the JsonHelper.h API (used at the current code for instance in Subscription::toJson() or Registrations::toJson()). The approach of composing the document to render using “add” methods, then getting the string corresponding to the final composition is very powerful in terms of simplicity of use and frees the developer of dealing directly with start/ending tags, commas and all that delicate and distracting stuff.

The second action is related with _optimization and performance_. JsonHelper implementation (JsonHelper.cpp) is based in an internal buffer, currently implemented using a std::ostringstream object.  However, there are alternatives:

* std::ostringstream (the one currently used)
* rapidjson (inspired in @aarranz's PR https://github.com/telefonicaid/fiware-orion/pull/2960)
* std::string (similar to the way old render() and toJson() code work)
* static char array
[...]

PR #2518, opened 2016-09-14
Metadata flags for notifications

fgalan:
Implements #2507 

@kzangeli @crbrox @jmcanterafonseca @mrutid (the more, the better)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -399,6 +399,16 @@ The following strings must not be used as attribute names:
 
 * Virtual attribute names (see specific section on "Virtual Attributes")
 
+## Metadata names restrictions
+
+The following strings must not be used as metadata names:
+
+* `ngsi:onArrival`, as it has an special use (see "Notification metadata marks" section).
```
kzangeli: onArrival ...  not 100% clear what is meant.
I assume it is onSubscriptionCreation ... ?

fgalan: I have tried to explain at the cited section... https://github.com/telefonicaid/fiware-orion/pull/2518/files#diff-d5ec9d3a1874c68d6328577a86cd5dc0R926.

If still not being clear, then documentation have to be improved. Any suggestion?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,21 @@ Note that if a custom payload is used for the notification (the field `payload`
 corresponding subscription), then a value of `custom` is used for the `Ngsiv2-AttrsFormat` header
 in the notification.
 
+## Notification metadata marks
+
+In the case of using `metadataFlags` to `true` in the subscription, the following metadata
+are automatically inserted in the notification messages:
```
kzangeli: inserted -> included

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,21 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+
+* For these attributes included in the update triggering the notification, no matter if the actually
```
kzangeli: the -> they ?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,21 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+  changed or not, `ngsi:onArrival` with boolean value `true` and type `"Boolean"` is inserted.
+
+* For these attributes included in the update triggering the notification that actually changed their
+  values, `ngsi:onChange` with boolean value `true` and type `"Boolean"` is inserted.
+
+* For these attributes notififed as a consecuence of an initial creation or update of the
+  subscription, `ngsi:onSubscriptionChange` with boolean value `true` and type `"Boolean"`
+  is inserted.
+
```
kzangeli: Just to see that I have understood this correctly.

Notifications due to attribute value update:
  Updated (triggering) attributes will have a metadata called ngsi:onChange set to TRUE
  All other included attributes will have a metadata called ngsi:onArrival

Notifications due to sub creation or update:
  ALL attributes will have a metadata called ngsi:onSubscriptionChange

If this is the case, it seems extremely verbose.
What if we have 1000 attributes?
Add a metadata (50 chars?) for each and every attribute?
 
Why not simply add a field to the Notification itself, informing about dueToAttrValueUpdate/dueToSubscriptionChange - valid for all attributes.
And then, the triggering attributes should have a metadata indicating that they are 'triggering'. The non-triggering attributes need no metadata at all. The absence of the metadata already indicates they're not triggering attributes.

fgalan: > Updated (triggering) attributes will have a metadata called ngsi:onChange set to TRUE

Note that "updated" doesn't necesarily mean "triggering". An attribute could have changed and not be part of the triggering ones.

> All other included attributes will have a metadata called ngsi:onArrival

Not all. Only the ones included in the triggering update request. Note that the notification may include attributes not included in the update.

The .test in this PR show the differences described above.

> Notifications due to sub creation or update:
> ALL attributes will have a metadata called ngsi:onSubscriptionChange
>
> If this is the case, it seems extremely verbose.
> What if we have 1000 attributes?
> Add a metadata (50 chars?) for each and every attribute?
>
> Why not simply add a field to the Notification itself, informing about 
> dueToAttrValueUpdate/dueToSubscriptionChange - valid for all attributes.

Initial (and "re-initial") notifications are a case of low interesent nowdays. Thus, I don't worry to be verbose there. Adding a new field in the notification message have two problems: 1) it cannot be aligned with NGSIv1 (where we don't have freedom to add fields), 2) it would be syntactically "weird". Maybe a HTTP header would be better, e.g. "Ngsiv2-notification-type: initial" or something like that.

> And then, the triggering attributes should have a metadata indicating that they are '
> triggering'. The non-triggering attributes need no metadata at all. The absence of the 
> metadata already indicates they're not triggering attributes.

We could add a "ngsiv1:triggering" metadata. However, this is not part of the scope of #2507. It would be a separate functionaly, not to include in this PR.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1740,6 +1742,32 @@ static bool processOnChangeConditionForUpdateContext
...
+      Metadata* newMdP = new Metadata(NGSI_MD_NOTIF_ONUPDATE, DEFAULT_ATTR_BOOL_TYPE, true);
+      caP->metadataVector.push_back(newMdP);
+    }
+
+    if (caP->onChange)
+    {
+      Metadata* newMdP = new Metadata(NGSI_MD_NOTIF_ONCHANGE, DEFAULT_ATTR_BOOL_TYPE, true);
+      caP->metadataVector.push_back(newMdP);
+    }
+  }
```
kzangeli: I believe the names (onUpdate and onChange) confuses me a little here.
Can both onChange and onUpdate be true at the same time?
I ask this as adding an `else` may make the code easier to read and avoid a possible bug.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2054,6 +2089,13 @@ static void updateAttrInNotifyCer
         caP->type = targetAttr->type;
       }
 
+      /* Metadata flags for notifications */
+      caP->onUpdate = true;
+      if (actualUpdate)
+      {
+        caP->onChange = true;
+      }
+
```
kzangeli: ok, I see here that both flags can be TRUE

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
```diff
@@ -0,0 +1,326 @@
...
+    "metadataFlags": false,
+    "attrs": [ ]
+  }
+}'
+orionCurl --url /v2/subscriptions/$SUB_ID -X PATCH --payload "$payload"
+echo
+echo
+
+
+echo "08. Get subscription and see metadataFlags: "
```
fgalan: You mean remove trailing `: `, right?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3496,6 +3556,9 @@ void processContextElement
         //
         ContextElementResponse* notifyCerP = new ContextElementResponse(ceP, apiVersion == "v2");
 
+        // Set metadata flags
+        setCreateAndUpdateFlags(notifyCerP);
```
kzangeli: The function name would depend on the name of the flags that the function sets.
Right now these flags are called onUpdate and onChange, so:

  setCreateAndUpdateFlags => setChangeAndUpdateFlags

However, change and update have a very similar meaning.
I believe we should try to find better names of the flags and after that rename the function accordingly.

I believe the variables have the following meaning:
  onChange:  the subscription has been modified
  onUpdate:   the value of the attribute has been modified

If this is true, perhaps **subUpdated** and **valueUpdated** are acceptable names?
The function could be called **setNotificationFlags** (or **setUpdatedFlags**).

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: As both variables can be TRUE, perhaps better leave it the way it is?

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
```diff
@@ -0,0 +1,326 @@
...
+                "type": "T"
+            }
+        ],
+        "condition": {
+            "attrs": [ ]
+         }
+    },
+    "notification": {
+        "http": {"url": "http://localhost:1234/notify"},
+        "metadataFlags": true,
```
kzangeli: Or simply `details`, like we use somewhere ...

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
kzangeli: Yeah, and adding the important word `no`.
Simply, the TOC line differs from the line used in the echo, and the TOC line seems more accurate.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: > Yes, probably using "else if" is better thant "if" in line 1759. I'll change that.

Lokking closer, we cannot do an else, as both flags are independent. Thus, I think the current logic is correct or I don't understanding what do you mean by "adding an else may make the code easier to read and avoid a possible bug" :). In the latter case, a code sketch would help to understand.

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
fgalan: "details" is too generic...

The name should be something that give the following idea "setting this to true you would get metadata marks in the notifications associated to this subscription". Which short and compact name you suggest for that?

kzangeli: The problem is that 'metadata marks' doesn't really say anything, it is almost as generic as 'details'. To me, metadataMarks says nothing :-(

To find a short and concise word for this seems impossible.
Perhaps verbose/details with a value?

"verbose": "notificationReason"   ?

[ In the future: "verbose": "notificationReason, XXX" ]

Because it is all about the reason that the attribute was included in the notification, more or less, right?

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yes, as I mentioned after seeing that both vars can be TRUE simultaneously.
NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: Yes, it is documented. That's OK. I just have a hard time understanding the name 'onArrival'. This is about the source code, not the documentation.

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
fgalan: The idea of having a field ("verbose" or "details"... in this case probably "details" fits) is good. It would be extensible if we can define "switches" for new details to be included.

```
"notification": {
   ...
   "details": ["A", "B", "C"]
}
```

or 

```
"notification": {
   ...
   "details": {
      "A": true,
      "B": true
   }
   ...
}
```

However, this proposal only moves the problem :). Now good names for A, B, etc. should be deciced.

A final note: note that this functionality (in the scope of this PR at least) has nothing to do with reasons for sending the notifications: onArrival is for attributes included in the update (not necesarily the triggering ones) and onChange is for attribute not only included in the updat but also meaning an actual change in the value (again, not necesarily the triggering ones).

kzangeli: True, it only moves the difficult naming problem … :-)

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Add: notification metadata marks ngsi:onArrival, ngsi:onChange and ngsi:onSubscriptionChange (#2507)
```
jmcanterafonseca: needs to be updated

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,22 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+* `previousValue`: the value and type of this metadata is the one of the attribute to which it
+  is attached previous to the processing of the request that triggered the notification. If the
+  type/value of `previousValue` is the same type/value than in the attribute in which it has been
+  included, then the attribute has not actually changed its value.
+
+* `actionType`: it is included if the attribute to which it is attached was included in the request
+  that triggered the notification. Its value depend on the request operation type: `update`
+  for updates, `append` for creation and `delete` for deletion. Its type is always `Text`.
+  (NOTE: currently Orion, reference implementation of NGSIv2 API, doesn't implement the append
+  and delete cases).
```
jmcanterafonseca: that's a note for Orion Documentation, not for the spec.

doc/manuals/user/metadata.md
```diff
@@ -10,7 +10,10 @@ You can use any name for your custom metadata except for a few reserved names, u
 for special metadata that are interpreted by Orion:
 
 -   [ID](#metadata-id-for-attributes)
--   [location], which is currently [deprecated](../deprecated.md), but still supported
+-   location, which is currently [deprecated](../deprecated.md), but still supported
+-   ngsi:onArrival, used by the "Notification metadata marks" functionality in NGSIv2
+-   ngsi:onChange, used by the "Notification metadata marks" functionality in NGSIv2
+-   ngsi:onSubscription, used by the "Notification metadata marks" functionality in NGSIv2
```
jmcanterafonseca: this is not needed

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -724,6 +724,12 @@ Let's examine in detail the different elements included in the payload:
     way only to show the enormous flexibility of subscriptions.
 -   You can leave `conditions.attrs` empty to make a notification
     trigger on any entity attribute change (regardless of the name of the attribute).
+-   You can make Orion mark the attributes in the notification that at the same
+    time were part of the triggering update. In addition, Orion can mark the attributes
+    that actually changed as a consequence of that update. These marks are based on
+    metadata, described in [the following piece of documentation](metadata.md). Have
+    a look at the "Notification metadata marks" section in the
```
jmcanterafonseca: not sure if it is a good idea to link to the spec from the manual, nonetheless take into account that this new feature is going to land in latest and not in stable

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: This part has been re-done so probably the comment is obsolete. NTC, review the new version.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: This part has been re-done so probably the comment is obsolete. NTC, review the new version.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: This part has been re-done so probably the comment is obsolete. NTC, review the new version.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: onSubscriptionChange has been referred out of the scope of this PR. NTC.

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
fgalan: New version of the spec proposes using `metadata` for this. Please re-review in the new version on the spec. This thread marked as NTC.

CHANGES_NEXT_RELEASE
fgalan: Fixed in 9a645cd8244a7c4023dc0eaaee610c87c3f351df

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Note moved to the ngsiv2_implementation_notes.md file. Fixed in 9a645cd8244a7c4023dc0eaaee610c87c3f351df.

doc/manuals/user/metadata.md
fgalan: Fixed in 9a645cd8244a7c4023dc0eaaee610c87c3f351df

doc/manuals/user/walkthrough_apiv2.md
fgalan: This has been changed. Now the reference is to the metadata section (another section in the Orion manual).

Fixed in 9a645cd8244a7c4023dc0eaaee610c87c3f351df

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -399,6 +399,16 @@ The following strings must not be used as attribute names:
...
 
+## Metadata names restrictions
+
+The following strings must not be used as metadata names:
+
+* `previousValue`, as it is a reserved metadata (see "Special metadata in notifications" section).
+
+* `actionType`, as it is a reserved metadata (see "Special metadata in notifications" section).
+
+* `*`, as it has an special meaning as "all the custom (user) metadata".
```
jmcanterafonseca: I'm wondering whether '*' is a valid metadata name.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,20 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+`notification` field) in subscription is used to do so.
+
+* `previousValue`: the value and type of this metadata is the one of the attribute to which it
+  is attached previous to the processing of the request that triggered the notification. If the
+  type/value of `previousValue` is the same type/value than in the attribute in which it has been
+  included, then the attribute has not actually changed its value.
+
+* `actionType`: it is included if the attribute to which it is attached was included in the request
+  that triggered the notification. Its value depend on the request operation type: `update`
+  for updates, `append` for creation and `delete` for deletion. Its type is always `Text`.
```
jmcanterafonseca: in the future we would need to align with subscription action types ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,20 @@ Note that if a custom payload is used for the notification (the field `payload`
 corresponding subscription), then a value of `custom` is used for the `Ngsiv2-AttrsFormat` header
 in the notification.
 
+## Special metadata in notifications
+
+The following metadata are not included in notifications by default, the `metadata` field (within
+`notification` field) in subscription is used to do so.
+
```
jmcanterafonseca: The `metadata` field (within the `notification` field) can be used to select a set of special attribute metadata included by notifications and aimed at providing more information about the nature of the event that triggered the notification. These metadata are:

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,20 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+* `previousValue`: the value and type of this metadata is the one of the attribute to which it
```
jmcanterafonseca: The value of this metadata must be the previous value of the associated attribute. The type of this metadata must be the previous type of the associated attribute.

jmcanterafonseca: what should happen if the attribute was not formerly defined. I think in that case this metadata should not appear. But we need to specify it.

fgalan: This only can happen in a actionType = append case, which is out of the scope of this PR. We can defer the decission until #1494 gets implemented. What do you think?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Yes. However, I understand that will be done in the PR covering #1494, not in this one. Thus, I guess this is NTC.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,20 @@ Note that if a custom payload is used for the notification (the field `payload`
...
+  is attached previous to the processing of the request that triggered the notification. If the
+  type/value of `previousValue` is the same type/value than in the attribute in which it has been
+  included, then the attribute has not actually changed its value.
+
```
jmcanterafonseca: the current state of implementations make this impossible because if there is no attribute value change no notification is sent ... but of course we are aiming at changing this in the near future

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Good point. However, as long as `*` is a valid ASCII char and it is not a control character, whitespace, `&`, `?, `/` or `#`, it fits the name requirements.

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: It would be interesting, future wise, to distinguish between update and change action. so my suggestion here is to use now `change' or 'valueChange' to identify the changes currently being notified by the Orion implementation.

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -97,9 +97,19 @@ The particular validations that Orion implements on NGSIv2 subscription payloads
...
 * **status**: optional (must be a valid status keyword)
 
+# actionType metadata
+
+From NGSIv2 specification section "Special metadata in notifications", regarding `actionType` metadata:
+
+> Its value depend on the request operation type: `update` for updates,
+> `append` for creation and `delete` for deletion. Its type is always `Text`.
+
+Current Orion implementation only supports the "update" case. Other cases will be
```
jmcanterafonseca: see comment above

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: After a careful thinking at notification level it makes no sense to be so precise because the programmer can always check whether there has been a real change by comparing current and prev value so, NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: good idea

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: this is not true as there can be attrs included in the notification payload that do not change, so NTC, sorry for the noise

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: NTC (deferred to #1494)

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
fgalan: This .test has been rewritten, thus the comment is obsolete. NTC.

test/functionalTest/cases/2507_metadata_flags_in_notifications/create_and_update_csubs_with_metadata_flags_field.test
```diff
@@ -0,0 +1,326 @@
...
+07. Update subscription with metadataFlags: false
+=================================================
+HTTP/1.1 204 No Content
+Content-Length: 0
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+
+
+08. Get subscription and see metadataFlags:
```
fgalan: This .test has been rewritten, thus the comment is obsolete. NTC.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: This part has change in deep, making the comment obsolete.

Thus, NTC.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: That function no longer exists after the re-work of the code. Thus, NTC.

test/functionalTest/cases/2507_metadata_flags_in_notifications/metadata_flags_in_notifications_ngsiv1.test
```diff
@@ -0,0 +1,887 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+accumulatorStart --pretty-print
+
+--SHELL--
+
+#
+# Same script than metadata_flags_in_notifications_ngsiv1.test but using "legacy" as attrsFormat
```
fgalan: Fixed in recent commits.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Having clarified this, I understand this is NTC.

doc/manuals/user/ngsiv2_implementation_notes.md
fgalan: Not sure what do you refer, but given that above comments has been already solved (either NTC or "Fixed in") I understand this is also NTC.

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -725,6 +725,10 @@ Let's examine in detail the different elements included in the payload:
     way only to show the enormous flexibility of subscriptions.
 -   You can leave `conditions.attrs` empty to make a notification
     trigger on any entity attribute change (regardless of the name of the attribute).
+-   Notifications include the attribute values *after* processing the update operation
+    triggering the notification. However, you can make Orion to include also the
+    *previous* value. This is achieved using metadata. Have a look to
+    [the following piece of documentation](metadata.md#metadata-in-notifications).
```
fgalan: We don't use to include links to PRs in documentation. However, I understand that the topic is well covered by the materials in metadata.md cited from this point.

Thus, I'd say this is NTC.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 6c4e9ea

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Fixed in 6c4e9ea

src/lib/jsonParseV2/utilsParse.h
```diff
@@ -38,6 +38,22 @@ struct Opt {
   bool ok() { return error.empty(); }
 };
 
+
+
+/*
+ * getBoolMust -  get a mandatory bool from the rapidjson node
+ */
+Opt<bool> getBoolMust(const rapidjson::Value& parent, const char* field,const std::string& description = "");
```
fgalan: `getBoolMust` and `getBoolOpt` were initially added when the approach was to use a true/false flag for metadata field. Now that the field has evolved to an array, they are no longer needed.

However, I think we can leave them as utility functions, to be used to process future boolean field at parsing stage.

Opinions?

src/lib/mongoBackend/mongoGetSubscriptions.cpp
```diff
@@ -139,6 +139,7 @@ static void setSubject(Subscription* s, const BSONObj& r)
 static void setNotification(Subscription* subP, const BSONObj& r, const std::string& tenant)
 {
   // Attributes
+  // FIXME: use setStringVectorF
```
fgalan: This (an other similar FIXME) will be addressed in a very next PR.

NTC (just informative)

src/lib/ngsi/Metadata.cpp
```diff
@@ -259,11 +259,14 @@ std::string Metadata::render(const std::string& indent, bool comma)
...
     {
+      // Note in this case we don't add the "value" key, the toJson()
+      // method does it for toplevel compound (a bit crazy... this deserves a FIXME mark)
+      compoundValueP->renderName = true;
       part = compoundValueP->toJson(true, false);
     }
     else if (compoundValueP->isVector())
     {
-      part = "[" + compoundValueP->toJson(true, false) + "]";
+      part = JSON_STR("value") + ": [" + compoundValueP->toJson(true, false) + "]";
```
fgalan: I wonder how metadata rendering could work in the case of vector values... Maybe we don't have any test covering that.

It worth the effort to evaluate the vector metadata rendering comparing develop with this branch results in order to know if we have fixed an existing bug in the process.

src/lib/ngsi/Metadata.cpp
```diff
@@ -428,8 +429,7 @@ std::string Metadata::toStringValue(void) const
     }
     else // regular number
     {
-      snprintf(buffer, sizeof(buffer), "%f", numberValue);
-      return std::string(buffer);
+      return toString(numberValue);
```
fgalan: The fix in this function is the one mentioned in CNR as:

> Fix: removed spurious decimals in metadata numbers in NGSIv1 rendering (a follow up of #2176)

We need this otherwise we would get thinks like this in NGSIv1 notifications:

```
                    {
                        "metadatas": [
                            {
                                "name": "previousValue",
                                "type": "Number",
                                "value": 20.0
                            }
                        ],
                        "name": "B",
                        "type": "Number",
                        "value": 20
                    }
```

which is a bit weird.

NTC (just informative).

test/functionalTest/cases/2507_csub_metadata_field/metadata_filter.test
```diff
@@ -0,0 +1,792 @@
...
+Content-Length: 0
+Location: /v2/entities/E?type=T
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+
+
+06. Dump and reset: see 4 notification (M1+M2+M3, M1+M2+M3+actionType, M1+M2, actionType+M3)
+============================================================================================
+#SORT_START
```
fgalan: There are three simultaneous subscriptions covering the same entity. Thus, notification order is not guaranteed and we need to use SORT_START/SORT_END in this .test.

NTC  (just informative).

test/functionalTest/cases/2507_csub_metadata_field/metadata_filter_existing_sub.test.DISABLED
```diff
@@ -0,0 +1,259 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: I haven't been able to finish this .test, due to some problems in update subscription logic to be solved in a very next PR. Thus, the .test file is added but in .DISABLED mode.

NTC (just informative).

test/functionalTest/cases/2496_mq_metadata_not_exist/mq_metadata_not_exist.test
```diff
@@ -0,0 +1,139 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This file shouldn't be part of this PR... Not sure how it has ended here...

fgalan: This is funny... it should be part of https://github.com/telefonicaid/fiware-orion/pull/2512 but it seems that it wasn't added in that moment.

So this PR now includes it. Problem fixed. NTC.

kzangeli:
LGTM

src/lib/ngsi/Metadata.cpp
kzangeli: Oh yes, we have plenty of tests.
The "value": is added before this function.
There is a risk we'll get a double "value": with this change.
I know this because I've been here seeing/fixing problems not so long ago.
However, I imagine all functests work so we should be safe.

In the case of an JSON Object, the rendering is different, no "value": there ...
We should go over this and clean it, amke Array and Object behave the same way.

I just wonder, is it ok to hardcode "value" here?
What if we have a name of the vector? For keyValues ...
In any case, if all functests are OK, all is OK. At least until we go deeper here

Perhaps we should add an issue about 'refactoring' this, unless we already have one (I might have created one not too long ago)

NTC

src/lib/ngsi/Metadata.cpp
```diff
@@ -259,16 +259,17 @@ std::string Metadata::render(const std::string& indent, bool comma)
...
     }
     else if (compoundValueP->isVector())
     {
-      part = "[" + compoundValueP->toJson(true, false) + "]";
+      part = JSON_STR("value") + ": [" + compoundValueP->toJson(true, false) + "]";
     }    
 
-    //out += part;
-    out += indent + "  " + JSON_STR("value") + ": " + part;
+    out += part;
```
kzangeli: This is where "value" was removed, to add it above

NTC

src/lib/ngsi/MetadataVector.cpp
```diff
@@ -87,6 +87,25 @@ std::string MetadataVector::render(const std::string& indent, bool comma)
...
+*
+*/
+bool MetadataVector::matchFilter(const std::string& mdName, const std::vector<std::string>& metadataFilter)
+{
+
+  /* Metadata filtering only in the case of actual metadata vector not conatining "*" */
+  if ((metadataFilter.size() == 0) || (std::find(metadataFilter.begin(), metadataFilter.end(), NGSI_MD_ALL) != metadataFilter.end()))
+  {
+    return true;
+  }
```
kzangeli: What about the two special metadata?
It seems here that TRUE is returned if metadataFilter is empty or contains "*",
but I understand this is not enough for those two, they must be explicitly there ...

Right?

test/functionalTest/cases/2507_csub_metadata_field/metadata_filter.test
kzangeli: However, you have written the expected outcome in 'chronological order', right? (order of sub creation)

The problem can be avoided by using three entities and three dumps (+accumulatorReset), however, the test would not be identical. Depends on exactly what you want to test.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -913,6 +923,21 @@ Note that if a custom payload is used for the notification (the field `payload`
 corresponding subscription), then a value of `custom` is used for the `Ngsiv2-AttrsFormat` header
 in the notification.
 
+## Special metadata in notifications
```
jmcanterafonseca: Should we name this as "System Metadata"?

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1670,6 +1695,9 @@ A `notification` object contains the following subfields:
   Accepted values are `normalized` (default), `keyValues` or `values`.
   If `attrsFormat` takes any value different
   than those, an error is raised. See detail in "Notification Messages" section.
++ `metadata` (optional): List of metadata to be included in notification messages. By default all
+  metadata are included except the special ones described at "Special metadata in notifications"
```
jmcanterafonseca: all user metadata are included (but not System Metadata)

doc/manuals/user/metadata.md
```diff
@@ -402,3 +403,34 @@ attributes using ID metadata:
     remove an specific attribute identified by ID (see DELETE attribute
     semantics [described in a previous
     section](append_and_delete.md#adding-and-removing-attributes-with-append-and-delete-in-updatecontext)).
+
+
+## Metadata in notifications
+
+By default, all custom (user) metadata are included in notifications. However, the field `metadata`
+can be used to filter the list. In addition, it can be used to specify that the following special
```
jmcanterafonseca: system metadata

doc/manuals/user/metadata.md
```diff
@@ -402,3 +403,34 @@ attributes using ID metadata:
...
+Details about their meaning can be found in the "Special metadata in notifications" section in
+the NGSIv2 specification).
+
+Note that using the following
+
+```
+"metadata": [ "previousValue" ]
+```
+
+will cause to include `previousValue` but will remove any other metadata that
```
jmcanterafonseca: but will exclude user metadata

doc/manuals/user/metadata.md
```diff
@@ -402,3 +403,34 @@ attributes using ID metadata:
...
+attributes in the notification may have. If you want to get `previousValue`
+*and* any other "regular" metadata then use:
+
+```
+"metadata": [ "previousValue", "*" ]
+```
+
+Note that you can also use `"metadata": [ "*" ]` although it doesn't make much sense, as
+it gives the same result as not including `metadata` at all (remember that the default
+behaviour is to include all custom metadata).
```
jmcanterafonseca: all user metadata

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Maybe. But I move that discussion outside this PR (it also affects to attributes, such as dataCreated, which call be named "System attribute").

doc/manuals/user/metadata.md
fgalan: See comment above.

src/lib/ngsi/Metadata.cpp
fgalan: We have plenty of test but... we have a test covering the metadata compound vector rendering in NGSIv1?

Note that "value" is added *but only in the case* of objects. Have a look to CompoundNode::toJson(), the check for "topmost" is done only in the object case.

I think it is not only ok to hardoce "value" here (for the case of compound vector) because no other is going to add it. Again, CompoundNode::toJson() only add it in the case of object, as far as I have seen).

I think we already have that issue on refactoring: https://github.com/telefonicaid/fiware-orion/issues/1298 (however, the issue currently mention only NGSIv2... we should consider if it worth the time/effort to refactor also all the NGSIv1 render logic).

test/functionalTest/cases/2507_csub_metadata_field/metadata_filter.test
fgalan: Not fully sure about the order... I think I used a reference execution (copy-paste :) to fill that part, then use SORT when I realize after some executions that the order actually was slightly different from .test to .test.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Disccused at skype. NTC.

doc/manuals/user/metadata.md
fgalan: NTC

src/lib/ngsi/MetadataVector.cpp
fgalan: The previousValue and actionType metadata has been previously added along with all the other "regular" metadata at processSubscription() function, in particular at this point: https://github.com/telefonicaid/fiware-orion/issues/1298

src/lib/jsonParseV2/utilsParse.h
kzangeli: Leave them be.
We will probably use them some day.
If you want, use #if 0 / #endif to not occupy and code in the executable

doc/apiary/v2/fiware-ngsiv2-reference.apib

doc/manuals/user/metadata.md
fgalan: Fixed in 7a2e365

doc/manuals/user/metadata.md
fgalan: Fixed in 7a2e365 (although we haven't decided yet about "System metadata" I think "user" is better than "custom" in this point)

kzangeli:
LGTM (again)

jmcanterafonseca:
my review was approved but LGTM, for the avoidance of doubt

PR #2528, opened 2016-09-22
Bug/2527 crash with empty uri path

### LINKED ISSUE #2527: Broker crashes with empty URI PATH
curl localhost:1026/ crashes the broker

kzangeli:
Fixes #2527 

Checking that URI PATH vector is not empty before using it.
This avoids a crash for empty URI PATH.


fgalan:
LGTM

PR #2534, opened 2016-09-26
[CHANGES] in Tests summary (table)

iariasleon:
Now in the new format, the summary is presented as a table

See a example in https://jirapdi.tid.es/secure/attachment/912737/regressiom_test.txt

### DIFF
--- a/test/acceptance/behave/environment.py
+++ b/test/acceptance/behave/environment.py
@@ -129,8 +129,8 @@ def __add_chars(text, before, after, char_to_add=" "):
     """
     append chars before or/and after in a text
     :param text: text to modify
-    :param before: number of blanks added to the beginning of the text
-    :param after: number of blanks added to the end of the text
+    :param before: number of chars added to the beginning of the text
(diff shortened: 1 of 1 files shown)


test/acceptance/behave/environment.py
```diff
@@ -125,6 +125,20 @@ def init_feature():
     scenarios_status = {"untested": 0, "skipped": 0, "passed": 0, "failed": 0}
 
 
+def __add_chars(text, before, after, char_to_add=" "):
+    """
+    append chars before or/and after in a text
+    :param text: text to modify
+    :param before: number of blanks added to the beginning of the text
```
fgalan: number of blanks -> number of chars

(You mention in char_to_add comment that blank is the default, but not the general case)

iariasleon: Yes, initially in this function only blank spaces are appended, now, it is used with any character.
Fixed in  	b5998b6

jcanonav:
LGTM

fgalan:
LGTM

PR #2537, opened 2016-09-26
Some fixes from bugs found during PR #2518

fgalan:
Related with fails in csub update logic and a bug in cache update logic when condtions.attrs is empty.


test/functionalTest/cases/2507_csub_metadata_field/metadata_filter_existing_sub.test.DISABLED
```diff
@@ -1,259 +0,0 @@
-# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: It seems that the github heuristic to detect that a file has been renamed didn't detect the .test.DISABLED -> .test in this PR...

NTC (just a remark)

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -618,15 +626,14 @@ void updateInCache
...
-    subCacheItemRemove(subCacheP);
-  }
-
   if (mscInsert == 0)  // 0: Insertion was really made
   {
     subCacheUpdateStatisticsIncrement();
+    if (subCacheP != NULL)
+    {
+      LM_T(LmtSubCache, ("Calling subCacheItemRemove"));
+      subCacheItemRemove(subCacheP);
```
fgalan: Moving the remove operation inside the `if (mscInsert == 0)` is not directly related with the fix implemented in this bug. However, I think is an important "safety measure" as in the future other conditions may fail at mongoSubCacheItemInsert() and we don't want in that case not only not updating the subscription but removing it from the cache.

kzangeli: Sure. Seems like a good idea

kzangeli:
LGTM

PR #2538, opened 2016-09-26
Refactor to use setStringVectorF() as much as possible

fgalan:
This PR doesn't deserve an entry in CNR. The change is a internal refactor.


kzangeli:
LGTM - but conflict must be resolved before merging

fgalan:
Conflict just solved

PR #2540, opened 2016-09-26
Bug/1068 render metadata value when compound

kzangeli:
Part of old and closed issue #1068.
Fixed the rendering of compound metadatas when the value is a vector.

An intent to rewrite/simplify that part (before I found the fix) has been kept inside #if 0/#endif.
I don't want to throw away this simplification as I believe we should use it in the near future.

Also, not sure if this fix should have an entry in the CNR ...

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,6 +2,7 @@
 - Add: filtering metadata in notifications (#2507)
 - Fix: removed spurious decimals in metadata numbers in NGSIv1 rendering (a follow up of #2176)
 - Fix: crash when creating entity with metadata when subscription is in place with "mq" on a different metadata (#2496)
+- Fix: correct rendering of metadata compound values when a vector
 - Fix: crash with request with empty URI PATH (Issue #2527)
 - Fix: supporting DateTime filters for metadata (#2443, #2445)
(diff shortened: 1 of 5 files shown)


src/lib/ngsi/Metadata.cpp
```diff
@@ -259,13 +259,18 @@ std::string Metadata::render(const std::string& indent, bool comma)
```
kzangeli: Of course.

Fixed in abfb123

CHANGES_NEXT_RELEASE
```diff
@@ -2,6 +2,7 @@
 - Add: filtering metadata in notifications (#2507)
 - Fix: removed spurious decimals in metadata numbers in NGSIv1 rendering (a follow up of #2176)
 - Fix: crash when creating entity with metadata when subscription is in place with "mq" on a different metadata (#2496)
+- Fix: correct rendering of metadata compound values when a vector
```
fgalan: I suggest to refer to #1068 here, e.g:

- Fix: correct rendering of metadata compound values when a vector (a bugfix for #1068 functionality)

kzangeli: ok

kzangeli: Fixed in be56f35

fgalan:
LGTM

PR #2545, opened 2016-09-28
FIX avoid usage of getFieldF() as much as possible

fgalan:
We have reduce the number of getFieldF() occurrences from 98 to 60. The 60 that remains seems to be necessary.

I have initially though to make much more changes, but the ones this PR includes are valuable anyway.


src/lib/mongoBackend/mongoGetSubscriptions.cpp
```diff
@@ -115,7 +115,7 @@ static void setSubject(Subscription* s, const BSONObj& r)
 
   if (r.hasField(CSUB_EXPR))
   {
-    mongo::BSONObj expression = getFieldF(r, CSUB_EXPR).Obj();
+    mongo::BSONObj expression = getObjectFieldF(r, CSUB_EXPR);
     std::string    q          = getStringFieldF(expression, CSUB_EXPR_Q);
```
kzangeli: Surprising to see earlier changes  from `getFieldF(...).embeddedObject()` to `getObjectFieldF()` but here we go from `getFieldF().Obj()` to `getObjectFieldF()` ...   Care to explain?

src/lib/mongoBackend/safeMongo.cpp
```diff
@@ -106,6 +106,34 @@ std::string getStringField(const BSONObj& b, const std::string& field, const std
...
+  if (!b.hasField(field))
+  {
+    LM_E(("Runtime Error (double field '%s' is missing in BSONObj <%s> from caller %s:%d)", field.c_str(), b.toString().c_str(), caller.c_str(), line));
+  }
+  else
+  {
+    LM_E(("Runtime Error (field '%s' was supposed to be an double but type=%d in BSONObj <%s> from caller %s:%d)",
+          field.c_str(), b.getField(field).type(), b.toString().c_str(), caller.c_str(), line));
+  }
+  return -1;
```
kzangeli: -1 is a perfectly valid `double`.
How does the caller detect this error?
Seems to be this function should return 0 on OK and send the double as an output parameter ... ?

src/lib/mongoBackend/mongoGetSubscriptions.cpp
fgalan: My theory is that `embeddedObject()` and `Obj()` results the same, i.e. one function is a kind of alias of the other.

The fact that test harness are ok seems to confirm this theory.

src/lib/mongoBackend/safeMongo.cpp
fgalan: In general for all the functions in safeMongo, the caller is not able to detect the error. That is intentional: before safeMongo fails of this kind make the CB crash. So the apporach now is to not interrupt the flow (we have to return something... -1 seems to be a good value as use to be very weird to see it, so the user can detect also in the payload) and mark the error in the logs.

However, looking to how getStringFieldF() is implemented, returning "" in this case, maybe we should return 0, which is more neutral.

src/lib/mongoBackend/mongoGetSubscriptions.cpp
kzangeli: Yeah, I assumed just that.
Wanted you to confirm it.

NTC

src/lib/mongoBackend/safeMongo.cpp
kzangeli: getStringFieldF returns a std::string I assume.
Had it been a char*, NULL would be the choice. With std::string you can't really do this.
"" seems like a good choice. Either that or returning an error description, which may be only confusing.

NTC IMO

PR #2555, opened 2016-09-29
feature/2073_uri_param_to_disable_type_detail

kzangeli:
This PR merges to develop, must be taken to master after that.

Implements #2073.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,2 @@
-- Add: new value for URI param 'options' (options=noAttrDetail) for lighter queries (Issue #2073) 
+- Add: new value for URI param 'options' (options=noAttrDetail) for NGSIv2 queries to exclude attribute detail, making the query faster (#2073) 
+- Fix: avoid unused attribute type detail in NGSIv1 queries (which was making the queries slower) (#2073)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,12 +1,1 @@
-- Add: notification metadata previousValue and actionType (except 'delete') (#2507)
-- Add: filtering metadata in notifications (#2507)
-- Fix: removed spurious decimals in metadata numbers in NGSIv1 rendering (a follow up of #2176)
-- Fix: wrong attributes included in notifications at subscription update time in some cases (both in blacklist and not blacklist cases)
-- Fix: csub cache wrong deletion of subscription at subscription update time when new subscription has empty condition attributes list
-- Fix: crash when creating entity with metadata when subscription is in place with "mq" on a different metadata (#2496)
-- Fix: correct rendering of metadata compound values when a vector (bugfix for issue #1068)
-- Fix: crash with request with empty URI PATH (Issue #2527)
-- Fix: supporting DateTime filters for metadata (#2443, #2445)
-- Fix: wrong matching in metadata existence and not existence filters with compounds
-- Fix: missing expression sub-fields (q, etc.) in csubs makes Orion to crash
-- Hardening: ensure to use type oriented functions when getting data from DB (safer, but more verbose in error log if DB is not perfectly aligned with what Orion is expecting)
+- Add: new value for URI param 'options' (options=noAttrDetail) for lighter queries (Issue #2073) 
--- a/doc/manuals/user/ngsiv2_implementation_notes.md
+++ b/doc/manuals/user/ngsiv2_implementation_notes.md
@@ -113,3 +113,9 @@ From NGSIv2 specification section "Special metadata in notifications", regarding
 
 Current Orion implementation supports "update" and "append". The "delete" case will be
 supported upon completion of [this issue](https://github.com/telefonicaid/fiware-orion/issues/1494).
+
+# URI parameters
+
+A special value of the URI param `options` (options=noAttrDetail) has been invented, for users of the
+broker to be able to avoid the CPU-intense query of the list of attribute types. `noAttrDetails` exists not
+for NGSIv2 but to improve performance in some operations.
--- a/doc/manuals/user/ngsiv2_implementation_notes.md
+++ b/doc/manuals/user/ngsiv2_implementation_notes.md
@@ -114,8 +114,12 @@ From NGSIv2 specification section "Special metadata in notifications", regarding
 Current Orion implementation supports "update" and "append". The "delete" case will be
 supported upon completion of [this issue](https://github.com/telefonicaid/fiware-orion/issues/1494).
 
-# URI parameters
+# `noAttrDetail` option
 
-A special value of the URI param `options` (options=noAttrDetail) has been invented, for users of the
-broker to be able to avoid the CPU-intense query of the list of attribute types. `noAttrDetails` exists not
-for NGSIv2 but to improve performance in some operations.
+The value `noAttrDetail` of the URI param `options` may be used in order to avoid NGSIv2 type browsing queries
+(`GET /v2/types` and `GET /v2/types/<type>`) to provide attribute type details.
+When used, the `types` list associated to each attribute name is set to `[]`.
+
+Using this option, Orion solves these queries much faster, especially in the case of a large number of attributes, each one with a different type.
+This can be very useful if your use case doesn't need the attribute type detail.
+In some cases savings from 30 seconds to 0.5 seconds with the `noAttrDetails` option have been detected.
--- a/src/lib/common/globals.h
+++ b/src/lib/common/globals.h
@@ -116,7 +116,7 @@
 #define OPT_UNIQUE_VALUES   "unique"
 #define OPT_DATE_CREATED    DATE_CREATED
 #define OPT_DATE_MODIFIED   DATE_MODIFIED
-
+#define OPT_NO_ATTR_DETAIL  "noAttrDetail"
 
  
 /* ****************************************************************************
--- a/src/lib/mongoBackend/mongoQueryTypes.cpp
+++ b/src/lib/mongoBackend/mongoQueryTypes.cpp
@@ -54,8 +54,8 @@ static void getAttributeTypes
 {
   std::string  idType         = std::string("_id.")    + ENT_ENTITY_TYPE;
   std::string  idServicePath  = std::string("_id.")    + ENT_SERVICE_PATH;
+  BSONObj      query;
 
-  BSONObj query;
   if (entityType == "")
   {
     query = BSON("$or"         << BSON_ARRAY(BSON(idType << entityType) << BSON(idType << BSON("$exists" << false)) ) <<
@@ -100,7 +100,7 @@ static void getAttributeTypes
      *   BSONObj attr  = getObjectFieldF(attrs, attrName);
      *   attrTypes->push_back(getStringFieldF(attr, ENT_ATTRS_TYPE));
      *
-     * However, it doesn't work when the attribute used metadata ID
+     * However, it doesn't work when the attribute uses metadata ID
      *
      */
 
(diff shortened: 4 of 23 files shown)


fgalan:
This PR is NOT TO MERGE YET, right?

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -434,24 +435,31 @@ HttpStatusCode mongoEntityTypes
         }
 
         /* Note that we need and extra query() to the database (inside attributeType() function) to get each attribute type.
-         * This could be unefficient, specially if the number of attributes is large */
-        std::vector<std::string> attrTypes;
-        getAttributeTypes(tenant, servicePathV, entityType->type , attrsArray[jx].str(), &attrTypes);
-
-        for (unsigned int kx = 0; kx < attrTypes.size(); ++kx)
+         * This could be inefficient, especially if the number of attributes is large */
+        if (!noAttrDetail)
```
fgalan: Just to see if I'm understanding correctly the modification... You have move the all logic under an `if (!noAttrDetail) and, in addition, adding a else clause. Is that correct?

src/lib/ngsi/ContextAttributeVector.cpp
```diff
@@ -99,16 +98,20 @@ std::string ContextAttributeVector::toJsonTypes()
...
     out += JSON_STR(attrName) + ":{" + JSON_STR("types") + ":[";
 
-    std::map<std::string, int>::iterator jt;
-    unsigned int                         jx;
-    for (jt = attrTypes.begin(), jx = 0; jt != attrTypes.end(); ++jt, ++jx)
+    if (noAttrDetail == false)
     {
-      std::string type = jt->first;
-      out += JSON_STR(type);
+      std::map<std::string, int>::iterator jt;
```
fgalan: I'll suggest to avoid modifying any toJson() signature detecting the "no type condition" directly in this function. I mean, if the types vector has only one element and that element is "" (which is not legal in NGSIv2) then print `[]`.

I can try to commit the change myself if you prefer or the strategy is not clear enough...

src/lib/mongoBackend/mongoQueryTypes.h
```diff
@@ -56,7 +59,8 @@ extern HttpStatusCode mongoEntityTypes
   const std::vector<std::string>&      servicePathV,
   std::map<std::string, std::string>&  uriParams,
   const std::string&                   apiVersion,
-  unsigned int*                        totalTypesP
+  unsigned int*                        totalTypesP,
+  bool                                 noAttrDetail = false
```
fgalan: I'd suggest not using default for `noAttrDetails` (and take the opportunity to remove also in apiVersion in the next function). The number of calls to this functions is small and that way we can ensure that it is correctly set in NGSIv1 calls (that are a bit "critial" due to its usage from IoTP).

src/lib/common/globals.h
```diff
@@ -116,7 +116,7 @@
 #define OPT_UNIQUE_VALUES   "unique"
```
fgalan: CNR entry should be added.

In addition, a new section in ngsiv2_implementation_notes.md should be created explaining the existence of this new implementation specific parameter (it some way, it is like "legacy": something that Orion support for performance reasons but not part of the NGSIv2 API spec).

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -607,20 +616,28 @@ HttpStatusCode mongoAttributesForEntityType
 
     /* Note that we need and extra query() to the database (inside attributeType() function) to get each attribute type.
      * This could be unefficient, specially if the number of attributes is large */
-    std::vector<std::string> attrTypes;
-    getAttributeTypes(tenant, servicePathV, entityType , idField.str(), &attrTypes);
-
-    for (unsigned int kx = 0; kx < attrTypes.size(); ++kx)
+    if (!noAttrDetail)
```
kzangeli: True (horrible diff - just for the change in indentation)

src/lib/mongoBackend/mongoQueryTypes.h
kzangeli: Sure, default values just complicate things.

src/lib/ngsi/ContextAttributeVector.cpp
kzangeli: ok, I understand the 'strategy'.
It makes the function a bit hard to understand though.
If we do this, it must be clearly documented in the code

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: NTC

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: NTC

src/lib/common/globals.h
kzangeli: This PR merges to develop. The CNR in develop is still full of stuff from the las release …
I'll empty it and add my line.

src/lib/mongoBackend/mongoQueryTypes.h
kzangeli: Fixed in be80dc8

src/lib/common/globals.h
kzangeli: Fixed in be80dc8

CHANGES_NEXT_RELEASE
```diff
@@ -1,12 +1 @@
...
-- Fix: wrong attributes included in notifications at subscription update time in some cases (both in blacklist and not blacklist cases)
-- Fix: csub cache wrong deletion of subscription at subscription update time when new subscription has empty condition attributes list
-- Fix: crash when creating entity with metadata when subscription is in place with "mq" on a different metadata (#2496)
-- Fix: correct rendering of metadata compound values when a vector (bugfix for issue #1068)
-- Fix: crash with request with empty URI PATH (Issue #2527)
-- Fix: supporting DateTime filters for metadata (#2443, #2445)
-- Fix: wrong matching in metadata existence and not existence filters with compounds
-- Fix: missing expression sub-fields (q, etc.) in csubs makes Orion to crash
-- Hardening: ensure to use type oriented functions when getting data from DB (safer, but more verbose in error log if DB is not perfectly aligned with what Orion is expecting)
+- Add: new value for URI param 'options' (options=noAttrDetail) for lighter queries (Issue #2073)
```
fgalan: I'd suggest to be more precise:

```
Add: new value for URI param 'options' (options=noAttrDetail) for NGSIv2 type browsing queries to avoid attribute detail, making the query faster (#2073) 
```

or something like that. With regards to NGSIv1, this PR also involves a fix (in a second CNR line):

```
Fix: avoid unused attribute type detail in NGSIv1 type browsing queries (which was making queries slower) (#2073)
```

or something like that (probably in this case my wording proposal is not too good... but I guess you get the idea).

Finally, I'll suggest to remove "Issue" before the number. As every #XXXX refers to an issue number, "Issue" is redundant.

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -113,3 +113,9 @@ From NGSIv2 specification section "Special metadata in notifications", regarding
 
 Current Orion implementation supports "update" and "append". The "delete" case will be
 supported upon completion of [this issue](https://github.com/telefonicaid/fiware-orion/issues/1494).
+
+# URI parameters
+
+A special value of the URI param `options` (options=noAttrDetail) has been invented, for users of the
+broker to be able to avoid the CPU-intense query of the list of attribute types. `noAttrDetails` exists not
+for NGSIv2 but to improve performance in some operations.
```
fgalan: I'd suggest the following alternative:

```
# `noAttrDetail` option

The `noAttrDetail` option may be used in order to avoid NGSIv2 type browing queries (`GET /v2/types` and `GET /v2/types/<type>`) to provide attribute type detail. In that case, the `types` list associated to each attribute name is set to `[]`.

Using this option, Orion can solve these queries much faster, specially in the cases of a large number of attributes, each one with a different type. This can be very useful if your use case doesn't need the attribute type detail. We have tested in some cases savings from 30 seconds to 0.5 seconds using the `noAttrDetails` option.
```

src/lib/ngsi/ContextAttributeVector.cpp
```diff
@@ -101,10 +100,23 @@ std::string ContextAttributeVector::toJsonTypes()
...
       std::string type = jt->first;
-      out += JSON_STR(type);
+      
+      //
+      // Special condition for 'options=noAttrDetail':
+      //   When the 'options' URI parameter contains 'noAttrDetail',
+      //   mongoBackend fills the attribute type vector with *just one item* (that is an empty string).
+      //   This special condition is checked for here, to produce a [] for the vector for the response.
+      //
+      // See the origin of this in mongoQueryTypes.cpp. Look for "NOTE: here we add", in two locations.
```
fgalan: Has been this "NOTE: here we add" actually added? I have looked in the PR but I haven't found it...

src/lib/serviceRoutines/getAttributesForEntityType.cpp
```diff
@@ -49,6 +49,7 @@
 *
 * URI parameters:
 *   - attributesFormat=object
+*   - options=noAttrDetail
```
fgalan: I understand this is the service routine for NGSIv1. In that case, `options=noAttrDetail` is not an option and its value is always "true" when invoking mongoAttributesForEntityType().

A comment shluld be added just above mongoAttributesForEntityType() to explain why is alwyas true, e.g. "Note that noAttrDetail is set to true in NGSIv1 operations, as NGSIv1 type browing doesn't actually provides attribute type detail in the API response" or something like that.

src/lib/serviceRoutines/getEntityTypes.cpp
```diff
@@ -48,6 +48,7 @@
 * URI parameters:
 *   - attributesFormat=object
 *   - collapse=true
+*   - options=noAttrDetail
```
fgalan: I understand this is the service routine for NGSIv1. In that case, `options=noAttrDetail` is not an option and its value is always "true" when invoking mongoAttributesForEntityType().

A comment shluld be added just above mongoAttributesForEntityType() to explain why is alwyas true, e.g. "Note that noAttrDetail is set to true in NGSIv1 operations, as NGSIv1 type browing doesn't actually provides attribute type detail in the API response" or something like that.

src/lib/serviceRoutines/getAttributesForEntityType.cpp
kzangeli: This was just added as the signature for mongoAttributesForEntityType was changed (adding the 'noAttrDetail' parameter).
If what you say is true, that the code for the attr-type-list is not even executed for v1 (and I don't doubt your word :-)), then this should change, removing the option as a possible URO param and false sent to mongoAttributesForEntityType

kzangeli: Fixed in 50ab227

doc/manuals/user/ngsiv2_implementation_notes.md
kzangeli: Fixed in 50ab227

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 50ab227

src/lib/ngsi/ContextAttributeVector.cpp
fgalan: NTC

src/lib/serviceRoutines/getAttributesForEntityType.cpp
fgalan: You don't need to trust on me if you can trust on the .test for NGSIv1 type browing operations ;)

src/lib/serviceRoutines/getEntityTypes.cpp
kzangeli: Fixed in 8cc952f

fgalan:
LGTM

PR #2573, opened 2016-10-06
NGSIv2 spec (latests): create/update subscriptions operation in NGSIv2 spec with regards to expires field equal to empty string

fgalan:
"Moved" from https://github.com/telefonicaid/fiware-orion/pull/2305

Comments on the original PR:

@jmcanterafonseca:
and what about null? I think it is the same. and what about " " ? 

@kzangeli:
null is not a string, so error right away there (I hope).
" " is a string but not a 'valid time', so that should give an error as well.
The empty string is accepted, but as an exception ...

@jmcanterafonseca:
null is equivalent as having no expire date, so it should not raise an error

@kzangeli:
If we wish to accept null as another way for permanent subscriptions, no problem.
Right now I believe we force the expiration to be a string, in the parsing stage and an error is returned if not a string. Very easy to modify.

I actually believe null is better than the empty string for a 'non-expiring subscription' ...

@jmcanterafonseca:
sure


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1807,6 +1807,9 @@ Response:
 Creates a new subscription.
 The subscription is represented by a JSON object as described at the beginning of this section.
 
+Using `expires` with value equal to empty string (`""`) equals to omit the field, i.e. creating
```
fgalan: On @kzangeli behalf, from the old PR:

equal-equals ... I don't like it very much.

I propose (with a few more 'tweaks'):

    Using `expires` with a value of the empty string (`""`) equals omitting the field, i.e. creating ...

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1901,6 +1904,9 @@ Response:
 
 Only the fields included in the request are updated in the subscription.
 
+Using `expires` with value equal to empty string (`""`) equals to removing the field, i.e.
```
fgalan: On @kzangeli behalf, from the old PR:

And same here:

    Using `expires` with a value of the empty string (`""`) equals removing the field, i.e. ...

fgalan:
This PR is pretty old and without significatives changes to be preserved. In other words, if the issue about empty/null/missing fields in NGSIv2 is addresse3d some day, it is better to restart the discussing (and modify the .apib based on the conclusiong) rather to trying to reuse this PR and its supporting branch.

The supporting branch will be deleted after closing this PR.

Issue #876, opened 2015-04-12
Implement a timestamp


gogos-venge:
Implement an automatic timestamp (edit: of last action, or let a user set at update/append). That would save lives considering syncing.
```javascript
{
  "contextElements": [
  {
    "type": "Room",
    "isPattern": "false",
    "id": "Room1",
    "timestamp": "1428799422",
    "attributes": [
    {
      "name": "temperature",
      "type": "float",
      "value": "26.5",
      "timestamp": "1428799841",
      "metadatas": [
      {
        "name": "average",
        "type": "float",
        "value": "22.4",
        "timestamp": "1428799831"
      },
      {
        "name": "accuracy",
        "type": "float",
        "value": "0.9",
        "timestamp": "1428799312"
      }
      ]
    }
    ]
  }
  ],
  "statusCode": {
    "code": "200",
    "reasonPhrase": "OK"
  }
}
```

fgalan:
Thanks for the feedback!

Actually, this is a planned feature since time ago, although implementing it in a slightly different way. Instead of including a `timestamp` field (which is not allowed by NGSI spec) to specify the timestamp of an attribute we plan to use a metadata. We don't plan to implement a timestamp for the metadata themselves, but we think that doing it at attribute level is enough (in some sense, metadata array is part of the value of the attribute).

gogos-venge:
Yes, I thought later after posting this feedback, NGSI would not allow such thing. But actually I already implemented exactly what you said for an API I'm working, which uses Orion for syncing. I figured out it would be impossible to set a timestamp for metadata since it's the bottom level, and using metadata pairs would not only be a VERY bad practice, but would eventually generate extra traffic and frustration. That disables metadatas for us though, but still attributes should be quite enough for the data storing.

fgalan:
Finally, we have defered this to v2 API. This version uses a simplified API such as:

```
  {
    "type": "Car",
    "id": "P-9873-K",
    "speed": {
      "value": 100,
      "accuracy": 2
      }
  }
```
In which including `creDate` and `modDate` (both at entity level and attribute level) would be easy:

```
  {
    "type": "Car",
    "id": "P-9873-K",
    "creDate": <ISODATE>,
    "modeDate": <ISODATE>,
    "speed": {
      "value": 100,
      "creDate": <ISODATE>,
      "modeDate": <ISODATE>,
      "accuracy": 2
      }
  }
```

[...]

Issue #2450, opened 2016-08-16
make simpler entity's attribute dates / metadata management


jmcanterafonseca:
to list entity attributes now we have two different options

options=dateCreated, dateModified , 

perhaps we can just have `options=entityDates` to simplify. Or even `options=entityMeta` to list all the attributes which are actually entity level metadata.

fgalan:
Related with this, virtual attribute probably should use its own URI param. Not sure if overloading options with this is a good idea. E.g:

`?virtAttrs=dateCreate`.

My only concern if that this breaks the stable NGSIv2 spec. It is clear this is a case of "slight modifications in the functionality currently specified" but not sure if this the change is "completely justified" or a "matter of taste" (in the later case, it shouldn't be implemented).

jmcanterafonseca:
yes, `extraAttrs` would be nice and would leave an extensible API

fgalan:
This issue has probably be surpassed by #2455. Maybe it should be closed.

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2590

PR #2590, opened 2016-10-07
Feature/876 attribute timestamps

### LINKED ISSUE #2455: by default list dateCreated and dateModified for entities?
I was wondering if we can save some syntax and API complexity if by default `dateCreated` and `dateModified` always appear. All in all the API user can use the `attrs` filter to list only the attributes she wants. It is a bit of an anomaly to have some attributes that are not listed by default, even if they are automatically generated by the implementation. Treating them as regular attributes would save us work.

kzangeli:
Fixes issue #876 [edit: in addition it fixes #2455 and #2450]


test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
```diff
@@ -0,0 +1,212 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: It would be interesting to include also a .test for the subscription/notification case

src/lib/ngsi/ContextElementResponse.cpp
```diff
@@ -268,7 +268,6 @@ ContextElementResponse::ContextElementResponse
     /* Setting custom metadata (if any) */
     if (attr.hasField(ENT_ATTRS_MD))
```
fgalan: Documentation missing (as discussed at skype... comenting in the PR to ensure we don't forget about it during weekend :)

CHANGES_NEXT_RELEASE
```diff
@@ -1,5 +1,6 @@
-- Add: Added URI parameter metadata=metadata-name-list for NGSIv2 retrieve requests (Issue #2544)
-- Add: Added 'metadata' as part of the payload for POST /v2/op/query (Issue #2544)
+- Add: Added URI parameter metadata=metadata-name-list for NGSIv2 retrieve requests (#2544)
+- Add: Added 'metadata' as part of the payload for POST /v2/op/query (#2544)
+- Add: timestamps (dateCreated + dateModified) for attributes (as for entities) (#876)
```
fgalan: I'd suggest mentioning metadata in the entry. Something like this:

> Add: timestamps (dateCreated + dateModified) metadata for attributes (as for entities) (#876)

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
```diff
@@ -0,0 +1,212 @@
...
+}'
+T_0=$(TZ=GMT date +"%Y-%m-%dT%H:%M:%S.00Z")
+orionCurl --url /v2/entities --payload "$payload"
+echo
+echo
+
+
+echo "02. GET E1/A1, see dateCreated == dateModified"
+echo "=============================================="
+orionCurl --url '/v2/entities/E1/attrs/A1?options=dateCreated'
```
fgalan: For metadata, this should be done in a different way. In particular:

```
GET /v2/entities/E1/attrs/A1?metadata=dateCreated
```

Similar to the future way in which this will be done for dateCreated/dateModifiied associated to entities:

```
GET /v2/entities/E1/attrs?attrs=dateCreated
```

Note that the `options=dateCreated` way will be deprecated at the end (although still supported, due to backward compatibility). See https://github.com/telefonicaid/fiware-orion/issues/2455#issuecomment-248883258

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
fgalan: I mean using this in the subscription:

```
{
  ...
 "metadata": [ "dateCreated", "dateModified" ]
  ...
}
```

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
kzangeli: Discussed on skype. only 'metadata=dateCreated,dateModified` to be supported (for metadata - and options=dateCreated,dateModified to be kept for backward compatibility but only for entity timestamps)

src/lib/ngsi/ContextElementResponse.cpp
```diff
@@ -134,6 +134,7 @@ ContextElementResponse::ContextElementResponse
 (
   const mongo::BSONObj&  entityDoc,
   const AttributeList&   attrL,
+  AttributeList*         metadataList,
```
kzangeli: Yeah, I tried to do that.
The compiler didn't agree with me so I made it a simple pointer

src/lib/serviceRoutinesV2/getEntityAttribute.cpp
```diff
@@ -74,7 +74,8 @@ std::string getEntityAttribute
 
   // 01. Fill in QueryContextRequest
   parseDataP->qcr.res.fill(compV[2], type, "false", EntityTypeEmptyOrNotEmpty, "");
-
+  parseDataP->qcr.res.metadataList.fill(ciP->uriParam["metadata"]);
```
fgalan: A define should be used instead of `"metadata"`, as done with URI_PARAM_ATTRIBUTES.

fgalan: Actually the parameter exists (URI_PARAM_METADATA). Easily fixable.

fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/pull/2590/commits/126e9506763c2389da6cac16d89e15da2a271fba

src/lib/ngsi/ContextElementResponse.cpp
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/pull/2590/commits/126e9506763c2389da6cac16d89e15da2a271fba

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/pull/2590/commits/126e9506763c2389da6cac16d89e15da2a271fba

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
fgalan: These test has been added in the recent commits today.

test/functionalTest/cases/0876_entity_dates/get_entity_dates.test
```diff
@@ -57,14 +57,14 @@ echo
 
 echo "02a. Get entity with dates"
 echo "=========================="
-orionCurl --url /v2/entities/E?options=dateCreated,dateModified
```
fgalan: The old copy of this .test has been moved to cases/0000_deprecated_functionality (or something like that)

NTC

test/functionalTest/cases/0876_entity_dates/notify_entity_dates_with_attr.test
```diff
@@ -0,0 +1,951 @@
...
+        {
+            "A": {
+                "metadata": {},
+                "type": "Number",
+                "value": 500
+            },
+            "dateCreated": {
+                "metadata": {},
+                "type": "DateTime",
+                "value": "REGEX(2\d\d\d\-.*)"
```
fgalan: This REGEX partern should be used instead of REGEX(.*) for values meaning creation or modification dates. It ensures that the time starts with "2xxx-" so it can detect fails due to dates has not been initialized (the not-initialized value is 0, which corresponds to a date in 1970, so the pattern will detect the problem).

NTC

test/functionalTest/cases/0876_attribute_dates/attribute_dates.test
fgalan: NTC (alrady solved)

src/lib/ngsi/ContextElementResponse.cpp
fgalan: Documentation added in the last commits. NTC.

fgalan:
I have self-approved :) my own review. I'm pasing the ball to @kzangeli for an additional review.

src/lib/apiTypesV2/Entity.cpp
```diff
@@ -237,12 +248,23 @@ void Entity::present(const std::string& indent)
...
 */
-void Entity::fill(const std::string& _id, const std::string& _type, const std::string& _isPattern, ContextAttributeVector* aVec)
+void Entity::fill
+(
+  const std::string&       _id,
+  const std::string&       _type,
+  const std::string&       _isPattern,
+  ContextAttributeVector*  aVec,
+  double                   _creDate,
+  double                   _modDate
```
kzangeli: Why use 'double' for dates?
Isn't `long long` a better type?

test/functionalTest/cases/0876_entity_dates/notify_entity_dates_with_attr.test
kzangeli: That is very true. We use REGEX(*) in many places where we could use something as little more intelligent instead.

NTC

kzangeli:
Entire PR revised. Just that doubt about using `double` for dates.

src/lib/apiTypesV2/Entity.cpp
fgalan: I'm just using the same that was used before this PR... See for example https://github.com/telefonicaid/fiware-orion/pull/2590/files#diff-41b80b5423da4a52cade90d9124bcb87L51

Changing to `long long` could be a better option... but we need to analyze the implications (including the ones at DB level). I'd suggest to open an issue about it.

kzangeli: I was thinking maybe it was to have decimals of a second ...
If that is not the case, let's open an issue. Doubles give rounding problems that integers don't have.

NTC

kzangeli:
LGTM

src/lib/apiTypesV2/Entity.cpp
fgalan: Issue created: https://github.com/telefonicaid/fiware-orion/issues/2611

PR #2615, opened 2016-10-25
Filtering by attr dateCreated/dateModified

fgalan:
Last work items related with #876 

This PR also include some adjustment in the "siblinb" .test for entity dates cases.


crbrox:
LGTM

Issue #2622, opened 2016-10-28
Threadpool is not used


fgalan:
Some problems has been recently found with regards to threadpool notification mode:

In particular, the threadpool is not actually used, due to a dynamic linking fail, i.e. transient mode is used even if `-notificationMode threadpool:n:q` is used at CB startup time.

fgalan:
Partial fix in PR https://github.com/telefonicaid/fiware-orion/pull/2623 for Orion 1.5.0. However, the work will continue for a definitive fix in 1.6.0.

fgalan:
The partial fix has been also included in release/1.4.0 branch, in PR https://github.com/telefonicaid/fiware-orion/pull/2625

iariasleon:
LGTM
Tested with 100 recurrent threads and only V2 requests

fgalan:
Definitive fix (master branch) PR https://github.com/telefonicaid/fiware-orion/pull/2636

fgalan:
Definitive fix (release/1.4.0) PR https://github.com/telefonicaid/fiware-orion/pull/2640

PR #2623, opened 2016-10-28
Bug/use pool

crbrox:
A partial fix for #2622 

@fgalan

### DIFF
--- a/src/lib/ngsiNotify/QueueNotifier.cpp
+++ b/src/lib/ngsiNotify/QueueNotifier.cpp
@@ -79,8 +79,8 @@ void QueueNotifier::sendNotifyContextRequest
   // Now, if it is a "custom" notification (with template)
   // we delegate to parent method, and do not use the pool
   // of threads. (To be changed in a future version)
+  // This is a workaround for issue #2622
 
-  //
   if (httpInfo.custom && !disableCusNotif)
   {
     return Notifier::sendNotifyContextRequest(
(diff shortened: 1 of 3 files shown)


src/lib/ngsiNotify/QueueNotifier.cpp
```diff
@@ -63,7 +63,7 @@ int QueueNotifier::start()
 void QueueNotifier::sendNotifyContextRequest
```
fgalan: CNR entry missing. Suggested line:

> Fix: threadpool mode was not actually used, even if -notificationMode is set so; this fix solves the problem for not-custom notifications (partial fix for #2622)

src/lib/ngsiNotify/QueueNotifier.cpp
```diff
@@ -75,6 +75,26 @@ void QueueNotifier::sendNotifyContextRequest
 {
   ConnectionInfo ci;
 
+
+  // Now, if it is a "custom" notification (with template)
+  // we delegate to parent method, and do not use the pool
+  // of threads. (To be changed in a future version)
```
fgalan: For completeness, a reference to #2622 could be a good idea.

crbrox: Fixed 48f6fc4

src/lib/ngsiNotify/QueueNotifier.cpp
crbrox: Fixed 1d399ab

fgalan:
LGTM

PR #2651, opened 2016-11-10
Missing prevValue when only type changes

### LINKED ISSUE #2588: previousValue is not included in notifications (and it should) when only attribute type changes
For subscriptions configured to include previousValue in notifications, this metadata (previousValue) is not included when only the attribute type changes. In other words, the event of changing type triggers the notification, but the notification doesn't include the previousValue corresponding to the attribute before changing type.

Example: the attribute is as follows

> type: T
> value: A

Then an update is sent with the following:

> type: T1 (it changes)
> value: A (it doesn't change)

This triggers a notification, that incluces A-T1. However, the previousValue with A-T is not included.

crbrox:
@fgalan 

close #2588


kzangeli:
LGTM

Issue #2632, opened 2016-11-01
[Bug] Context broker does not respect DateTime timezone


jmcanterafonseca:
how to reproduce:

create an entity with an attribute of type `DateTime` coded as an ISO8601 string, for instance

2016-11-01T19:00:00+01:00

Perform a query which lists the referred entity. The attribute value is now expressed incorrectly in UTC:

2016-11-01T19:00:00.00Z

fgalan:
As far as I understand by https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators, the following timezone designators should be supported

```
<time>Z
<time>±hh:mm
<time>±hhmm
<time>±hh
```

But not strings like "UTC", "CET", "PDT", etc.

Is my interpretation correct?

fgalan:
Interesting link with a similar problem: http://stackoverflow.com/questions/26895428/how-do-i-parse-an-iso-8601-date-with-optional-milliseconds-to-a-struct-tm-in-c

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2660

PR #2660, opened 2016-11-11
Timezone and partial date times support

### LINKED ISSUE #2631: If no time is specified in date filters nothing is returned
Example

curl -s -S --header 'fiware-service:AirQuality' --header 'fiware-servicepath:/Spain/Barcelona' http://130.206.83.68:1026/v2/entities/?q=dateObserved>2016-10-31

returns 0 records

however 

curl -s -S --header 'fiware-service:AirQuality' --header 'fiware-servicepath:/Spain/Barcelona' http://130.206.83.68:1026/v2/entities/?q=dateObserved>2016-10-31T0:0:0

returns a result set. 

We need to define clearly the behaviour when no time specified. Options are:

A/ Consider the whole day, so, in the example above, we will provide all the entities which have a date starting from 1st of November. 

B/ Default to, when no time is specified, 0 hours UTC of the corresponding day 

I tend to prefer option B/ as we will dealing seamlessly with use cases like `dateObserved>=2016-11-01` (all the records from the 1st of November on).

fgalan:
Fixes #2631 and #2632

### DIFF
--- a/doc/manuals/user/ngsiv2_implementation_notes.md
+++ b/doc/manuals/user/ngsiv2_implementation_notes.md
@@ -89,8 +89,9 @@ The following considerations have to be taken into account at attribute creation
   * `MM`: month (two digits)
   * `DD`: day (two digits)
 * Regarding `<time>` it must follow any of the patterns described in [the ISO8601 specification](https://en.wikipedia.org/wiki/ISO_8601#Times):
-  * `Thh:mm:ss.sss` or `Thhmmss.sss`. At the present moment, Orion is able to process times including microseconds although
-    internally they are stored as `.000`. However, this may change in the future (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2670)).
+  * `Thh:mm:ss.sss` or `Thhmmss.sss`. At the present moment, Orion is able to process times including microseconds (or even
+    smaller resolutions) although internally they are stored as `.00`. However, this may change in the future
+    (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2670)).
   * `Thh:mm:ss` or `Thhmmss`.
   * `Thh:mm` or `Thhmm`. Seconds are set to `00` in this case.
   * `Thh`. Minutes and seconds are set to `00` in this case.
@@ -104,7 +105,7 @@ The following considerations have to be taken into account at attribute creation
   However, this is ambiguous when client and server are in different zones. Thus, in order to solve this ambiguety, Orion will always
   assume timezone `Z` when timezone designator is ommited.
 
-Orion always provides datetime attributes/metadata using the format `YYYY-MM-DDThh:mm:ss.sssZ`. Note it uses UTC/Zulu
+Orion always provides datetime attributes/metadata using the format `YYYY-MM-DDThh:mm:ss.ssZ`. Note it uses UTC/Zulu
 timezone (which is the best default option, as clients/receivers may be running in any timezone). This may change in the
 future (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2663)).
 
--- a/doc/manuals/user/ngsiv2_implementation_notes.md
+++ b/doc/manuals/user/ngsiv2_implementation_notes.md
@@ -79,7 +79,7 @@ enhancements of NGSIv2 subscriptions (e.g. filtering) with NGSIv1 legacy notific
 From "Special Attribute Types" section at NGSIv2 specification:
 
 > DateTime: identifies dates, in ISO8601 format. These attributes can be used with the query operators greater-than,
-> less-than, greater-or-equal, less-or-equal and range
+> less-than, greater-or-equal, less-or-equal and range.
 
 The following considerations have to be taken into account at attribute creation/update time or when used in `q` and `mq` filters:
 
@@ -88,24 +88,25 @@ The following considerations have to be taken into account at attribute creation
   * `YYYY`: year (four digits)
   * `MM`: month (two digits)
   * `DD`: day (two digits)
-* Regarding `<time> it must follow any of the patterns described in [the ISO8601 specification](https://en.wikipedia.org/wiki/ISO_8601#Times):
+* Regarding `<time>` it must follow any of the patterns described in [the ISO8601 specification](https://en.wikipedia.org/wiki/ISO_8601#Times):
   * `Thh:mm:ss.sss` or `Thhmmss.sss`. At the present moment, Orion is able to process times including microseconds although
     internally they are stored as `.000`. However, this may change in the future (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2670)).
   * `Thh:mm:ss` or `Thhmmss`.
   * `Thh:mm` or `Thhmm`. Seconds are set to `00` in this case.
   * `Thh`. Minutes and seconds are set to `00` in this case.
   * If `<time>` is ommited, then hours, minutes and seconds are set to `00`.
-* Regarding `<timezones> it must follow any of the patterns described in [the ISO8601 specification](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators):
-  * Z
-  * ±hh:mm
-  * ±hhmm
-  * ±hh
-* ISO8601 specifies that *"if no UTC relation information is given with a time representation, the time is assumed to be in local time*.
-  However, this in ambiguous when client and server are in different zones. Thus, in order to solve this ambiguety, Orion will always
+* Regarding `<timezones>` it must follow any of the patterns described in [the ISO8601 specification](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators):
+  * `Z`
+  * `±hh:mm`
+  * `±hhmm`
+  * `±hh`
+* ISO8601 specifies that *"if no UTC relation information is given with a time representation, the time is assumed to be in local time"*.
+  However, this is ambiguous when client and server are in different zones. Thus, in order to solve this ambiguety, Orion will always
   assume timezone `Z` when timezone designator is ommited.
 
-Orion always provides datetime attributes/metadata in UTC/Zulu (which is the best default option, as clients/receivers may be running
-in any timezone). This may change in the future (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2663)).
+Orion always provides datetime attributes/metadata using the format `YYYY-MM-DDThh:mm:ss.sssZ`. Note it uses UTC/Zulu
+timezone (which is the best default option, as clients/receivers may be running in any timezone). This may change in the
+future (see [related issue](https://github.com/telefonicaid/fiware-orion/issues/2663)).
 
 [Top](#top)
 
--- a/doc/manuals/user/ngsiv2_implementation_notes.md
+++ b/doc/manuals/user/ngsiv2_implementation_notes.md
@@ -5,6 +5,7 @@
 * [Option to disable custom notifications](#option-to-disable-custom-notifications)
 * [Limit to attributes for entity location](#limit-to-attributes-for-entity-location)
 * [Legacy attribute format in notifications](#legacy-attribute-format-in-notifications)
+* [Datetime support](#datetime-support)
 * [Scope functionality](#scope-functionality)
 * [Error responses](#error-responses)
 * [Subscription payload validations](#subscription-payload-validations)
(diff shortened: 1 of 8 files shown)


test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
```diff
@@ -0,0 +1,148 @@
...
+--NAME--
+Support partial date time filters (only date, not time)
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
```
fgalan: @jmcanterafonseca please have a look to this test case in order to be sure it suffices your case.

test/functionalTest/cases/2632_timezones/timezones.test
```diff
@@ -0,0 +1,181 @@
...
+Support to timezones in DateTime attributes
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create entity with several timezones, all meaning the same date
```
fgalan: @jmcanterafonseca please have a look to this test case in order to be sure it suffices your case.

src/lib/common/globals.h
```diff
@@ -283,7 +283,7 @@ extern int64_t parse8601(const std::string& s);
 * This is common code for Duration and Throttling (at least)
 *
 */
-int64_t parse8601Time(const std::string& s);
+extern int64_t parse8601Time(const std::string& s);
```
fgalan: Not directly related with this PR, but I have take the oportunity to add `extern` to some functions that  should have it.

NTC (just informative)

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,139 @@ int64_t parse8601(const std::string& s)
...
+    return -1;
+  }
+
+  if (n == 3)
+  {
+    // Asumming 0:0:0
+    h = 0;
+    m = 0;
+    s = 0;
+    snprintf(tz, sizeof(tz), "%s", "Z");
```
fgalan: Maybe there are easier ways of filling the `char[]` with a string. Suggestions welcome.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
```diff
@@ -0,0 +1,91 @@
...
+# 02. Check that it has been created corresponding to 00:00:00 that day
+#
+
+echo "01. Creating date time using only date"
+echo "======================================"
+payload='{
+  "id": "E",
+  "type": "T",
+  "D01": {
+    "value": "2016-10-01",
```
jmcanterafonseca: are you assuming  that a timeless date string refers to 0 hours UTC? A timezoneless datestring refers to the local time ...

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
```diff
@@ -0,0 +1,91 @@
...
+Content-Length: 95
+Content-Type: application/json
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+{
+    "D01": {
+        "metadata": {},
+        "type": "DateTime",
+        "value": "2016-10-01T00:00:00.00Z"
```
jmcanterafonseca: Z is incorrect as per my observation above. If I create a timezoneless date string it should still refer to local time.

test/functionalTest/cases/2632_timezones/timezones.test
```diff
@@ -0,0 +1,181 @@
...
+    },
+    "D10": {
+        "metadata": {},
+        "type": "DateTime",
+        "value": "2016-10-01T01:23:45.00Z"
+    },
+    "D11": {
+        "metadata": {},
+        "type": "DateTime",
+        "value": "2016-10-01T01:23:45.00Z"
```
jmcanterafonseca: From this test case I can see that date string will always be listed referred to UTC, right?

jmcanterafonseca:
are partial times supported as well, for instance:

2016-10-01T11:00 --> 2016-10-01T11:00:00 

or

2016-10-01T11 --> 2016-10-01T11:00:00

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
fgalan: You mean the right string should be `2016-10-01Z`? (I didn't see mentions to timezone at https://github.com/telefonicaid/fiware-orion/issues/2631 so maybe I wrongly interpret this was correct).

In addition, should timezones in general be supported in partial representations? I mean, things like:

2016-10-01+01:30
2016-10-01-02
etc.

test/functionalTest/cases/2632_timezones/timezones.test
fgalan: That's correct. I think the following fragment at http://stackoverflow.com/questions/40213520/orion-context-broker-timezone describes it pretty well:

> EDIT: Orion timezone support will improve in version 1.6.0. That version will support timezone interpretation at creation/update time (so you can use for example `2017-06-17T07:21:24.238Z` and `2017-06-17T08:21:24.238+0100` meaning the same). However, in order to set a common ground for clients accesing to the same entities from different timezones (as @LeonanCarvalho mentions) date time retrieval is in UTC/Zulu, no matter how it was stored.

fgalan:
Regarding partial times, i.e. including only hour+minutes orr only hour, they are not supported. Only partial dates are supported.

Should partial times hour+minutes or hour to be supported? Are the allowed by the ISO8601 standard? Have we relevant use cases behind it?

jmcanterafonseca:
According to

https://en.wikipedia.org/wiki/ISO_8601#Times it is perfectly possible to omit minutes or seconds. I think this would make developer's life easier specially when dealing with events which happen periodically, particularly on an hour basis, for instance air quality measurements in cities. In that case it would suffice to say give me what happen at 11 and omit the minute and second component

test/functionalTest/cases/2632_timezones/timezones.test
jmcanterafonseca: Unfortunately there is no yet a standard way to specify requested timezone in HTTP requests, although some attempts have been made for instance

https://tools.ietf.org/html/draft-sharhalakis-httptz-05

so providing datestrings referred to UTC could work.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
jmcanterafonseca: To sum up:

Timezoneless datestrings refer to a local time, what we cannot assume to be Zulu. 

So my proposal is not to accept date strings without the timezone component to avoid errors. 

What do you think?

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
jmcanterafonseca: As per my comment below, a timeless datestring will be assumed to be 00:00:00 Zulu time, so this would be correct.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
```diff
@@ -0,0 +1,148 @@
...
+echo "03. Get entities after 2016-10-01, get E2"
+echo "===========================0============="
+orionCurl --url '/v2/entities?q=D>2016-10-01'
+echo
+echo
+
+
+echo "04. Get entities before 2016-10-01, get E1"
+echo "=========================================="
+orionCurl --url '/v2/entities?q=D<2016-10-01'
```
jmcanterafonseca: can we add tests that exercise times as well and particularly times which include timezone offsets?

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
fgalan: >is not to accept date strings without the timezone component to avoid errors

I agree.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
fgalan: Not sure what you mean... could you provide an example, please?

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
fgalan: After analyzing this deeper, I have changed my mind... I cannot agree on "not to accept date strings without the timezone componente" because doing so will break backward compatibility.

Currently (i.e. in current master before merging this PR) we are using the following filter to process DateTime at creation/update time (https://github.com/telefonicaid/fiware-orion/blob/b445e84ec05046bbf83e2e366056c86583c38dae/src/lib/common/globals.cpp#L441):

```
p = strptime (s.c_str(), " %Y-%m-%dT%T", &tm);
```

Thus, currently strings like `2016-11-30T10:11:12` *which doesn't include timezone* are supported. And, given that we are forced to support it, the most clever options seems to set default to UTC. This is described in documentation in this commit added to the PRhttps://github.com/telefonicaid/fiware-orion/pull/2660/commits/bb83cb66063c27121cf28dc83ff6c2dbfa6c2ddb:

> ISO8601 specifies that *"if no UTC relation information is given with a time representation, the time is assumed to be in local time"*. However, this in ambiguous when client and server are in different zones. Thus, in order to solve this ambiguety, Orion will always assume timezone `Z` when timezone designator is ommited.

You could argue that although we are forced to support timezone in `YYYY-MM-DDThh:dd` pattern we are not forced to support it in others such as `YYYY-MM-DDThhddss` or `YYYY-MM-DD`. However, doing that (i.e. supporting default timezone in some patterns while not in others) would be very confusing.

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -73,6 +74,41 @@ enhancements of NGSIv2 subscriptions (e.g. filtering) with NGSIv1 legacy notific
 
 [Top](#top)
 
+## Datetime support
```
fgalan: @iariasleon pay attention to this.

In the past, we were discussing that the particular implementation supports of ISO8601 was not clear and that situation was precluding the implementation of test. This documentation should be precise enough (so the impediment to develop new tests disappear) but if you find it not clear, please raise a comment.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
jmcanterafonseca: queries on which the datestring includes a timezone offset

test/functionalTest/cases/2632_timezones/timezones.test
jmcanterafonseca: that's interesting information to be added to https://github.com/telefonicaid/fiware-orion/issues/2663 @LeonanCarvalho

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
jmcanterafonseca: No problem, I think this is aligned with what is suggested in https://github.com/telefonicaid/fiware-orion/issues/2663

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
fgalan: Review done (see other comments). NTC

test/functionalTest/cases/2632_timezones/timezones.test
fgalan: Review done (see other comments). NTC.

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
fgalan: Fixed in 91e0fc4

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_filter.test
fgalan: Fixed in 91e0fc4, with timezones_filters.test file including these cases.

doc/manuals/user/ngsiv2_implementation_notes.md
iariasleon: LGTM

test/functionalTest/cases/2631_partial_datetimes/partial_datetime_creation.test
```diff
@@ -0,0 +1,226 @@
...
+echo "=================================================================================================="
+payload='{
+  "id": "E",
+  "type": "T",
+  "D01": {
+    "value": "2016-10-01",
+    "type": "DateTime"
+  },
+  "D02": {
+    "value": "2016-10-01Z",
```
jmcanterafonseca: not sure if timezone can appear without time component ...

fgalan: I understand that the concern if is "2016-10-01Z", "2016-10-01+01:00", etc. are legal or if they should use also the time part.

ISO8601 says: "If a time zone designator is required, it follows the combined date and time". My interpretation is that "2016-10-01Z" is *not* correct.

Morevoer, taking into account from https://en.wikipedia.org/wiki/ISO_8601#Times that the minimal time expression is `hh` (i.e. no "null" time expressions are allowed) them "2016-10-01TZ", "2016-10-01T+01:00" are also incorrect.

Thus, the "minimal" date expression with timezone is "2016-10-01T00Z", "2016-10-01T00+01:00".

Related with this (but applies to the whole implementation), ISO8601 also says "It is permitted to omit the 'T' character by mutual agreement". However, given that in general Orion and the client are unknown to each other (worst case assumption) I think we cannot omit the T.

fgalan: Fixed in ce2c14c5bdc7b8d99ec75f880d2550879d513261

test/functionalTest/cases/2632_timezones/timezones.test
fgalan: I have moved @LeonanCarvalho 's feedback to #2663, to have it at hand after this PR merge.

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,162 @@ int64_t parse8601(const std::string& s)
...
+  }
+
+  if (positive)
+  {
+    return offset;
+  }
+  else
+  {
+    return -offset;
+  }
```
kzangeli: I would suggest a one-liner here:

    return positive? offset : -offset;

Or, make positive an int, with value 1 or -1 and call it 'sign'

    int sign = (tz[0] == '+')? 1 : -1;
￼
and then return like this:

    return sign * offset;

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,162 @@ int64_t parse8601(const std::string& s)
...
 */
-int64_t parse8601Time(const std::string& s)
+int64_t parse8601Time(const std::string& ss)
 {
-  struct tm   tm = {0};
-  const char* p;
+  int y, M, d, h, m;
+  char tz[10];
+  float s;
+
```
kzangeli: Our style guide says: one variable per line.
And, horizontal indents not to be forgotten.

Also (just a matter of taste), why not initialize h, m and s to zero to avoid doing that in each `if`  ?
Lots of lines to be saved there. The entire if/else if/ ,,, could become a if (() || () ||() ...

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,162 @@ int64_t parse8601(const std::string& s)
...
+  // Lenght check, to avoid buffer overflow in tz[]. Calculation is as follows:
```
kzangeli: Typo: Length

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,162 @@ int64_t parse8601(const std::string& s)
...
   }
-  return (int64_t) timegm(&tm);
+
+  int offset = timezoneOffset(tz);
+  if (offset == -1)
+  {
+    return -1;
+  }
+
+  // Note that at the present moment we are not doing anything with mileseconds, but
```
kzangeli: Typo: milliseconds

src/lib/common/globals.cpp
fgalan: Fixed in 438ba016bc6ecd2aa80ab2a6dd8c6a098c583b15

src/lib/common/globals.cpp
```diff
@@ -426,24 +427,162 @@ int64_t parse8601(const std::string& s)
...
+    offset = h * 60 * 60 + m * 60;
+  }
+  else if (sscanf(tz + 1, "%2d", &h) == 1)         // Trying the <time>±hh format
+  {
+    offset = h * 60 * 60;
+  }
+
+  if (offset == -1)
+  {
+    // wrong timezone
```
fgalan: Fixed in 438ba016bc6ecd2aa80ab2a6dd8c6a098c583b15

src/lib/common/globals.cpp
fgalan: Fixed in 438ba016bc6ecd2aa80ab2a6dd8c6a098c583b15

src/lib/common/globals.cpp
fgalan: Fixed in 438ba016bc6ecd2aa80ab2a6dd8c6a098c583b15

kzangeli:
LGTM

Issue #2676, opened 2016-11-15
in POST v2/op/update (batch op) ID validations fail for md_type


fgalan:
Similar to #2673

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2679.

iariasleon:
LGTM

Issue #2673, opened 2016-11-15
in POST v2/op/update (batch op) ID validations fail for attr_name


iariasleon:
## Case 1 - not plain ascii chars are allowed in attribute name
in POST v2/op/update (batch op) not plain ascii chars are allowed in attribute name. See similar issues: #2658, #2664.

#### Dataset used
```
      | entities_id |
      |-------------|
      | habitación  |
      | españa      |
      | barça       |
```
#### update batch operation request
```
url: POST http://localhost:1026/v2/op/update
headers:
    Content-Type: application/json
    Fiware-Service: test_op_update_entitiies_type_error
    Fiware-ServicePath: /test
payload:  {"entities": [{"barça": {"value": "34", "metadata": {"warning": {"type": "alarm", "value": "hot"}}}, "type": "house", "id": "room1"}], "actionType": "APPEND"}
```

#### Response
```
http code: 204 - No Content
headers:
   Connection: Keep-Alive
   Content-Length: 0
   Fiware-Correlator: a4910862-ab21-11e6-a0c8-005056a20feb
   Date: Tue, 15 Nov 2016 10:52:46 GMT
payload:
```

#### Expected response
```
http code: 400 - Bad Request
{"error":"BadRequest","description":"Invalid characters in attribute name"}
```

#### doc in mongo
```
{ "_id" : { "id" : "room1", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "habitación", "españa", "barça" ], "attrs" : { "habitación" : { "type" : "Text", "creDate" : 1479207166, "modDate" : 1479207166, "value" : "34", "md" : { "warning" : { "type" : "alarm", "value" : "hot" } }, "mdNames" : [ "warning" ] }, "españa" : { "value" : "34", "type" : "Text", "md" : { "warning" : { "type" : "alarm", "value" : "hot" } }, "mdNames" : [ "warning" ], "creDate" : 1479207166, "modDate" : 1479207166 }, "barça" : { "value" : "34", "type" : "Text", "md" : { "warning" : { "type" : "alarm", "value" : "hot" } }, "mdNames" : [ "warning" ], "creDate" : 1479207166, "modDate" : 1479207166 } }, "creDate" : 1479207166, "modDate" : 1479207166 }
```

## Case 2 - length longer than 256 chars in attribute name
in POST v2/op/update (batch op) length longer than 256 chars in attribute name. See similar issues: #2658, #2664.

#### update batch operation request
```
[...]

Issue #2666, opened 2016-11-14
in POST v2/op/update forbidden chars are allowed in the "type" field


iariasleon:
in POST v2/op/update forbidden chars are allowed in the "type" field. See similar issue: #2661.

#### Dataset used
```
      | entities_type |
      |---------------|
      | house<flat>   |
      | house=flat    |
      | house"flat"   |
      | house'flat'   |
      | house;flat    |
      | house(flat)   |
      | house_?       |
      | house_&       |
      | house_/       |
      | house_#       |
```

#### Update batch operation request
```
url: POST http://qa-orion-fe-02:1026/v2/op/update
headers:
    Content-Type: application/json
    Fiware-Service: test_op_update_entities_type_error
    Fiware-ServicePath: /test
payload: {"entities": [{"type": "house<flat>", "temperature": {"type": "celsius", "value": "4", "metadata": {"alarm": {"type": "hot", "value": "warning"}}}, "id": "room_5"}], "actionType": "APPEND"}
```

#### Update batch operation response
```
http code: 204 - No Content
headers:
   Connection: Keep-Alive
   Content-Length: 0
   Fiware-Correlator: 73cd9c3c-aa58-11e6-89e6-005056a20feb
   Date: Mon, 14 Nov 2016 10:52:35 GMT
payload:
```

#### Expected response
```
http code: 400 - Bad Request
{"error":"BadRequest","description":"Invalid characters in entity type"}
```

#### docs in mongo
```
{ "_id" : { "id" : "room_5", "type" : "house<flat>", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "celsius", "creDate" : 1479120755, "modDate" : 1479120755, "value" : "4", "md" : { "alarm" : { "type" : "hot", "value" : "warning" } }, "mdNames" : [ "alarm" ] } }, "creDate" : 1479120755, "modDate" : 1479120755 }
{ "_id" : { "id" : "room_5", "type" : "house=flat", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "celsius", "creDate" : 1479120755, "modDate" : 1479120755, "value" : "4", "md" : { "alarm" : { "type" : "hot", "value" : "warning" } }, "mdNames" : [ "alarm" ] } }, "creDate" : 1479120755, "modDate" : 1479120755 }
[...]

Issue #2665, opened 2016-11-14
in POST v2/op/update (batch op) entity type allows length longer than 256 chars and length equals to zero


iariasleon:
## Case 1:  length longer than 256 chars

in POST v2/op/update (batch op) entity id allows length longer than 256 chars. See similar issue: #2659.

#### update bacth operation request
```
url: POST http://localhost:1026/v2/op/update
headers:
    Content-Type: application/json
    Fiware-Service: test_op_update_entities_type_error
    Fiware-ServicePath: /test
payload: {"entities": [{"type": "zzSzwDmHxb6FqW1C159MhUvlBFH3mbVhoJ8c7NUzJjBL58smi9cI0Q0HmIEhFxeb3QNRXdJ8uWzu9SGW6SFmmV7WnIXjp76wITnecGAj2NUN0fAK4Io8wvD1kvvWL4KNa2RrF3ayNUGt6aLWvLKaiaML6fF8CEkbGmNtdIWslOUfE7bdc6IiZeYRID5mNZkAD5ssxDmvDfjBZsFscfqvpw3KmCCwUq14zJucjmKSG8ZviFV1FNH46MqlHBk0G6TuN", "temperature": {"type": "celsius", "value": "4", "metadata": {"alarm": {"type": "hot", "value": "warning"}}}, "id": "room_3"}], "actionType": "APPEND"}

```

#### response
```
http code: 204 - No Content
headers:
   Connection: Keep-Alive
   Content-Length: 0
   Fiware-Correlator: efb46f44-aa56-11e6-ba14-005056a20feb
   Date: Mon, 14 Nov 2016 10:41:44 GMT
payload:
```

#### Expected response
```
http code: 400 - Bad Request
{"error":"BadRequest","description":"entity type length: 257, max length supported: 256"}
```

#### doc in mongo
```
[...]

PR #2679, opened 2016-11-15
Several validation bugfixes for the /v2/op/update

### LINKED ISSUE #2653: in POST v2/op/update, if we use an unknown action the op is successful
in POST v2/op/update, if we use an unknown action `"actionType": "dfgdf"` the op is successfully

#### update bach operation request
```
POST localhost:1026/v2/op/update
headers:
    Fiware-Service: test_op_update_happy_path
    Fiware-ServicePath: /test
payload: {"entities": [{"type": "house", "temperature": {"value": "34"}, "id": "room1_0"}, {"type": "house", "temperature": {"value": "34"}, "id": "room1_1"}, {"type": "house", "temperature": {"value": "34"}, "id": "room1_2"}, {"type": "house", "temperature": {"value": "34"}, "id": "room1_3"}, {"type": "house", "temperature": {"value": "34"}, "id": "room1_4"}], "actionType": "dfgdf"}
```

#### update bach operation response
```
http code: 204 - No Content
headers:
   Connection: Keep-Alive
   Content-Length: 0
   Fiware-Correlator: 84769862-a766-11e6-98a0-005056a20feb
   Date: Thu, 10 Nov 2016 16:55:43 GMT
payload:
```

#### docs in mongo
```
{ "_id" : { "id" : "room1_0", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "Text", "creDate" : 1478797236, "modDate" : 1478797236, "value" : "34", "mdNames" : [ ] } }, "creDate" : 1478797236, "modDate" : 1478797236 }
{ "_id" : { "id" : "room1_1", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "Text", "creDate" : 1478797236, "modDate" : 1478797236, "value" : "34", "mdNames" : [ ] } }, "creDate" : 1478797236, "modDate" : 1478797236 }
{ "_id" : { "id" : "room1_2", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "Text", "creDate" : 1478797236, "modDate" : 1478797236, "value" : "34", "mdNames" : [ ] } }, "creDate" : 1478797236, "modDate" : 1478797236 }
{ "_id" : { "id" : "room1_3", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "Text", "creDate" : 1478797236, "modDate" : 1478797236, "value" : "34", "mdNames" : [ ] } }, "creDate" : 1478797236, "modDate" : 1478797236 }
{ "_id" : { "id" : "room1_4", "type" : "house", "servicePath" : "/test" }, "attrNames" : [ "temperature" ], "attrs" : { "temperature" : { "type" : "Text", "creDate" : 1478797236, "modDate" : 1478797236, "value" : "34", "mdNames" : [ ] } }, "creDate" : 1478797236, "modDate" : 1478797236 }

```

fgalan:
Fixes #2653, #2658, #2659, #2661, #2664, #2665, #2666, #2673, #2674, #2675, #2676

*Update*: it happens that this PR also solves #2681 and #2682, although the .test covering them will be added in next PR in order not to make this more complex now that we are near of final LGTM.


src/lib/apiTypesV2/BatchQuery.cpp
```diff
@@ -53,38 +53,6 @@ BatchQuery::~BatchQuery()
 
 /* ****************************************************************************
 *
-* BatchQuery::check - 
-*/
-std::string BatchQuery::check(ConnectionInfo* ciP, RequestType requestType)
```
fgalan: Probably this check() method (and a similar one in BatchUpdate) was designed to to the whole validation in one shot... However, by the way the validation code has evolved this is not actually used (maybe it was never used?).

Thus, let's remove now. If in the future we want doing this in this way, it makes more sense to re-implement at that moment that keep a dead piece of code which original purpose has lost in the mist of the past.

src/lib/apiTypesV2/BatchQuery.h
```diff
@@ -51,7 +51,7 @@ class BatchQuery
   BatchQuery();
   ~BatchQuery();
 
-  std::string    check(ConnectionInfo* ciP, RequestType requestType);
+  //std::string    check(ConnectionInfo* ciP, RequestType requestType);
```
iariasleon: this commented line is leftover?

src/lib/ngsi/UpdateActionType.cpp
```diff
@@ -38,13 +38,7 @@
 *
 * UpdateActionType::check - 
 */
-std::string UpdateActionType::check
```
fgalan: Four parameters not actually needed and making the function call complex. Removing.

test/functionalTest/cases/0970_create_entity/errors_at_creation.test
```diff
@@ -235,13 +235,13 @@ Date: REGEX(.*)
...
 HTTP/1.1 400 Bad Request
-Content-Length: 69
+Content-Length: 88
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
-    "description": "no 'name' for ContextAttribute", 
+    "description": "attribute name length: 0, min length supported: 1",
```
fgalan: Changes in description are minor things a not considered as a breaking in backward compatibility.

test/functionalTest/cases/1601_restrict_chars_in_ids/restrict_len_ids.test
```diff
@@ -51,7 +51,7 @@ echo "========================================="
 payload='{
   "id": "VERY_LONG",
   "type": "T1",
-  "España": {
+  "A": {
```
fgalan: After this PR, the check about forbidden chars in attributes was done before than the one in entity id length covered by this .test. Thus I have changed the attribute name to keep the .test result the same.

As long as the code change only involves a change in the order in which error conditions are raised and that doesn't have an impact in the response (appart from description, i.e. both are 400 Bad Request) I think this doesn't break backward compatibility.

However, I have preferred to describe the case in this comment in the case somebody have a different opinion :)

src/lib/apiTypesV2/BatchQuery.h
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/pull/2679/commits/7bcc3414bae24dbce4d999882838acc219088515

src/lib/jsonParseV2/parseBatchUpdate.cpp
```diff
@@ -113,7 +113,19 @@ std::string parseBatchUpdate(ConnectionInfo* ciP, BatchUpdate* burP)
     }
     else if (name == "actionType")
     {
-      burP->updateActionType.set(iter->value.GetString());
+      UpdateActionType uat;
+      uat.set(iter->value.GetString());
+
```
kzangeli: Why create an additional vartiable of type `UpdateActionType` ?
`burP->updateActionType` could be used instead.
Less code and slightly faster.

Also, the set() method of UpdateActionType could call its check() method and return false if check() fails:

    if (burP->updateActionType.set(iter->value.GetString(), &err) == false)
    {
      alarmMgr.badInput(clientIp, err);
      ...
    }

src/lib/jsonParseV2/parseContextAttribute.cpp
```diff
@@ -272,16 +273,19 @@ std::string parseContextAttribute(ConnectionInfo* ciP, const Value::ConstMemberI
...
   {
-    alarmMgr.badInput(clientIp, "no 'name' for ContextAttribute");
-    ciP->httpStatusCode = SccBadRequest;
-    return "no 'name' for ContextAttribute";
+    caP->type = (compoundVector)? defaultType(orion::ValueTypeVector) : defaultType(caP->valueType);
   }
 
-  if (!caP->typeGiven)
+  // FIXME P2: weird... one argument is a sub-argument of the other. ciP should be expanded to the minimum set
+  // of needed arguments
```
kzangeli: +1

src/lib/jsonParseV2/parseEntityVector.cpp
```diff
@@ -53,11 +53,16 @@ std::string parseEntityVector(ConnectionInfo* ciP, const Value::ConstMemberItera
...
     std::string  r;
     Entity*      eP = new Entity();
 
     evP->vec.push_back(eP);
 
+    // FIXME PR: is the logic of this function correct? I mean, parseEntityObject() could result in error, so the push_back maybe
+    // should be done after the parseEntityObject() call (and delete the eP just before return r). Or maybe it doesn't matter... as in the
+    // case of error the error is propagated back to result in an OrionError and the evP is not actually used (and properly destroyed during
+    // the release step)
     r = parseEntityObject(ciP, iter2, eP, attributesAllowed);
```
kzangeli: I imagine the push is done before so that evP->release() is able to delete the allocated Entity.
Apart from that, yes, it would be better not to push (and to delete 'explicitly') in case of error.

Can't say I care too much ...

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -967,6 +967,13 @@ std::string ContextAttribute::check
   size_t len;
   char errorMsg[128];
 
+  if (((ciP->apiVersion == "v2") && (len = strlen(name.c_str())) < MIN_ID_LEN) && (requestType != EntityAttributeValueRequest))
+  {
+    snprintf(errorMsg, sizeof errorMsg, "attribute name length: %zd, min length supported: %d", len, MIN_ID_LEN);
+    alarmMgr.badInput(clientIp, errorMsg);
+    return std::string(errorMsg);
+  }
```
kzangeli: Is this check not implemented elsewhere?
If it is, that part should be removed now that the check is done here 'instead'.

test/functionalTest/cases/0970_create_entity/errors_at_creation.test
kzangeli: NTC

test/functionalTest/cases/1601_restrict_chars_in_ids/restrict_len_ids.test
kzangeli: NTC

src/lib/ngsi/ContextAttribute.cpp
fgalan: Looking for EMPTY_ATTR_NAME and "attribute name length" it seems it is not implemented in any other place, so I think is ok as it is now... or my search heuristic failed :)

src/lib/jsonParseV2/parseBatchUpdate.cpp
fgalan: Good point. Fixed in 32bf17d (without modiying set()... I prefer not to change a method that others points of the code may be using).

src/lib/ngsi/ContextAttribute.cpp
fgalan: No occurrence of `no 'name' for ContextAttribute` in the PRs branch (in code, documentation or tests).

kzangeli: Strange ... the modified functest had that string. Perhaps 'name' is inserted via `%s` or streams or something ...

New search string: `' for ContextAttribute` and after that we might as well give up :-)

src/lib/apiTypesV2/BatchQuery.cpp
kzangeli: It was probably never used, which is a bit funny.
The check methods form a tree and a father calls the check methods of its 'sons'.
This will not change.
Perhaps/Probably all the fixes would be covered by simply calling the check method ...

Is there any check whatsoever on scopes, in the current code?

For background, in v1 parsing, all check methods are called after parsing the payload, in one unique place in the source code. V2 parsing has no unique point and I believe we've lost the use of the check methods.
We should try to get that very important mechanism for v2 as well.

I would not remove these methods but instead try to call them, after parsing the incoming payload.

src/lib/jsonParseV2/parseContextAttribute.cpp
kzangeli: ok,

NTC

src/lib/apiTypesV2/BatchQuery.h
iariasleon: LGTM

src/lib/apiTypesV2/BatchQuery.cpp
fgalan: This needs a careful examination... I have the feeling that righ now most of the checking is done "inline" as the parsing is done. In that case, the strategy of "parsing all, them check" would not easy to apply as it would require big changes in the code. And actually I'm not sure if it is the best one... I mean, maybe is better to early detect error conditions with inline checking.

Note the *maybe* :) I mean, I'm not sure which option is better. What I'm sure is that we would need time to discuss so I'm leave that appart from that PR.

Thus, in sum, what I can do is to keep the code of these methods under `#if 0` instead of removing them, with a FIXME comment about "a discussion is pending if we should implement a check-as-we-parse or parse-all-check-after strategy in NGSIv2 parsing"

(Note the "in NGSIv2 parsing". NGSIv1 parsing doen't deserves changes in the strategy or big changes, from my point of view)

src/lib/ngsi/ContextAttribute.cpp
fgalan: > Strange ... the modified functest had that string. Perhaps 'name' is inserted via %s or streams or something ...

Correctly, `had`. Note the "in the PR branch". I mean, after the changes I did in .test and code, no occurrences.

Let's see now for the other pattern you suggest...

fgalan: This ones has been captured with the last patterns:

```
LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
alarmMgr.badInput(clientIp, "bad type for ContextAttribute");
return "invalid JSON type for ContextAttribute";
alarmMgr.badInput(clientIp, "no 'value' for ContextAttribute without keyValues");
return "no 'value' for ContextAttribute without keyValues";
```

kzangeli: So, after some greps and skype conversation we've seen that the old texts `no 'name' for ContextAttribute` were removed in this very PR.

NTC

src/lib/apiTypesV2/BatchQuery.cpp
kzangeli: Yeah, I thought about  this a little bit also ...
In V2 parsing all requests that need to parse i.e. an attribute will use one unique function for that, same same with metadatas etc. And, all checks are done within that function.

Conclusion, no need for the check() methods in V2 parsing. 
So, those check methods that are not used by V1 parsing **should be removed**, just as is in this PR.

NTC

src/lib/jsonParseV2/parseEntityVector.cpp
fgalan: Comment changed in PR 35fa9a3.

crbrox: It looks good to me, a simple comment remarking "it doesn't matter" could be enough. In fact when one entity fails to be parsed, it can be N previous parsed entities in the vector in that moment and they will be deleted when the error is returned and the whole vector is discarded (or so I think). So an empty entity does not make bad, al most a little performance improvement

crbrox:
LGTM

kzangeli:
LGTM

PR #2694, opened 2016-11-18
Refactor rendering logic

fgalan:
This PR includes several simplifications:

* Unify startTag() and startTag1() methods into only one
* Unify valueTag(), valueTag1() and valueTag2() into only one
* Simplifying argument in suviving startTag() and valueTag() and also in endTag()
* Remove unneeded tag/key management (a leftover that makes more sense in the times of XML)

*Edit*: again, important to note: this PR doesn't touch any .test, so 100% regresion is kept with this refactor.

### DIFF
--- a/src/lib/common/tag.cpp
+++ b/src/lib/common/tag.cpp
@@ -269,6 +269,8 @@ std::string endTag
 *
 * valueTag -  
 *
+* Function version for string values
+*
 */
 std::string valueTag
 (
@@ -335,7 +337,10 @@ std::string valueTag
 
 /* ****************************************************************************
 *
-* valueTag -  
+* valueTag -
+*
+* Function version for integer values
+*
 */
 std::string valueTag
 (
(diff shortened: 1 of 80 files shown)


src/lib/common/tag.cpp
```diff
@@ -208,48 +208,20 @@ std::string jsonInvalidCharsTransformation(const std::string& input)
...
-    if (showKey == false)
-    {
-      return indent + "{\n" + indent + "  {\n";
-    }
- 
-    return indent + "{\n" + indent + "  " + "\"" + key + "\" : {\n";
-  }
+  // Empty key is legal JSON. However, doesn't use that kind of JSON in Orion,
+  // so we can this instead of a showkey boolean parameter, keeping the function
+  // signature simpler
```
kzangeli: This comment is strange. Words are missing ...
My guess:

    //
    // Empty key is legal JSON. However, Orion doesn't use that 'feature',
    // so instead of a showkey boolean parameter we check the JSON key to be empty
    // to decide whether the key should be printed or not.
    // This makes the function signature a bit lighter
    //

src/lib/common/tag.cpp
```diff
@@ -208,48 +208,20 @@ std::string jsonInvalidCharsTransformation(const std::string& input)
...
 *
-* startTag1 -
+* startTag -
 */
-std::string startTag1
+std::string startTag
 (
   const std::string&  indent,
   const std::string&  key,
-  bool                showKey,
```
kzangeli: About replacing 'bool showKey' with `key == ""` ...
I imagine showKey is FALSE if within vector and while it may be true that orion forces keys to be not empty, what about compound values?
Would be good to have a functest with compound values testing this.

src/lib/jsonParseV2/parseEntity.cpp
```diff
@@ -184,30 +177,29 @@ std::string parseEntity(ConnectionInfo* ciP, Entity* eP, bool eidInURL)
...
+        OrionError oe(SccBadRequest, r, "BadRequest");
+        return oe.toJson();
       }
     }
   }
 
   if (membersFound == 0)
   {
-    eP->oe.fill(SccBadRequest, "empty payload", "BadRequest");
+    alarmMgr.badInput(clientIp, "empty payload");
```
kzangeli: Possible 'overlogging'

kzangeli:
Nice cleanup!

src/lib/jsonParseV2/parseEntity.cpp
fgalan: I have had a look to this part of code but I don't find the log logic that overlaps with this...

src/lib/common/tag.cpp
fgalan: Fixed in b2d3d6b58232c5ad869e697f761aeec1af35b98b

src/lib/common/tag.cpp
kzangeli: Perfect. All tested

kzangeli:
LGTM

PR #2697, opened 2016-11-21
Error string extraction

fgalan:
Note that as cited in the code itself (https://github.com/telefonicaid/fiware-orion/compare/hardening/error_string?expand=1#diff-c2640173c861df9babf636b17a6b3755R40) this PR doesn't do an exhaustive extraction. However, it puts some order in some existing and quite common error situations, as you can see in the .test modifications (only in error description texts).

### DIFF
--- a/test/functionalTest/cases/0970_create_entity/errors_at_creation.test
+++ b/test/functionalTest/cases/0970_create_entity/errors_at_creation.test
@@ -151,13 +151,13 @@ echo
 01. POST /v2/entities without Entity::id
 ========================================
 HTTP/1.1 400 Bad Request
-Content-Length: 61
+Content-Length: 83
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
-    "description": "no entity id specified", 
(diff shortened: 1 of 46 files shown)


test/functionalTest/cases/0970_create_entity/errors_at_creation.test
```diff
@@ -151,27 +151,27 @@ echo
...
 HTTP/1.1 400 Bad Request
-Content-Length: 61
+Content-Length: 83
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
-    "description": "no entity id specified", 
+    "description": "entity id length: 0, min length supported: 1",
```
kzangeli: Not that it is **really** important but here we lose some information.
Non-present and present-but-empty now give the same error description.

crbrox:
LGTM

test/functionalTest/cases/0970_create_entity/errors_at_creation.test
kzangeli: Sure, NTC

kzangeli:
LGTM

PR #2699, opened 2016-11-22
apiVersion from std::string to int

fgalan:
Apart from the apiVersion from std::string to int refactor, some other minor things has been included (e.g. removing old FIXME PR leftover code, remove unneded parameters in some functions, simplyfing arguments with asJsonObject, etc).

### DIFF
--- a/src/lib/parseArgs/parseTest.cpp
+++ b/src/lib/parseArgs/parseTest.cpp
@@ -104,7 +104,7 @@ PaArgument paArgs[] =
   { "-f1",      &f1,     "F",      PaFloat,  PaOpt, _i 3.1,   0,       100,     "float"               },
   { "-d1",      &d1,     "D",      PaDouble, PaOpt, _i 3.3,   0,       100,     "double"              },
   { "--lmTest", &lmTest, "",       PaBool,   PaOpt, FALSE,    FALSE,   TRUE,    "test logMsg"         },
-  { NULL,       &v2,     "V2",     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
+  { NULL,       &v2,     2,     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
 
   PA_END_OF_ARGS
 };
--- a/src/lib/parseArgs/parseTest.cpp
+++ b/src/lib/parseArgs/parseTest.cpp
@@ -104,7 +104,7 @@ PaArgument paArgs[] =
   { "-f1",      &f1,     "F",      PaFloat,  PaOpt, _i 3.1,   0,       100,     "float"               },
   { "-d1",      &d1,     "D",      PaDouble, PaOpt, _i 3.3,   0,       100,     "double"              },
   { "--lmTest", &lmTest, "",       PaBool,   PaOpt, FALSE,    FALSE,   TRUE,    "test logMsg"         },
-  { NULL,       &v2,     2,     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
+  { NULL,       &v2,     "v2",     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
 
   PA_END_OF_ARGS
 };
--- a/src/lib/rest/RestService.cpp
+++ b/src/lib/rest/RestService.cpp
@@ -488,19 +488,6 @@ std::string restService(ConnectionInfo* ciP, RestService* serviceV)
     }
   }
 
-
-  //
-  // Detect API version
-  // FIXME PR: we have a method for this...
-  //
-  int apiVersion = 1;
-
-  if ((compV.size() != 0) && strcasecmp(compV[0].c_str(), "v2") == 0)
-  {
-    apiVersion = 2;
-  }
-
-
   //
   // Lookup the requested service
   //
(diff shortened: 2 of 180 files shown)


fgalan:
As others PRs in the "check/renders refactor", no .test has been touched, ensuring 100% functional regression.

src/lib/rest/RestService.cpp
```diff
@@ -443,7 +444,7 @@ static bool compErrorDetect
 * we need to know if it's v1 or v2.
 *
 * This is very easy, we just look at the first component of the URI PATH, and
-* if it is "v2" or "V2", then case-sensitive checks are used.
+* if it is 2 or 2, then case-sensitive checks are used.
```
kzangeli: This string must change, but it must also be thoroughly read to understand what has happened here ...

src/lib/rest/RestService.cpp
```diff
@@ -375,14 +375,15 @@ static bool compCheck(int components, const std::vector<std::string>& compV)
...
 (
+  int                              apiVersion,
   int                              components,
   const std::vector<std::string>&  compV,
   OrionError*                      oeP
 )
 {
   std::string  details; 
 
-  if ((compV[0] == "v2") && (compV[1] == "entities"))
```
kzangeli: Not sure this line should be modified.
Perhaps, just not sure

src/lib/rest/RestService.cpp
```diff
@@ -487,18 +488,6 @@ std::string restService(ConnectionInfo* ciP, RestService* serviceV)
...
-  //
-  std::string apiVersion = "v1";
-
-  if ((compV.size() != 0) && strcasecmp(compV[0].c_str(), "v2") == 0)
-  {
-    apiVersion = "v2";
-  }
-
-
   //
```
kzangeli: Where is the API version detected in the new implementation?

src/lib/rest/rest.cpp
```diff
@@ -906,32 +906,33 @@ bool urlCheck(ConnectionInfo* ciP, const std::string& url)
...
-  if (strcmp(path, "/admin/log") == 0)
+  if (strncmp("/admin", path, strlen("/admin")) == 0)
   {
-    return "admin";
+    return 0;
   }
 
-  return "v1";
+  return -1;
 }
```
kzangeli: Here I think we need to take a look at contextBroker.cpp to see the list of all requests the broker currently supports.
And then decide what to do with all of them.

What do we do with /statistics?  v1 ?
And /v1/admin ...  apiVersion 0 ?

etc.

src/lib/parseArgs/parseTest.cpp
```diff
@@ -104,7 +104,7 @@ PaArgument paArgs[] =
   { "-f1",      &f1,     "F",      PaFloat,  PaOpt, _i 3.1,   0,       100,     "float"               },
   { "-d1",      &d1,     "D",      PaDouble, PaOpt, _i 3.3,   0,       100,     "double"              },
   { "--lmTest", &lmTest, "",       PaBool,   PaOpt, FALSE,    FALSE,   TRUE,    "test logMsg"         },
-  { NULL,       &v2,     "V2",     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
+  { NULL,       &v2,     2,     PaInt,    PaOpt, 19,       PaNL,    1000,    "env var 2"           },
```
fgalan: Most of the "v2" usage in this kind of structures (specially in the URL routes for service routines) has been correctly ignored by the replace... however, it seems I miss this.

Fixing soon.

src/lib/rest/RestService.cpp
fgalan: The whole comment preamble can be removed. API detection has been unified in just one place: `apiVersionGet`. In that function is where this should be explained (actually, it is, which simpler comments).

In fact, note that before this PR, the restService() function was using at the same time the ciP->apiVersion in some places and the apiVersion caculated looking to comp[] in another place. Crazy... :)

src/lib/rest/RestService.cpp
fgalan: Addressed in my comment above. NTC.

src/lib/rest/rest.cpp
fgalan: I think the criteria should be:

```
/v2 -> 2
/v1 -> 1
other admin ops -> 0
uknown -> -1

Looking to `apiVersionGet()` operation

```
  if (strncmp("/admin", path, strlen("/admin")) == 0)
  {
    return 0;
  }
```

it happens that `/version` and `/statistics` are outside `/admin` so they get -1. However, it seems it doesn't have any effect (as .test regresion works 100%) so I understand that the version number is not used in that cases.

Worth the time to research on it?

src/lib/parseArgs/parseTest.cpp
fgalan: Fixed in 11b4cb09c9a6b481138a72ee33ea5736fd0c6f15

src/lib/rest/rest.cpp
kzangeli: Yeah, I think it is worth to spend 15 minutes.
I'm doing it right now.
For example /ngsi9/xxx and /ngsi10/xxx (we still support those) should give api version 1

kzangeli: /log should also give v1 (  /log/trace == /v1/admin/log/trace )

crbrox:
LGTM

src/lib/rest/rest.cpp
fgalan: Fixed in 4731051ebd59791eee2ce3b3a13e6f016ce58c12

kzangeli:
LGTM

PR #2707, opened 2016-11-23
Fix op/update, omitted type

### LINKED ISSUE #2685: in POST v2/op/update (batch op) if we do not use entity type (optional) it is assumed as empty
in POST v2/op/update (batch op) if we do not use entity type (optional) it is assumed as empty. It is related to #2665.

#### update batch op request
```
url: POST http://localhost:1026/v2/op/update
headers:
    Fiware-Service: test_op_update_entities_type_error
    Fiware-ServicePath: /test
payload: {"entities": [{"temperature": {"type": "celsius", "value": "4", "metadata": {"alarm": {"type": "hot", "value": "warning"}}}, "id": "room_4"}], "actionType": "APPEND"}
```

#### update response
```
http code: 400 - Bad Request
headers:
   Connection: Keep-Alive
   Content-Length: 85
   Fiware-Correlator: 443fdb4a-ac1c-11e6-9947-005056a20feb
   Date: Wed, 16 Nov 2016 16:46:48 GMT
payload: {"error":"BadRequest","description":"entity type length: 0, min length supported: 1"}
```

#### expected response
```
http code: 204 - No Content
```

crbrox:
Fixes #2685 

@fgalan 
@kzangeli 

FYI: @iariasleon

### DIFF
--- a/src/lib/apiTypesV2/Entity.cpp
+++ b/src/lib/apiTypesV2/Entity.cpp
@@ -212,11 +212,16 @@ std::string Entity::check(ApiVersion apiVersion, RequestType requestType)
     return std::string(errorMsg);
   }
 
-  if ((apiVersion == V2) && ((len = strlen(type.c_str())) < MIN_ID_LEN) && (requestType != BatchQueryRequest))
+
+  if (!((requestType == BatchQueryRequest) || (requestType == BatchUpdateRequest && !typeGiven)))
+
   {
-    snprintf(errorMsg, sizeof errorMsg, "entity type length: %zd, min length supported: %d", len, MIN_ID_LEN);
-    alarmMgr.badInput(clientIp, errorMsg);
-    return std::string(errorMsg);
+    if ((apiVersion == V2) && ((len = strlen(type.c_str())) < MIN_ID_LEN))
+    {
+      snprintf(errorMsg, sizeof errorMsg, "entity type length: %zd, min length supported: %d", len, MIN_ID_LEN);
+      alarmMgr.badInput(clientIp, errorMsg);
+      return std::string(errorMsg);
+    }
   }
 
   if (forbiddenIdChars(apiVersion, type.c_str()))
--- a/test/functionalTest/cases/2685_batch_op_empty_type/2685_batch_op_empty_type.test
+++ b/test/functionalTest/cases/2685_batch_op_empty_type/2685_batch_op_empty_type.test
@@ -1,0 +1,108 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


test/functionalTest/cases/2685_batch_op_empty_type/2685_batch_op_empty_type.test
```diff
@@ -0,0 +1,108 @@
...
+Cover the bugfix for omitted entity type in POST /v2/op/update
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Update using empty entity type and see error
```
fgalan: Should be:

01. Update using empty entity type

crbrox: Fixed in b0d6629fe0f40d7ee01476110c3bb835bed5597a

src/lib/apiTypesV2/Entity.cpp
```diff
@@ -212,11 +212,16 @@ std::string Entity::check(ApiVersion apiVersion, RequestType requestType)
...
   {
-    snprintf(errorMsg, sizeof errorMsg, "entity type length: %zd, min length supported: %d", len, MIN_ID_LEN);
-    alarmMgr.badInput(clientIp, errorMsg);
-    return std::string(errorMsg);
+    if ((apiVersion == V2) && ((len = strlen(type.c_str())) < MIN_ID_LEN))
+    {
+      snprintf(errorMsg, sizeof errorMsg, "entity type length: %zd, min length supported: %d", len, MIN_ID_LEN);
+      alarmMgr.badInput(clientIp, errorMsg);
+      return std::string(errorMsg);
+    }
```
kzangeli: Just one doubt:
We are in apiTypesV2/Entity.cpp.  Do we really need to check for apiVersion == V2 ?
I understand we will *only* get here for V2 requests ...

crbrox: Good point, not sure why it is there. Maybe the class is used in some other way ...

fgalan: Makes sense... I'd suggest to solve this in one of the following alternatives:

1. Just put a "FIXME Px: probably we don't need to check for V2 here as this method is only called from the NGSIv2 processing logic" (x could be "5", for instance) and left this as it is now.
2. Remove the apiVersion == V2 checking and run the .test regression to check if it is safe removing.

I prefer 1, as I don't think this deserve to wait ~1 hour to see how it goes :). However, up to you @crbrox

crbrox: Finally , it is used in the call to forbiddenIdChars, so it would be a refactor issue more general. I'd leave it as it is now

fgalan: You mean option 1? (adding the FIXME comment and leave it as it is now)

crbrox: Not even, it seems a very general issue to me. I wouldn't  "bury" the FIXME inside this function, but I don`t have a strong opinion, only a suggestion.

fgalan: NTC from my point of view.

fgalan:
LGTM

src/lib/apiTypesV2/Entity.cpp
kzangeli: To be fixed later,

NTC for now

kzangeli:
LGTM

PR #2708, opened 2016-11-23
Feature/1960 last success and status failed

kzangeli:
Removed `timesFailed` and added `lastSuccess` for subscriptions.

FIXED:
_One thing is missing in the PR: it seems that lastFailure and lastSuccess have become mutually exclusive. Everything still works (as the newer of the two is always present in the database) but this was not my intention and I would like to ask for some help to fix this problem (or we decide that it is OK as is).
Because of this problem, the documentation is unfinished.
To be decided ..._

### DIFF
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1759,8 +1759,6 @@ A `notification` object contains the following subfields:
 + `timesSent` (not editable, only present in GET operations): Number of notifications sent due to
    this subscription.
 + `lastNotification` (not editable, only present in GET operations): Last notification timestamp in ISO8601 format.
-+ `timesFailed` (not editable, only present in GET operations): Number of consecutive errors.
-  Not present if subscription is notifying correctly.
 + `lastFailure` (not editable, only present in GET operations): Last failure timestamp in ISO8601 format.
   Not present if subscription has never has a problem with notifications.
 
@@ -1847,7 +1845,6 @@ Response:
               "attrs": ["temperature", "humidity"],
               "timesSent": 12,
               "lastNotification": "2015-10-05T16:00:00.00Z",
-              "timesFailed": 4
               "lastFailure": "2015-10-06T16:00:00.00Z"
             },
             "expires": "2016-04-05T14:00:00.00Z",
--- a/doc/apiary/v2/fiware-ngsiv2-reference.apib
+++ b/doc/apiary/v2/fiware-ngsiv2-reference.apib
@@ -1761,6 +1761,7 @@ A `notification` object contains the following subfields:
 + `lastNotification` (not editable, only present in GET operations): Last notification timestamp in ISO8601 format.
 + `lastFailure` (not editable, only present in GET operations): Last failure timestamp in ISO8601 format.
   Not present if subscription has never has a problem with notifications.
++ `lastSuccess` (not editable, only present in GET operations): Timestamp in ISO8601 format for last successful notification.
 
 An `http` object contains the following subfields:
 
@@ -1942,6 +1943,7 @@ Response:
             "attrs": ["temperature", "humidity"],
             "timesSent": 12,
             "lastNotification": "2015-10-05T16:00:00.00Z"
+            "lastSuccess": "2015-10-05T16:00:00.00Z"
           },
           "expires": "2016-04-05T14:00:00.00Z",
           "status": "active",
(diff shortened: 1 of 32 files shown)


src/lib/ngsiNotify/QueueNotifier.cpp
```diff
@@ -62,29 +62,39 @@ int QueueNotifier::start()
 */
 void QueueNotifier::sendNotifyContextRequest
```
kzangeli: Only style-guide fixes in this file

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -480,88 +481,154 @@ void mongoSubCacheRefresh(const std::string& database)
...
+/* ****************************************************************************
+*
+* mongoSubCountersUpdate - update subscription counters and timestamps in mongo
+*
+* FIXME P4:
+*   This function the way it is implemented right now might perform FIVE consecutive
+*   updates in mongo.
+*   There must be some faster way to do this:
+*     1. Read first, update only what's necessary OR
+*     2. Do a 'mega update' with all five fields (have the functions concatenate to final mongo 'query')
```
fgalan: I was thinking on this at the very begining, to "package" count and lastNofication updates into just one operation instead 2...

The fact is that given the filtering condition for each update needs to be evaluated "locally" for each lastX field, you need to split in several places. 

Thus, probably this FIXME should be removed.

src/lib/ngsiNotify/QueueNotifier.cpp
fgalan: Ack. NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1759,10 +1759,9 @@ A `notification` object contains the following subfields:
 + `timesSent` (not editable, only present in GET operations): Number of notifications sent due to
```
kzangeli: Fixed in c44e821

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -488,20 +488,28 @@ static long long setLastFailure(const BSONObj& subOrig, CachedSubscription* subC
 
 /* ****************************************************************************
 *
-* setTimesFailed - 
+* setLastSuccess -
 */
-static void setTimesFailed(const BSONObj& subOrig, CachedSubscription* subCacheP, BSONObjBuilder* b)
+static long long setLastSuccess(const BSONObj& subOrig, CachedSubscription* subCacheP, BSONObjBuilder* b)
```
kzangeli: Actually, lastNotification is added to before sending the notification while the other two (lastSuccess/lastFailure) need to wait until after to know the status of the notification and the resulting values are stored in the sub-cache only, to be added to mongo when a sub cache refresh is performed.  lastNotification works very differently.

I could change setLastNotification to return the value of lastNotification but that value is not used, so I really don't see any meaning with it ...

To be further discussed perhaps, but, in my opinion, NTC

fgalan: Maybe a comment to setLastNotification() preamble should be added, e.g.:

"Note that this function returns void, different from setLastFailure() and setLastSucces() functions, which use long long. This is due to <a fragment based in your excelent explanation above>"

kzangeli: ok, no probs

kzangeli: Fixed in 2cfde59

doc/manuals/admin/database_model.md
```diff
@@ -309,8 +308,8 @@ Fields:
 -   **payload**: optional field to store the payload for notification customization functionality in NGSIv2.
 -   **lastFailure**: the time when last notification failure occurred.
     Not present if the subscription has never failed.
--   **timesFailed**: number of consecutive notification errors for the subscription.
-    Not present if the subscription has never failed.
+-   **lastSuccess**: the time when last successful notification occurred.
+    Not present if the subscription has never provoked a notification.
```
fgalan: Maybe is more precise

"Not present if the subscription has never provoked a sucessful notification"

fgalan: Fixed in a3b71b891be0b333a5bb84b68a6f8151b1459f7e

fgalan:
LGTM

PR #2709, opened 2016-11-24
Hotfix about being less strict on Accept header

kzangeli:
Hotfix about being less strict on Accept header in release 1.4.0


fgalan:
LGTM

PR #2713, opened 2016-11-24
New limits for -maxConnections

kzangeli:
Upper limit has been removed, lower limit set to 1 (to avoid confusion with 0 changing to 1020).
Default value is set to 10240 connections.


fgalan:
LGTM

PR #2729, opened 2016-11-28
Hardening/mem leak fix for 1 6

kzangeli:
Adding a call to release in ContextAttribute destructor.
Unfortunately, this makes unit tests crash, so a number of modifications have been needed in unit tests as well.


fgalan:
LGTM

PR #2730, opened 2016-11-28
Removed an extra std::string r that cleared out the error description

kzangeli:
Using the same branch name as previous PR ... sorry about that


fgalan:
LGTM

Issue #2720, opened 2016-11-25
in /v2/op/update if entities is empty CB crashes


iariasleon:
in /v2/op/update if entities is empty CB crashes

#### CB version
```
  "version" : "1.5.0-next"
  "git_hash" : "d943fb6827fdee0d7e89501c343dc252f6757f60"
```

#### Request
```
curl http://qa-orion-fe-02:1026/v2/op/update -H "Content-Type: application/json" -d '{"entities": [], "actionType": "APPEND"}' 
```
#### Response
```
curl: (52) Empty reply from server
```

#### log
````
time=2016-11-25T13:07:27.918Z | lvl=DEBUG | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=rest.cpp[1123]:connectionTreat | msg=
time=2016-11-25T13:07:27.919Z | lvl=DEBUG | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=rest.cpp[1125]:connectionTreat | msg=--------------------- Serving request POST /v2/op/update -----------------
time=2016-11-25T13:07:27.919Z | lvl=DEBUG | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=rest.cpp[500]:httpHeaderGet | msg=HTTP Header:   User-Agent: curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.19.1 Basic ECC zlib/1.2.3 libidn/1.18 libssh2/1.4.2
time=2016-11-25T13:07:27.919Z | lvl=DEBUG | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=rest.cpp[500]:httpHeaderGet | msg=HTTP Header:   Host: qa-orion-fe-02:1026
time=2016-11-25T13:07:27.919Z | lvl=DEBUG | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=rest.cpp[500]:httpHeaderGet | msg=HTTP Header:   Accept: */*
[...]

PR #2733, opened 2016-11-28
FIX status code OK -> Bad Request when op update uses empty entities vector

fgalan:
A little missing part in the fix for #2720


kzangeli:
LGTM

PR #2734, opened 2016-11-28
FIX improving performance avoiding unneeded calls to gettid in log library

### LINKED ISSUE #2716: Many calls to gettid from the logMsg library
I just found out that a call to gettid() has mistakedly been put inside a loop, causing many many calls (108 in the case of the broker) to gettid() which is a system call, for each log message.
This must be removed, as to only call gettid(), when needed.

Actually, it should be called only once and then saved as a thread variable.

fgalan:
Fix #2716


kzangeli:
LGTM

PR #2735, opened 2016-11-28
Hardening/failing functests with no cache

kzangeli:
Fixed the crash with -noCache.
Transformed a functest to 'new layout'.

### DIFF
--- a/src/lib/mongoBackend/mongoSubCache.cpp
+++ b/src/lib/mongoBackend/mongoSubCache.cpp
@@ -608,7 +608,7 @@ void mongoSubCountersUpdate
 
   if (subId == "")
   {
-    // LM_E(("Runtime Error (no subscription id)"));
+    LM_E(("Runtime Error (empty subscription id)"));
     return;
   }
 
--- a/src/lib/mongoBackend/mongoSubCache.cpp
+++ b/src/lib/mongoBackend/mongoSubCache.cpp
@@ -606,6 +606,12 @@ void mongoSubCountersUpdate
 {
   std::string  collection = getSubscribeContextCollectionName(tenant);
 
+  if (subId == "")
+  {
+    // LM_E(("Runtime Error (no subscription id)"));
+    return;
+  }
+
   if (count > 0)
   {
     mongoSubCountersUpdateCount(collection, subId, count);
--- a/src/lib/ngsiNotify/Notifier.cpp
+++ b/src/lib/ngsiNotify/Notifier.cpp
@@ -148,6 +148,7 @@ void Notifier::sendNotifyContextAvailabilityRequest
     params->mimeType         = JSON;
     params->fiwareCorrelator = fiwareCorrelator;
     params->renderFormat     = renderFormatToString(renderFormat);
+    params->registration     = true;
 
     strncpy(params->transactionId, transactionId, sizeof(params->transactionId));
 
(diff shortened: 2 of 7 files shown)


src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -606,6 +606,12 @@ void mongoSubCountersUpdate
 {
   std::string  collection = getSubscribeContextCollectionName(tenant);
 
+  if (subId == "")
+  {
+    // LM_E(("Runtime Error (no subscription id)"));
```
fgalan: Discussed at skype.

This situation should be detected previously to call this function, i.e. in the startSenderThread() function, based in a new field in SenderThreadParams to identify the origin (either NGSI10 or NGSI9) of the notification.

kzangeli: Fixed in 7419ad4

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -338,6 +337,7 @@ static std::vector<SenderThreadParams*>* buildSenderParamsCustom
     params->renderFormat     = renderFormatToString(renderFormat);
     params->fiwareCorrelator = fiwareCorrelator;
     params->extraHeaders     = headers;
+    params->registration     = false;
```
fgalan: Maybe a better name for this param (and closer to its semantic) would be `persistErrorStatusInCache` ("inverting" the current logic).

Just a suggestion...

fgalan: ...or something shorter but with the same idea.

kzangeli: Not that I particularly like the name I picked, but my intention was to be generic. Probably soon registrations will have the same functionality as well and a variable with a very specific name will then be worthless. A variable to just differentiate ngsi9 from ngsi10 seems more useful.

Let's sleep on it ... :-)

kzangeli: What about inventing an `enum RequestFamily`, with values like NormalRequest, AdminRequest, and RegistrationRequest, and using a variable of that type ?

fgalan: > Probably soon registrations will have the same functionality as well and a variable with a very specific name will then be worthless

Well... actually I think that a most probable evolution is to remove comopletely the NGSI9 registration notification stuff. It is a mostly useless feature (nobody uses it). Thus, I think is improvable that NGSI9 registration notifications use cache some day.

In this same line, creating an enum for this would increase compleixty with little gain.

Let's go with the current name of the bool field. It is no so important, as long as we remember what it is and doesn't confuse ourselves if we need to change this part of the code in the future.

NTC

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -606,6 +606,12 @@ void mongoSubCountersUpdate
 {
   std::string  collection = getSubscribeContextCollectionName(tenant);
 
+  if (subId == "")
+  {
+    LM_E(("Runtime Error (empty subscription id)"));
```
fgalan: I  understand that after the `if (params->registration == false)` ward this is not really needed... However, I like the idea of having it, so in the case of bug and the program flow ends here we have at least a Runtime Error in the logs to signal the problem.

NTC

fgalan:
LGTM

PR #2742, opened 2016-11-30
FIX improve 'Checking ID fields' section in the documentation

fgalan:
Some improvements in that section in order to make it more precise.

It would be safe to wait to merge after a review of the `strictNgsiv1Ids` functionality has been done, as we have discussed recently.


crbrox:
LGTM

PR #2748, opened 2016-12-02
Feature/metrics manager

fgalan:
**DO NOT MERGE YET** (although the modified code doesn't have impact in CB, it is safer to wait to 1.6 releasing after soak)


kzangeli:
LGTM

Issue #2166, opened 2016-05-20
Broker fails to close used connections and they remain in CLOSE_WAIT


kzangeli:
It seems, as discovered by a client, that the broker sometimes leaves used connections without closing them, causing all fds to be used and consequently the broker stops working.

The broker uses _libmicrohttpd (MHD)_ to respond to requests and _libcurl_ for notifications and forwarding of messages and it seems libcurl might be responsible for the detected problem.

kzangeli:
Similar problem described in SOF:
http://stackoverflow.com/questions/12730477/close-is-not-closing-socket-properly

fgalan:
I understand that the problem has been discoverd in Orion 0.26.1. Reviewing the changelog from Orion 0.26.1 to Orion 1.1 (the current one) note the following in [Orion 0.28.0 release notes](https://github.com/telefonicaid/fiware-orion/releases/tag/0.28.0):

> Fix: libmicrohttpd 0.9.48 included in contextBroker as static lib (previous Orion versions used 0.9.22 as dynamic library) (Issue #1675)

If the problem cause points to libmicrohttpd, it should be reviewed if 0.9.48 has solved it.

crbrox:
Finally, it seems an expected behavior if the limit of file descriptors for the process is not high enough. With notification mode set as `threadpool` , each thread can hold up to 5 connections (libcurl pool). So, for outgoing request, 5*(number of pool's threads) file descriptors are needed to avoid exhausting the available fds

The total limit of  fds should include the number of simultaneous incoming request in MHD, plus an amount for logs, listening sockets and file descriptors used by libraries
`maxCon + 5 * no of threads + X` 

Or the number of threads in the pool can be lowered consequently

fgalan:
After analyzing this, it seem it isn't a problem in the code, but something that need to be documented. Thus, removing "bug" label and adding "doc" label.

fgalan:
Doc modifications in PR #2214

mrutid:
It's happening...

fgalan:
Recovering this issue for the 1.7.0 milestone.

fgalan:
After researching on this, CLOSE_WAIT are not the root of the problem: the root of the problem is fd exahustion when select() was used in MHD (which is the scope of issue #2724).

However, documentation should be improved to explain this better so this issue will remain opened.

fgalan:
[...]

PR #2751, opened 2016-12-05
Hardening/2724 mhd to use poll

### LINKED ISSUE #2724: Broker to use poll and not select
We need to modify the broker (**restInit** in `rest.cpp`) so that MHD uses `poll` and not `select`.
Select only supports 1024 simultaneous connections and some configurations might need more than that.
Also, there is a possibility we gain some performance using poll/epoll instead of select.

A few unfruitful attempts have been made, but select is still used (check done with `strace contextBroker -fg`) and not poll ...

kzangeli:
Made MHD use `poll` and not `select`.
Also, changed a little bit how serverMode was set - made it cleaner.

Fixes #2724.


CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix:  libmicrohttpd now uses poll() and not select() (#2724)
```
fgalan: Maybe epoll and https://github.com/telefonicaid/fiware-orion/issues/2166 should be mentionces here. In addition, explain the "functional" nature of the change. Suggestion:

```
- Fix:  libmicrohttpd now uses poll()/epoll() and not select() so fds greater than 1024 can be used for incomming connections  (#2724)
```

src/lib/rest/rest.cpp
```diff
@@ -1435,7 +1441,16 @@ static int restStart(IpVersion ipVersion, const char* httpsKey = NULL, const cha
 
   if (threadPoolSize != 0)
   {
-    serverMode = MHD_USE_SELECT_INTERNALLY;
+    //
+    // To use poll() instead of select(), MHD 0.9.48 has the define MHD_USE_EPOLL_LINUX_ONLY,
+    // while in MHD 0.9.51, the name of the define has changed to MHD_USE_EPOLL.
+    // So, to support both names, we need a ifdef/else cpp directive here.
+    //
+#ifdef MHD_USE_EPOLL
```
fgalan: Looks smart, as this will work if we some day upgrade MHD. However, it wouldn't be a problem also to "stick" to 0.9.48 now and change the code in the future :)

NTC

src/app/contextBroker/contextBroker.cpp
```diff
@@ -379,7 +369,7 @@ PaArgument paArgs[] =
   { "-subCacheIval",     &subCacheInterval, "SUBCACHE_IVAL",     PaInt,    PaOpt, 60,             0,     3600,     SUB_CACHE_IVAL_DESC    },
   { "-noCache",          &noCache,          "NOCACHE",           PaBool,   PaOpt, false,          false, true,     NO_CACHE               },
   { "-connectionMemory", &connectionMemory, "CONN_MEMORY",       PaUInt,   PaOpt, 64,             0,     1024,     CONN_MEMORY_DESC       },  
-  { "-maxConnections",   &maxConnections,   "MAX_CONN",          PaUInt,   PaOpt, FD_SETSIZE - 4, 0,     FD_SETSIZE - 4, MAX_CONN_DESC    },
+  { "-maxConnections",   &maxConnections,   "MAX_CONN",          PaUInt,   PaOpt, 1020,           0,     PaNL,     MAX_CONN_DESC          },
```
fgalan: Maybe a comment about that 1020 comes from old implementation based in select and it has been kept for legacy reasons could be useful (in the future, we could easily forget where this magic 1020 came from).

CHANGES_NEXT_RELEASE
kzangeli: Good.
Fixed in 3614f3b

src/app/contextBroker/contextBroker.cpp
kzangeli: Fixed in 39ca80e

crbrox:
LGTM

fgalan:
LGTM

Issue #940, opened 2015-05-28
Clean and refactor connectionTreat()


fgalan:
During the devels workshop meeting in May 2015, we found that the connectionTreat() function could be cleanned and refactored. It was originally written in a moment we don't know to much about MHD, but now that we have a better understanding, we could improve.

**Effort: 3 man day**

kzangeli:
This callback function is called 2,3 or more times by MHD in three different 'modes':
1.   init - get http headers and url parameters
2*.  gather payload data (can be called N times)
3.   last call, everything is prepared

So, the idea is to split the implementation of connectionTreat into three functions, one for each 'invocation-mode'. connectionTreat will stay, but just acting as a 'gateway' for the three different modes so in total where will be four functions.

fgalan:
Changed to nice-to-have (in my opinion, it is not so critical to be "should").

PR #2759, opened 2016-12-09
Feature/2750 common metrics reset and more counters

kzangeli:
reset() method
DELETE /admin/metrics
int => unsigned long long
serviceTime
Initial '/' in service path
release() method

### DIFF
--- a/src/app/contextBroker/contextBroker.cpp
+++ b/src/app/contextBroker/contextBroker.cpp
@@ -1735,7 +1735,6 @@ int main(int argC, char* argV[])
   contextBrokerInit(dbName, mtenant);
   curl_global_init(CURL_GLOBAL_NOTHING);
   alarmMgr.init(relogAlarms);
-
   metricsMgr.init(!noMetrics);
   logSummaryInit(&lsPeriod);
 
--- a/src/app/contextBroker/contextBroker.cpp
+++ b/src/app/contextBroker/contextBroker.cpp
@@ -1345,6 +1345,8 @@ void exitFunc(void)
   subCacheDestroy();
 #endif
 
+  metricsMgr.release();
+
   curl_context_cleanup();
   curl_global_cleanup();
 
--- a/src/lib/metricsMgr/MetricsManager.h
+++ b/src/lib/metricsMgr/MetricsManager.h
@@ -36,18 +36,21 @@
 * Metrics - 
 *
 * NOTE
-*   METRIC_TOTAL_SERVICE_TIME is in the metrics map but excluded from
+*   _METRIC_TOTAL_SERVICE_TIME is in the metrics map but excluded from
 *   the metric response, while METRIC_SERVICE_TIME is NOT in the metrics map, but
 *   include in the metric response.
-*   METRIC_SERVICE_TIME == METRIC_TOTAL_SERVICE_TIME / METRIC_TRANS_IN
+*   METRIC_SERVICE_TIME == _METRIC_TOTAL_SERVICE_TIME / METRIC_TRANS_IN
 *
+*   All 'help-counters' in the set, like _METRIC_TOTAL_SERVICE_TIME will have the
+*   prefix '_', to help remember it's a help measure.
+*   A prefix'_' in both macro name and in its string translation
 */
 #define METRIC_TRANS_IN                            "incomingTransactions"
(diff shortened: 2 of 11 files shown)


src/lib/rest/rest.cpp
```diff
@@ -1410,7 +1410,22 @@ static int connectionTreat
   }
   else
   {
-    serveFunction(ciP);
+    if (metricsMgr.isOn())
+    {
+      struct timeval  start;
+      struct timeval  end;
+
+      gettimeofday(&start, NULL);
```
fgalan: Not sure if this is teh best strategy due to:

* The transaction has actually started much before (the start of `connectionTreat()` is around 200 lines above :)
* Before reaching this point, the transaction could have ended (due to error conditions). Note the `restReply()` cases abvoe this point.

Maybe the current implementation of `connectionTreat()` make this difficult. Maybe it is the moment to deal with https://github.com/telefonicaid/fiware-orion/issues/940 to easy things.

We can discuss this at skype on Monday.

kzangeli: Fixed in c02e15a

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1345,6 +1345,8 @@ void exitFunc(void)
   subCacheDestroy();
 #endif
 
+  metricsMgr.release();
+
   curl_context_cleanup();
```
kzangeli: Test

src/app/contextBroker/contextBroker.cpp
```diff
@@ -1735,7 +1737,6 @@ int main(int argC, char* argV[])
   contextBrokerInit(dbName, mtenant);
```
kzangeli: test 2

src/lib/rest/rest.cpp
fgalan: We cannot go inside MHD but a good approach is at least to measure that time the transaction passes in CB code, i.e. the start should be the moment just after MHD call us and the end should be just before return to MHD.

src/lib/metricsMgr/MetricsManager.h
```diff
@@ -34,12 +34,20 @@
 /* ****************************************************************************
 *
 * Metrics - 
+*
+* NOTE
+*   METRIC_TOTAL_SERVICE_TIME is in the metrics map but excluded from
```
fgalan: Maybe we could use some syntax for those that are not indended to be printed. This would make easier debugging work, if needed. E.g. `__totalServiceTime`

src/lib/rest/rest.cpp
fgalan: Looking after `c02e15a` it seems that the "start time" is in the righ place. However, not sure about the "end time". The connectionTreat() function has many exit points (consisting in restReply + return, e.g. https://github.com/telefonicaid/fiware-orion/blob/b143fe974eaa58003c7488fc3197999eca9a9289/src/lib/rest/rest.cpp#L1385) before the point in which this "end time" is done. There are also two "if branches" apart from the one in which the "end time" is located (here https://github.com/telefonicaid/fiware-orion/blob/b143fe974eaa58003c7488fc3197999eca9a9289/src/lib/rest/rest.cpp#L1418 and here https://github.com/telefonicaid/fiware-orion/blob/b143fe974eaa58003c7488fc3197999eca9a9289/src/lib/rest/rest.cpp#L1423).

(Maybe a function which so many exit points as connectionTreat() if difficult to manage... I have that "feeling" when I refere to #940. Now I see that #940 refers to a different kind of refactor, but I hope you get my point. Having say that, I'm not fully sure that a refactor help... I only have the "feeling" that cleaning things could make things easier).

src/lib/metricsMgr/MetricsManager.h
kzangeli: Some prefix for 'internal' counters.
Sure, seems like a good idea

kzangeli: Fixed in dea4f85

src/lib/rest/rest.cpp
kzangeli: The end time moved to the end of the function requestCompleted, which is the last point possible for the broker. Only MHD knows more.

Fixed in b95465e

fgalan:
LGTM

crbrox:
LGTM

Issue #2762, opened 2016-12-12
hardcoded mhd connection timeout


kzangeli:
Currently (Dec 14 2016), the broker has no connection timeout for MHD, which means that clients will wait forever if the broker doesn't respond to their requests.

As of PR #2764, a connection timeout for MHD is added, but hardcoded to 6 seconds.
With this timeout, MHD closes the connection after the timeout expires and at least the clients now gets the 'connection closed', instead of waiting forever for a 'non-coming' response.

We might want to use a CLI option for this in the future.

[ 6 seconds is used as we use 5 seconds as default for 'internal timeouts', e.g. forwarding of messages ]

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2779

kzangeli:
This seems to not have been such a good idea after all.
We need to be a little smarter, so that the connection is not closed while the broker is still working on the response.
Perhaps what we should do is to start measuring time after the broker has sent the response ans if the connection is still not closed after a few seconds, the broker closes it ...

First step would be to go back to 'as it was' => default value 0 for -reqTimeout

kzangeli:
After making sure that the broker (really MHD does it) closes the connection regardless of what the client is doing after sending a response, it is enough to set back the MHD request timeout to infinite (zero).
We leave the CLI parameter -reqTimeout to give freedom for our users that use their own copy of the broker to play with this timeout, but the default value must be zero (wait forever).

fgalan:
Re-fixed in PR https://github.com/telefonicaid/fiware-orion/pull/2851

Issue #2726, opened 2016-11-25
Not same building mode for unit_test and regular code


fgalan:
Do the following comparison.

First, look for this line in Entity::check()

```
if (((apiVersion == V2) && (len = strlen(id.c_str())) < MIN_ID_LEN) && (requestType != EntityRequest))
```

Change `V2` by `2` and run `make debug` or `make`. Everything is ok.

Now, look for the following line in mongoQueryTypes_test.cpp

```
ms = mongoEntityTypes(&res, "", servicePathVector, uriParams, V1, NULL, false);
```` 

Change `V1` by `1` and run `make unit_test`. You should get the following compiler error:

```
/home/fermin/src/fiware-orion/test/unittests/mongoBackend/mongoQueryTypes_test.cpp: In member function ‘virtual void mongoQueryTypes_queryAllType_Test::TestBody()’:
/home/fermin/src/fiware-orion/test/unittests/mongoBackend/mongoQueryTypes_test.cpp:204:81: error: invalid conversion from ‘int’ to ‘ApiVersion’ [-fpermissive]
     ms = mongoEntityTypes(&res, "", servicePathVector, uriParams, 1, NULL, false);
                                                                                 ^
In file included from /home/fermin/src/fiware-orion/test/unittests/mongoBackend/mongoQueryTypes_test.cpp:32:0:
/home/fermin/src/fiware-orion/src/lib/mongoBackend/mongoQueryTypes.h:55:23: note: initializing argument 5 of ‘HttpStatusCode mongoEntityTypes(EntityTypeVectorResponse*, const string&, const std::vector<std::basic_string<char> >&, std::map<std::basic_string<char>, std::basic_string<char> >&, ApiVersion, unsigned int*, bool)’
 extern HttpStatusCode mongoEntityTypes
                       ^
```

My theory is that CMakeList.txt are in a way in which unit test and regular build is not the same. The unit test build is more strict, so we should change regular build to be the same.

**Comment by fgalan** [comment]
More info:

Running with VERBOSE=1 we have, for `debug` target (random file):

```
cd /home/fermin/src/fiware-orion/BUILD_DEBUG/src/lib/parseArgs && /usr/bin/c++   -DDEBUG 
-DDEBUG_fermin -DLINUX -DLM_ON -Wall -Wno-unknown-pragmas -D_LARGEFILE64_SOURCE 
-D_GNU_SOURCE -Werror -fno-var-tracking-assignments -DLINUX -g
[...]

PR #2764, opened 2016-12-14
Feature/2750 common metrics erroneous transaction

kzangeli:
* metricsMgr style-checked (had to make 'non-const reference' MINOR error)
* hardcoded MHD timeout of 6 seconds
* Exclude the first '/' from the Sub Service ONLY	if it starts with a '/' (erroneous requests may not start with '/')
* Refactored MetricsManager::toJson()
* Added the metrics counter 'incomingTransactionErrors'
* Added the sums in metrics
* Added the metrics counter 'incomingTransactionResponseSize'
* Fixed functests to NOT USE "GET /version" as the number of bytes in response is not constant between machines/users


src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -130,16 +133,22 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
...
   // Exclude the first '/' from the Sub Service
+  // But, only if it starts with a '/'
   //
-  const char* subService = &subServ.c_str()[1];
+  const char* subService = subServ.c_str();
+
+  if (subService[0] == '/')
+  {
+    subService = &subService[1];
+  }
```
kzangeli: Service path not starting with '/' is erroneous, and the request is saved in the metrics map with the erroneous service path. If we strip the first letter for this kind of requests it gets weird ...

NTC

scripts/style_check.sh
```diff
@@ -199,7 +199,7 @@ sev[9]=MAYOR;  cat[9]='Weird number of spaces at line-start'
 sev[10]=MAYOR; cat[10]='Blank line at the end of a code block'
 sev[11]=MAYOR; cat[11]='Include the directory when naming'
 sev[12]=MAYOR; cat[12]='Do not use namespace using-directives'
-sev[13]=MAYOR; cat[13]='Is this a non-const reference'
+sev[13]=MINOR; cat[13]='Is this a non-const reference'
```
kzangeli: See https://github.com/telefonicaid/fiware-orion/pull/2764/files#diff-e098492fc7346834f3ab6905d32d9b9bR232

src/lib/rest/rest.cpp
```diff
@@ -1435,15 +1457,25 @@ static int connectionTreat
...
+    metricsMgr.add(ciP->httpHeaders.tenant, ciP->httpHeaders.servicePath, METRIC_TRANS_IN_ERRORS, 1);
   }
   else
   {
     serveFunction(ciP);
+
+    //
+    // If a service function sets the httpStatusCode to something above the set of 200s, an error has occurred
+    //
+    if (ciP->httpStatusCode >= SccBadRequest)
```
kzangeli: Not ideal, but `>= 400` works for us as 300s aren't used

CHANGES_NEXT_RELEASE
```diff
@@ -3,4 +3,5 @@
 - Fix: proper 500 Internal Server Error response with descriptive text when "sort fail" at DB occur (#2752)
 - Add: metrics REST API (#2750)
 - Add: new CLI '-disableMetrics' to disable metrics (#2750)
+- Add: timeout of 6 seconds (hardcoded for now) for all incoming REST requests (#2762)
```
fgalan: As this is not a definitive solution, my suggestion is change: (#2762) -> (temporal fix for #2762) or something similar.

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -216,6 +227,57 @@ void MetricsManager::reset(void)
 
 /* ****************************************************************************
 *
+* metricsRender - 
+*/
+static std::string metricsRender(std::map<std::string, uint64_t>& metricsMap)
```
fgalan: Maybe StackOverflow may help.

scripts/style_check.sh
fgalan: If this is kept at the end, I understand that a FIXME mark should be added.

src/lib/metricsMgr/MetricsManager.cpp
fgalan: Another alternative to avoid a "scattered" of wrong service paths should be to use the same group for all them. Maybe the root service path (although the problem in this case is that the metrics get mixed with the ones using correctly the root service path) or use an special keyword.

Not for this PR. For this PR I'd suggest just include the issue in the ".h FIXME list" as something to fix before concluding the functionality. Probably the whole IoTP team should be involved in the discussion.

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -216,6 +227,57 @@ void MetricsManager::reset(void)
...
+{
+  std::map<std::string, uint64_t>::iterator  it;
+  uint64_t                                   incomingTransactions = 0;
+  uint64_t                                   totalServiceTime     = 0;
+  int                                        metricsAdded         = 0;
```
fgalan: It seems this counter is not used at the end (I mean, I see the increments, but now who is using the final value).

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -216,6 +227,57 @@ void MetricsManager::reset(void)
...
+  return jh.str();
+}
+
+
+
+/* ****************************************************************************
+*
 * MetricsManager::toJson -
 *
 * FIXME PR: needs a refactor (see .h)
```
fgalan: If you are happy with the refactor, remove this line.

src/lib/rest/rest.cpp
```diff
@@ -1245,10 +1247,19 @@ static int connectionTreat
...
     //
     // If the HTTP header says the request is bigger than our PAYLOAD_MAX_SIZE,
-    // just silently "eat" the entire message
+    // just silently "eat" the entire message.
+    // 
+    // The problem occurs when the broker is lied to and there aren't ciP->httpHeaders.contentLength
+    // bytes to read.
+    // When this happens, MHD blocks until it times out (MHD_OPTION_CONNECTION_TIMEOUT defaults to 5 seconds),
+    // and the broker isn't able to respond. MHD just closes the connection.
+    // Question asked in mhd mailing list.
```
fgalan: I think we have an issue number for this... should be referred here.

src/lib/rest/rest.cpp
fgalan: The condition is good but, as discussed at skype, probably all them add METRIC_TRANS_IN_ERROR with 1 could be moved to requestCompleted() function.

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
```diff
@@ -30,21 +30,21 @@ brokerStart CB
...
+# 01. Send five GET /v2/entities
 # 02. Ask for Metrics, see 5 Incoming Transactions
 #
 
-echo "01. Send five GET /version"
-echo "=========================="
-orionCurl --url /version --tenant S1 --servicePath /SP1
+echo "01. Send five GET /v2/entities"
+echo "=============================="
+orionCurl --url /v2/entities --tenant S1 --servicePath /entities1
```
fgalan: I understand the change in the URL but... why to change the service path string?  I think that `/SPx` looks better...

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
```diff
@@ -0,0 +1,519 @@
...
+# 10. Send a request with an invalid character in a URI parameter
+# 11. Send a request with too many components in ServicePath
+# 12. Send a PUT /v2/entities/<id>/attrs/<attrName>/value for an non-existing entity E_NOPE
+# 13. Ask for metrics and see 10 errors
+#
+
+
+echo "00. GET /v2/entities"
+echo "===================="
+orionCurl --url /v2/entities --servicePath /entities
```
fgalan: In general, it is better to use "abstract" SPs, like `/SPx`. See an `"entities"` in the metric payload could be a bit confusing (as the same field is uses in NGSIv2 payloads).

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
```diff
@@ -0,0 +1,519 @@
...
+
+09. Send a request with no acceptable mime-type in accept header
+================================================================
+HTTP/1.1 406 Not Acceptable
+Content-Length: 145
+Content-Type: application/json
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+{
```
fgalan: I understand this NGSIv1-style error response is ok. If isn't, then a reference to the issue number should be include in a FIXME mark in the test step comment preamble.

test/functionalTest/cases/2750_common_metrics/2750_totalPayloadAsMetric.test
```diff
@@ -30,17 +30,19 @@ brokerStart CB
...
 #
-# 01. Send a GET /version
+# 01. Send a GET /v2/entities
 # 02. Ask for Metrics, see 1 Incoming Transaction, but nothing about incomingTransactionRequestSize
 # 03. Create an entity (with payload of exactly 50 bytes)
 # 04. Ask for Metrics, see 4 Incoming Transactions, and 50 bytes in incomingTransactionRequestSize
 # 05. Create an entity (with payload of exactly 100 bytes)
 # 06. Ask for Metrics, see 6 Incoming Transactions, and 150 bytes in incomingTransactionRequestSize
+# 07. GET /v2/entities - to provoke payload in response
+# 08. Ask for Metrics, see 8, and XXX bytes as incomingTransactionResponseSize
```
fgalan: I guess your plan was to replace XXX by an actual bytes number. I think is not a good idea, as this may change from time to time and, while we can fix in the test step itself, maybe we forget in the preamble. So better to use neutral messages for preamble/echos.

(Maybe the same problem happens in other .test that I haven't realized)

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Yes, I was thinking about that too.
However, not sure it's a good idea to cancel the cause of the error, which in this case in an erroneous service path ... ?

We would need a special keyword, but, the info about the erroneous service path goes away. That's a pity ...

I'll include this in the ToDo in the header file as a discussion item. Perhaps we can find a better solution

src/lib/rest/rest.cpp
kzangeli: Fixed in 8a2a59b

CHANGES_NEXT_RELEASE
kzangeli: Fixed in f29aaf4

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 216e26c

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 493c33d
Thanks!
Pity the compiler doesn't warn about this ...

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
kzangeli: I started using those 'descriptive' service path names in newer functests, to make it easier to make sure the responses (metrics) were OK.
I like it and I would like to keep it ...

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: This is the same comment as the one above.
If you really don't like this (just, take a look at the functests without diff-noise first, please), I can change it.
However, I prefer to keep it as it is (which is why I implemented it this way).

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: FIXME added in db86eeb

test/functionalTest/cases/2750_common_metrics/2750_totalPayloadAsMetric.test
kzangeli: The whole idea is to make sure the adding of bytes in payload is working (in mean the previous 50, 100 and 150 in steps 03-06).
This last one in step 08 can be removed as 'exact number', no probs

kzangeli: Fixed in 1f267e1 (changed the text of step 08)

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 796a135

test/functionalTest/cases/2750_common_metrics/2750_totalPayloadAsMetric.test
```diff
@@ -116,10 +123,21 @@ Date: REGEX(.*)
...
                     "serviceTime": REGEX(.*)
                 }
+            },
+            "sum": {
+                "incomingTransactionResponseSize": 2,
+                "incomingTransactions": 1,
+                "serviceTime": REGEX(.*)
             }
+        },
+        "sum": {
```
fgalan: This is not correct as "sum" must not be at "S1" level, but a "services" level.

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
fgalan: I don't like, but maybe we can find a tradeoff. What about `/SP_entities`?

src/lib/metricsMgr/MetricsManager.cpp
fgalan: (Out-topic regarding compiler: it could be our fault... we have weird things happening at built, see https://github.com/telefonicaid/fiware-orion/issues/2726)

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: Sure, /SP_xxx is OK for me

test/functionalTest/cases/2750_common_metrics/2750_totalPayloadAsMetric.test
kzangeli: Fixed in a3e5109

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: Picked /SP/xxx instead (looks better). In cd4e7a7

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
```diff
@@ -30,21 +30,21 @@ brokerStart CB
...
+# 01. Send five GET /v2/entities
 # 02. Ask for Metrics, see 5 Incoming Transactions
 #
 
-echo "01. Send five GET /version"
-echo "=========================="
-orionCurl --url /version --tenant S1 --servicePath /SP1
+echo "01. Send five GET /v2/entities"
+echo "=============================="
+orionCurl --url /v2/entities --tenant S1 --servicePath /SP/entities1
```
fgalan: For the future, better to use `/SP_entities1`, as hierarchical SPs are not an use case at IoTP.

For this PR, it doesn't worth the time to change it. NTC.

src/lib/metricsMgr/MetricsManager.cpp
fgalan: NTC

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
kzangeli: Ah, as it is about metrics ...
OK, will try to remember

fgalan:
LGTM

Issue #2750, opened 2016-12-02
Metrics API: Common Metrics


fgalan:
This issue is about implementing the basic API metrics operations, along with the first batch of counters (the common ones for all the IoTP components).

Operations to implement:

```
GET <host:port>/admin/metrics
DELETE <host:port>/admin/metrics
GET <host:port>/admin/metrics?reset=true (maybe, yet under discussion)
```

Syntax:

```
{
    "services": {
        "service1": {
            "subservs": {
                "subserv1": {},
                "subserv2": {},
                "subserv3": {}
            },
            "sum": {}
        },
        "service2": {
            "subservs": {
                "subserv1": {},
                "subserv2": {},
                "subserv3": {}
            },
            "sum": {}
        },
        "service3": {
            "subservs": {
                "subserv1": {},
                "subserv2": {},
                "subserv3": {}
            },
            "sum": {}
        }
    },
    "sum": {
        "subservs": {
            "subserv1": {},
            "subserv2": {},
            "subserv3": {}
        },
        "sum": {}
    }
}
```

Metrics to implement:

- **incomingTransactions**: number of requests consumed by the component.
- **incomingTransactionRequestSize**: total size (bytes) in requests associated to incoming transactions (“in” from the point of view of the component).
- **incomingTransactionResponseSize**: total size (bytes) in responses associated to incoming transactions (“out” from the point of view of the component).
- **incomingTransactionErrors**: number of incoming transactions resulting in error.
- **serviceTime**: average time to serve a transaction. 
- **outgoingTransactions**: number of requests sent by the component.
- **outgoingTransactionRequestSize**: total size (bytes) in requests associated to outgoing transactions (“out” from the point of view of the component).
- **outgoingTransactionResponseSize**: total size (bytes) in responses associated to outgoing transactions (“in” from the point of view of the component).
- **outgoingTransactionErrors**: number of outgoing transactions resulting in error.

Note: all incoming/outgoing channels count for the above metrics.

fgalan:
Implementation will be based in the AlarmManager class, which PoC is provided in PR https://github.com/telefonicaid/fiware-orion/pull/2748

kzangeli:
A CLI parameter '-metrics' will be supplied to **not++ offer metrics, as the calculation of metrics lowers the efficiency of the broker.

kzangeli:
About sizes, only payload size is accumulated

kzangeli:
[...]

PR #2767, opened 2016-12-16
Feature/2750 cross tenant metrics for service path

kzangeli:
Implemented the cross-tenant sums for service paths
Also, tenants names from uppercase to lowercase

### DIFF
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -142,23 +142,6 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     subService = &subService[1];
   }
 
-  //
-  // Convert tenant/service to lowercase
-  //
-  char* sP   = (char*) srv.c_str();
-  char  diff = 'a' - 'A';  // To add to char to go from uppercase to lower case
-
-  while (*sP != 0)
-  {
-    if ((*sP >= 'A') && (*sP <= 'Z'))
-    {
-      *sP += diff;
-    }
-
-    ++sP;
-  }
-
-
   semTake();
 
 
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -381,7 +381,7 @@ std::string MetricsManager::toJson(void)
     jhSubServ.addRaw(subService, subServiceString);
   }
 
-  lastSum.addRaw("subServs", jhSubServ.str());
+  lastSum.addRaw("subservs", jhSubServ.str());
 
   std::string  sumString = metricsRender(&sum);
   lastSum.addRaw("sum", sumString);
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -142,7 +142,25 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     subService = &subService[1];
   }
 
+  //
+  // Convert tenant/service to lowercase
+  //
+  char* sP   = (char*) srv.c_str();
+  char  diff = 'a' - 'A';  // To add to char to go from uppercase to lower case
+
+  while (*sP != 0)
+  {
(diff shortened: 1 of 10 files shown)


src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -142,8 +142,26 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     subService = &subService[1];
   }
 
+  //
+  // Convert tenant/service to lowercase
```
fgalan: This seems to be a rather generic string processing... Don't we have some functional already for this (probably in the commons/ library)? In negative case, shouldn't we move this to a generic function?

fgalan: Moreover... maybe the lower-casing should be done at the very beginning (just after reading it from the HTTP header) so the lowercase version would be used in all the logic (both metric related code and "regular" processing)?

src/lib/metricsMgr/MetricsManager.h
```diff
@@ -107,24 +107,15 @@
 *
 * FIXME PR: summary of things to improve
 *
-* 01. Check alternatives to "triple std::map" from the point of view of performance
```
fgalan: If at the end we declare "remove" for this item I think we should remove some "FIXME PR" code existing code related with this.

Please, review the "FIXME PR" marks in MetricssManager.cpp in case someone could be removed.

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
```diff
@@ -159,6 +162,28 @@ Date: REGEX(.*)
         }
     },
     "sum": {
+        "subServs": {
```
fgalan: Note (at https://github.com/telefonicaid/fiware-orion/issues/2750) that the right one is `"subservs", not "subServs"...

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
```diff
@@ -295,25 +302,44 @@ Date: REGEX(.*)
                     "serviceTime": REGEX(.*)
                 },
                 "SP/get/metrics": {
-                    "incomingTransactionResponseSize": 337,
+                    "incomingTransactionResponseSize": 457,
```
fgalan: I guess this is due to the new `subservs` item within `sum` tenant...

Very interesting (and funny :) way to check that the modification has worked in the metric API system itself.

kzangeli: Funny ... any change in Content-Length in these test cases is recursively spread to many other test cases ...
I would use the word nightmare, not fun ... :-D

fgalan: Remember the `fixContentLengths.py` script can be used to automatically adjust Content-Length (when the only difference between .test and .out is Content-Lenght lines). Your nightmare can become in a sweet dream xD

test/functionalTest/cases/2750_common_metrics/2750_incomingTransactions.test
kzangeli: Oops ...

Fixed in 258473d

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 13595f5

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: True, that would solve some of the problems.

src/lib/metricsMgr/MetricsManager.h
kzangeli: FIXMEs removed in 9423e2a

fgalan:
LGTM

crbrox:
LGTM

PR #2771, opened 2016-12-19
Feature/2750 notifications

kzangeli:
Metrics for outgoing requests.
Last part of #2750.

### DIFF
--- a/src/lib/rest/httpRequestSend.cpp
+++ b/src/lib/rest/httpRequestSend.cpp
@@ -44,6 +44,7 @@
 #include "common/sem.h"
 #include "common/limits.h"
 #include "alarmMgr/alarmMgr.h"
+#include "metricsMgr/metricsMgr.h"
 #include "rest/ConnectionInfo.h"
 #include "rest/httpRequestSend.h"
 #include "rest/rest.h"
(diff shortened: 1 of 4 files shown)


src/lib/rest/httpRequestSend.cpp
```diff
@@ -521,12 +537,34 @@ int httpRequestSendWithCurl
...
     *outP = "notification failure";
+
+    metricsMgr.add(tenant, servicePath, METRIC_TRANS_OUT_ERRORS, 1);
   }
   else
   {
+    //
     // The Response is here
+    //
+    // To get the size of the payload, simply search for Content-Len inside the HTTP headers and extract the number
```
fgalan: I'd suggest to move this "length extraction logic" to a separate function.

src/lib/rest/httpRequestSend.cpp
```diff
@@ -418,7 +427,9 @@ int httpRequestSendWithCurl
   httpHeaderAdd(&headers, "Content-length", headerContentLength, &outgoingMsgSize, extraHeaders, usedExtraHeaders);
 
   // Add the size of the actual payload
```
fgalan: I understand the difference between payloadSize and outgoingMsgSize is that the later is the total size of the message, including also headers and so. It would be useful to explain in a comment... it is not direct (I have had to look to the code a bit in order to realize about the difference).

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in 17a5c5d

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in 17a5c5d

fgalan:
LGTM

Issue #2761, opened 2016-12-12
broker doesn't respond if not able to read all of the incoming payload


kzangeli:
In a functest where we needed to check error for TOO BIG payload size, we got the problem that the broker is unable to respond with a 'nice error' when this occurs.

To provoke the problem, we simply lie to the broker, saying that the incoming payload is bigger than it really is:

    payload='{ "id": "E1", "type": "T1" }'
    orionCurl --url /v2/entities --payload "$payload" -H "Content-Length: 6000000"

In `connectionTreat`, an attempt to send a decent response is done, but due to a limitation in libmicrohttpd, the response isn't sent.

A question to be sent to the mailing list of MHD.

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2774

PR #2774, opened 2016-12-19
Bug/2761 no response if not all payload is read

kzangeli:
Checking HTTP header 'Content-Length' for **too long** message before even starting to read the message body, and returning error if the body is too long (according to Content-Length).

Fixes issue #2761

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -1216,6 +1216,18 @@ static int connectionTreat
     ciP->httpHeader.push_back("Fiware-Correlator");
     ciP->httpHeaderValue.push_back(ciP->httpHeaders.correlator);
 
+    if (ciP->httpHeaders.contentLength > PAYLOAD_MAX_SIZE)
+    {
+      char details[256];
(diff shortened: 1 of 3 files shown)


src/lib/rest/rest.cpp
```diff
@@ -1220,6 +1220,18 @@ static int connectionTreat
     ciP->httpHeader.push_back("Fiware-Correlator");
     ciP->httpHeaderValue.push_back(ciP->httpHeaders.correlator);
 
+    if ((ciP->httpHeaders.contentLength > PAYLOAD_MAX_SIZE) && (ciP->apiVersion == V2))
```
fgalan: What about not NGSIv1 requests or out-of-NGSI requests?

kzangeli: NGSI v1 requests are tricky.
Each type of request has a different response payload.
Not sure it is worth the effort to implement this for NGSI v1
Out-of-NGSI requests have very small if any payload at all.
So, not very interesting, could be implemented though. Easily.

fgalan: It doesn't worth the effort.

NTC

fgalan:
LGTM

PR #2776, opened 2016-12-19
feature/2750_get_and_delete

kzangeli:
One more PR for issue #2750.

Added URI param 'reset' to GET /admin/metrics

### DIFF
--- a/src/lib/serviceRoutinesV2/getMetrics.cpp
+++ b/src/lib/serviceRoutinesV2/getMetrics.cpp
@@ -63,13 +63,8 @@ std::string getMetrics
     return oe.toJson();
   }
 
-  std::string  reset   = ciP->uriParam["reset"];
-  std::string  payload = metricsMgr.toJson();
-
-  if (reset == "true")
-  {
-    metricsMgr.reset();
-  }
+  bool         doReset  = (ciP->uriParam["reset"] == "true")? true : false;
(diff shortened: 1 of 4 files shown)


src/lib/serviceRoutinesV2/getMetrics.cpp
```diff
@@ -58,5 +63,13 @@ std::string getMetrics
     return oe.toJson();
   }
 
-  return metricsMgr.toJson();
+  std::string  reset   = ciP->uriParam["reset"];
```
fgalan: Note that the requests that arrive in parallel between toJson() call and reset() call (i.e. while this thread progress from line 67 to line 71) will be lost for the next toJson(). In other words, the operation wouldn't lost atomicity, which is one of its requirements. Thus, I'd suggest a refactor in MetricsManager class to implement this, in the following way (pseudocode):

```
toJson()
{
  semTake();
  __toJson();
  semGive();
}

reset()
{
  semTake();
  __reset();
  semGive();
}

toJsonAndReset()
{
  semTake();
  __toJson();
  __reset();
  semGive():
}
```

The `__toJson()` and `__reset()` method correspond to the current `toJson()` and `reset()` without the sem management functions. Of course, better names could be defined. They should be declared as class private methods are they are not supposed to be called by class users.

crbrox:
LGTM

Issue #2781, opened 2016-12-21
More efficient implementation for skipping metrics when invalid service or service-path


kzangeli:
We've decided to not record any metrics for requests with invalid `tenant` (service) or `service-path`.
The implementation is not simple. The best implementation from an efficiency point-of-view makes the code 'ugly' and we've decided to not go that way.

Instead, we let the call get through to `MetricsMgr::add()` where we use the functions `tenantCheck` and `servicePathCheck` to decide inside `add` whether to add to metrics or not. 

We might have another look at this solution, to gain efficiency.

One possible solution would be to simply add everything to metrics but, in `toJson` filter out metrics origining from requests with invalid tenants/service-paths.

This is also not very efficient of course ...
Especially as system calls (time measurements and semaphores) are involved and a lot of adding/removing from the complex C++ maps.

The only 'fast' solution' is the ugly one, so far ...

fgalan:
> This is also not very efficient of course ...

Two things has to be taken into account:

* toJson() is not going to be called very frecuent. Thus, the filtering step that we would need there hasn't a significant impact in performance
* add() is goging to be called very frecuent. However, the probablilty of getting an invalidad service/subservice is low. Thus, the time wasted in these cases hasn't a significant impact in performance.

In sum, no significant impact in performance with the proposed solution after all :)

kzangeli:
For the initial solution (that we will probably keep), `tenantCheck` has been put in rest.h (should be a static function in rest.cpp, except that it isn't for unit test purposes).
Also, `servicePathCheck'  was static before this. So, now it is extern and present in RestService.h.

**If** we change the current solution, we should rollback these two changes (about static/extern tenantCheck and servicePathCheck).

PR #2785, opened 2016-12-21
Feature/2750 ignore metrics for error in service and service path

kzangeli:
Fixed last "known and decided part" of #2750:
* No metrics for erroneous tenants and service paths
* Special names for default tenant and service path in metrics
* Complex Service Paths to use a simplification of the service path for metrics

### DIFF
--- a/src/lib/common/defaultValues.h
+++ b/src/lib/common/defaultValues.h
@@ -28,8 +28,9 @@
 
 
 
-#define  DEFAULT_SERVICE_PATH_UPDATES  "/"
-#define  DEFAULT_SERVICE_PATH_QUERIES  "/#"
-
+#define  DEFAULT_SERVICE_PATH_UPDATES         "/"
+#define  DEFAULT_SERVICE_PATH_QUERIES         "/#"
+#define  DEFAULT_SERVICE_KEY_FOR_METRICS      "default-service"
+#define  DEFAULT_SUB_SERVICE_KEY_FOR_METRICS  "root-subserv"
 
 #endif  // SRC_LIB_COMMON_DEFAULTVALUES_H_
--- a/src/lib/common/defaultValues.h
+++ b/src/lib/common/defaultValues.h
@@ -30,7 +30,5 @@
 
 #define  DEFAULT_SERVICE_PATH_UPDATES         "/"
 #define  DEFAULT_SERVICE_PATH_QUERIES         "/#"
-#define  DEFAULT_SERVICE_KEY_FOR_METRICS      "default-service"
-#define  DEFAULT_SUB_SERVICE_KEY_FOR_METRICS  "root-subserv"
 
 #endif  // SRC_LIB_COMMON_DEFAULTVALUES_H_
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -152,11 +152,36 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
   // Exclude the first '/' from the Sub Service
   // But, only if it starts with a '/'
   //
-  const char* subService = subServ.c_str();
+  // Also, if service path ends in '/#', cancel out that part
+  //
+  // Note that for the 'cancel out' part we need to allocate a copy of
+  // the service path (and frre it after usage).
+  //
+  char* subService;
 
-  if (subService[0] == '/')
+  if (subServ.c_str()[0] == '/')
   {
-    subService = &subService[1];
+    subService = strdup(&subServ.c_str()[1]);
+  }
+  else
+  {
+    subService = strdup(subServ.c_str());
+  }
+
+  //
+  // Now, if service path ends in '/#', cancel out that part
+  //
+  int subServiceLen = strlen(subService);
+
+  if ((subServiceLen >= 2) && (subService[subServiceLen - 1] == '#') && (subService[subServiceLen - 2] == '/'))
+  {
+    subService[subServiceLen - 2] = 0;
+  }
+
+  // Need to cover the '/#' case as well (only '#' left as the first '/' has been skipped already)
+  if ((subService[0] == '#') && (subService[1] == 0))
+  {
+    subService[0] = 0;
   }
 
   semTake();
@@ -192,6 +217,8 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
   metrics[srv]->at(subService)->at(metric) += value;
 
   semGive();
+
+  free(subService);
 }
 
 
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -60,6 +60,42 @@ MetricsManager::MetricsManager(): on(false), semWaitStatistics(false), semWaitTi
 
 /* ****************************************************************************
 *
+* MetricsManager::serviceValid - 
+*/
+bool MetricsManager::serviceValid(const std::string& srv)
+{
+  if (tenantCheck(srv) != "OK")
+  {
+    return false;
+  }
+
+  return true;
+}
+
+
+
+/* ****************************************************************************
+*
+* MetricsManager::subServiceValid - 
+*/
+bool MetricsManager::subServiceValid(const std::string& subsrv)
+{
+  ConnectionInfo ci;
+
+  ci.httpHeaders.servicePathReceived = true;
+
+  if (servicePathCheck(&ci, subsrv.c_str()) != 0)
+  {
+    return false;
+  }
+
+  return true;
+}
+
+
+
+/* ****************************************************************************
+*
 * MetricsManager::init -
 *
 * NOTE
@@ -134,6 +170,11 @@ int64_t MetricsManager::semWaitTimeGet(void)
 /* ****************************************************************************
 *
 * MetricsManager::add -
+*
+* FIXME P4: About the calls to serviceValid and subServiceValid:
+*   we check that 'srv' and 'subServ' are legal names, and if not, metrics are skipped.
+*   Doing this each time add() is called is not optimal for performance.
+*   The github issue #2781 is about better solutions for this.
 */
 void MetricsManager::add(const std::string& srv, const std::string& subServ, const std::string& metric, uint64_t value)
 {
@@ -142,13 +183,12 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     return;
   }
 
-  // FIXME P4: See github issue #2781
-  if (tenantCheck(srv) != "OK")
+  if (serviceValid(srv) == false)
   {
     return;
   }
 
-  char* toFree = strdup(subServ.c_str());
+  char* toFree      = strdup(subServ.c_str());
   char* subServiceP = toFree;
 
   //
@@ -160,10 +200,7 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     *commaP = 0;
   }
 
-  // FIXME P4: See github issue #2781
-  ConnectionInfo ci;
-  ci.httpHeaders.servicePathReceived = true;
-  if (servicePathCheck(&ci, subServiceP) != 0)
+  if (subServiceValid(subServiceP) == false)
   {
     free(toFree);
     return;
--- a/src/lib/metricsMgr/MetricsManager.cpp
+++ b/src/lib/metricsMgr/MetricsManager.cpp
@@ -33,10 +33,18 @@
 #include "logMsg/traceLevels.h"
 
 #include "common/JsonHelper.h"
-#include "common/defaultValues.h"
 #include "rest/rest.h"
 #include "rest/RestService.h"
 #include "metricsMgr/MetricsManager.h"
+
+
+
+/* ****************************************************************************
+*
+* Default values for metrics - 
+*/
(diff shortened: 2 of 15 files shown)


src/lib/common/defaultValues.h
```diff
@@ -28,8 +28,9 @@
 
 
 
-#define  DEFAULT_SERVICE_PATH_UPDATES  "/"
-#define  DEFAULT_SERVICE_PATH_QUERIES  "/#"
-
+#define  DEFAULT_SERVICE_PATH_UPDATES         "/"
+#define  DEFAULT_SERVICE_PATH_QUERIES         "/#"
+#define  DEFAULT_SERVICE_KEY_FOR_METRICS      "default-service"
```
fgalan: Probably metrics defaults should be at MetricsManager.h (or even MetricsManager.cpp?) as per encapsulation/modularity it doesn't make too much sense to be used outise there.

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -131,15 +134,54 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     return;
   }
 
+  // FIXME P4: See github issue #2781
```
fgalan: Probably an elaboration of this beyond the issue number itself could be useful. I mean:

"FIXME P4: we check service and subsrv are legal names, otherwise metrics are not accounted. Doing this each time add() is called could have performance issues, #2781 is about better solutions for this"

or something similar.

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -131,15 +134,54 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
...
+    subService = strdup(subServ.c_str());
+  }
+
+  //
+  // Now, if service path ends in '/#', cancel out that part
+  //
+  int subServiceLen = strlen(subService);
 
-  if (subService[0] == '/')
+  if ((subServiceLen >= 2) && (subService[subServiceLen - 1] == '#') && (subService[subServiceLen - 2] == '/'))
```
fgalan: Side-comment, but meanifull for this PR anyway: 

What about "/A#"? They are previously detected as invalid service paths at the very beginning (just after reading it from the header)? What documentation says about the correct syntax?

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -339,7 +398,15 @@ std::string MetricsManager::_toJson(void)
...
     subServiceString = metricsRender(&it->second);
-    jhSubServ.addRaw(subService, subServiceString);
+
+    if (subService != "")
+    {
+      jhSubServ.addRaw(subService, subServiceString);
+    }
+    else
+    {
+      jhSubServ.addRaw(DEFAULT_SUB_SERVICE_KEY_FOR_METRICS, subServiceString);
```
fgalan: Probably ROOT_SUB_SERVICE_KEY_FOR_METRICS is a better name.

src/lib/rest/RestService.cpp
```diff
@@ -149,7 +149,7 @@ std::string payloadParse
 *
 * tenantCheck - 
 */
-static std::string tenantCheck(const std::string& tenant)
+extern std::string tenantCheck(const std::string& tenant)
```
fgalan: Maybe a note about reverting this citing #2781 should be added?

(In fact #2781 cite also another function, where this "rollback remark" could also make sense)

fgalan: Do we really need "extern" here? (typically, when we use "extern" in the .h, we use nothing in the .cpp...)

src/lib/rest/rest.cpp
```diff
@@ -835,6 +834,14 @@ int servicePathSplit(ConnectionInfo* ciP)
     LM_I(("Service Path %d: '%s'", ix, ciP->servicePathV[ix].c_str()));
   }
 
+  //
+  // stringSplit destroys ciP->servicePath.
```
fgalan: This is obscure... instead of manipulating `ciP->httpHeaders.servicePath` after servicePathSplit() processing so it points to the first element to be prepared for add() to work, we should directly pass `ciP->httpHeaders.servicePath[0]` to add().

test/functionalTest/cases/2750_common_metrics/2750_complex_service_path.test
```diff
@@ -0,0 +1,184 @@
...
+HTTP/1.1 200 OK
+Content-Length: 2
+Content-Type: application/json
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+[]
+
+
+05. Ask for Metrics, see A1 and A2 as service paths, as well as two 'default service path'
```
fgalan: Maybe you refer to 1 default service and 2 root service paths?

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
```diff
@@ -90,12 +90,13 @@ brokerStart CB
 # 05. Send a request with an Invalid Service Path
 # 06. Send a request with an Invalid Content-Type
 # 07. Send a request with invalid value for URI param /options/
-# 08. Send a request with Content-Length > PAYLOAD_MAX_SIZE
+# 08. Send a request with a component-name too long in ServicePath
```
fgalan: This seems not being related with the PR... I understand that is a side-fix...

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Yes, I tried those. The broker says 'illegal character' for a spath of /A#

src/lib/rest/RestService.cpp
kzangeli: Yeah, 'extern' is not needed. I left it to be more clear.
Look at the issue #2781. All the info is there.
Enough, right?

src/lib/rest/rest.cpp
kzangeli: Wish it was that easy. That was my first attempt, of course
From httpRequestSend() we don't have that information, unfortunately.

test/functionalTest/cases/2750_common_metrics/2750_complex_service_path.test
kzangeli: I refer to 'A1', 'A2', and '/#' twice.   Once explicit and once implicit

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
kzangeli: Yeah, this is an older change. The ToC wasn't changed when it was supposed to be

src/lib/metricsMgr/MetricsManager.cpp
fgalan: Thus, controlled. NTC

src/lib/rest/RestService.cpp
fgalan: If our common style in other parts is not including extern in this case (as I understand it is) I'll suggest to do it in the same way also here.

test/functionalTest/cases/2750_common_metrics/2750_complex_service_path.test
fgalan: It is a bit confusing as we have de *default* service and the  *root* service path. The *default service path* is an ambiguous concept... (default depends on the request, either queries or updates). So, I'll suggest be clearer here and use "root service path".

test/functionalTest/cases/2750_common_metrics/2750_incoming_errors.test
fgalan: NTC

src/lib/metricsMgr/MetricsManager.cpp
fgalan: In addition, I'd suggest to move these checking to an additional (class private?) function, so add() could be kept simple.

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in faf8469

src/lib/common/defaultValues.h
kzangeli: Fixed in faf8469

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 07f1987

src/lib/rest/RestService.cpp
kzangeli: Removed the 'extern' keyword in f3a230e.

kzangeli: Added comment about the function in 8a61075

test/functionalTest/cases/2750_common_metrics/2750_complex_service_path.test
kzangeli: Sure, I still see it as *default* service path (as it is used if no service path is given), but it is true we have two different 'default service paths'

kzangeli: Fixed in 500958c

src/lib/rest/rest.cpp
kzangeli: Fixed in acc0506
Using ciP->servicePath[0] wherever possible in calls to metricsMgr.add()

src/lib/rest/httpRequestSend.cpp
```diff
@@ -266,8 +267,21 @@ int httpRequestSendWithCurl
   int                             outgoingMsgSize       = 0;
   std::string                     content_type(orig_content_type);
   std::map<std::string, bool>     usedExtraHeaders;
+  char                            servicePath0[64];  // 64 > 50 (max component length in service path)
```
fgalan: Including slashesh which separate components?

By the way, we should have the max for SP string in limits.h... Otherwise, it should be added.

kzangeli: Ah, of course. I meant to change that 64, of course!

src/lib/metricsMgr/MetricsManager.cpp
```diff
@@ -131,15 +183,57 @@ void MetricsManager::add(const std::string& srv, const std::string& subServ, con
     return;
   }
 
+  if (serviceValid(srv) == false)
+  {
+    return;
+  }
+
+  char* toFree      = strdup(subServ.c_str());
```
fgalan: Discussed at skype: try to move the "SP calculation logic" as part of subServiceValid() funciton (changing the function a bit... it should return the pointer to the SP char-string to use, probably the name should be also adjusted).

src/lib/rest/httpRequestSend.cpp
```diff
@@ -266,8 +267,21 @@ int httpRequestSendWithCurl
...
+  char*                           spEnd;
 
-  metricsMgr.add(tenant, servicePath, METRIC_TRANS_OUT, 1);
+  memset(servicePath0, 0, sizeof(servicePath0));
+
+  if ((spEnd = strchr((char*) servicePath.c_str(), ',')) != NULL)
```
fgalan: Discussed at skype to move this part to a separate function.

src/lib/rest/httpRequestSend.cpp
```diff
@@ -652,6 +666,12 @@ int httpRequestSend
   get_curl_context(_ip, &cc);
   if (cc.curl == NULL)
   {
+    //
+    // FIXME P4: servicePath may contain more than one service path, i.e.  /A,/B.
+    //           Only if coming from a forward of a query (postQueryContext())
+    //           This is taken care of in metricsMgr.add(), but could be moved here instead
+    //           as it is the only place where this can happen (99% sure about this).
+    //
```
fgalan: Discussed at skype, to use the funtion mentioned here https://github.com/telefonicaid/fiware-orion/pull/2785/files#r93621406

src/lib/metricsMgr/MetricsManager.cpp
kzangeli: Fixed in 0c19c82

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in 0c19c82

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in 0c19c82

src/lib/rest/httpRequestSend.cpp
kzangeli: Fixed in 0c19c82

src/lib/rest/httpRequestSend.cpp
```diff
@@ -212,6 +213,28 @@ static int contentLenParse(char* s)
 
 /* ****************************************************************************
 *
+* servicePathForMetrics - extract first component of service-path
+*/
+static void servicePathForMetrics(const char* servicePath, char* servicePath0, int servicePath0Len)
```
fgalan: Discussed at skype.

* New name: firstServicePath
* Move to the library which deals with other service path manitpulations, e.g. spliting service path header string into vector. It seems that library is rest.cpp/h

kzangeli: Fixed in e2b1d24

fgalan:
LGTM

PR #2787, opened 2016-12-22
service path cleanup

### LINKED ISSUE #2786: More order in service path
Right now, ConnectionInfo has three different fields to hold the service path:

1. ciP->httpHeaders.servicePath (The incoming value of the HTTP header)
2. ciP->servicePathV  (vector of service paths)
3. ciP->servicePath     (a copy of httpHeaders.servicePath that we don't really need)

This 'disaster' makes the broker more complex that it needs to be and it should be fixed.
Item 3 should be removed and item 1, that is destroyed when item 2 is created should not be destroyed.

Item 1 is needed for forwarding requests, *an untouched service path*, while for metrics we need item 2, to be specific, the first member of the vector: ciP->servicePathV[0].

kzangeli:
Service path cleanup
Fixes #2786.

This PR contains only internal changes, so no entry in CNR.

### DIFF
--- a/src/lib/rest/ConnectionInfo.h
+++ b/src/lib/rest/ConnectionInfo.h
@@ -59,7 +59,6 @@ public:
     inMimeType             (JSON),
     outMimeType            (JSON),
     tenant                 (""),
-    servicePath            (""),
     payload                (NULL),
     payloadSize            (0),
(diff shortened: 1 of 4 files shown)


fgalan:
LGTM

Issue #2691, opened 2016-11-17
Csub cache improvements


fgalan:
From https://github.com/telefonicaid/fiware-orion/pull/2638#discussion_r88472767:

> Yes... I remember that. These were the days of research on weird crashed that we suspected were in due to cache implementation but and the were (most probably) caused by a bug in the usage of mongo driver and we over-simplified the cache trying to find the bug.
>
> In my opinion that 50% of char* should be "recovered" into std::string to align with the rest of the CB code. In addition, the linked-list structure could be probably be replaced by a std::map (using subscription id) as key to improve search time (O(log n) vs O(n)).

Thus, in sum, csub cache should be modified to:

* Use std::string instead of char* in the places we typically use in the other places of CB code
* Use std::map instead of a linked list of items

fgalan:
Implemented by PR #2790

fgalan:
PR #2760 was closed at the end. We have reasonable doubts about the change to std::map is a good idea, as the subscription id is not the field that is used to search for triggered subscriptions (we look for entities and attributes, basically). Thas item has been removed from the issue.

But PR #2760 has also a part regarding chaning from char* to std::string that could be useful. Thus, the branch hardening/2691_sub_cache_using_std_map is still there, in the case it could be useful for that (maybe not, if the csub cache implementation has changed so far since the PR was branched from master). To be evaluated by the one which implement this issue at the end (if anybody ;)

Issue #2216, opened 2016-05-31
Duplicated subscription in the subscription cache


kzangeli:
We have seen that sometimes a subscription is present TWICE in the subscription cache.
Probably at subscription update, the subscription is added and not updated.
Remember that the subscription-update in the cache is done this way:
* insert new subscription
* remove the old one

Perhaps that removal of the old one fails sometimes ...
This is a bug and it must be fixed.

The functional test `1048_subscription_cache/csub_update_with_notif.test` can be used to provoke the bug, If the synchronization of the subscription cache encounters more than one item of the same subscription in the cache, there you have the bug reproduced. Look for the FIXME with the text 
(Issue XXX) in subCache.cpp (function subCacheSync).

fgalan:
Maybe the recent fixes in #1406 also solved this problem? A re-check worth the effort in my opinion...

kzangeli:
Trying to reproduce this issue I modified `src/lib/cache/subCache.cpp`:

```
    if (savedSubV[cSubP->subscriptionId] != NULL)
    {
#if 0
      // This is the workaround
      cSubP = cSubP->next;
      continue;
#else
      // This is new, to detect the bug
      LM_E(("duplicated subscription: %s", cSubP->subscriptionId));
      exit(73);
#endif
```

And, then I used a simple script:

```
typeset -i tNo
tNo=0
while /bin/true
do
  testHarness.sh csub_update_with_notif.test
  if [ $? != 0 ]
  then
    echo error detected
    exit 1
  fi

  tNo=$tNo+1
  echo '*** ' $tNo tests
  echo
  echo
done
```

After 100 iterations, still no error detected ...

kzangeli:
Made a try in the branch `hardening/2788_expired_subscriptions_not_in_sub_cache`, which is a slightly modified `hardening/2691_sub_cache_using_std_map` (for PR #2790 ), like this:
```
kz@hoppe:fiware-orion> git diff
diff --git a/src/lib/cache/subCache.cpp b/src/lib/cache/subCache.cpp
index 92c0ca4..0533a65 100644
--- a/src/lib/cache/subCache.cpp
+++ b/src/lib/cache/subCache.cpp
@@ -1070,6 +1070,8 @@ void subCacheSync(void)
     //
     if (savedSubV[cSubP->tenant + cSubP->subscriptionId] != NULL)
     {
[...]

PR #2790, opened 2016-12-29
Hardening/2691 sub cache using std map

kzangeli:
Fixes issue  #2691.
Implemented the subscription cache using a std::map instead of the linked list that was previously used.
Also, changed all use of 'char*' to use 'std::string' instead.
Full valgrind test after implementing this. 0 leaks in the branch.
Tested the branch with the script proposed for issue #2216, and the issue is gone.


src/lib/cache/subCache.cpp
```diff
@@ -331,41 +337,38 @@ static bool attributeMatch(CachedSubscription* cSubP, const std::vector<std::str
...
+  if (servicePath == "/#")
   {
     return true;
   }
 
   //
   // Empty service path?
   //
-  if ((strlen(servicePath) == 0) && (strlen(cSubP->servicePath) == 0))
+  if ((servicePath == "") && (cSubP->servicePath == ""))
```
fgalan: Maybe a more efficient way of checking string emptyness is to use the length() method, i.e.

```
if ((servicePath.length() == 0) && (cSubP->servicePath,lengt() == 0))
```

However, I'd suggest not chaning that in this PR, but move it to an external issue, as it could impact other places in the code.  In addition, if we agree on that, we should include it as rule in the coding style guide. 

(Probably the library implementation is clever enough so not sure if this could make any real difference... but probably worth the change anyway as it is easdy to do, i.e. just grep for `""` along the code).

src/lib/common/limits.h
```diff
@@ -94,6 +95,7 @@
 */
 #define  LOCAL_IP_V6  "::"
 #define  LOCAL_IP_V4  "0.0.0.0"
+#define  MAX_LEN_IP   64
```
fgalan: The reational behind this limit calculation should be added as comment around this line... It also works for IPv6 addresses (it seems so, but think about it for a moment...)?

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -683,40 +685,55 @@ void updateInCache
...
-    subCacheUpdateStatisticsIncrement();
-    if (subCacheP != NULL)
-    {
-      LM_T(LmtSubCache, ("Calling subCacheItemRemove"));
-      subCacheItemRemove(subCacheP);
-    }
+    subCacheStatisticsIncrementUpdates();
+
+    //
+    // To simulate the old way of updating:
```
fgalan: Not sure about the purpose of this comment... You are commenting the way in which the implementation previously worked? Or the way in which currently work due to "legacy"?

src/lib/cache/subCache.cpp
```diff
@@ -33,6 +33,7 @@
```
fgalan: Although there shouldn't be any functional change with regard this PR, I think a CNR entry should be included, due to:

1. It may have a potential impact (let's hope that benefitial) in performance
2. The change is critical enougth to appear in future release notes. This is the kind of things that could be tricky to realize when unknown problems occur :)

Something like this:

```
Hardening: csub cache implemenetation now uses hashmap instead of linked list, thus potentially improving search performance
```

In addition, I understand this PR could be the solution for #2216. In that case, another additional entry (of kind "Fix") should be also added.

test/functionalTest/cases/0000_json_parse/jsonParsePostUpdateContextSubscription.test
```diff
@@ -28,8 +28,15 @@ dbInit CB
 brokerStart CB
```
fgalan: Just to check: is this PR including any "real" change in .test?

I have seen some improvements (e.g. typo fixing, steps preamble in comments, attribute value changes in order to align them with step number) but not sure if is there anything else appart fom that.

src/lib/cache/subCache.cpp
fgalan: I have asked about this at SOF: http://stackoverflow.com/questions/41425569/checking-for-empty-string-in-c-alternatives

fgalan: So, after looking to the anwers, it seems that the recommended way is not one nor the other :) but this:

```
if ((servicePath.empty()) && (cSubP->servicePath.empty()))
```

fgalan: Moved to issue https://github.com/telefonicaid/fiware-orion/issues/2791

NTC

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: Option 2: this is the way it works currently due to legacy.

If we change this, then the stats of the sub-cache will change and I'm not sure we want that.
It's just about a counter, not so important.

test/functionalTest/cases/0000_json_parse/jsonParsePostUpdateContextSubscription.test
kzangeli: Only beautifying old tests

src/lib/common/limits.h
kzangeli: Actually I only moved this line from another place.
64 should be enough for IPv6. For IPv4 you only need 3+1+3+1+3+1+3+1 == 16 (terminating zero included)

src/lib/cache/subCache.cpp
kzangeli: Sure.
True, I did a pretty thorough test of #2216 and was unable to reproduce it.

kzangeli: About performance ... I'm afraid we will not gain but lose a little.
The only operation that will benefit from the new implementation should be
`GET /v2/subscriptions/{SUB-ID}`.

fgalan: Thus, the "thus potentially improving search performance" part should be removed and maybe mentioning something about code cleaness or similar...

src/lib/common/limits.h
fgalan: For IPv4 the calculation is perfect... What about a similar calculation for IPv6?

src/lib/mongoBackend/mongoUpdateSubscription.cpp
fgalan: sub-cache stats were designed to help debugging the cache part. If the cache way of working has severely changed, the stats should "follow" the new implementation, i.e. maybe some fields doens't make sense any longer ans dome others need to be added.

Thus, I'd "break" legacy in this case and change it.

test/functionalTest/cases/0000_json_parse/jsonParsePostUpdateContextSubscription.test
fgalan: NTC

src/lib/common/limits.h
kzangeli: OK, looked up the format of IPv6:
```
2001:0db8:85a3:08d3:1319:8a2e:0370:7334
```

As a string, 8 chunks of 4 chars separated by 7 colons (and the ending 0): 8 * 5 == 40.
64 is more than enough

src/lib/cache/subCache.cpp
kzangeli: Fixed in 1f9b497

src/lib/common/limits.h
kzangeli: Fixed in 1f9b497

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: Fixed in 1f9b497

src/lib/common/limits.h
```diff
@@ -91,9 +92,21 @@
...
+* About MAX_LEN_IP:
+* An IPv4 address has the following format:
+*   xxx.xxx.xxx.xxx
+* As a string, 4x4 (16) chars are needed to hold the address (including the zero termination char)
+*
+* An IPv6 address has the following format:
+*   xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:
+* As a string, 8x5 (40) chars are needed to hold theaddress(including the zero termination char)
+*
+* 64 is more than sufficient to holsd any numerical IP address as a string
```
fgalan: Typo: to holsd

fgalan: Fixed in b8c7818077bfe554c3d572ef25369627f83754a3

fgalan:
LGTM, but waiting to @iariasleon report before merging this.

src/lib/cache/subCache.cpp
```diff
@@ -859,74 +790,46 @@ void subCacheItemInsert
...
-      //
-      // If "msg" and ", " has no room in the "list", then no list is shown.
-      // (the '+ 2' if for the string ", ")
-      //
-      if (strlen(msg) + 2 > bytesLeft)
+      if (*listP != "")
       {
-        snprintf(list, listSize, "too many subscriptions");
-        break;
+        *listP += ", ";  // FIXME PR: This SPACE is here for historical reasons. Do we want it?
```
kzangeli: Not important. If we remove it, some functests will fail, I imagine

src/lib/cache/subCache.cpp
```diff
@@ -1242,23 +1119,25 @@ void subCacheSync(void)
...
+    CachedSubSaved*     cssP  = savedSubV[cSubP->tenant + cSubP->subscriptionId];
 
     if (cssP != NULL)
     {
-      std::string tenant = (cSubP->tenant == NULL)? "" : cSubP->tenant;
+      // FIXME PR: the original line before conflict solving was the one commented.
+      // I leave this mark here, so we can think about this at merge time
+      //
+      //std::string tenant = (cSubP->tenant == NULL)? "" : cSubP->tenant;
+      std::string tenant = cSubP->tenant;
```
kzangeli: The change is good. cSubP->tenant was `char*` but now is a `std::string`

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -283,15 +282,22 @@ int mongoSubCacheItemInsert
   //
   // 01. Check incoming parameters
   //
-  if ((subscriptionId == NULL) || (subscriptionId[0] == 0))
+  // FIXME PR: check the commented lines. They correspond to conflict solving
+  //
+  //if ((subscriptionId == NULL) || (subscriptionId[0] == 0))
+  if (subscriptionId.empty())
```
kzangeli: OK (subscriptionId was `char*` but now is a `std::string`)

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -283,15 +282,22 @@ int mongoSubCacheItemInsert
...
-    return -2;
-  }
+  // FIXME PR: empty servicepath is legal, in means "default".
+  // Thuss this checking is not meaninfull anymore and -2 removed
+  // from the list (and probably offset -3, -4, etc. a step)
+  //if ((servicePath == NULL) || (servicePath[0] == 0))
+  //if (servicePath.empty())
+  //{
+  //  return -2;
+  //}
```
kzangeli: empty servicepath may be legal in the "beginning of the request", but here ... not so sure.
I mean, if no service path is given (or empty service path), then we "force" the default service path on the request.
But in this point of the execution, if service path is empty, something has gone wrong, I believe.
Then again, checking for this may be a little paranoid, except for when dealing with pointers as NULL pointers may crash the system if used. 

OK to remove the entire check

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -803,15 +802,16 @@ void updateInCache
...
-    if (subCacheP != NULL)
+    subCacheStatisticsIncrementUpdates();
+  // FIXME PR: I understand this block is not needed with std::map based
+  // logic
+  /*  if (subCacheP != NULL)
     {
       LM_T(LmtSubCache, ("Calling subCacheItemRemove"));
       subCacheItemRemove(subCacheP);
     }
+   */
```
kzangeli: I think this part needs to be as it was.

subCacheP comes from a lookup in the cache (a call to subCacheItemLookup) and it can perfectly be NULL, which only means that the subscription didn't exist in the cache (new creation).

If it didn't previously exist, it cannot be removed.
subCacheItemRemove may even crash the system is called with a NULL pointer ...

fgalan: Look to this part in the very first commit of this PR: https://github.com/telefonicaid/fiware-orion/pull/2790/commits/91ab399c4de7816cf876155efb7fc0666b3b7c83#diff-dfc89b08d3094a4b23a8ac8b936b6d98L712

In fact, I think that not doing this change was crashing CB in some functest.

src/lib/cache/subCache.cpp
kzangeli: Actually, let's keep this space. Backward compatibility ...

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: Yeah, looking closer at it, it becomes clear. The cache now works differently and the remove should never be done. We no longer "add then remove" items but they are overwritten

src/lib/cache/subCache.cpp
kzangeli: Fixed (removed comment) in b6f39b1

src/lib/cache/subCache.cpp
kzangeli: Fixed (removed comment) in b6f39b1

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed (removed comment) in b6f39b1

src/lib/mongoBackend/mongoSubCache.cpp
kzangeli: Fixed (removed comment) in b6f39b1

src/lib/mongoBackend/mongoUpdateSubscription.cpp
kzangeli: Fixed (removed comment) in b6f39b1

kzangeli:
I'd propose to close this PR

fgalan:
Almost a year withoug activity... yes, probably is better to close. Specially when the root issue will remain opened and a note about this can be added to it.

PR #2795, opened 2017-01-03
[CATCH] exception in remote connection

iariasleon:


test/loadTest/connections_stress_tests/connections_stress_tests.py
```diff
@@ -148,31 +148,35 @@ def __get_connection_info(self):
              can have a name like "contextBroker-1.4.3..." and the logic will still work
         :return tuple (established_conn, close_wait_conn)
         """
-        conn = rpyc.classic.connect(self.host)
-        connections_est = """def get_number_of_connections():
```
fgalan: Better to keep `connections_est` where it is now and move only the `conn` sentence inside the try. That way, the try block is shorter and, at the same time, the PR simpler to review.

fgalan: where it is now -> I mean, "now" at master (i.e. previus to this PR proposed changes)

iariasleon: Fixed in  	fdb3329

fgalan:
LGTM

Issue #2760, opened 2016-12-09
Clean render()/check() NGSIv1 method from indent argument passing


fgalan:
The removal work has been already done at branch haderding/remove_ngsiv1_indent (note the typo 'haderding' in the branch name). 

However, in order to have a "safe measure" in the case that some API client in IoTP were (wrongly) processing the result in text-positional way (instead of the right way in JSON way) we should implement a formatting step just before response NGSIv1 payloads in order to keep 100% backward compatibility (the existing .test and unit test will guarantee that). This formatting step should be activable by configuration (e.g. `-legacyNgsiv1Indent`) and it is thought as a provisional measure to be removed (including the CLI) once we check that no IoTP case breaks.

The builtin PrettyWriter from rapidjson could do the trick (see https://github.com/telefonicaid/fiware-orion/blob/43e96f984826bb7ad9c1b17a32acc8d891f63a86/src/lib/rest/restReply.cpp#L70). However, it seems that we are using an extra whitespace before the `:` (different to the usual way in any JSON pretty printer out there) so probably we would need to re-implement a special pretty printer ourselves :(

fgalan:
We are now (*) in and "almost" finished status with regards to this piece of work, and the situation is as follows.

Branch haderding/remove_ngsiv1_indent implements a refactor in render()/check() code to remove all the whitespace and all the newlines in all the NGSIv1 renders. 

In addition, the `-paranoidV1Indent` CLI parameter can be used to get the *exact* NGSIv1 renderings that we had before haderding/remove_ngsiv1_indent (at least as far we have checked with existing .test suite) *except* in the following cases:

* Case A: `GET /v1/contextEntities/<type>` response is correctly idented (before haderding/remove_ngsiv1_indent some elements were incorrently over-indent by 2 spaces). Example on how it was before haderding/remove_ngsiv1_indent:

```
{
    "name" : "Thing",
    "attributes" : [
      "A1"
    ],
  "statusCode" : {
    "code" : "200",
    "reasonPhrase" : "OK"
  }
```

* Case B: correct compound metadata values rendered. Previously it was without indent, printing the `"value"` plus the value in one line, with trailing `}` in the same line. Example on how it was before haderding/remove_ngsiv1_indent:

```
{
  "contextResponses" : [
    {
      "contextElement" : {
        ...
        "attributes" : [
          {
            "name" : "A1",
            "type" : "Number",
            "value" : 1,
            "metadatas" : [
              {
[...]

PR #2798, opened 2017-01-05
FIX use --prety-print in accumulstorStart in all .test in all cases where it makes sense

fgalan:
I needed this as I worked on #2760. Otherwise, all the all cases in NGSIv1 gets collapsed to one line after v indent refactoring.

Note that althoguth the PR is pretty large, no Content-Lenght has been changed, ensuring that it isn't any actual change in the .test. This is the only aspect relevant to be reviewed in this PR, the actual changes in the payload shown by the diff are not relevant.


crbrox:
LGTM

PR #2799, opened 2017-01-05
Prepare SUB_ID, REG_ID, etc. grab code in .test to work even when NGSIv1 responses are one-liners instaead of multi-line pretty-print

fgalan:
Note there aren't actual changes in .test (only in the way in which REG_ID is grabbed in some cases)

### DIFF
--- a/test/functionalTest/cases/0000_log_operation/log_rest.test
+++ b/test/functionalTest/cases/0000_log_operation/log_rest.test
@@ -136,6 +136,7 @@ Date: REGEX(.*)
 
 
 
+
 === 5. PUT trace level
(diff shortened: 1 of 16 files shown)


test/functionalTest/cases/0000_log_operation/log_rest.test
```diff
@@ -136,6 +136,7 @@ Date: REGEX(.*)
 
 
 
+
```
fgalan: Not sure why the new way of dealing `_response` requires this .test to add an empty line to work... However, it doesn't seem to be important.

crbrox:
LGTM

Issue #2796, opened 2017-01-03
Overlogging metadata abscense in csub docs


fgalan:
Orion shows this message in logs:

```
time=2016-11-28T15:43:51.854Z | lvl=ERROR | corr=8539470c-e112-42ab-8ff6-160062f172d8 | 
trans=1480318521-946-00000025921 | from=10.0.0.1 | srv=bar| subsrv=/foo | comp=Orion | 
op=safeMongo.cpp[297]:setStringVector | msg=Runtime Error (object field 'metadata' is 
missing in BSONObj <> from caller setNotification:141)
```

(The setNotification:141 and safeMongo.cpp:[297] line references work at 1.4.6 85ffcff5bdd506cbb6de1fd348bbbbf1293af65c)

This is not actually a lvl=ERROR trace. In fact, is shouldn't even appear in the logs. It is just that csub document lack of the metadata field (which is optional).

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2800

fgalan:
Ready for QA Comp testing, @iariasleon

iariasleon:
LGTM...Tested in CB version:
```
  "version" : "1.6.0-next",
  "git_hash" : "6232159b397f40aff97abd497b1e142a510e3e6d",
```

PR #2800, opened 2017-01-05
FIX wrongly overlogging metadata abscense in csub docs as Runtime Error

fgalan:
Fixes https://github.com/telefonicaid/fiware-orion/issues/2796

Tested:

* 100% unit test regression ok
* 100% ft regression ok
* Manual test:
  * Create a subscription
  * Remove `metatada` field in DB
  * `GET /v2/subscriptions` and check that no log output is generated


kzangeli:
LGTM

Issue #2807, opened 2017-01-11
Check functionality of quotes in string filters


kzangeli:
We should do a review of use cases with quote (') described in the "Simple Query Language" section and check that all of them are correctly implemented.

[ But not create new test cases for already covered cases ]

fgalan:
This issue is an spin-off of https://github.com/telefonicaid/fiware-orion/issues/2805.

We have the following cases

* Number as string
* Bool as string
* Null as string

According to https://github.com/telefonicaid/fiware-orion/issues/2805#issuecomment-271856667, the fist one is already covered by 1595_query_with_q_and_string_values/query_with_q_and_string_values.test so this issue should focus in second and third case.

fgalan:
.test to cover all cases in PR https://github.com/telefonicaid/fiware-orion/pull/2810

PR #2810, opened 2017-01-12
Functest for #2807

kzangeli:
Assures that #2807 works correctly.

I added a test with int/string-int, but did not add a test for false ('true' is present).
Seems enough as it is, if anybody wants changes, it is very easy.


crbrox:
LGTM

fgalan:
LGTM

Issue #2811, opened 2017-01-13
Possible strange behaviour on too long tenants


kzangeli:
Some strange behaviour has been detected, creating a subscription with a long Service/Subservice, via 'portal' without  getting any error but the subscription is later not found.

Also, in the traces of the broker, the following line is detected:

```
op=mongoSubCache.cpp[468]:mongoSubCacheRefresh | msg=Runtime Error (exception in nextSafe(): nextSafe(): { $err: "Invalid ns [orion-dddddddddddddddddddddddddddddddddddddddddddddddddd/.csubs]", code: 16256 } at mongoSubCacheRefresh:466 - query: {})
```

The trace line is seen once every ~60 seconds and all points to the refresh of the sub cache.
#### Note
this trace line may not have anything to do with the problem of not creating the subscription, but, an error is detected and a strange trace is there ... might not be a coincidence ...

kzangeli:
Doing tests creating a subscription with a long name of tenant, I have seen that the broker accepts tenants of up to 50 chars in length, but after that some internal buffer of only 51 chars (50 + zero termination) is too small and the sub creation seems to fail, even though the broker responds with `200 OK`.

If the tenant is > 50 chars long, the broker correctly responds with an error.

Now, remember that the tenant name is added to the database name (which is "orion" by default but that can be changed with the `-db` CLI option).

Tenant 'T1' yields the database name "orion-T1".

So, that buffer of 51 chars needs to be larger:
```
50 + 1 + strlen(database-name)
```

Again, this may have nothing to do with the problem described in this issue (the error in the log file is a different one), but ...

This here is the error in the orion log file by the way:
```
... comp=Orion | op=string.cpp[631]:strToLower | msg=Runtime Error (cannot copy 52 bytes into a buffer of 51 bytes)
```

fgalan:
PR solving the sub-issue at https://github.com/telefonicaid/fiware-orion/issues/2811#issuecomment-272406993: https://github.com/telefonicaid/fiware-orion/pull/2812

fgalan:
After skype discussion, to be closed by @ jcanonav with about brief comment about the reason for doing so.

jcanonav:
Check in tenant with 50 chars long. Buffer size has increased  to 61 ("db-prefix"+"-"+"tenant_name") and works rigth

fgalan:
Closing on behalf @jcanonav

PR #2812, opened 2017-01-13
Bug/2811 long service names in subscriptions

kzangeli:
Fixes a bug that was discovered looking into issue #2811.

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -1320,10 +1320,29 @@ static int connectionTreat
       lmTransactionSetFrom(ip);
     }
 
-    char tenant[SERVICE_NAME_MAX_LEN + 1];
-    ciP->tenantFromHttpHeader = strToLower(tenant, ciP->httpHeaders.tenant.c_str(), sizeof(tenant));
+    //
+    // Max name length of a database name for a tenant:
+    // Remember the database name is created like this:
+    //   orion-<tenant>
+    //
+    // 'orion' is the default database name prefix, but this
+    // string is configurable via CLI.
+    // However, its maximum allowed length is SERVICE_NAME_MAX_LEN, so
+    // this we can always use.
+    //
+    // Now:
+    //   DB_NAME_MAX_LEN + 1 + SERVICE_NAME_MAX_LEN + 1:
+    //     DB_NAME_MAX_LEN:      'orion'
+    //     1:                    '-'
+    //     SERVICE_NAME_MAX_LEN: tenant name, via HTTP header 'Fiware-Service' 
+    //     1:                    zero-termination of the string
+    // 
+    //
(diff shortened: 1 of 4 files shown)


crbrox:
LGTM

src/lib/rest/rest.cpp
```diff
@@ -1320,10 +1320,29 @@ static int connectionTreat
...
+    //
+    // Now:
+    //   DB_NAME_MAX_LEN + 1 + SERVICE_NAME_MAX_LEN + 1:
+    //     DB_NAME_MAX_LEN:      'orion'
+    //     1:                    '-'
+    //     SERVICE_NAME_MAX_LEN: tenant name, via HTTP header 'Fiware-Service' 
+    //     1:                    zero-termination of the string
+    // 
+    //
+    char tenant[DB_NAME_MAX_LEN + 1 + SERVICE_NAME_MAX_LEN + 1];
```
fgalan: It would be simpler with `DB_NAME_MAX_LEN + 1 + SERVICE_NAME_MAX_LEN + 1` as a new constant in the limits.h file (moving also the above explanation in the comment to limits.h).

src/lib/rest/rest.cpp
```diff
@@ -1320,10 +1320,29 @@ static int connectionTreat
...
+
     ciP->tenantFromHttpHeader = strToLower(tenant, ciP->httpHeaders.tenant.c_str(), sizeof(tenant));
+    ciP->outMimeType          = mimeTypeSelect(ciP, "Normal");
```
fgalan: Not actually related with this PR but... which is the meaning of "Normal" outMimeType?

kzangeli: Not a clue, I just indented the line :-)

Looked it up:

```
static MimeType mimeTypeSelect(ConnectionInfo* ciP, const std::string& what)
{
  if ((what == "Error") ||   // JSON Error to be returned
      (what == "Normal"))    // Normal handling - same as Error
  {
    if (ciP->httpHeaders.accepted("application/json"))
    {
      return JSON;
    }
    if (ciP->httpHeaders.accepted("text/plain"))
    {
      return TEXT;
    }

    return JSON;
  }

  return JSON;
}
```

As it seems, it is not used. "Error" and "Normal" are treated the same way.

I can fix that entire thing inside this PR, if you like

src/lib/rest/rest.cpp
kzangeli: Yes, not a bad idea.
Name of the new constant:
DB_AND_SERVICE_NAME_MAX_LEN ?

src/lib/rest/rest.cpp
kzangeli: Fixed in 7d37094

src/lib/rest/rest.cpp
kzangeli: Fixed in 7d37094

fgalan:
LGTM

Issue #2280, opened 2016-06-15
in POST v2/subscriptions if notification httpCustom url has invalid value the subsc is created


iariasleon:
in POST v2/subscriptions if notification httpCustom url has invalid value the subsc is created
#### dataset used (replace these values into `url` field in request below)

```
      | url               |
      |-------------------|
      | ws://             |
      | ://localhost      |
      | http//localhost   |
      | http:/localhost   |
      | http:localhost    |
      | http://localhost: |
      | http://           |
      | //a               |
      | ///a              |
      | foo.com           |
      | rdar://1234       |
      | h://test          |
      | http://           |
      | ://               |
      | ftps://foo.bar/   |
```

#### Second dataset with the same behavior

```
      | url                        |
      |----------------------------| 
      | http://localhost:900000    |
      | http://localhost:dsfsdf    |
      | http://localhost\my_path   |
      | http://e34.56.45.34        |
      | http://34,56.45.34         |
      | http://34.56.45.34;3454    |
      | http://34.56:3454          |
      | http://.                   |
      | http://..                  |
      | http://../                 |
      | http://?                   |
      | http://??/                 |
      | http://#                   |
      | http://##/                 |
      | http://foo.bar/foo(bar)baz |
      | http://-error-.invalid/    |
      | http://a.b--c.de/          |
      | http://-a.b.co             |
      | http://a.b-.co             |
      | http://0.0.0.0             |
      | http://1.1.1.1.1           |
      | http://3441.2344.1231.1123 |
      | http://123.123.123         |
      | http://3628126748          |
      | http://.www.foo.bar/       |
      | http://www.foo.bar./       |
      | http://.www.foo.bar./      |
```

#### Third dataset:

```
      | url       |
      |-----------|
      | //        |
      | ///       |
      | http:///a |
```
#### subscription request

```
url: POST http://qa-orion-fe-02:1026/v2/subscriptions
headers:
    Content-Type: application/json
    Fiware-Service: test_notif_http_custom_url_error
    Fiware-ServicePath: /test
payload: {"notification": {"httpCustom": {"url": "ftps://foo.bar/"}, "attrs": ["temperature"]}, "expires": "2016-04-05T14:00:00.00Z", "subject": {"entities": [{"idPattern": ".*"}], "condition": {"attrs": ["temperature"]}}}
```
#### request response

```
http code: 201 - Created
headers:
   date: Wed, 15 Jun 2016 11:23:21 GMT
[...]

PR #2815, opened 2017-01-16
Bug/2280 invalid url in subscription creation

kzangeli:
Partial fix of #2280 (actually, everything that we can possibly do at sub-creation time)

Custom URLs in subscription had no sanity check whatsoever.
It is impossible to do a sanity check if the URL contains variable substitutions, with values coming from perhaps not yet created entities.
However, if the URL contains no var substitutions, the same sanity check as for 'normal', n on-custom URLs can be used.
This PR implements this.


fgalan:
LGTM

crbrox:
LGTM

Issue #1994, opened 2016-04-04
[BUG] in POST v2/subscriptions with "q" condition having invalid chars


iariasleon:
in POST v2/subscriptions with "q" condition having invalid chars
#### dataset

```
      | q           |
      |-------------|
      | house<flat> |
      | house=flat  |
      | house"flat" |
      | house'flat' |
      | house;flat  |
      | house(flat) |
      | house_?     |
      | house_&     |
      | house_/     |
      | house_#     |
      | my house    |
```
#### subscription request

```
POST http://localhost:1026/v2/subscriptions
Content-Type: application/json
Fiware-Service: test_condition_expression_error
Fiware-ServicePath: /test
{"notification": {"callback": "http://localhost:1234", "attributes": ["temperature"]}, "expires": "2016-04-05T14:00:00.00Z", "subject": {"entities": [{"idPattern": ".*", "type": "room"}], "condition": {"attributes": ["temperature"], "expression": {"q": "house;flat"}}}}
```
#### subscription response

```
http code: 201
date: Mon, 04 Apr 2016 10:50:40 GMT
connection: Keep-Alive
content-length: 0
location: /v2/subscriptions/570247000f6b2ac1b6346f02
```
#### expected response

```
http code: 400
{"error":"BadRequest","description":"Invalid characters in condition expression"}
```

```
```

iariasleon:
In the version:
```
  "version" : "1.0.0-next"
  "git_hash" : "fee3d035442e78c1a2e19720f68bdeba083ca560",
```
And they respond:
```

Only in this case `house=flat`, it returns a payload malformed (only description) and 200 in http code.
```
http code: 200
date: Fri, 29 Apr 2016 12:02:28 GMT
fiware-correlator: 3ecb594a-0e02-11e6-bf64-005056a20feb
connection: Keep-Alive
content-type: application/json
content-length: 40
invalid character found in URI param /q/
```

**Comment by iariasleon** [comment]
In CB version:
```
 "version" : "1.1.0-next",
  "git_hash" : "301c258a9ef8f05cabdc276b7e21365e8f4d5913"
```

These cases respond correctly, with a json object and the description is `invalid character found in URI param /q/`
```
      | q           |
      |-------------| 
      | house=flat  |
      | house"flat" |
      | house'flat' |    
      | house(flat) |     
```

But these case are responding `201 - OK`
```
      | q           |
      |-------------|
      | house<flat> | 
      | house;flat  |    
[...]

PR #2817, opened 2017-01-16
bug/1994_invalid_chars_in_q_at_sub_creation

kzangeli:
Fixes at least part of #1994 

Checking for forbidden chars inside components of string filters

### DIFF
--- a/src/lib/rest/StringFilter.cpp
+++ b/src/lib/rest/StringFilter.cpp
@@ -442,7 +442,7 @@ bool StringFilterItem::valueGet
     *stringP = s;
   }
 
-  if (*valueTypeP == SfvtString)  // FIXME PR: perhaps not calling forbiddenChars if pattern matching ...
+  if ((*valueTypeP == SfvtString) && (op != SfopMatchPattern))
   {
     if (forbiddenChars(s, ""))
     {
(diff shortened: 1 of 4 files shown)


src/lib/rest/StringFilter.cpp
```diff
@@ -441,6 +442,15 @@ bool StringFilterItem::valueGet
     *stringP = s;
   }
 
+  if (*valueTypeP == SfvtString)  // FIXME PR: perhaps not calling forbiddenChars if pattern matching ...
```
fgalan: FIXME PR should be removed before merging?

kzangeli: Not before discussing it first.

It seems to be we need to exclude the RHS of pattern matching from the forbiddenChars check.
To do that, we will need the OP of the string filter down in StringFilterItem::valueGet.

kzangeli: Discussed over skype and decicion made to send OP down to StringFilterItem::valueGet to avoid doing forbiddenChars for pattern matching

kzangeli: Fixed in 4395fe8

fgalan:
LGTM

Issue #2819, opened 2017-01-17
Forbidden Char checks in /v2/op/query and NOT in typePattern as URI param


kzangeli:
Searching for the usage of forbidden chars, we have found two obvious mistakes:

- POST /v2/op/query  doesn't use forbiddenChars at all
- ?typePattern=XXX as URI param suffers forbiddenChars which it should not

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2820

Issue #2791, opened 2017-01-02
Use empty() for checking zero-length strings


fgalan:
(Raised at https://github.com/telefonicaid/fiware-orion/pull/2790#discussion_r94306733)

As discussed in http://stackoverflow.com/questions/41425569/checking-for-empty-string-in-c-alternatives, `empty()` is the best way of checking for zero-length strings.

Thus:

1. Add that rule in the coding style guide
2. Search (e.g. grep) for all checkings with `""` and usages of `length()`, replacing with `empty()`

kzangeli:
As the issue in SOF doesn't give much info, I did a quick test myself.
This is the source code:

```
#include <stdio.h>
#include <time.h>
#include <string>

// -----------------------------------------------------------------------------
//
// clock_difftime -
//
void clock_difftime(struct timespec* start, struct timespec* end, struct timespec* diff)
{
  diff->tv_sec  = end->tv_sec  - start->tv_sec;
  diff->tv_nsec = end->tv_nsec - start->tv_nsec;

  if (diff->tv_nsec < 0)
  {
    diff->tv_sec  -= 1;
    diff->tv_nsec += 1000000000;
  }
}

// -----------------------------------------------------------------------------
//
// main -
//
// Testing 5 ways of checking for empty string:
//
// 1. (s == "")
// 2. (s.length == 0)
// 3. (s.empty())
// 4. (s.c_str()[0] == 0)
// 5. (*cP == 0)
// 6. nothing - just setting 'b = true', which is 'extra' in each test case
//
int main(int argC, char* argV[])
{
  std::string      s  = "";
  char*            cP = (char*) "";
  struct timespec  startTime[6];
  struct timespec  endTime[6];
  struct timespec  diffTime[6];
  bool             b;
  const char*      testDescription[] =
    {
      "s == \"\"",
      "s.length == 0",
      "s.empty()",
      "s.c_str()[0] == 0",
      "*cP == 0",
      "b = true"
    };

  clock_gettime(CLOCK_REALTIME, &startTime[0]);
  b = (s == "");
  clock_gettime(CLOCK_REALTIME, &endTime[0]);

  clock_gettime(CLOCK_REALTIME, &startTime[1]);
  b = (s.length() == 0);
  clock_gettime(CLOCK_REALTIME, &endTime[1]);

  clock_gettime(CLOCK_REALTIME, &startTime[2]);
  b = (s.empty());
[...]

PR #2820, opened 2017-01-17
Bug/2819 forbidden chars in batch query and not in uri param type pattern

kzangeli:
Fixes issue #2819.
Also, instead of checking for forbidden chars for Entity::isPattern, now a simple check for "true" or "false" is done. Not sure what to do about this CNR wise. I'm pretty sure there is no real change for the users. The broker is simply a tiny little bit faster like this.


src/lib/apiTypesV2/Entity.cpp
```diff
@@ -199,10 +199,24 @@ std::string Entity::check(ApiVersion apiVersion, RequestType requestType)
     return std::string(errorMsg);
   }
 
-  if (forbiddenIdChars(apiVersion, id.c_str()))
+  if (isPattern == "")
```
fgalan: Although we have an issue for an overall fix for this (https://github.com/telefonicaid/fiware-orion/issues/2791) let's try to start using `empty()` in new/modified code from now on.

kzangeli: Not gonna be easy to change the chip ...
But, sure, let's try at least :-)

kzangeli: Fixed in 8baaa28

fgalan:
LGTM

Issue #2452, opened 2016-08-16
Unary statements fails in GET v2/entities?q= if attribute has ":"


LeonanCarvalho:
The character `:` isn't a [forbidden character](http://fiware-orion.readthedocs.io/en/stable/user/forbidden_characters/) but when it's used to a unary statements on simple query fails because 

> "The symbol `:`  can be used instead of `=`"

For example:

`GET http://localhost:1026/v2/entities?q=!device:location`

OR

`GET http://localhost:1026/v2/entities?q=device:location`

Both will return a empty array because of this rule.

OCB Info:

```
{
  "orion" : {
  "version" : "1.2.0-next",
  "uptime" : "9 d, 17 h, 27 m, 37 s",
  "git_hash" : "c1a224db08086751f05d471a8b52c0eaec15f3a5",
  "compile_time" : "Thu Aug 4 09:04:12 BRT 2016",
  "compiled_by" : "root"
}
}
```

fgalan:
So you mean that 

```
GET http://localhost:1026/v2/entities?q=device:location
```

returns a "forbidden character error" ?

LeonanCarvalho:
Not, If I have a attribute like "device:location" and I try to query entities with that attribute  it will return same as this request: 
GET http://localhost:1026/v2/entities?q=device=location

So, I can't use on `q`  attribute with `:` composing their name, maybe it should be a forbidden character.

fgalan:
Probably `:` to `==` is extending to all the URL query string, not only to the cases which are mean to be operators. Based on that, it seems to be a bug, to be fixed in a next release.

(Note: you mention `=` as an equivalente for `:`, but according to documentation it should be `==`. Is there a typo in your report?)

Thanks!

LeonanCarvalho:
Yes, there is a typo in my report, please consider `== `instead a single `=`.

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/2830. Note that ' need to be used to avoid `:` to be interpreted as part of the filter. Thus, the @LeonanCarvalho 's examples would be as follow:

```
GET http://localhost:1026/v2/entities?q='device:location'
GET http://localhost:1026/v2/entities?q='!device:location'
```

fgalan:
[...]

PR #2830, opened 2017-01-20
bug/2452_q_for_attr_with_colon_in_name

kzangeli:
Fixed issue #2452 with a MUCH better op-lookup that is a lot faster and also quote aware.

This means that, e.g. the string filter `q='a:12'>11` will no longer be misinterpreted as an `'a EQUALS 12'>11` (due to the colon inside quotes) but as `a:12  GT 11`, a:12 being the name of an attribute.

About being faster, instead of calling a complex function as `strstr` up to 8 times, we now do **one single fast loop** over the input.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -18,6 +18,7 @@
 - Fix: Partial sanity check for custon url, only if the url contains no replacements using ${xxx} (#2280, #2279)
 - Fix: if entity::type is given but as an empty string, in payload of POST /v2/subscriptions or PATCH /v2/subscriptions/{sub}, an error is returned (#1985)
 - Fix: more intelligent, quote-aware, operator lookup inside q/mq string filters (#2452)
+- Fix: attribute names that include ':' now are supported, with the single-quote syntax. E.g. q='devices:location' (#2452)
 - Fix: admin requests (such as GET /version) with errors now return '400 Bad Request', not '200 OK'
 - Fix: attempt to create an entity with an error in service path (and probably more types of errors) now returns '400 Bad Request', not '200 OK' (#2763)
 - Fix: not calling regfree when regcomp fails. Potential fix for a crash (#2769)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -17,6 +17,7 @@
 - Add: Support for null values in string filters (#2359)
 - Fix: Partial sanity check for custon url, only if the url contains no replacements using ${xxx} (#2280, #2279)
 - Fix: if entity::type is given but as an empty string, in payload of POST /v2/subscriptions or PATCH /v2/subscriptions/{sub}, an error is returned (#1985)
+- Fix: more intelligent, quote-aware, operator lookup inside q/mq string filters (#2452)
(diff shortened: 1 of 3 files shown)


src/lib/rest/StringFilter.cpp
```diff
@@ -458,6 +458,58 @@ bool StringFilterItem::valueGet
 
 /* ****************************************************************************
 *
+* opFind - 
+*/
+static StringFilterOp opFind(char* s, char** restPP)
```
fgalan: Hard to understand parameters... An explatation in the function prebable would be great.

CHANGES_NEXT_RELEASE
```diff
@@ -17,6 +17,7 @@
 - Add: Support for null values in string filters (#2359)
 - Fix: Partial sanity check for custon url, only if the url contains no replacements using ${xxx} (#2280, #2279)
 - Fix: if entity::type is given but as an empty string, in payload of POST /v2/subscriptions or PATCH /v2/subscriptions/{sub}, an error is returned (#1985)
+- Fix: more intelligent, quote-aware, operator lookup inside q/mq string filters (#2452)
```
fgalan: What about the actual fix? I mean 

"Fix: attribute names including ':' now are supported with the single-quote syntax, e.g. q='devices:location' (#2452)"

test/functionalTest/cases/2452_q_for_attr_with_colon_in_name/q_for_attr_with_colon_in_name.test
```diff
@@ -0,0 +1,108 @@
...
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create entity E with attribute A:1
+# 02. GET /v2/entities?q=A:1, no hits (no attr named A)
+# 03. GET /v2/entities?q='A:1', see the entity E
```
fgalan: Looking to @LeonanCarvalho report in #2452 , it would be great to include a case to ensure it works with negation, i.e.

GET /v2/entities?q=!'A:2'

works

kzangeli: My pleasure :-)

CHANGES_NEXT_RELEASE
kzangeli: Sure.
[ I didn't really see this fix as a 'fix for 2452', it is much more. #2452 just happens to get a way to work thanks to this PR (if quotes are used, of course) ]

kzangeli: Fixed in b59c7d7

test/functionalTest/cases/2452_q_for_attr_with_colon_in_name/q_for_attr_with_colon_in_name.test
kzangeli: Fixed in b59c7d7

src/lib/rest/StringFilter.cpp
kzangeli: Fixed in b59c7d7 (that and a little more)

fgalan:
LGTM

PR #2832, opened 2017-01-22
Fix header file guard #define case inconsistencies

pekkanikander:
A number of header files had a differing cases in their
    #ifdef guards vs. the following #define statements

    This patch fixes those jound when compiling with LLVM instead of GCC.


kzangeli:
Nice!
Funny we haven't seen any problems compiling ...
The lowercase v2 most probably come from automatic global changes v2 -> V2.
The other errors are simply human errors. When the compiler finds them (or helps to find them), they're fixed.
Now, no need for that :-)
Thank you Pekka for your contribution!

LGTM

PR #2833, opened 2017-01-22
Fix strncat usage in src/lib/logMsg/logMsg.cpp

pekkanikander:
Another bug found by LLVM.  strncat is particularly
    tricky to use.  This commit attempts to fix the errors
    found by LLVM.  Careful inspection is needed, this has
    not been tested.


kzangeli:
Great!
This error I actually had on my personal ToDo list, just very far down.
A quick fix, bumping LINE_MAX up to 32k made it less urgent.
But, now it's fixed. Thank you!

LGTM

PR #2834, opened 2017-01-22
Change paClearAt type in src/lib/parseArgs

pekkanikander:
Yet another bug found by LLVM.  paClearAt is clearly
    used as an integer, and declared as an integer in the
    options list, but syntactically declared as a bool.

    This commit simply changed the type from bool to int

### DIFF
--- a/src/lib/parseArgs/paBuiltin.cpp
+++ b/src/lib/parseArgs/paBuiltin.cpp
@@ -47,7 +47,7 @@ bool            paHelpVar;      /* the -help option                   */
 bool            paVersion;
 char            paHome[512];
 bool            paNoClear;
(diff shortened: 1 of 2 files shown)


kzangeli:
ok!
Very strange that the C++ compiler does not complain in paConfig.cpp where paClearAt is used:
```
lmClearAt(paClearAt, paKeepLines, paLastLines);
```

This whole thing is a feature of liblm that we don't use in the broker. It's about shrinking the logfile, removing 'uninteresting' loglines. For the broker we use logrotate instead (this library was created probably *long* before logrotate was invented).

Again, thank you for your valuable contributions!

LGTM

Issue #2831, opened 2017-01-20
Weird md: { } added at DB at PUT v2/entities/<entity>/attrs operation


fgalan:
Having a look to https://github.com/telefonicaid/fiware-orion/issues/2829, the following has been found (probably the order should be first fix #2829, then check if this problem stills happening).

Initial situation: an entity created, with a "position" attribute. In DB it looks like:

```
> db.entities.findOne()
{
	"_id" : {
		"id" : "contenedor1",
		"type" : "WasteContainer",
		"servicePath" : "/"
	},
	"attrNames" : [
		"position"
	],
	"attrs" : {
		"position" : {
			"value" : "32, 4",
			"type" : "coords",
			"mdNames" : [ ],
			"creDate" : 1484934630,
			"modDate" : 1484934720
		}
	},
	"creDate" : 1484934061,
	"modDate" : 1484934720
}
```

The, the following PUT update for the entity (attribute replace, but with an attribute which name is the same) is done:

```
PUT /v2/entities/contenedor1/attrs

{
        "position": {
            "type": "coords",
            "value": "37, 5",
            "metadata": {}
            }
        }
}
```

The result is:

```
> db.entities.findOne()
{
	"_id" : {
		"id" : "contenedor1",
		"type" : "WasteContainer",
		"servicePath" : "/"
	},
	"attrNames" : [
		"position"
	],
	"attrs" : {
		"position" : {
			"type" : "coords",
			"creDate" : 1484934837,
			"modDate" : 1484934837,
			"value" : "37, 5",
			"md" : {
				
			},
			"mdNames" : [ ]
		}
	},
	"creDate" : 1484934061,
	"modDate" : 1484934837
}
```

The weird thing is:

```
			"md" : {
				
			}
```

Same behaviour if:

* not including `"metadata": {}` at all in the PUT request.
* using an enterily new attribute in the PUT request (eg. "new_attr")

fgalan:
It might be solved by PR https://github.com/telefonicaid/fiware-orion/pull/2835. Need to re-check (and adjust CNR in positive case).

fgalan:
As suspected, this has been already solved. Tested in master at (2f457fde3c8fd8d6f93cd520f34a518aec1ba002), just after PUT operation we have:

```
> db.entities.findOne()
{
        "_id" : {
                "id" : "contenedor1",
                "type" : "WasteContainer",
                "servicePath" : "/"
        },
        "attrNames" : [
                "position"
        ],
[...]

PR #2835, opened 2017-01-22
Remove a semicolon from the end of an if statement

pekkanikander:
Yet another bug found with LLVM.

    This appears to be a bug, but since I don't know the
    program logic, the extra semicolon just *might* be
    a very dubious feature.

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -754,7 +754,7 @@ static bool updateAttribute
     /* Custom metadata */
     BSONObj    md;
     BSONArray  mdNames;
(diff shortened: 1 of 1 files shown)


kzangeli:
Really really looks like a bug.
However, this is in the middle of real important software, the guts of the broker. I will run our test suite on this change before merging.

Thanks!

kzangeli:
All tests passed.

Seems we don't have a functest to catch this bug (of course - if we did, the bug would have been caught right away :-)).

Great, thanks for the contribution!

LGTM

fgalan:
After having a look, the fix seems to be ok from the point of view of program logic. Moreover, this PR could be a fix for issue https://github.com/telefonicaid/fiware-orion/issues/2831. 

Thanks @pekkanikander !

PR #2836, opened 2017-01-22
Change printing of 64bit ints to use %lld instead of %ld

pekkanikander:
Another issue found when compiling with LLVM.

    At least under Mac OS X, when printing 64-bit ints,
    the printf string must be %lld as they are long long
    while long is just another name for 32-bit ints.

    Of course, this may be different in Linux and I don't
    have the ability to compile under 64-bit Linux.

    Please check that this works fine under 64-bit Linux
    before merging.


kzangeli:
This PR I cannot let into master, sorry.
Doesn't compile in CentOS.
Seems to be a mess about `int64_t` being typedeffed to either `long int` or `long long int`.
Will look into this, creating an issue for it.
Must discard the PR though.

PR #2837, opened 2017-01-22
Change the recognition of MHD_USE_EPOLL support

pekkanikander:
In some newer versions of microhttpd.h, MHD_USE_EPOLL is
   an enum instead of being a #define.  This commit changes
   the #if guarding its use


kzangeli:
Tested OK (not that I doubted very much ... :)

LGTM

PR #2838, opened 2017-01-22
Add #include <string> to src/lib/jsonParseV2/utilsParse.h

pekkanikander:
Without this patch compilation fails with LLVM under Mac OS X with the
   following and other related errors:

   In file included from fiware-orion/src/lib/jsonParseV2/utilsParse.cpp:27:
   fiware-orion/src/lib/jsonParseV2/utilsParse.h:41:15: error:
      implicit instantiation of undefined template
      'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
   std::string error;

### DIFF
--- a/src/lib/jsonParseV2/utilsParse.h
+++ b/src/lib/jsonParseV2/utilsParse.h
@@ -26,6 +26,7 @@
 * Author: Orion dev team
 */
 
+#include <string>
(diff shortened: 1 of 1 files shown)


kzangeli:
LGTM (no need to test here ... :-))

PR #2849, opened 2017-01-26
Slight change in MHD EPOLL detection

kzangeli:
Slight modification of the PR so kindly offered by @pekkanikander 

Adding a line of explanation and changing the constant of MHD version to coincide 'exactly' with the comment: 0x00095102 => 0x00095100 (as comment says MHD 0.9.51 ...).

As I am not 100% sure that the change 0x00095102 => 0x00095100 might have some negative impact in mac ... Pekka, could you test in your machine and let me know if you have any problem with this change, please? And if so, propose a fix that works for you too ...

### DIFF
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -1593,7 +1593,10 @@ static int restStart(IpVersion ipVersion, const char* httpsKey = NULL, const cha
     // while in MHD 0.9.51, the name of the define has changed to MHD_USE_EPOLL.
     // So, to support both names, we need a ifdef/else cpp directive here.
     //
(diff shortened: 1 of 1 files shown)


pekkanikander:
Tested on Mac, compiles and runs fine.   Thanks for figuring out the exact version where the changes was made, I was too lazy/busy for that.

(The rest of the Mac version are not ready for merging yet.  I'm working on it, but right now I need to get a demo working.  Once I have the demo up and running, I'll return to the rest of the changes, clean then up, and make a series of pull requests.  In the meanwhile you can roughly see the work in progress in our repo fork.  Some of the uglier stuff I still have only on my laptop.)

fgalan:
Thanks for the feecback @pekkanikander!

LGTM (Look Good To Me)

Issue #2009, opened 2016-04-07
in POST v2/subscriptions, using "notification http url" accepts url malfomed, returning 201 Created


iariasleon:
in POST v2/subscriptions, using notification callback accepts url malfomed
#### Dataset

```
      | url                        |
      |----------------------------|
      | http://localhost:900000    |
      | http://localhost:dsfsdf    |
      | http://localhost\my_path   |
      | http://e34.56.45.34        |
      | http://34,56.45.34         |
      | http://34.56.45.34;3454    |
      | http://34.56:3454          |
      | http://.                   |
      | http://..                  |
      | http://../                 |
      | http://?                   |
      | http://??/                 |
      | http://#                   |
      | http://##/                 |
      | http://foo.bar/foo(bar)baz |
      | http://-error-.invalid/    |
      | http://a.b--c.de/          |
      | http://-a.b.co             |
      | http://a.b-.co             |
      | http://0.0.0.0             |
      | http://1.1.1.1.1           |
      | http://3441.2344.1231.1123 |
      | http://123.123.123         |
      | http://3628126748          |
      | http://.www.foo.bar/       |
      | http://www.foo.bar./       |
      | http://.www.foo.bar./      |
```
#### subscription request

```
http code: 201
date: Thu, 07 Apr 2016 16:11:16 GMT
connection: Keep-Alive
content-length: 0
location: /v2/subscriptions/570686a3ee5e5afde9131ba5
```
#### expected response

```
http code: 400
{"error":"BadRequest","description":"Invalid URL in notification callback"}
```

fortizc:
I check this when I develop the WS support. If I use "ws://" Orion response me with a "Invalid URI" message, then I had to add some code in order to use "ws://" first here 5932dc20c62c932ee2052df6d07285ca5107ebf2 and then improved here ea76cbb1ccda285b22ae9c2a79f1e60fb121eab6

iariasleon:
Re-tested in the CB version. It issue still fails and returns `201-Created`
```
  "version" : "1.2.0-next",
  "git_hash" : "d81dcf28e

**Comment by iariasleon** [comment]
Re-tested with CB version:
```
  "version" : "1.3.0-next"
  "git_hash" : "b7f1fb9f7c8918deb0faa298b8370a0d1d132d8f"
```

[...]

PR #2855, opened 2017-01-30
hardening/2009_check_of_notification_http_url_in_post_v2_subscriptions

kzangeli:
Fixed issue #2009, for IPv4 only.
FIXME P4 added about doing the same check for IPv6 addresses (I am not 100% sure about this distinction between IPv4 and 6 ... I imagine the hostname 'myhost' could be used for an IPv6 address as well as an IPv4 address ... ? So, ... let's discuss/investigate this a little!)


src/lib/common/string.cpp
```diff
@@ -213,6 +214,63 @@ int stringSplit(const std::string& in, char delimiter, std::vector<std::string>&
...
+*
+* hostnameIsValid - check a hostname for validity
+*
+* See https://en.wikipedia.org/wiki/Hostname
+*/
+static bool hostnameIsValid(const char* hostname, bool isIPv6)
+{
+  if (isIPv6 == true)  // No checks for IPv6
+  {
+    // FIXME P4: Implement hostnameIsValid for IPv6 addresses
```
fgalan: Hostnames are the same for IPv4 or IPv6. It is the name resolver (e.g. `/etc/hosts`, the DNS subsystem) the one which will provide either a IPv4 or IPv6 (or even both... I don't remember right now the details).

Thus, is this is an actual hostname checking function, I think whe whole IPV6 hasn't nothing to do.

src/lib/common/string.cpp
```diff
@@ -345,6 +405,21 @@ bool parseUrl(const std::string& url, std::string& host, int& port, std::string&
     }
   }
 
+  //
+  // Is 'port' a valid number?
+  // Remember the port is really an 'unsigned short', so numbers over MAXSHORT (0xFFFF) are
+  // not valid. Negative port numbers cannot exist and 0 is reserved.
+  //
+  if ((port > MAXSHORT) || (port <= 0))
```
fgalan: Valid port are a range betweeh 0 and 65365 (by desing of the TCP/IP stack). In our case it happens that MAXSHORT matches 65365 but it is an "implementation side effect". I think that using MAX_PORT (defined in limits.h) would be more explicit.

kzangeli: Ah, if MAX_PORT exists (probably defined to MAXSHORT :-)) then I will use it, of course.

Zero on the other hand I prefer not to let pass, as it seems to be a reserved value and cannot be used.
See https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

[ Also, conveniently enough: atoi("invalid string") == 0 ]

src/lib/common/string.cpp
kzangeli: ok
I will trust you on your word, I myself am clueless :-)

src/lib/common/string.cpp
fgalan: Zero is reserved. Fine.

If MAX_PORT exists then use it, but if doesn't exist, please define it and use it ;)

kzangeli: Fixed in 900322f

src/lib/common/string.cpp
kzangeli: Fixed in a8e4057

fgalan:
LGTM

PR #2864, opened 2017-02-01
Bug/2280 better hostname validity check

kzangeli:
Fixes part of issue #2280 
Hostnames are now split into labels and the validity of the labels are checked.


fgalan:
LGTM

PR #2865, opened 2017-02-01
ADD openssl to .spec file requirements

fgalan:


rpm/SPECS/contextBroker.spec
```diff
@@ -35,8 +35,8 @@ Packager:   Fermín Galán <fermin.galanmarquez@telefonica.com>
 URL:        http://catalogue.fiware.org/enablers/publishsubscribe-context-broker-orion-context-broker
 Source:     %{name}-%{broker_version}.tar.gz
 BuildRoot: /var/tmp/%{name}-buildroot
-Requires:  libstdc++, boost-thread, boost-filesystem, gnutls, libgcrypt, libcurl, logrotate, libuuid
-Buildrequires: gcc, cmake, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, boost-devel, libuuid-devel
+Requires:  libstdc++, boost-thread, boost-filesystem, gnutls, libgcrypt, libcurl, openssl, logrotate, libuuid
+Buildrequires: gcc, cmake, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
```
fgalan: As I mentioned in a previos PR, probably this is not actually needed as openssl may be installed due to a transitive dependency of another requirement (maybe gnutls or libgrypt). However, let's be strict adding it also to the .spec.

NTC

kzangeli:
LGTM

Issue #2877, opened 2017-04-06
option to return ServicePath in entity response


cdupont:
Hi,
this is a feature request to add an option to return the servicePath in the entity response.
For example:

```
$ curl http://localhost:1026/v2/entities?options=servicePath
```
would return:
```
[
  {
    "id": "Room1",
    "type": "Room",
    "servicePath": "/MY/APP"
  }
]
```

Including the service path in the response would allow the user to perform actions on the returned entities.

kzangeli:
This would definitely be a nice feature.
+1

fgalan:
Thanks for the contribution. However, I'd suggest to implement it as a virtual attribute, i.e.

```
$ curl http://localhost:1026/v2/entities?attrs=servicePath
```

or (if you want the servicePath, along any other attribute, following the rules for virtual attributes):

```
$ curl http://localhost:1026/v2/entities?attrs=*,servicePath
```

cdupont:
@fgalan yes, that would be perfect

cdupont:
Hi, I tried to implement it, but I have a small problem.
In the function `Entity::render`, the field `servicePath` is empty:
https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/apiTypesV2/Entity.cpp#L73

When I read it, the service path is the empty string.

fgalan:
The `servicePath` field in the Entity class is not actually used. You can check it commenting it out in Entity.h and noticing that the code can be built without problems. Probably is a leftover of a time in which it was actually used for something.

Anyway, the servicePath virtual attribute should be implemented in a similar way other virtual attributes. Have a look to how DATE_CREATED (corresponding to `dateCreated` virtual attribute) is implemented, paying attention not only how this attribute is added to the response, but also that it may be used for sorting and filtering logic.

cdupont:
@fgalan for me it's not clear where I should get the service path from. In the database it's at the same level than id and type.

fgalan:
I don't think so. Probably the best approach is to modify the ContextElementResponse constructor.

```
ContextElementResponse::ContextElementResponse
(
  const mongo::BSONObj&  entityDoc,
  const AttributeList&   attrL,
  bool                   includeEmpty,
  bool                   includeCreDate,
  bool                   includeModDate,
  const std::string&     apiVersion
)
```

[...]

PR #2880, opened 2017-04-13
Add service path as a virtual attribute

cdupont:


kzangeli:
Nice feature, thanks for the effort!

However, we don't ever add/modify features without also adding a functional test to make sure the change does what it should.
@fgalan or myself would be pleased to guide in how to implement a func test

kzangeli:
Discussion point:

The service path should probably also be returned in other requests, not only `GET /v2/entities`, perhaps even in notifications ...

cdupont:
@kzangeli that would be with pleasure.
(however let's do it after the 25 April...)

fgalan:
> The service path should probably also be returned in other requests, not only `GET /v2/entities`, perhaps even in notifications ...

I agree. Probably, most of the GET operations on the NGSIv2 will work, as the PR include changes in the Entity::render() method, along with "service path passing" from NGSIv1 to NGSIv2 structures in the Entity::fill(QueryContextResponse*) method. Functional test will tell the exact situation.

Regarding notifications, the information is already available in the EntityId in the ContextElementResponse of the notification (see https://github.com/telefonicaid/fiware-orion/issues/2877#issuecomment-294774266). In fact, this information is currently used to fill the `Fiware-ServicePath`, so it is a matter of using also to render a prover attribute (based on the attrs filter in the subscriptions, as with other virtual attribute). But I think it can be assumed as a current limitation (as part of issue #2877) without impacting directly in this PR. Actually it is not crucial in this case as, as mentioned before, the service path information is already included in the notification (as a HTTP header).

src/lib/parse/forbiddenChars.cpp
```diff
@@ -44,7 +44,7 @@ inline static bool commonForbidden(char c)
   case '>':
   case '"':
   case '\'':
-  case '=':
+//  case '=':
```
fgalan: This breaks the forbidden characters functionality (see https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html). Unfortunatelly, that means that this PR cannot be merged in its current form into the official Orion master (although of course the branch of this PR is usable for special addapations of Orion you may need, as I understand by your question at http://stackoverflow.com/questions/44049667/orion-notifications-with-basic-auth)

cdupont: It's very strange, because I never meant to update this pull request... All I did was making some experiments in my own fork and push it to my repo. How did it land here?

fgalan: Don't worry... sometimes github does unexpected things ;)

cdupont: I understand now, I should have used a topic branch for this pull request!

cdupont:
Hi,
what is the status with this pull request?
I would close it since it is old, and re-open another if needed.

fgalan:
Yes, this PR is old... I'd suggest the following actions.

With regards to docker/README.md modifications, I'd suggest to create a PR with only that change, due to is very specific and not related with servicepath builtin attribute.

With regards to servicepath bultin atribute, in April 2017 it wasn't clear if service/servicepath concept will be include at the end in NGSIv2, given specification was in draft mode (probably this was the cause we not priorize this PR in that moment... sorry for that). Now that NGSIv2 has been closed and the spec is final (since September 2018) we know that service/servicepath are not part of the spec. However, servicePath built stills making sense as Orion implementation specific aspect.

So, we can do the following:

1. Update the branch on this PR with master or close and cast another one if it is easier.
2. Add tests covering the functionality. This is the receipt: https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/cookbook.md#adding-a-functional-test-case
3. Add documentation explaining the functionality. I think the best place would be doc/manuals/user/ngsiv2_implementation_notes.md in a new section "`servicePath` builtin attribute".

What do you think @cdupont ?

PR #2892, opened 2017-05-19
Hardening/1368 style in mongo backend 01

kzangeli:
* Changed makefile, to remove strly checks from all rules exept the ones about *style*, and added a rule *style* (do we need to modify any doc because of this?)
* Added rule M12bis to style guide doc.
* Fixed style_check script to include mongoBackend and made a rule MINOR for mongoBackend only (mongodb lib forces this, unfortunately
* Fixed style in MongoCommonRegister.cpp

### DIFF
--- a/doc/manuals/contribution_guidelines.md
+++ b/doc/manuals/contribution_guidelines.md
@@ -407,11 +407,11 @@ directly or indirectly are forced to have that construct as well.
 
 #### M12bis (`using namespace` in source code files)
 
-*Rule 1*: the construct ```using namespace XXX``` must not be used in source code files.
+*Rule*: the construct ```using namespace XXX``` must not be used in source code files. The construct using XXX::YYY may be used in source code files.
 
-*Rationale 1*: Much clearer where functions/variables/types come from when `using namespace XXX` is not used
+*Rationale*: Much clearer where functions/variables/types come from when `using namespace XXX` is not used.
 
-*How to check 1*: **Rule 1** is checked using the internal script ```scripts/style_check.sh```
+*How to check*: checked using the internal script ```scripts/style_check.sh```
 
 ### ‘SHOULD follow’ rules:
 
--- a/doc/manuals/contribution_guidelines.md
+++ b/doc/manuals/contribution_guidelines.md
@@ -408,13 +408,10 @@ directly or indirectly are forced to have that construct as well.
 #### M12bis (`using namespace` in source code files)
 
 *Rule 1*: the construct ```using namespace XXX``` must not be used in source code files.
-*Rule 2*: the construct ```using XXX::YYY" **may** be used in source code files, but even better is to **not** use `using` but to write out the namespace (call it prefix if you want) wherever needed.
 
 *Rationale 1*: Much clearer where functions/variables/types come from when `using namespace XXX` is not used
-*Rationale 2*: Even clearer if each function/variable/type contains the namespace everywgere in the code
 
 *How to check 1*: **Rule 1** is checked using the internal script ```scripts/style_check.sh```
-*How to check 2*: **Rule 2**: simply remove all "using" constructs and compile
 
 ### ‘SHOULD follow’ rules:
 
--- a/doc/manuals/contribution_guidelines.md
+++ b/doc/manuals/contribution_guidelines.md
@@ -405,13 +405,13 @@ directly or indirectly are forced to have that construct as well.
 
 *How to check*: using an intelligent ```grep```.
 
-#### M12bis (`using namespace` in source code files)
+#### M12bis (`using namespace` in source files)
 
-*Rule*: the construct ```using namespace XXX``` must not be used in source code files. The construct using XXX::YYY may be used in source code files.
+*Rule*: the construct `using namespace XXX` must not be used in source files (i.e. `.cpp`). The construct `using XXX::YYY` may be used in source files.
 
 *Rationale*: Much clearer where functions/variables/types come from when `using namespace XXX` is not used.
 
-*How to check*: checked using the internal script ```scripts/style_check.sh```
+*How to check*: checked using the internal script `scripts/style_check.sh`
 
 ### ‘SHOULD follow’ rules:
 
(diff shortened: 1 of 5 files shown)


doc/manuals/contribution_guidelines.md
```diff
@@ -405,6 +405,17 @@ directly or indirectly are forced to have that construct as well.
 
 *How to check*: using an intelligent ```grep```.
 
+#### M12bis (`using namespace` in source code files)
+
+*Rule 1*: the construct ```using namespace XXX``` must not be used in source code files.
```
fgalan: We don't use "Rules 1", "Rule 2" style. As alternative, keep the "Rule", "Rationale", etc. subscection and use a bullet list with two items. This would make this rule homogenoeus among the others.

doc/manuals/contribution_guidelines.md
```diff
@@ -405,6 +405,17 @@ directly or indirectly are forced to have that construct as well.
...
+*Rule 2*: the construct ```using XXX::YYY" **may** be used in source code files, but even better is to **not** use `using` but to write out the namespace (call it prefix if you want) wherever needed.
```
fgalan: > but even better is to **not** use `using` but to write out the namespace (call it prefix if you want) wherever needed

That's subjetive... I guess @kzangeli think this is important (as his code uses that), @fgalan doesn't think is so important (as his code uses that). 

Let's remove that part and make that a free choice for the programmer. I mean, the program chose either `ns::myFunct()`` or `using ns::myFunc` + `myFunc()`.

doc/manuals/contribution_guidelines.md
```diff
@@ -405,6 +405,17 @@ directly or indirectly are forced to have that construct as well.
...
+
+*Rationale 1*: Much clearer where functions/variables/types come from when `using namespace XXX` is not used
+*Rationale 2*: Even clearer if each function/variable/type contains the namespace everywgere in the code
+
+*How to check 1*: **Rule 1** is checked using the internal script ```scripts/style_check.sh```
+*How to check 2*: **Rule 2**: simply remove all "using" constructs and compile
```
fgalan: According to my prvious comment, I think rule 2 is not needed. It should be removed.

scripts/style_check.sh
```diff
@@ -235,6 +235,24 @@ sev[47]=MAYOR; cat[47]='Extra space before )'
 sev[48]=NONE;  cat[48]="LAST"
 
 
+
+# -----------------------------------------------------------------------------
+#
+# Exceptions per library
+#
+# o Rule 6 for mongoBackend
```
fgalan: Rule 6... This is an internal numbering you use for modifications in the script? Or does it correspond with some Mx or Sx in our rules set for coding style?

src/lib/mongoBackend/MongoCommonRegister.cpp
```diff
@@ -173,31 +168,25 @@ static bool addTriggeredSubscriptions
...
-
   for (unsigned int ix = 0; ix < idJsV.size(); ++ix)
   {
+    idJsString += "\"" + idJsV[ix] + "\"";
+
     if (ix != idJsV.size() - 1)
     {
-      idJsString += "\""+idJsV[ix]+ "\" ,";
-    }
-    else
```
fgalan: This goes beyond a style check, but I understand has the same effect and the code is more concise an clearer...

Regression test (functest) will ensure that.

NTC (just informative)

kzangeli: Yeah, this I just changed as I saw it right in front of me ...

A "make test" has already finshed OK

scripts/style_check.sh
kzangeli: The 6 is the index in the internal vectors in style_check.sh. Only.
I picked "Rule 6" instead of writing the entire text about "int64/int32 etc".
If you prefer, we can use the text instead. Longer but clearer

doc/manuals/contribution_guidelines.md
kzangeli: ok, that is another alternative to get the same result

doc/manuals/contribution_guidelines.md
kzangeli: Subjective ... perhaps a little. I don't think anybody could oppose to what I say though. It *is* clearer to write out the namespace everywhere. That's more than an opinion (in my opinion :-D)

The rule 2 actually says 'using XXX::YYY" **may** be used', leaving it open to the programmer to decide.

That said, I'll remove rule 2 to not spend any more time on this

scripts/style_check.sh
kzangeli: Fixed in 6bf015c

doc/manuals/contribution_guidelines.md
kzangeli: Fixed in 6bf015c

doc/manuals/contribution_guidelines.md
kzangeli: Fixed in 6bf015c

doc/manuals/contribution_guidelines.md
kzangeli: Fixed in 6bf015c

doc/manuals/contribution_guidelines.md
```diff
@@ -405,6 +405,14 @@ directly or indirectly are forced to have that construct as well.
 
 *How to check*: using an intelligent ```grep```.
 
+#### M12bis (`using namespace` in source code files)
```
fgalan: Rule 1 -> Rule
Rationale 1 -> Rationale
How to check 1 -> how to check

doc/manuals/contribution_guidelines.md
```diff
@@ -405,6 +405,14 @@ directly or indirectly are forced to have that construct as well.
...
+
+*Rule 1*: the construct ```using namespace XXX``` must not be used in source code files.
```
fgalan: Suggestion:

> The construct ```using namespace XXX``` must not be used in source code files. The construct ```using XXX::YYY``` **may** be used in source code files.

kzangeli: Fixed in 46cf62e

doc/manuals/contribution_guidelines.md
kzangeli: Fixed in 46cf62e

fgalan:
LGTM

PR #2898, opened 2017-05-22
Hardening/1368 style in mongo backend 03

kzangeli:
* Removed all "using" from header files, in mongoBackend + **one** in StringFilter.h
* Indentaton changed as a result of the longer type names (including the namespace) in header files
* Added	all necessary "using namespace::item" in source	files
* Temporary "using namespace mongo" in test/unittests/commonMocks.h
* Some other	sporadic fixes:
  * Three empty lines before each function header
  * Order in include files
  * Correct comment in function headers
  * Correct multiline external declaration


fgalan:
LGTM

PR #2899, opened 2017-05-23
hardening/1368_style_in_mongoBackend_04

kzangeli:
style guide in MongoCommonUpdate and compoundValueBson

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1,4 +1,5 @@
-﻿/*
+﻿
+/*
 *
 * Copyright 2013 Telefonica Investigacion y Desarrollo, S.A.U
 *
@@ -21,7 +22,6 @@
 * iot_support at tid dot es
 *
 * Author: Fermín Galán
-*
 */
 #include <utility>
 #include <map>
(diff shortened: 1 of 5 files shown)


src/lib/mongoBackend/compoundValueBson.h
```diff
@@ -37,14 +37,14 @@
 *
 * compoundValueBson (for objects) -
 */
-extern void compoundValueBson(std::vector<orion::CompoundValueNode*> children, mongo::BSONObjBuilder& b);
+extern void compoundValueBson(const std::vector<orion::CompoundValueNode*>& children, mongo::BSONObjBuilder& b);
```
fgalan: Different from the general rule with style fixing, maybe this kind of change (X -> const X&) should be mentioned in the CNR, e.g.:

"Fix: several changes in argument passing in mongoBackend library to align style, from X to const X&"

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -130,15 +127,12 @@ static bool hasMetadata(std::string name, std::string type, ContextAttribute* ca
...
 
 
-
 /* ****************************************************************************
 *
 * equalMetadataValues -
-*
 */
-static bool equalMetadataValues(BSONObj& md1, BSONObj& md2)
+static bool equalMetadataValues(const BSONObj& md1, const BSONObj& md2)
```
fgalan: Different from the general rule with style fixing, maybe this kind of change (X& -> const X&) should be mentioned in the CNR, e.g.:

"Hardening: several changes in argument passing in mongoBackend library to align style, from X& to const X&"

src/lib/mongoBackend/compoundValueBson.h
kzangeli: It is true this is an actual change in code, not just style ...
However it is not a bug fix, *only* a fix to save space on the stack.
Do we really want to go this low level in CNR ?

fgalan: > However it is not a bug fix, only a fix to save space on the stack.

That's why I say to use "Hardening:" and not "Fix:" in the CNR line ;)

"Low level will" be going change by change mentioning every changed function in CNR. Adding a general comment like the one I proposed doesn't hurt.

kzangeli: Sure, no probs.
Just wanted to make sure you had thought it over ... :-)

kzangeli: Fixed in 3b07d1c

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Fixed in 3b07d1c

kzangeli: Actually: c56d91e

src/lib/mongoBackend/compoundValueBson.h
kzangeli: Actually: c56d91e

fgalan:
LGTM

PR #2909, opened 2017-05-26
Hardening/1368 style in mongo backend 13

kzangeli:
Style guide fixes for modules:
* mongoUnsubscribeContextAvailability,
* mongoUpdateContext,
* mongoUpdateContextAvailabilitySubscription,
* safeMongo, and
* mongoUpdateContextSubscription

+ A fix for failing unit tests


src/lib/mongoBackend/mongoRegisterContext.cpp
```diff
@@ -109,9 +109,10 @@ HttpStatusCode mongoRegisterContext
     return SccOk;
   }
 
-  const char* colName = getRegistrationsCollectionName(tenant).c_str();
+  const std::string    colName  = getRegistrationsCollectionName(tenant);
+  const mongo::BSONObj bson     = BSON("_id" << id << REG_SERVICE_PATH << sPath);
 
-  if (!collectionFindOne(colName, BSON("_id" << id << REG_SERVICE_PATH << sPath), &reg, &err))
+  if (!collectionFindOne(colName, bson, &reg, &err))
```
kzangeli: Here is the change that fixes the local problem I had with unittests.
The problem I saw was that `colName` was overwritten with "servicePath", a text string that belongs to the BSON() macro call.
I created a help variable for the BSON stuff and the problem went away.
Not 100% sure this is the fix but ... that BSON call as a parameter (accepted as a const `mongo::BSONObj&` by `collectionFindOne`) is not very nice ...

fgalan: What about the constributions_guide.md rule about this? Is expected in this PR or in the next one?

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
```diff
@@ -155,38 +168,71 @@ HttpStatusCode mongoUpdateContextAvailabilitySubscription
...
+  //
+
   /* Adding format to use in notifications */
   newSub.append(CASUB_FORMAT, "JSON");
 
   /* Update document in MongoDB */
-  if (!collectionUpdate(getSubscribeContextAvailabilityCollectionName(tenant), BSON("_id" << OID(requestP->subscriptionId.get())), newSub.obj(), false, &err))
+
+  //
+  // FIXME P4: Strange errors in two unit tests:
```
fgalan: Maybe the same solution you used for https://github.com/telefonicaid/fiware-orion/pull/2909/files#r118671472 also solves this case?

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
```diff
@@ -76,11 +76,8 @@ extern void setMongoConnectionForUnitTest(DBClientBase*);
 * registrations collection.
 *
 */
-static void prepareDatabase(void) {
-
-  /* Set database */
-  setupDatabase();
```
fgalan: The `setupDatabase()` call in the original code has been moved or removed? It seems the second case, in which case this goes beyond just style check... or maybe I'm missing something.

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
```diff
@@ -908,11 +910,6 @@ TEST(mongoRegisterContext_update, NotifyContextAvailability1)
           .Times(0);
   setNotifier(notifierMock);
 
-  TimerMock* timerMock = new TimerMock();
```
fgalan: This seems to be a change in the unit test logic... Or maybe is a mock not needed in this test?

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
```diff
@@ -975,11 +977,6 @@ TEST(mongoRegisterContext_update, NotifyContextAvailability2)
           .Times(1);
   setNotifier(notifierMock);
 
-  TimerMock* timerMock = new TimerMock();
```
fgalan: Same here... Maybe the (newly added) `uInit()` call is doing this, but I prefer you to confirm.

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
```diff
@@ -1041,11 +1041,6 @@ TEST(mongoRegisterContext_update, NotifyContextAvailability3)
           .Times(0);
   setNotifier(notifierMock);
 
-  TimerMock* timerMock = new TimerMock();
```
fgalan: Same here... Maybe the (newly added) `uInit()` call is doing this, but I prefer you to confirm.

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: I removed it, as all functions in this module now call `utInit()`, which calls `setupDatabase()`.
Actually, it was just one of many desperate tests trying to fix the problems in the unit tests.
However, the change is good and I let it stay.

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: utInit() creates the TimerMock. Not sure what would happen if we had two timer mocks ...

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: Yes, that's it

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
fgalan: NTC

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
kzangeli: That's it

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
fgalan: > Not sure what would happen if we had two timer mocks ...

Most probably bad things :)

NTC

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
fgalan: NTC

test/unittests/mongoBackend/mongoRegisterContext_update_test.cpp
fgalan: NTC

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
kzangeli: I thought I had removed this FIXME (and added the help var for BSON) ... in the last commit ...

kzangeli: Found it in commit ca1a4bb11bcc68ce990eb16b3fe246a1e6fff28e, that I had forgotten to push ...
However, as discussed over Skype, the FIXME is still there and we need a decision on what exactly to do with it

src/lib/mongoBackend/mongoRegisterContext.cpp
kzangeli: I haven't really thought about that ...
But sure, we can add something about this in the style guide

kzangeli: Fixed in e294564 (and the two following commits that fixed mistakes: 8923a76 and a9ed96b)

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
kzangeli: Unable to reproduce the problem so, FIXME removed in 970e360

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
```diff
@@ -111,29 +115,36 @@ HttpStatusCode mongoUpdateContextAvailabilitySubscription
    * FIXME: a better implementation strategy could be doing an findAndModify() query to do the
    * update, so detecting if the document was not found, instead of using findOne() + update()
    * with $set operation. One operations to MongoDb. vs two operations.
+  std::string collectionName = colName;
```
fgalan: Missplaced?

src/lib/mongoBackend/mongoUnsubscribeContextAvailability.cpp
```diff
@@ -76,45 +76,57 @@ HttpStatusCode mongoUnsubscribeContextAvailability
...
   /* Remove document in MongoDB */
+
+  //
   // FIXME: I would prefer to do the find and remove in a single operation. Is the some similar
-  // to findAndModify for this?  
-  if (!collectionRemove(getSubscribeContextAvailabilityCollectionName(tenant), BSON("_id" << OID(requestP->subscriptionId.get())), &err))
+  // to findAndModify for this?
+  //
+
+  std::string colName = getSubscribeContextAvailabilityCollectionName(tenant);
```
fgalan: This is the other place in which S14 rule has been applied: https://github.com/telefonicaid/fiware-orion/pull/2909/files#r118671472

To confirm: I undestand that S14 has been applied 2 times in this PR, but there are more places yet (out of the scope of this PR) in which the S14 rule should be applied.

src/lib/mongoBackend/mongoUpdateContextAvailabilitySubscription.cpp
kzangeli: Yeah :-D
Fixed in 6c75e7d

src/lib/mongoBackend/mongoUnsubscribeContextAvailability.cpp
kzangeli: That is true, probably in many many places ...

fgalan: As discussed in our last talk, being a "should" rule we don't need to apply to everyplace, although probably it would be a good idea in all the cases we use BSON() macro in a non trivial way (e.g. trivial way `BSON ("_id" << id)`).

fgalan:
LGTM

PR #2912, opened 2017-05-31
setNewSubscriptionId

kzangeli:
Made `setNewSubscriptionId()` not return a copy of `std::string` on the stack

### DIFF
--- a/src/lib/mongoBackend/mongoCreateSubscription.cpp
+++ b/src/lib/mongoBackend/mongoCreateSubscription.cpp
@@ -144,13 +144,15 @@ std::string mongoCreateSubscription
 
   reqSemTake(__FUNCTION__, "ngsiv2 create subscription request", SemWriteOp, &reqSemTaken);
 
-  BSONObjBuilder     b;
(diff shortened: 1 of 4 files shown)


src/lib/mongoBackend/mongoCreateSubscription.cpp
```diff
@@ -144,13 +144,15 @@ std::string mongoCreateSubscription
...
-  long long          lastFailure      = 0;
-  long long          lastSuccess      = 0;
-  const std::string  subId            = setNewSubscriptionId(&b);
+  BSONObjBuilder  b;
+  std::string     servicePath      = servicePathV[0] == "" ? DEFAULT_SERVICE_PATH_QUERIES : servicePathV[0];
+  bool            notificationDone = false;
+  long long       lastNotification = 0;
+  long long       lastFailure      = 0;
+  long long       lastSuccess      = 0;
+  std::string     subId;
```
fgalan: With the new approach we loss the "const-ability" of `subId`.

kzangeli: Yes, a small price to pay to not pass objects on the stack

PR #2916, opened 2017-06-01
Hardening/1368 style in unittest framework

kzangeli:
Style guide fixes in unit test framework


test/unittests/CMakeLists.txt
```diff
@@ -228,6 +228,7 @@ include_directories("${PROJECT_SOURCE_DIR}")
 include_directories("${PROJECT_SOURCE_DIR}/src/lib")
 include_directories("${PROJECT_SOURCE_DIR}/src/app")
 include_directories("${PROJECT_SOURCE_DIR}/test/unittests")
+include_directories("${PROJECT_SOURCE_DIR}/test")
```
fgalan: Interesting... how this change is related with code style?

test/unittests/commonMocks.h
```diff
@@ -36,133 +40,163 @@
 
 /* ****************************************************************************
 *
-* USING 
+* USING
 */
 using ::testing::_;
```
fgalan: Keeping this here doesn't follow [M12 rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#m12-using-namespace-in-header-files). However, I think we should make an exceptio here. Otherwise I guess that:

```
ON_CALL(*this, count(_, _, _, _, _))
```
will become in

```
ON_CALL(*this, count(::testing::_,::testing::_,::testing::_,::testing::_,::testing::_))
```

An that's crazy.

However, maybe the using statement could be moved inside the `DBClientConnectionMock` class in order to have a narrower scope?

test/unittests/CMakeLists.txt
kzangeli: This is for the rule about including the directory in the #include directive.
```
#include "unittest.h"
```
Must be:
```
#include "unittest/unittest.h"
```
And for this, we need the proper `-I`, which this modification gives us.

test/unittests/commonMocks.h
kzangeli: Oh yes, that is the intention, no using in header files, only `using xx::yy` in cpp files.
Must have missed this one.

Unfortunately, the script doesn't flag this as error, so it is done manually.

Good catch! Must be fixed!

test/unittests/CMakeLists.txt
fgalan: NTC

test/unittests/unittest.cpp
```diff
@@ -22,25 +22,36 @@
...
+* debugging -
 *
 * FIXME P4 - these counters are useful (only visible if UT_DEBUG is defined)
 *            to detect calls to utInit without respective call to utExit.
 *            All of it should be removed as soon as we implement the new
 *            more complete TRST macro (I_TEST?)
-* 
 */
-//#define UT_DEBUG
+// #define UT_DEBUG
```
fgalan: Leftover?

kzangeli: No, this is old.  I only added the space, as the style guide asks for it
This line is to be uncommented to run in debug mode

fgalan: NTC

fgalan:
LGTM

PR #2932, opened 2017-06-08
Hardening/1368 cppcheck script

kzangeli:
Script to run the prepared directories.
Not all are 100% prepared though, there seems to be some bug in cppcheck:

```
  std::string s = "abc";
  LM_M(("s == %s", s.c_str()));
```

cppchecks flag `s` as "set but not used". Seems like a bug in cppcheck.

So, this must somehow be avoided ...

Also, *one* cppcheck warning was still there. Fixed it


fgalan:
LGTM

PR #2896, opened 2017-05-22
Add Development Manual

fgalan:
**DO NOT MERGE**. This PR is just for getting feedback and to provide the final version of hardening/design_spec_01, but probably the actual merge to master will be done in a different way.

PR #2935, opened 2017-06-09
ADD Developers Manual documentation

fgalan:
A "fresh" version of the PR in https://github.com/telefonicaid/fiware-orion/pull/2896 in just one commit (instead of <strike>111</strike> 112). <strike>The only different seems to be in the logSummary.cpp file, which in the present PR is not touched. This explains why this PR is 63 files with +3,434 -33 while the other is 64 +3,345 -34</strike>. After refreshing old PR with master, both PRs seems to be exactly equal from a diff point of view.


kzangeli:
LGTM

PR #2939, opened 2017-06-12
hardening/cppcheck_without_LM_OFF

kzangeli:
Avoiding cppcheck for the define LM_OFF as it provokes non-desired errors

### DIFF
--- a/scripts/cppcheck_in_makefile.sh
+++ b/scripts/cppcheck_in_makefile.sh
@@ -41,7 +41,7 @@ function cpp_check
 {
   dir=$1
   echo "cppcheck $dir"
-  cppcheck --error-exitcode=42 -j 8 --enable=all -I src/lib -I test/ -I test/unittests $dir > /dev/null 2> /tmp/cppcheck.log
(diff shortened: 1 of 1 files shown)


scripts/cppcheck_in_makefile.sh
```diff
@@ -41,7 +41,7 @@ function cpp_check
 {
   dir=$1
   echo "cppcheck $dir"
-  cppcheck --error-exitcode=42 -j 8 --enable=all -I src/lib -I test/ -I test/unittests $dir > /dev/null 2> /tmp/cppcheck.log
+  cppcheck -ULM_OFF --error-exitcode=42 -j 8 --enable=all -I src/lib -I test/ -I test/unittests $dir > /dev/null 2> /tmp/cppcheck.log
```
fgalan: Maybe a comment explaining why this addiotion is needed would be useful. Something like "We need -ULM_OFF in order cppcheck to process on LM_X macros etc etc"

kzangeli: Sure.
Fixed in 3dbdbfd

fgalan:
LGTM

PR #2945, opened 2017-06-13
Bug/error in macro substitute

kzangeli:
Fixed a bug in the function that does the substitution of ${xxx} in custom notification payloads.
After reallocating a buffer (when > 1024 bytes), the new part of the buffer needs to be reset for the algorithm to work correctly.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,4 +2,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to avoid passing entire objects on the stack, from "X x" to "const X& x"
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
-- Fix: bug in variable substitution of custom notifications that limited the size of a custom notification to 1024 bytes
+- Fix: bug in variable substitution of custom notifications that limited the size of the payload of a custom notification to 1024 bytes (new limit: 8Mb)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,4 +2,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to avoid passing entire objects on the stack, from "X x" to "const X& x"
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
-- Fix: bug in variable substitution of custom notifications that limited the size of the payload of a custom notification to 1024 bytes (new limit: 8Mb)
+- Fix: bug in variable substitution of custom notifications that limited the size of the payload of a custom notification to 1024 bytes (new limit: 8MB)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,3 +2,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to avoid passing entire objects on the stack, from "X x" to "const X& x"
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
+- Fix: bug in variable substitution of custom notifications that limited the size of a custom notification to 1024 bytes
--- a/src/lib/common/macroSubstitute.cpp
+++ b/src/lib/common/macroSubstitute.cpp
@@ -123,7 +123,22 @@ static void attributeValue(std::string* valueP, const std::vector<ContextAttribu
 #define CHUNK_SIZE 1024
 bool macroSubstitute(std::string* to, const std::string& from, const ContextElement& ce)
 {
+  //
   // Initial size check: is the string to convert too big?
+  //
+  // If the string to convert is bigger than the maximum allowed buffer size (MAX_DYN_MSG_SIZE),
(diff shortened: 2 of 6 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -2,3 +2,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to avoid passing entire objects on the stack, from "X x" to "const X& x"
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
+- Fix: bug in variable substitution of custom notifications that limited the size of a custom notification to 1024 bytes
```
fgalan: "Fix: bug in variable substitution of custom notifications that limited the size of a custom notification payload to 1024 bytes (new limit: 8MB)"

(An of course the change to implement that new limit of 8MB ;))

Note the "payload" word. As far as I understand the bug was limiting the payload size, not the notification size (including headers, etc.).

kzangeli: Fixed in b04a82a

CHANGES_NEXT_RELEASE
```diff
@@ -2,3 +2,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to avoid passing entire objects on the stack, from "X x" to "const X& x"
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
+- Fix: bug in variable substitution of custom notifications that limited the size of the payload of a custom notification to 1024 bytes (new limit: 8Mb)
```
fgalan: 8 Mb -> 8MB, as discussed at Skype

src/lib/common/macroSubstitute.cpp
```diff
@@ -108,7 +108,7 @@ static void attributeValue(std::string* valueP, const std::vector<ContextAttribu
 * out-buffer is not big enough, it is realloced with another CHUNK_SIZE.
 *
 * 1024 is set as CHUNK_SIZE and hopefully for most substitutions it will be enough with ONE allocation.
-* Allocation is very costly operations ...
+* Allocation a is very costly operation ...
```
fgalan: a?

src/lib/common/macroSubstitute.cpp
```diff
@@ -121,8 +121,16 @@ static void attributeValue(std::string* valueP, const std::vector<ContextAttribu
 *
 */
 #define CHUNK_SIZE 1024
-void macroSubstitute(std::string* to, const std::string& from, const ContextElement& ce)
+bool macroSubstitute(std::string* to, const std::string& from, const ContextElement& ce)
 {
+  // Initial size check: is the string to convert too big?
```
fgalan: Although I understand it is not really a big problem mark as overpass the case in which the template is >8MB but the actual payload after substitution <8MB, is this really needed? I mean, if this check is removed, does the rest of the logic breaks?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -235,7 +234,14 @@ static std::vector<SenderThreadParams*>* buildSenderParamsCustom
     }
     else
     {
-      macroSubstitute(&payload, httpInfo.payload, ce);
+      bool b = macroSubstitute(&payload, httpInfo.payload, ce);
+
+      if (b == false)
```
fgalan: A more compact alternative (used in other places in our code sometimes):

```
      if (!macroSubstitute(&payload, httpInfo.payload, ce))
      {
        LM_E(("Bad Input (not sending NotifyContextRequest: payload too large)"));
        return paramsV;  // empty vector
      }
```

Just a suggestion. It is ok also the current way.

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -235,7 +234,14 @@ static std::vector<SenderThreadParams*>* buildSenderParamsCustom
...
+      {
+        LM_E(("Bad Input (not sending NotifyContextRequest: payload too large)"));
```
fgalan: I think is better to use LM_W. It is not a bug in CB, but a kind of consequence of bad usage on the users side.

CHANGES_NEXT_RELEASE
kzangeli: Fixed in 3eb90b9

src/lib/common/macroSubstitute.cpp
kzangeli: ah!

src/lib/common/macroSubstitute.cpp
kzangeli: The initial check is not 100% necessary, but imagine a buffer of 16MB enters the function. The algorithm starts and copies and copies, only to finally after copying 8MB decide to return an error.
This initial check avoids that, which was the entire reason for the 8MB limit in macroSubstitute()

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 209e19a

src/lib/ngsiNotify/Notifier.cpp
kzangeli: Fixed in 209e19a

src/lib/common/macroSubstitute.cpp
kzangeli: Fixed in 209e19a

src/lib/common/macroSubstitute.cpp
kzangeli: Discussed over Skype

NTC, but with an additional comment from 1b3be59

fgalan:
LGTM

PR #2946, opened 2017-06-19
memory-leaks fixed in macroSubstitute.cpp

fgalan:
Although macroSubstitude() could be replaced soon by another kind of implementation, let's have the curresnt one as error-free as we can. Thus, this PR should be merged into master.


kzangeli:
LGTM

PR #2947, opened 2017-06-19
FIX macroSubstitute() reimplemented

fgalan:


CHANGES_NEXT_RELEASE
```diff
@@ -3,3 +3,4 @@
 - Hardening: Several changes in argument passing in mongoBackend library to add 'const' in references to objects that are not altered by the function
 - Fix: bug in parseArg lib that may cause problem printing the error message for wrong CLI usage (#2926)
 - Fix: bug in variable substitution of custom notifications that limited the size of the payload of a custom notification to 1024 bytes (new limit: 8MB)
+- Fix: "request payload too large" (>1MB) as Bad Input alarm (WARN log level)
```
fgalan: I have found that there was two places in which this is checked in rest.cpp. In one of them, we were already using alarmMgr.badInput(). Thus, I see this as a kind of error fixing, so mentioning it in CNR.

NTC (just informative)

src/lib/common/macroSubstitute.cpp
```diff
@@ -104,26 +103,29 @@ static void attributeValue(std::string* valueP, const std::vector<ContextAttribu
 *
 * macroSubstitute -
```
fgalan: It is almost possible to review the new version based on the diff. I'd recommend to look to the file directly ("View" buttom above).

NTC (just informative)

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -236,15 +236,15 @@ static std::vector<SenderThreadParams*>* buildSenderParamsCustom
...
+        LM_W(("Runtime Error (not sending NotifyContextRequest: payload too large)"));
         return paramsV;  // empty vector
       }
 
       char* pload  = curl_unescape(payload.c_str(), payload.length());
       payload      = std::string(pload);
       renderFormat = NGSI_V2_CUSTOM;
       mimeType     = "text/plain";  // May be overridden by 'Content-Type' in 'headers'
-      free(pload);
+      curl_free(pload);
```
fgalan: I have take advantage of this PR to do this little fix we recently discover.

NTC (just informative)

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -236,15 +236,15 @@ static std::vector<SenderThreadParams*>* buildSenderParamsCustom
     {
       if (macroSubstitute(&payload, httpInfo.payload, ce) == false)
       {
-        LM_W(("Bad Input (not sending NotifyContextRequest: payload too large)"));
+        LM_W(("Runtime Error (not sending NotifyContextRequest: payload too large)"));
```
kzangeli: In the case of missing ending '}', it is not payload too large ...
Perhaps macroSubstitute() should return enum values that describe the cause of the error ...
Actually (IMHO), a missing ending '}' should *not* cause an error, simply not being replaced and kept as is ...

fgalan: Regarding 

> In the case of missing ending '}', it is not payload too large ...

You are right. However, instead of using an enum as return value (complex) a simpler approach is to move the log message to macroSubstitude().

Regarding 

> a missing ending '}' should not cause an error, simply not being replaced and kept as is ...

Although from a theoretical point of view a user could decice to print `${temperature` in its payload, the 99.99% cases this would corresond to human error situation in which the user missed `}`. Thus I think we should keep the current error detection in that case.

kzangeli: 99.99%  ... Yeah, can't argue with that.
It's a pity to not give 100% liberty to our users, but in this case you are probably right. Not worth it.

fgalan: Fixed in f46f516cfd9849d74dd7b36d7c4ab8e08e8324c6

kzangeli:
LGTM

PR #2949, opened 2017-06-27
Hardening/errors not only leaks in valgrind suite

kzangeli:
Adding detection of "Invalid read/write" to the valgrind test suite

### DIFF
--- a/test/valgrind/valgrindTestSuite.sh
+++ b/test/valgrind/valgrindTestSuite.sh
@@ -485,7 +485,7 @@ function memErrorFound()
 
   echo "FAILED (mem errors: $_memErrors). Check $_valgrindFile for clues"
 
-  memErrorV[$memErrorNo]="_errorList"
+  memErrorV[$memErrorNo]="$_file shows $_memErrors memory errors"
   memErrorNo=$memErrorNo+1
 }
 
@@ -620,7 +620,11 @@ then
 
         typeset -i headEndLine1
         typeset -i headEndLine2
+
+        lost=0
         leakInfo ${NAME}.out
+
+        memErrors=0
         memErrorInfo ${NAME}.out
         failText=''
       fi
@@ -631,7 +635,7 @@ then
     if [ "$lost" != "0" ]
     then
       leakFound "test/valgrind/$vtest.*" $vtest $lost "valgrind" $xTestNo
-    elif [ "$memErrors" != "" ]
+    elif [ "$memErrors" != "0" ]
     then
       memErrorFound "test/valgrind/$vtest.*" $vtest "$errorLines"
     elif [ "$vTestResult" == 0 ]
@@ -746,10 +750,12 @@ then
       typeset -i headEndLine1
       typeset -i headEndLine2
       vMsg processing $directory/$htest.valgrind.out in $(pwd)
-      vMsg "calling leakInfo"
+
+      lost=0
       leakInfo test/functionalTest/$CASES_DIR/$directory/$htest.valgrind.out
+
+      memErrors=0
       memErrorInfo test/functionalTest/$CASES_DIR/$directory/$htest.valgrind.out
-      vMsg "called leakInfo"
     else
       if [ "$dryLeaks" == "on" ]
       then
@@ -806,14 +812,14 @@ if [ $memErrorNo != 0 ]
 then
   echo
   echo
-  echo "$memErrorNo memory errors:"
+  echo "$memErrorNo test cases show memory errors:"
   typeset -i ix
   ix=0
 
   while [ $ix -ne $memErrorNo ]
   do
+    echo "o ${memErrorV[$ix]}"
     ix=$ix+1
-    echo "  " ${memErrorV[$ix]}
   done
 
   echo "---------------------------------------"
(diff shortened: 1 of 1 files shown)


test/valgrind/valgrindTestSuite.sh
```diff
@@ -352,6 +353,21 @@ function processResult()
...
 #
+# Extract error info from valgrind output file
+#
+# It uses as argument the .out file to process
+#
+function memErrorInfo()
+{
+  filename=$1
+  # Get info from valgrind
+  memErrors=$(grep -n "Invalid write of size" $filename | wc -l)
```
fgalan: Maybe could even add a "invalid read" to the check? Not sure if we have some case like that but better to be prepared for that...

test/valgrind/valgrindTestSuite.sh
```diff
@@ -456,6 +472,25 @@ function failedTest()
...
+#
+# memErrorFound
+#
+function memErrorFound()
+{
+  _valgrindFile=$1
+  _file=$2
+  _memErrors=$3
+
+  echo "FAILED (mem errors: $_memErrors). Check $_valgrindFile for clues"
```
fgalan: What would happen if at the same time a given test fails due to leaks *and* invalid writes/reads? Which sort of output would we see?

test/valgrind/valgrindTestSuite.sh
```diff
@@ -590,7 +634,10 @@ then
 
     if [ "$lost" != "0" ]
     then
-      failedTest "test/valgrind/$vtest.*" $vtest $lost "valgrind" $xTestNo
+      leakFound "test/valgrind/$vtest.*" $vtest $lost "valgrind" $xTestNo
```
fgalan: Does the if here mean that the script detec either leak fails or mem error fails? (leak seems to take precedence in the case of errors of both types)

test/valgrind/valgrindTestSuite.sh
```diff
@@ -575,13 +614,18 @@ then
         mv /tmp/accumulator_$LISTENER_PORT       $vtest.accumulator_$LISTENER_PORT
         mv /tmp/accumulator_$LISTENER2_PORT      $vtest.accumulator_$LISTENER2_PORT
 
-        failText=$(failedTest "$vtest.valgrind.out" $vtest 0 "valgrind" $xTestNo)
+        failText=$(leakFound "$vtest.valgrind.out" $vtest 0 "valgrind" $xTestNo)
```
fgalan: This means that .vtest are only look for leaks. Is that correct?

(No problem with that, .vtest are not very important at the present time; just to know)

test/valgrind/valgrindTestSuite.sh
kzangeli: I really prefer to do an as exact match as possible. That way false positives are avoided as much as possible. If valgrind changes its syntax, we'll deal with it then.

Never seen an "invalid read", but that is probably easy to provoke. I'll try ...

test/valgrind/valgrindTestSuite.sh
kzangeli: Both `leakInfo` and `memErrorInfo` are called, so we will see both error types if present

test/valgrind/valgrindTestSuite.sh
kzangeli: Actually, in the vtest section, both functions are called as well.
I didn't pay very much attention to that part as I would like to remove it completely.
`failText` is just the text that is written after each case.

I haven't tried the vtest. It just may work perfectly.
I vote for removing it instead of making it "perfect".

test/valgrind/valgrindTestSuite.sh
kzangeli: Ah, this is bad.
Before there was only two options, now there are three.
This I will need to change

fgalan: Yes, probably it is better to remove the whole .vtest thing. However, not in this PR.

In the meanwhile (i.e. in the context of this PR) a warning comment could be useful, e.g. "The .vtest test are deprectaed, so memory corruption is not done for them, only memory errors. Probably this is going to be removed compeltely in a next refactor".

test/valgrind/valgrindTestSuite.sh
fgalan: Yes, probably it is better to remove the whole .vtest thing. However, not in this PR.

In the meanwhile (i.e. in the context of this PR) a warning comment could be useful, e.g. "The .vtest test are deprectaed, so memory corruption is not done for them, only memory errors. Probably this is going to be removed compeltely in a next refactor".

test/valgrind/valgrindTestSuite.sh
kzangeli: Fixed in b6502d8

test/valgrind/valgrindTestSuite.sh
kzangeli: Fixed in b6502d8

test/valgrind/valgrindTestSuite.sh
kzangeli: NTC

test/valgrind/valgrindTestSuite.sh
kzangeli: Fixed in b6502d8

fgalan:
LGTM

PR #2952, opened 2017-07-07
hardening/removing_compilation_flags

kzangeli:
Removed compiler flags about stack check and protection


CMakeLists.txt
```diff
@@ -76,14 +76,25 @@ if (${CMAKE_BUILD_TYPE} STREQUAL DEBUG)
...
+    #   This is due that this directive makes the compiler introduce code at the beginning/end of each functions,
+    #   so it has a pretty big impact on performance
     #
-    # Our unit tests don't pass the compilation with "stack protection", so we turn it off using
-    # remove_definitions().
+    #   Our unit tests don't pass the compilation with "stack protection", so we turn it off using
+    #   remove_definitions().
     #
-    add_definitions(-fstack-check -fstack-protector)
+    #add_definitions(-fstack-check -fstack-protector)
```
fgalan: It seems the bad one is `-fstac-check`... Maybe we should keep `-fstack-protector`, shouldn't we? Jenkins will complaint if we are wrong...

kzangeli: Fixed in c33478c

fgalan:
LGTM

PR #2953, opened 2017-07-11
hardening/memerrors

kzangeli:
Now ALL types of valgrind errors are caught. Fixed a few errors but we still have errors in 6 functests


fgalan:
LGTM

Issue #2955, opened 2017-07-12
Invalid JSON (quotes not escaped)


Hawkings:
In some cases, Orion returns an HTTP response with Content-Type: application/json but with the body being invalid json.

For example, the response for the query `/v2/entities?foo"` (note the `"` at the end) is:

    HTTP/1.1 400 Bad Request
    Connection: Keep-Alive
    Content-Length: 81
    Content-Type: application/json
    Fiware-Correlator: 2a1d9306-6332-11e7-a995-080027f6529d
    Date: Fri, 07 Jul 2017 16:34:36 GMT

    {"error":"BadRequest","description":"Empty right-hand-side for URI param /foo"/"}

Whose body is not valid JSON. I think this problem could be solved by escaping the quotes.

fgalan:
Thanks for the feedback! It seems that the logic that avoids rendering fordidden chars (note that `"` is one of them, see https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html) is failing in this particular case.

fgalan:
The function `htmlEscape()` is the one that translates forbidden chars into allowed chars. Have a look around the code to see how it is being used. 

In this issue it has to be found why this is not being applied to the "description" field in the case of the response shown above. And fix it.

* Impact on documentation: none
* Impact of unittest: *almost* sure none, but not *fully* sure. I don't remember any unit test related with this, but let's see...
* Impact of functional tests:
  * Existing .test should work without any expected change. If the modification break some existing test, please comment on this issue (in that case we would probably need to reevaluate the solution approach)
  * A new test mirroring the use case described by @Hawkings in his original post should be added using (for instance) the directory cases/2855_quote_in_url_not_escaped. Only one step reuired in that .test. You could start with some existng .test  (the case is so simple that any one including a GET request should work).

arigliano:
@fgalan Thank you for the suggestion. I've found the problem in the **uriArgumentGet** method. 
In particular, it appends "**ckey**" to "Empty right-hand-side for URI param..." string, in order to pass it as "details" in the OrionError constructor. The real problem is in **OrionError::toJson**() method, because, unlike the **render**(), it prints out **JSON_STR(details)** , without perfoming any check about forbidden chars. 
I propose two ways to fix this issue:

- The immediate solution would be to wrap the "ckey", when creating the details string, with either **jsonInvalidCharsTransformation**() or better **JSONHelper::toJsonString**() (rest.cpp:114)

[...]

Issue #2938, opened 2017-06-12
Bad encoding of end of line characters 


aarranz:
Orion allows to insert end of line characters inside string values of attributes, and that is working as expected. But the context broker is not correctly encoding back that EOL characters, that results in invalid JSON response from the context broker.

Example:

```
$ curl -X POST http://orion.server:1026/v2/entities -d '{"id": "deleteme", "type": "deleteme", "attribute": {"type": "text", "value": "a\nb"}}' -H "Content-Type: application/json"

$ curl -X GET http://orion.server:1026/v2/entities/deleteme                                                                                                                                            
{"id":"deleteme","type":"deleteme","attribute":{"type":"text","value":"a
b","metadata":{}}}
```

The same happens to the ' \r' character used also as EOL character.

Tested with:
```
{
"orion" : {
  "version" : "1.7.0-next",
  "uptime" : "16 d, 16 h, 45 m, 13 s",
  "git_hash" : "7ac993b3c00b3b1dccd629c1ced380623d487b16",
  "compile_time" : "Fri May 26 10:13:54 UTC 2017",
  "compiled_by" : "root",
  "compiled_in" : "b99744612d0b"
}
```

kzangeli:
I'm not so sure this is an error ...
`\n` is *one* character, with the value 10 (0xA) and when printed on a screen it provokes a newline. Similar with `\r`.
I think that if you want *two* characters: a backslash and an `n`, you need to escape the backslash: `\\n`.

This needs some thorough thought though :-)

fgalan:
I understand the problem is in the round trip. I mean, if a given attribute at CB has been set to:

```
"value": "foo \n bar"
```

what the user wants to be retrieved is


```
"value": "foo \n bar"
```

and not


```
"value": "foo 
 bar"
```

which, in addition, is ilegal JSON (have a look with https://jsonlint.com)

aarranz:
Thanks @fgalan, that's the point. In JSON the character with value 10 (0xA, line feed) is encoded as '\n' and the character with value 13 (0xD, carriage return) is encoded as '\r'. Using a line feed or a carriage return character directly inside an string is ilegal in JSON.

kzangeli:
Reading a little about this, i.e.
https://stackoverflow.com/questions/2392766/multiline-strings-in-json

it seems that orion should modify the incoming one byte `\n` into two bytes `\` and `n`.

A bit strange to modify the value that the user has given ...
[...]

PR #2960, opened 2017-07-20
Unify json serialization by using rapidjson

aarranz:
Current implementation of the Context Broker has a mixture of code for serializing JSON structures. I'm proposing to change it to use rapidjson.

Current pull request code is confirmed to fix bug #2955 and it's passing all the unit test (including a new test for stressing bug #2955). I'm currently passing the valgrind tests (at least 0000_https_support/https and 0000_ipv6_support/ipv4_ipv6_both are failing) so probably this pull request has to be updated.

Anyway, it also simplifies current context broker code by removing the need to deal with JSON commas, different function for serializing JSON, ...

It probably fix bug #2938, and I can easily be extended to fix bugs #2959 as well as other bugs regarding JSON serialization.


fgalan:
Impressive work... +/- 6000 changed lines ;)

One quick question, before analyzing the work and providing more in-deep feedback. You say:

> I'm currently passing the valgrind tests [...]

Do you mean valgrind suite (i.e. `make valgrind`) or functional test (i.e. `make ft`) ?

aarranz:
I mean `make valgrind` and it's taking a lot of time (not finished yet, currently running test 440). I know that there are functional tests not passed when using the `make valgrind` command. I will run `make ft` after this first round using `make valgrind`.

These are the reported errors until now:
```
Test 023/927: 0000_https_support/https .............................................................................................................. OK (15.10 seconds) (no leak but ftest error 11)
Test 024/927: 0000_ipv6_support/ipv4_ipv6_both ...................................................................................................... OK (76.73 seconds) (no leak but ftest error 1)
Test 144/927: 0000_statistics_operation/statistics_with_semaphores .................................................................................. OK (55.44 seconds) (no leak but ftest error 1)
Test 369/927: 0876_attribute_dates/filter_attr_datecreated_sub ...................................................................................... OK (91.25 seconds) (no leak but ftest error 1)
Test 370/927: 0876_attribute_dates/filter_attr_datemodified_sub ..................................................................................... OK (82.35 seconds) (no leak but ftest error 1)
```

fgalan:
The valgrind regression is time intensive. As figure, it could take around 2-3 hours in our reference CI platform.

For the future, do `make ft` first. The valgring regression is typically done after all the functional test (and unit test) are ok so the only thing that is pending is to check is about memory leaks and memory access errors.

fgalan:
> It probably fix (or can easily be extended to fix) bugs #2938, #2425, #2959 as well as other bugs regarding JSON serialization.

I think the reference to #2959 is a typo. Actually #2959 is a fix in the valgrind test framework...

fgalan:
Finally, we have had some time to have a look to this PR a bit closer.

First, thanks for your contribution! In some sense, it addresses the same old technical debt topic related with improving responses payload rendering described in #1298. Moreover, the PR has adopted a smarter approach, removing some structures (iostreams) that raised concern in some developers about performance in #1298. In addition, the removal of many internal returning of std::string between render methods (in favor of JsonHelper/writer reference passing) is also a good point.

Let’s consider the two different families in the NGSI API, i.e. NGSIv1 and NGSIv2.

Regarding NGSIv2, I understand this PR doesn’t change anything from a functional point of view. In other words, all the NGSIv2 .test should pass regression ok, even concerning Content-Length in response payload. However, a rendering-intensive performance test is required, in order to see that performance is not decreasing (or even better, is increasing :)

Regarding NGSIv1, it is a bit more complicated… The work in this PR collides with the payload simplification work already started at haderding/remove_ngsiv1_indent branch (yes, branch name has a typo ;). Although probably the work done in the present PR would reach at the end a similar result to the one in that branch, we have paid an extreme care to backward compatibility in it, including the “NGSIv1 indent paranoid” CLI flag, that we don’t want to lose. In fact, most probably current PR is not passing NGSIv1 .tests as they are now. Have a look to #2760 for more detail on haderding/remove_ngsiv1_indent work.

Thus, our proposal is to reduce the scope of this PR (maybe in a new fresh PR, leaving the present one to be closed), limiting to NGSIv2 renders by the moment. Once you have done that, let’s ensure that: 1) all .test (both NGSIv1 and NGSIv2) pass without changes, 2) the result of an NGSIv2 rendering-intensive performance test is positive (that test have to be carefully designed), 3) there is not conflict with haderding/remove_ngsiv1_indent work.

In a second iteration, once haderding/remove_ngsiv1_indent gets consolidated into master, the same approach (i.e. using rapidjson to render NGSIv1 responses) could be done. But that’s a separate piece of work for the future.

In addition to the above, the PR has some code style issues (see [contribution guidelines](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md) for more detail). @kzangeli  could elaborate on that… he knows more on style than me :)

One final note… the PR is HUGE. It is very difficult to review it in deep. In fact, we are not sure of having capture all the valuable feedback or missing potential problems with it :) Reducing the scope to just NGSIv2 could be also a good idea to make the review task easier. Moreover, maybe is a good idea to start with a first little PR addressing the NGSIv2 (covering just a small subset of render methods, corresponding to just one NGSIv2 operation) and progressing step by step in successive easy to review PRs.

src/lib/common/dtoa.h
```diff
@@ -0,0 +1,239 @@
+/*
```
fgalan: This file is already part of radidjson (typically, /usr/local/include/rapidjson/internal/dtoa.h). Maybe it could be used from the library instead of including it as Orion library?

src/lib/common/JsonHelper.cpp
```diff
@@ -33,233 +34,242 @@
...
-  } //end-for
-  ss << '"';
-  return ss.str();
+    indent = DEFAULT_JSON_INDENT;    
+  }
+  pwriter.SetIndent(' ', indent);
 }
 
 
+void JsonHelper::Null() {
```
kzangeli: First of all, great job!

A minor detail, this line here (and many lines that follow), doesn't follow the Orion style guide (see https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md).

As you will see in that doc, and throughout the code, we put brackets (`{}`) on a line of its own, all functions have a function header and functions/methods are camelCase, starting with lowercase (`null`, `int`, etc may give problems with the compiler for being reserved words ... - if so, an exception-decision needs to be made, and added to the style guide).

Please read the style guide and fix these minor details.
[ I wont add a github-comment for each occurrence here, but instead do a re-check after this has been fixed ]

src/lib/common/JsonHelper.cpp
```diff
@@ -33,233 +34,242 @@
...
-      os << ',' << toJsonString(list[i]);
-    }
-    os << ']';
-    return os.str();
+
+void JsonHelper::Bool(bool value) {
+  if (indent == 0) {
+    writer.Bool(value);
+  } else {
+    pwriter.Bool(value);
```
kzangeli: Not sure it is possible, but it would be great to replace all these `if (indent == 0) { writer } else { pwriter }` for a single `if` upon creation of the writer/pwriter object. Polymorhphism ...  (this wont change during the rendering of one JSON object - it's either pretty or not). If this _can be done_, the code would gain a lot in readability and number of source code lines.

kzangeli: Also, a doubt about writer/pwriter.
This code now seems to be prepared to either do "pretty printing" or "no whitespace".
How/where is the mode selected?

It would be great (personal opinion) to have a URI parameter like `?output=formated` or something similar to let the client select. This is something we've been thinking about for a long time ...

aarranz: I also hate those `if` sentences. The problem is that polymorhphism cannot be used in this case due rapidjson providing the  `Writer` and `PrettyWriter` classes using class templates. Moreover, currently, rapidjson is not marking the required methods as virtual (and it is very difficult to use that scheme using class templates, due how class templates are implemented in C++).

Regarding supporting the `?output=formated` seems easy to implement. But I also agree with you regarding implementing it after cleaning and merging this PR.

src/lib/common/JsonHelper.cpp
aarranz: ok, I will fix those problems.

src/lib/common/JsonHelper.cpp
fgalan: @aarranz remember we have a pending talk. Don't continue coding on this PR without talking with me first ;)

src/lib/common/dtoa.h
aarranz: I have included this file to change how `rapidjson` render numbers (I am working now on creating a PR in `rapidjson`). I think is a bug to use a different syntax when rendering float and integer numbers. Take into account that JSON only uses one type for numbers (numbers are always floating numbers, there is not a different syntax for integer numbers).

aarranz: The rendering problem is exposed when rendering an integer number. E.g. if you use the `Double` method provided by `rapidjson` to render a `4` number it is rendered as: `4.0`. This syntax is required by some languages (like C/C++) to mark it as a floating number, but in JSON it is not required, so it would be better written as `4`, and is what is usually done by other libraries (in fact, `rapidjson` is the only library I know that serialises numbers using this C/C++ pattern)

fgalan:
This PR cannot be considered stale. It shows how rapidjson can be used to render JSON payload and, in that sense, it is useful.

PR #2963, opened 2017-08-18
Fixes sending update with v1 API and generating v2 notifications

giovannicuriel:
There was a problem when sending v1 attribute updates and v2 notification. As processing v1 updates will always set the value type as 'string', the v2 part of the notification generation process checks whether the value type is in accordance to the 'condition' of the subscription being analyzed (in StringFilterItem::compatibleType). If the attribute is anything but a string, this will fail.

This PR fixes the value type attribution while parsing v1 updates.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,8 @@ add_definitions(-fPIC)
 
 # Baseline compiler flags, any change here will affect all build types
 #set (CMAKE_CXX_FLAGS "-Wall -Wno-unknown-pragmas -D_LARGEFILE64_SOURCE -D_GNU_SOURCE -Werror")
-set (CMAKE_CXX_FLAGS "-Wall -Wno-unknown-pragmas -D_LARGEFILE64_SOURCE -D_GNU_SOURCE -Werror -fno-var-tracking-assignments")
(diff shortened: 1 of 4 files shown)


fgalan:
I have seen this PR were closed just after being created. Could you elaborate on the reason, please? 

I haven't reviewed the contribution in deep, but if it solves a bug currently existing in CB, it may be valuable.

Thanks!

giovannicuriel:
Hi! Github newbie here.

Actually, my intention was to create a PR to my own repository (in a different branch), but I didn't realized that the base branch was pointing to the one from the original repository. 

And this patch fixes a specific situation we found in some of our tests related to update an attribute using /v1/ API and receiving notifications created with /v2/ (we found that Orion was not generating the notifications if a condition is added to the subscription creation request). This code makes it work, but we don't feel particularly comfortable to propose it as a fix to this issue, as there might be a more generic solution to it. We would like to further analyze it so that we can provide a proper fix or, at least, gather more information about it.

(I say 'we' because there are more people involved in the project I'm working on, as you might see in our fork)

fgalan:
Now it's clear... :)

Thanks!

PR #2975, opened 2017-09-14
FIX missing lastsuccess and lastfailure

### LINKED ISSUE #2974: Missing lastSuccess/lastFailure in GET /v2/subscriptions in some cases
We have tests that in some cases (e.g. 1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications_stop_start_accumulator.test) the lastSucess/lastFailure field is missing in the GET /v2/subscription operation, although lastNotification and timesSent seem to be ok.

The problem is a bit annoying as it doesn't happend in all systems (e.g. we have a jenkins running regression that is showing the problem... while another jenkins with the same setup is not showing it!). It could be related with cache (in some systems we have observed the problem arises when cache is enabled and doesn't happen with noCache enabled).

Research in progress...

fgalan:
Fixes #2974 

Regression work in my local Debian machine. Jenkins regression (more complete, due to it does two phases, with and without cache) is on the way right now.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: missing lastSuccess/lastFailure associated to initial notification on subscription creation some times when csub cache is in use (#2974)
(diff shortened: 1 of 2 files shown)


kzangeli:
LGTM

fgalan:
Result in jenkins was ok

PR #2982, opened 2017-09-21
ADD doc decimal numbers rounding with functional tests

fgalan:


test/functionalTest/README.md
```diff
@@ -0,0 +1,24 @@
+## How to run functional tests
+
+TBD
+
+## Known issues
+
+### Decimal numbers rounding
+
+You may find fails likes this one when you run the tests:
```
kzangeli: fails => failures (or errors)

test/functionalTest/README.md
```diff
@@ -0,0 +1,24 @@
...
+                           "f": 3.1400000000000001,
+does not match ref line:
+                           "f": 3.14,
+```
+
+This is not an actual problem with the test or Orion, but a rounding problem in the version of the json module that 
+comes with Python 2.6 (the testHarness.sh program typically uses `python -mjson.tool` to beautify responses).
+
+The solution is easy: don't use Python 2.6. The recommended version is Python 2.7. Note that CentOS 6 comes with Python 2.6 at
+system level, but you can use [virtualenv](https://virtualenv.pypa.io/en/stable/) to use Python 2.7 in a easy way.
```
kzangeli: in **an** easy way

test/functionalTest/README.md
fgalan: Fixed in ee1d946dcc1ae74d5d1692a615919c32efccb625

test/functionalTest/README.md
fgalan: Fixes in ee1d946dcc1ae74d5d1692a615919c32efccb625

AlvaroVega:
LGTM

PR #2990, opened 2017-10-02
Enable NGSI9 using the JSON v2 binding

jmcanterafonseca:


jmcanterafonseca:
@kzangeli @fgalan r?

fgalan:
I think the more registrations get similar to subscriptions the better. Conceptually, they are pretty similar. 

I'd suggest to to the exercise in the following way:

* Copy the subscription section into registration section
* Look for fields that are not valid in the case of registrations.

jmcanterafonseca:
well, they are not similar, and we received comments from participants in
ETSI ISG CIM that subject was not a proper way of naming the "entities"
section. The PR is aligned with what we are proposing in ETSI ...

2017-10-02 13:34 GMT+02:00 Fermín Galán Márquez <notifications@github.com>:

kzangeli:
Are registrations (ngsi9) ever used except for forwarding of requests to context providers?
If not (which is my understanding), perhaps we can add a special metadata to the attributes describing the Context Provider (and a special attribute describing the default context provider for attrs of an entity).

Also, it would be nice to have a some kind of mechanism of Primary/Backup Context Providers.

I don't know about you guys, but I would like to redo the entire forwarding mechanism, and not "copy" the old mechanism based on ngsi9).

Especially, we need to avoid the 'shadowing problem':

1. Create an entity/attr in CP1
2. Register it in CB
3. Query the entity/attr in CB => CB consults CP1 and returns that entity/attr

Now the shadow:
4. Create (APPEND) an entity/attr (same entity id and attr name) in CB
5. Query the entity/attr in CB => CB returns its local entity/attr

It would be nice to remove this limitation ...

fgalan:
I think is is a good idea to keep separated the context availability management functionality (the old "NGSI9" part of the API) and the context management functionaliity (the old "NGSI10"), as they are in NGSIv1. I think is cleaner that way and having them "coupled" using metadata could make the usage of the API (and probably the implementation) harder.

Primary/backup CPrs is an interesting idea, but out of the scope of this PR in my opinion. In other words, let's first "solve" the easy case (without backup CPrs), then include that in the future.

Regarding shadowing, it has many aspects... It could be even considered a feature (me as an user want to "bypass" the forwarding to a CP by means of defining that context information locally, as a kind of "privileged cache"). As with primary/backup, I think that out of the "basic" registration and forwarding mechanism that we have to define as first step.

kzangeli:
"coupled" using metadata ...
No, that's not really what I propose. No coupling between ngsi9/10.

This is all about updates and queries. Pure ngsi10, but we decided to use the registrations to find a Context Provider for the entities/attributes.

What I propose is to have ngsi10 + context providers standalone, 100% decoupled from ngsi9.

I am not even sure that we want registration in APIv2. As far as I know, its usage is almost null.
We only use it in "NGSI10", for the context providers ...

About shadowing, sure it can be seen as a "feature" (not a very good one at that), but, a feature should be something that we let the users choose if they want to use or not. Right now there is no way of choosing. 
Somebody creates a local attribute in the broker and there is no longer any way to reach that attr in its context provider, possibly (probably) destroying the old functionality.

And, finally, if we believe Primary/Backup is something we want (I certainly thing so), we'd better keep that in mind while we decide the "forwarding mechanism"; if not, it may not be possible to add it afterwards.

fgalan:
Not sure if I'm using correctly the term "coupled"... anyway, what I mean is that I think is better to have CPr information *outside* of the context information itself.

In fact, this have important advantages:

* It is easy to deal with over-covering scenarios. E.g. the same entity covered by several registrations, so CB uses a given algorithm/heuristic to decide what to do (e.g. ask all CPrs or pick the best one according a given metric). In fact, this can be useful to implement the backup feature you mention.
* Scopes (e.g. geographical scopes for registrations, filters, etc.) can be associated to registrations, so a single registration covers several entities. It is easier to define a registration for all the entities closer to a given point in space that to re-define the CPr each time an entity moves.
* User can manage registration independently of entities themselves (e.g. create new registrations, delete them, etc.).

We have a paralelism: context subscriptions. When we defined subscriptions for NGSIv2, we could have opted to implemented a solution like the one you propose based on metadata for subscriptions, so the URLs to be notified for a given attribute were defined as metadata of that attribute. 

However, we didn't. Under my point if view is clear that having a separated resource for that (i.e. the "subscription resource") is better, in the same way I think having a separated resource (i.e. the "registration resource") is better also in this case.

fgalan:
> I am not even sure that we want registration in APIv2. As far as I know, its usage is almost null.

Interesting point...

You are right that registrations is a not very used feature from an *application user* point of view (at least comparing with context management). But, from a *platform-as-an user* point of view, registrations and forwarding are the enabling use case for IOTAgents passive attributes and commands.

kzangeli:
The dependency of NGSI9 in NGSI10 forwarding is exactly what I want to remove ...
If we have another mechanism for storing the context providers, NGSI10 would no longer depend on NGSI9 for forwarding of NGSI10 requests.

What @fgalan said earlier about having the CP info separate makes a lot of sense, what if we want to use REGEX for attrs/entities that a CP provides ...
Of course, my initial idea of CP-info as metadata of an attr falls.
However, this CP info doesn't have to be NGSI9, it should be "part of ngsi10" ...

_[ Thinking out loud ... what if the last updater of an attr automatically becomes the primary CP
   of the attr? No additional message required, only that the "updater" would have to support
  "POST /v2/entities ...", and to avoid being seen as a CP, we could invent a URI param, or something l 
  like that ... ]_

fgalan:
We (@kzangeli and @fgalan) have been talking by skype. We have agreed on this:

1. To have registrations (or a better name) as a resource (in RESTful sense) in the API, similar to subscriptions
2. It would be possible in the future (future = not for the first version) operations to simplify operations, e.g. to create and entity a register a context provider for it in the same operation. This "future" will also include dealing with shadowing, primary/backup CPrs, etc. aspects.
3. How to structure the regisitration information (JSON structure, field naming, etc.) is the main remaining point at the present moment.

jmcanterafonseca:
proposed name from ETSI is /csources (context sources). please provide
detailed feedback on:

data needed to register a new csource
the specific name of the fields

Wrt point 2, I don't think it is a good idea to do it in one operation, at
least we are not going to propose that to ETSI, but perhaps a proprietary
extension could make it.



2017-10-03 10:54 GMT+02:00 Fermín Galán Márquez <notifications@github.com>:

fgalan:
> data needed to register a new csource 
> the specific name of the fields

In order to provide precise feedback on that, could you add a JSON example of your proposed format? It is difficult to have an idea just looking to the noisy diff tab on this PR :)

jmcanterafonseca:
```
{
  "dataProvided": {
    "entities": [
      {
        "id": "room2",
        "type": "Room"
      }
    ],
    "attributes": [
      "relativeHumidity"
    ]
  },
  "source": {
    "endPoint": "http://localhost:1234",
    "name": "Relative Humidity Context Source"
  }
}
```

fgalan:
@jmcanterafonseca I have edited your comment to include fields were missing in the example, but included in the .apib. In particular, `expires`, `source.scope` and `source.description`.

fgalan:
Now, this  is an example of full-flegued subscription:

```
{
  "id": "abcdef",
  "description": "One subscription to rule them all",
  "subject": {
    "entities": [
      {
        "idPattern": ".*",
        "type": "Room"
      }
    ],
    "condition": {
      "attrs": [ "temperature " ],
      "expression": {
        "q": "temperature>40"
      }
    }
  },
  "notification": {
    "http": {
      "url": "http://localhost:1234"
    },
    "attrs": ["temperature", "humidity"],
    "timesSent": 12,
    "lastNotification": "2015-10-05T16:00:00.00Z"
    "lastSuccess": "2015-10-05T16:00:00.00Z"
  },
  "expires": "2016-04-05T14:00:00.00Z",
  "status": "active",
  "throttling": 5
}
```

fgalan:
Comparing both:

* regs don't have an `status` field, but probably it would be a good idea to have it, in order to stop/resume registrations and to state when the data source has a failure (i.e. a forwarded operation has failed).
* `throttling` seems not being meaninful in the case of registrations (in fact, probably is also a bad idea in the case of subscriptions, but too late to change, due to backward compatibility :) )
* The `subject` (subs) and `dataProvided` (regs) are roughtly similar. The difference is that attributes is in one depper level in the case of subs (within `conditions`) and regs lack of `expressions` (althought it shouldn't... more on this later, when talking about `scope`).
* `name` is redundant, as we already have `description`
* `description` in regs is inside `dataSource`. In fact, I think that `description` should apply to the whole reg resource, so comming at first level (in the same sense it works for subs).
* `timesSent`, `lastNotification`, `lastSuccess/lastFailure` could be also fields meaniful for a registration.
* Subs use a per-protocol endpoint structure, e.g `http` by the time being, but extensible to other possible mechanisms in the future. It would be clever to have the same for registrations.
* I understand `scope` in the same way `expression` works for subscription, i.e. narrowing a forwarding space with geolocation or other possible filters. In that sense is a filter that applies to entities+attrs covered by the registration in order to decide if the update/query has to be forwared. Thus, linked to entities+attrs and, in that way, part of dataProvide and not dataSource.
* `notification.attrs` (subs) probably doesn't have too much sense in regs case, as in the case of forwareding there isn't differenciation between triggering attributes and the attribute that has to be in the forwarded request.
* We need to think about some way of having "legacy" forwarding, in the same way we have legacy NGSIv1 notifications. This would not be part of the .apib, but of the corresponding .md.

jmcanterafonseca:
The operations that have to be supported by an NGSIv2 context provider are:

```
/v2/entities?type=XX&id=YY&attrs=at1,atn
```

I assume that at this stage we don't want to forward full query requests i.e. `q` attribute, 

We need essentially to support the attribute PATCH method so that we can forward update requests properly.   

```
/v2/entities/<entityId>/attrs
```

What do you think?

jmcanterafonseca:
general comment regarding the proposal made to align context providers with subscriptions, I don't think it is such a good idea to couple the API so much with forwarding as forwarding is a very specific feature implemented by Orion but out of the essential concept addressed by the API which is registering external providers of information that can be queried later. 

Nonetheless, I can understand that the current focus of Orion Broker for context providers is forwarding and particularly actuation over devices through the IOT Agents, so that can be an extra feature offered and that can be added later during a second phase.

fgalan:
Regarding the operations I guess you mean (in your previous comment verb is missing):

```
GET /v2/entities?type=XX,JJ&id=YY,ZZ&attrs=at1,atn
PATCH /v2/entities/<entityId>/attrs
```

I think it may sense. Next step in this line would be to include that "CPr requirement" in the specification in this PR along with a functional explanation on how forwarding works (paying special attention to "multi forwarding" case, e.g. a query on entity E ends in a forwared query for E-A1, E-A2 and E-A3 as E-A{1|2|3} belong to different CPrs).

This specification could be either in the .apib (if we agree forwarding is part of the NGSIv2 API) or specific .md in the Orion manual (if we agree on the opposite), which I see by your last comment https://github.com/telefonicaid/fiware-orion/pull/2990#issuecomment-334728474 is yet a matter of discussion.

PD. I agree `?q` shouldn't be a requirement for CPrs.

fgalan:
>  forwarding is a very specific feature implemented by Orion but out of the essential concept addressed by the API which is registering external providers of information that can be queried later.

I think they (registration & forwarding) are very coupled features. I mean, context availability registration is quite useless without the forwarding complementing feature. Registering CPrs just for the purpose of registering (or, in other words, using Orion as a kind of database for registrations, just a CRUD of them) is not very useful...

Any other opinion (apparts from @jmcanterafonseca and @fgalan)?

fgalan:
@jmcanterafonseca and me have been discussing about the definitive format for NGSIv2 registrations. a At the end we have agreed in something like this.

```
{
  "description": "The foobar provider",
  "dataProvided": {
    "entities": [
      {
        "id": "room2",
        "type": "Room"
      }
    ],
    "attrs": [
      "relativeHumidity"
    ],
    "expression": { [we are not sure if other expression elements are useful, e.g. q/mq]
         "georel": "near;maxDistance:100000",
         "geometry": "point",
         "coords": "40.418889,-3.691944"
    }
  },
  "provider": {
    "http": {
       "url": "http://localhost:1234"
    },
    "legacy": true|false,  [not in the NGSIv2 spec, but as Orion extension ]
    "forwarding": "query|update|all|none",
    <we could extend this provier parts with new properties we can need in the future, e.g. backup provider nature, etc.>
    }
  },
  "forwarding”: {
    "timesForwarded": 12,
    "lastForwarded": "2015-10-05T16:00:00.00Z",
    "lastSuccess": "2015-10-05T16:00:00.00Z",
    "lastFailure": "2015-10-04T16:00:00.00Z"
   },
  "status": "active|inactive|expired|failed",
  "expires": "2016-04-05T14:00:00.00Z"
}
```

AP for @jmcanterafonseca is to update PR according to this.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1990,30 +1990,76 @@ Response:
 
 # Group Registrations
 
-Context Registration allows to associate external services to context data. One of the main
-use cases of this functionality is the association of Context Providers.
+A Context Registration allows to bind external context information sources so that they can be
+play the role of providers
+of certain subsets (entities, attributes) of the context information space, including those located
+at specific geographical areas.
+A compliant Context Information provider shall expose an NGSI10 (legacy) or an NGSIv2 interface, in particular:
```
fgalan: I would adopt the same approach of the legacy mode for notifications so I don't describe legacy forwarding in the NGSIv2 .apib. In this sense, I'd suggest to remove this sentence.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1990,30 +1990,76 @@ Response:
...
++ `forwardingInformation`: Information related to the forwarding operations made against the provider.
+Automatically provided by the implementation. 
+
+The `provider` field contains the following subfields:
++ `http` : It is used to convey parameters for providers which deliver information through the HTTP protocol.
+(Only protocol supported nowadays).
+It must contain a subfield named `url` with the URL that serves as the endpoint that offers the providing interface.
+The endpooint must *not* include the protocol
+specific part (for instance `/v2/entities`). 
++ `legacy` : `true` if this provider only supports NGSIv1 protocol. `false` if this provider supports NGSIv2.
```
fgalan: I'd move this bullet outside the .apib, to the Orion .md. In fact, I'd suggest to add 2 sections similar to this ones (and in each one in the corresponding .md document):

https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/user/v1_v2_coexistence.md#ngsiv1-notification-with-ngsiv2-subscriptions
https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/user/ngsiv2_implementation_notes.md#legacy-attribute-format-in-notifications

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -2032,36 +2078,39 @@ Response:
...
-                 "value": "weather.example.com",
-                 "type": "Text"
-               },
-              "providingAuthority": {
-                 "value": "AEMET - Spain",
-                 "type": "Text"
-               }
+            "provider": {
+              "http": {
+                "url": "http://contextsource.example.org/v2/entities"
```
fgalan: Note 

> The endpoint must *not* include the protocol specific part (for instance `/v2/entities`)

Thus, I understand examples (this one an all the others in the .apib) should remove the `/v2/entities` part)

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -2032,36 +2078,39 @@ Response:
...
+              },
+              "legacy": false,
```
fgalan: Taking into account what I said at https://github.com/telefonicaid/fiware-orion/pull/2990/files#r144080050, `"legacy": ...` should be removed in all examples.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -2032,36 +2078,39 @@ Response:
...
+              "supportedForwardingMode": "all"
             },
-            "duration": "PT1M"
+            "expires": "2017-10-31T12:00:00",
+            "status": "active",
+            "forwardingInformation": {
+              "timesSent": 12,
+              "lastForwarding": "2017-10-05T16:00:00.00Z",
+              "lastFailure": "2017-10-06T16:00:00.00Z"
```
fgalan: `"status": "active"` and `lastFailure` equal to `lastForwarding` is an impossible combination :)

Simple solution: change `lastFailure` to `lastSuccess`. Not only here, but in all the examples with the same problem.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1990,30 +1990,76 @@ Response:
...
+  As soon as the forwarding operations start working again, the status is changed back to `active`.
++ `expires` : Registration expiration date in ISO8601 format. Permanent registrations must omit this field.
++ `forwardingInformation`: Information related to the forwarding operations made against the provider.
+Automatically provided by the implementation. 
+
+The `provider` field contains the following subfields:
++ `http` : It is used to convey parameters for providers which deliver information through the HTTP protocol.
+(Only protocol supported nowadays).
+It must contain a subfield named `url` with the URL that serves as the endpoint that offers the providing interface.
+The endpooint must *not* include the protocol
```
fgalan: Typo: endpoint

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: that could be part of a follow-up PR once this feature is implemented

removed at 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: fixed 78ce71dfa1b2c6952018ede3a9ef640c65abc1dd

fgalan:
LGTM

PR #3012, opened 2017-10-10
FIX Invalid JSON (quotes not escaped) #2955

### LINKED ISSUE #2955: Invalid JSON (quotes not escaped)

arigliano:
This PR fixes #2955 
As described in issue comments, this fix  consists of:

> The immediate solution would be to wrap the "ckey", when creating the details string, with either
> jsonInvalidCharsTransformation() or better JSONHelper::toJsonString()  (rest.cpp:114)

Let me know if this is ok or alternative approaches are preferable.

In addition, it was implemented a new functional test,  which passes, as well as existing ones.


src/lib/rest/rest.cpp
```diff
@@ -111,7 +111,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
 
   if (val == NULL || *val == 0)
   {
-    std::string  errorString = std::string("Empty right-hand-side for URI param /") + ckey + "/";
+    std::string  errorString = std::string("Empty right-hand-side for URI param /") + jsonInvalidCharsTransformation(ckey) + "/";
```
fgalan: As I mention in https://github.com/telefonicaid/fiware-orion/issues/2955#issuecomment-335530732, I think is better to fix this at OrionError::toJson(). I guess it should be easy to move the fix.

src/lib/rest/rest.cpp
```diff
@@ -111,7 +111,7 @@ static int uriArgumentGet(void* cbDataP, MHD_ValueKind kind, const char* ckey, c
```
fgalan: Please add a line to the CHANGE_NEXT_RELEASE file about this fix. Proposal:

```
- Fix: broken JSON due to unscaped quotes (") in NGSIv2 error description field (#2955)
```

src/lib/rest/OrionError.cpp
```diff
@@ -125,7 +125,22 @@ std::string OrionError::setStatusCodeAndSmartRender(ApiVersion apiVersion, HttpS
...
+
+  out += "{" + JSON_VALUE("error", reasonPhraseEscaped);
+  out += ",";
+  out += JSON_VALUE("description", detailsEscaped) + "}";
+
+  free(reasonPhraseEscaped);
+  free(detailsEscaped);
+
+  return out;
+//	return "{" + JSON_STR("error") + ":" + JSON_STR(reasonPhrase) + "," + JSON_STR("description") + ":" + JSON_STR(details) + "}";
```
fgalan: Leftover comment and 3 blank lines should be removed, pls.

fgalan:
It looks good! 

Howerver, note you have conflict with master branch that avoid merging this PR. I guess that is a small line disalingment in the CHANGES_NEXT_RELEASE file. 

The solution is to upgrade your branch with master and solve the conflict.

fgalan:
LGTM. Thx!

For the next PR, please end each line comment thread with a `Fixed in <commit_hash>` message, as detailed in the "pull request protocol" at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#pull-request-protocol:

>After discussion, the comment thread ends in one of the following ways:
>  * Fixed in <commit hash> in case the discussion involves a fix in the PR branch (which commit hash is included as reference)
>  * NTC, if finally nothing needs to be done (NTC = Nothing To Change)

kzangeli:
LGTM

fgalan:
@arigliano for next PRs please pay attention to execute a *full* regression of functional tests (i.e. `make ft`). I mean, not only check that the new or modified .test in the PR are working, but the full set of all them.

I'm saying this because after the changes merged in this PR, the following test were failing:

```
  o  1080_batch_query_context/batch_query_context_error_conditions.test
  o  1328_pause_subscriptions/pause_subscriptions_errors.test
  o  1607_query_with_q_and_ranges_but_invalid_operator/query_with_q_and_ranges_but_invalid_operator.test
  o  1678_geosubscriptions/geosubscriptions_errors.test
  o  1705_csub_cache_objects/csub_cache_objects_errors.test
```

Not a big problem, as they have been quickly fixed in PR https://github.com/telefonicaid/fiware-orion/pull/3016, but please take this into account in the future.

Thank you very much for your collaboration!

kzangeli:
For a full regression,better to use `make test`, as unit tests are not included in `make ft`.

Another way to run the functional tests is to use the `testHarness.sh` script directly, especially interesting when testing individual test cases, or a set of test cases.

The script `testHarness.sh` resides in test/functionsTest, and it has a number of arguments, use the `-u` argument to see a list of them:
```
testHarness.sh -u
Usage: testHarness.sh [-u (usage)]
                      [-v (verbose)]
                      [--filter <test filter>]
                      [--match <string for test to match>]
                      [--keep (don't remove output files)]
                      [--dryrun (don't execute any tests)]
                      [--dir <directory>]
                      [--fromIx <index of test where to start>]
                      [--ixList <list of test indexes>]
                      [--stopOnError (stop at first error encountered)]
                      [--no-duration (removes duration mark on successful tests)]
                      [--noCache (force broker to be started with the option --noCache)]
                      [--cache (force broker to be started without the option --noCache)]
                      [--noThreadpool (do not use a threadpool, unless specified by a test case. If not set, a thread pool of 200:20 is used by default in test cases which do not set notificationMode options)]
                      [ <directory or file> ]
```

I often use `--ixList "1 2 3 n"` after seeing errors in functests. As you might suspect, it executes only those functests with an index corresponding to one in the 'ixList'.

fgalan:
> Another way to run the functional tests is to use the testHarness.sh script directly

In this case, please ensure you have the contextBroker binary correctly installed in your execution path. If you forget that, you can get crazy after doing some changes in your code that compile (i.e. `make` works) but with test result that doesn't behave as if the changed were done because the compiled binary was not installed in the path... ;)

PR #3019, opened 2017-10-17
ADD Registration.[h|cpp] and mongoGetRegistrations.[h|cpp] modules

fgalan:
Note base branch is not master, but feature/ngsiv2_registrations

### DIFF
--- a/src/lib/apiTypesV2/EntID.h
+++ b/src/lib/apiTypesV2/EntID.h
@@ -1,0 +1,59 @@
+#ifndef SRC_LIB_APITYPESV2_ENTID_H_
+#define SRC_LIB_APITYPESV2_ENTID_H_
+
+/*
+*
+* Copyright 2017 Telefonica Investigacion y Desarrollo, S.A.U
+*
+* This file is part of Orion Context Broker.
+*
+* Orion Context Broker is free software: you can redistribute it and/or
+* modify it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* Orion Context Broker is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+* General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+*
+* For those usages not covered by this license please contact with
+* iot_support at tid dot es
+*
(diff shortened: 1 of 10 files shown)


src/lib/apiTypesV2/EntID.h
```diff
@@ -0,0 +1,59 @@
...
+* Author: Fermín Galán Márquez
+*/
+#include <string>
+
+
+namespace ngsiv2
+{
+/* ****************************************************************************
+*
+* EntID -
```
fgalan: This class was originally in Subscription.h, but now it will be also used by Registration.h, so better to have it separated.

A side advantage of alinging registrations an subscriptions in NGSIv2 :)

NTC (just informative)

src/lib/mongoBackend/mongoGetRegistrations.h
```diff
@@ -0,0 +1,71 @@
...
+*/
+extern void mongoGetRegistrations
+(
+  ngsiv2::Registration*                sub,
+  OrionError*                          oe,
+  const std::string&                   idReg,
+  std::map<std::string, std::string>&  uriParam,
+  const std::string&                   tenant
+);
+
```
kzangeli: Why not two files for two functionalities?

* mongoGetRegistrations.h/cpp
* mongoListRegistrations.h/cpp

Also, I would give them a common start in the name (same with the functions, of course):

* mongoRegistrationsGet.h/cpp
* mongoRegistrationsList.h/cpp

[ Subscriptions work like that is not a good enough excuse! :-) ]

fgalan: I agree as soon as subscrittion get and list get splited in the same way ;)

Do you want the split also in this PR? It can be done when mongoListRegistrations() get really used, i.e. the #if 0 removed.

kzangeli: If you want to include that split in the PR, I don't mind! Not at all :)
It's a good fix. More order and easier to find functions

fgalan: After looking in detail how it is done in the case of subscription, I think is better to have get and list operations in the same module. They are using common functionality:

```
    setNewSubscriptionId(sub, r);
    setDescription(sub, r);
    setSubject(sub, r);
    setNotification(sub, r, tenant);
    setStatus(sub, r);
```

I prefer it is better to have a common module (for get, list and common functions), that having 3 modules (one for get, other for list and other for common functions).

Under this new information, I think is better NTC.

kzangeli: Sure, I don't agree, but this is very subjective, and not very important.

NTC

kzangeli:
LGTM

PR #3020, opened 2017-10-17
FIX NGSIv1 responses don't use pretty-print JSON format any longer, as NGSIv2 responses work (potentially saving around 50% size)

fgalan:
Long-awaited implementation for #2760 :)

This is a clean re-work of the branch haderding/remove_ngsiv1_indent, in just one commit instead of 38. Easy to merge, easy to rollback if needed.

### DIFF
--- a/src/app/contextBroker/contextBroker.cpp
+++ b/src/app/contextBroker/contextBroker.cpp
@@ -280,6 +280,9 @@ bool            disableMetrics;
 int             reqTimeout;
 bool            insecureNotif;
 
+#ifdef PARANOID_JSON_INDENT
+bool            paranoidV1Indent;
+#endif
 
 
 /* ****************************************************************************
@@ -405,6 +408,10 @@ PaArgument paArgs[] =
   { "-disableMetrics", &disableMetrics,  "DISABLE_METRICS",    PaBool, PaOpt, false, false, true,             METRICS_DESC        },
 
   { "-insecureNotif", &insecureNotif, "INSECURE_NOTIF", PaBool, PaOpt, false, false, true, INSECURE_NOTIF },
+
+#ifdef PARANOID_JSON_INDENT
+  { "-paranoidV1Indent", &paranoidV1Indent, "PARANOID_V1_INDENT", PaBool, PaHid, false, false, true, "you shouldn't use this ;)" },
+#endif
 
   PA_END_OF_ARGS
 };
--- a/src/lib/common/globals.h
+++ b/src/lib/common/globals.h
@@ -32,7 +32,11 @@
 #include "common/Timer.h"
 #include "common/sem.h"
 
-
+/* ****************************************************************************
+*
+* FIXME P10: temporal definition, it should be removed soon
+*/
+#define PARANOID_JSON_INDENT
 
 /* ****************************************************************************
 *
@@ -221,6 +225,9 @@ extern bool               disableCusNotif;
 
 extern bool               insecureNotif;
 
+#ifdef PARANOID_JSON_INDENT
+extern bool               paranoidV1Indent;
+#endif
 
 
 /* ****************************************************************************
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -1082,3 +1082,103 @@ void toLowercase(char* s)
     ++s;
   }
 }
+
+
+#ifdef PARANOID_JSON_INDENT
+// -----------------------------------------------------------------------------
+//
+// jsonFix - replace all ':' with ' :' in a JSON buffer (as some other fixes)
+//
+// 01. Count number of replacements - to get necessary size of out buffer
+// 02. Go over the in buffer and copy one by one in -> out, avoiding replacements
+//     while inside strings.
+//
+char* jsonFix(const char* in)
+{
+  int          outSize   = 1;  // Room for zero-termination char
+  const char*  inP       = in;
+  char*        out;
+  bool         inString  = false;
+
+  while (*inP != 0)
+  {
+    // If backslash, step over it AND the next char
+    if (*inP == '\\')
+    {
+      outSize += 2;
+      inP     += 2;
+      continue;
+    }
+
+    // If double-quote, change state (inString false/true)
+    if (*inP == '"')
+    {
+      inString = (inString == true)? false : true;
+      ++outSize;
+      ++inP;
+      continue;
+    }
+
+    // If not inside a string, detect ':' and add an extra byte if found
+    if (!inString)
+    {
+      if (*inP == ':')
+      {
+        ++outSize;  // make room for an extra space
+      }
+    }
+
+    // Step over current char and continue
+    ++outSize;
+    ++inP;
+  }
+  //printf("in-size: %lu, out-size: %d\n", strlen(in), outSize);
+
+  // +1 for \n
+  ++outSize;
+
+  // Allocate room for the outgoing string
+  out = (char*) malloc(outSize);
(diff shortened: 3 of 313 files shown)


src/app/contextBroker/contextBroker.cpp
```diff
@@ -280,6 +280,9 @@ bool            disableMetrics;
 int             reqTimeout;
 bool            insecureNotif;
 
+#ifdef PARANOID_JSON_INDENT
```
kzangeli: paranoid ...
Perhaps look for a better name for the define.

JSON_NO_WS, for example

src/app/contextBroker/contextBroker.cpp
```diff
@@ -406,6 +409,10 @@ PaArgument paArgs[] =
 
   { "-insecureNotif", &insecureNotif, "INSECURE_NOTIF", PaBool, PaOpt, false, false, true, INSECURE_NOTIF },
 
+#ifdef PARANOID_JSON_INDENT
+  { "-paranoidV1Indent", &paranoidV1Indent, "PARANOID_V1_INDENT", PaBool, PaHid, false, false, true, "you shouldn't use this ;)" },
```
kzangeli: same same, nothing paranoid with giving a human readable format.
That actually might be a good name for it: "Human Readable Format":

So, I'd propose:
```
#ifdef HUMAN_READABLE_FORMAT ...
```

src/lib/common/globals.h
```diff
@@ -32,7 +32,11 @@
 #include "common/Timer.h"
 #include "common/sem.h"
 
-
+/* ****************************************************************************
+*
+* FIXME P10: temporal definition, it should be removed soon
+*/
+#define PARANOID_JSON_INDENT
```
kzangeli: Ah, it will be removed?
So, perhaps this PR should include the removal of the define ... ?

src/lib/common/globals.h
```diff
@@ -221,6 +225,9 @@ extern bool               disableCusNotif;
 
 extern bool               insecureNotif;
 
+#ifdef PARANOID_JSON_INDENT
+extern bool               paranoidV1Indent;
```
kzangeli: If this var is to be maintained (I kind of doubt it), same comment as before, look for a name in the line of 'human readable format', which is what the whole thing is about

src/lib/common/string.cpp
```diff
@@ -1082,3 +1082,103 @@ void toLowercase(char* s)
...
+      {
+        ++outSize;  // make room for an extra space
+      }
+    }
+
+    // Step over current char and continue
+    ++outSize;
+    ++inP;
+  }
+  //printf("in-size: %lu, out-size: %d\n", strlen(in), outSize);
```
kzangeli: This seems like an old line of debug that should be removed

src/lib/common/string.h
```diff
@@ -31,6 +31,7 @@
 #include <vector>
 
 #include "common/limits.h"
+#include "common/globals.h"  // FIXME P10: PARANOID_JSON_INDENT
```
kzangeli: FIXME P10 ... ?
If it's a P10, it should be fixed before we merge the PR

src/lib/convenience/AppendContextElementRequest.cpp
```diff
@@ -121,7 +125,7 @@ std::string AppendContextElementRequest::check
 *
 * present - 
 */
-void AppendContextElementRequest::present(std::string indent)
+void AppendContextElementRequest::present(const std::string&  indent)
```
kzangeli: Good!  One less ...

NTC

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -492,19 +493,18 @@ std::string ContextAttribute::getLocation(ApiVersion apiVersion) const
...
 )
 {
   std::string  out                    = "";
   bool         commaAfterContextValue = metadataVector.size() != 0;
   bool         commaAfterType         = !omitValue || commaAfterContextValue;
 
-  out += startTag(indent, name, false);
-  out += valueTag(indent + "  ", "type",         type,  commaAfterType);
+  out += startTag(name, false);
+  out += valueTag("type",         type,  commaAfterType);
```
kzangeli: Old indent no longer necessary

src/app/contextBroker/contextBroker.cpp
fgalan: The name is not really important, it will be removed soon if everything goes as expected, see https://github.com/telefonicaid/fiware-orion/issues/2760#issuecomment-337334053.

src/app/contextBroker/contextBroker.cpp
fgalan: It is not about being human readable... it is a about to be prepare in the rare case some NGSIv1 client breaks, so the "paranoid" about that. However, as I said in a previous comment, the name of the define is not really important.

src/lib/common/globals.h
fgalan: No, this PR cannot include that, but a soon one will do. See the plan in https://github.com/telefonicaid/fiware-orion/issues/2760#issuecomment-337334053.

src/lib/common/globals.h
fgalan: I will not be mantained. See https://github.com/telefonicaid/fiware-orion/issues/2760#issuecomment-337334053.

src/lib/common/string.h
fgalan: Not in this PR, but in a soon next one. See https://github.com/telefonicaid/fiware-orion/issues/2760#issuecomment-337334053.

src/lib/ngsi/ContextAttribute.cpp
fgalan: You are right, but I prefer not adding new commits to this PR except if really needed (i.e. if you find some criticial bug or problem). The less commits, the easy to rollback if we need so in the future (let's hope no, but better to be conservative). Of course we always can do squash&merge, but I prefer to add unneded complexity. 

Another reason is to be sure that this PR, when merged, mirrors exactly the same changes that the old branch haderding/remove_ngsiv1_indent (yet existing) has.

src/lib/common/string.cpp
fgalan: This is within a #ifdef PARANOID_JSON_INDENT block, so it will be removed soon.

src/app/contextBroker/contextBroker.cpp
kzangeli: Ok

NTC

src/app/contextBroker/contextBroker.cpp
kzangeli: Ok

NTC

src/lib/common/globals.h
kzangeli: Ok

NTC

src/lib/common/globals.h
kzangeli: Ok

NTC

src/lib/common/string.cpp
kzangeli: No probs

NTC

src/lib/common/string.h
kzangeli: Ok

NTC

src/lib/ngsi/ContextAttribute.cpp
kzangeli: No probs

NTC

kzangeli:
LGTM

Issue #3030, opened 2017-11-02
Add support for accessing response headers when making CORS requests


aarranz:
Current CORS support provided in #501 (although it is still WIP) doesn't allow reading headers from context broker responses. This will affect clients trying to access not only to the `Fiware-Correlator` header that is always returned by orion, but also to other headers used in the different API endpoints (e.g. The `Fiware-Total-Count` header returned when using the `count` option).

So, currently, listing entities by making a `GET` request to `/v2/entities` is working (using CORS), but if the JavaScript client tries to access the `Fiware-correlator` header the following error is raised (using Chrome):

```
Refused to get unsafe header "Fiware-correlator"
```

As, said before, this error is also raised when trying to access the `Fiware-Total-Count` header when using `options=count`.

McMutton:
Thanks for the suggestion @aarranz, although initially planned to be discussed, I had completely forgot about this. 

We can fix this by adding Access-Control-Exposed-Headers to the response with the header values the client would access.

jmcanterafonseca:
yes, totally agree. We need this. @McMutton please ensure you fix that in the last CORS PR.

fgalan:
@McMutton once PR https://github.com/telefonicaid/fiware-orion/pull/3032 is finished, do you plan continue with this issue? I think it would be great, in order to complete (as much as we now) the CORS issues for next Orion release (1.10.0).

McMutton:
@fgalan Now that #3032 is merged, I will be working on this, you can assign the issue to me.

fgalan:
Assigned. Thx!

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/3043

@aarranz as original issue reported, do you want to test before closing the issue, please? You can compile Orion from master branch or use "fiware/orion:latest" from docker hub.

aarranz:
I will test ASAP, probably today :)

aarranz:
I have make several tests creating, deleting and listing entities using services, servicepaths, correlators, ... and seems that everything is working as expected 👍.

PR #3032, opened 2017-11-03
ISSUE: 501. CORS Support.

McMutton:
Implements #501 

Rest of the /v2 resources now support CORS. New unit and functional tests are added.


McMutton:
About @jmcanterafonseca 's comment

> In order to have better extensibility or maintainability of the code it would be great to define the headers as an array of constants, so that when new headers are needed it would be enough to only touch the constant array.

about the line below:

    MHD_add_response_header(response, "Access-Control-Allow-Headers", "Content-Type, Fiware-Service, Fiware-Servicepath, Ngsiv2-AttrsFormat, Fiware-Correlator, X-Forwarded-For, X-Real-IP, X-Auth-Token");

I did what was suggested but then, storing these values in an array then looping over it to form this string again felt a bit of an overkill.

Should we instead have a string constant for each HTTP header and them create a constant string concatenating them in `restReply.cpp` as `cors_allowed_headers`?

What is the standard way of handling this kind of issues in the context of Orion? I will do another commit for this after your feedback.

test/functionalTest/cases/0501_cors/allowed_headers.test
```diff
@@ -40,6 +40,42 @@ orionCurl --url /v2/entities -X OPTIONS
 echo
 echo
 
+echo "03. Allowed headers: GET DELETE OPTIONS"
```
fgalan: As far as I undertand, https://github.com/telefonicaid/fiware-orion/pull/3032/files#diff-c7a50c3b97af231671cfad8674374d21R844 is adding 10 new routes (i.e. 10 new URLs supporting OPTIONS).

However, only 6 new URL/steps have been included...

Maybe I'm missing something? :)

test/functionalTest/cases/0501_cors/any_origin.test
```diff
@@ -40,6 +40,12 @@ orionCurl --url /v2 -X OPTIONS
 echo
 echo
 
+echo "03. Bad Verb Entry Points"
```
fgalan: This is already being tested https://github.com/telefonicaid/fiware-orion/pull/3032/files#diff-a375d95dcfcfe65f4a0395db510119dbR33, isn't it?

In that case, maybe this step is redundand in this .test and could be removed...

fgalan:
>I did what was suggested but then, storing these values in an array then looping over it to form this string again felt a bit of an overkill.
>
>Should we instead have a string constant for each HTTP header and them create a constant string concatenating them in restReply.cpp as cors_allowed_headers?
>
> What is the standard way of handling this kind of issues in the context of Orion? I will do another commit for this after your feedback.

I don't remember if we have an standard way for that... however, the approach of have a string constant for each HTTP header and them create a constant string concatenating them in restReply.cpp as cors_allowed_headers sound good. However, instead of using string constants for each HTTP header maybe a `#define` could be use.

I'd suggest you try something in that line a let's see how it looks like.

fgalan:
Would it take advantage of this PR to consider https://github.com/telefonicaid/fiware-orion/issues/3030 ? What do you think @McMutton ?

test/functionalTest/cases/0501_cors/allowed_headers.test
McMutton: We had discussed about this more than a few times, it is normal to get confused :)

Right now we have 12 different routes for /v2 which have 8 different verb sets. 

As an example, by having `optionsGetPostOnly` handler, we cover both `/v2/entities` and `/v2/subscriptions`. So that's why we have less handlers than the number of routes, similar to bad verb handlers.

test/functionalTest/cases/0501_cors/any_origin.test
McMutton: Fixed in 88eeebb

McMutton:
> Would it take advantage of this PR to consider #3030 ? What do you think @McMutton ?

I would prefer having a separate PR for this since a new response header (`Access-Control-Expose-Headers`) is going to be needed so having its discussion in a separate thread would be easier for the community and for us :) 

I am not sure if the new PR should belong to #3030 or #501 though, what do you think?

fgalan:
> I am not sure if the new PR should belong to #3030 or #501 though, what do you think?

I think it would be better to make that PR belongs to #3030

McMutton:
@fgalan, one more question about the HTTP header definitions.

 I thought one of these might be suitable to hold the definitions but I am not sure which: `httpHeaders.h`, `globals.h` and `rest.h`

test/functionalTest/cases/0501_cors/allowed_headers.test
fgalan: Ok, I think I see now your point...

Note functional tests should be designed from usage/API perspective, not implementation perspective. I mean, I agree with you that in this particular case and with the current implementation, with just the 8 operations you include in the test you can cover the 8 methods options handlers.

However, imagine that at some point and for some unknown reason at the present moment the implementation changes (but not the API). You would need to re-think the test in order to align with the new implementation. Thus, if the test is designed with API perspective from the very beginning, you would change the implementation without changing test... which at the end is the actual purpose of functional testing ;)

Moreover, in this case the modification in the .test is pretty simple. Just adding the 4 missing cases.

fgalan: It seems a similar situation occurs with cors_bad_verbs.test. It would be great (and easy :) to pass from 8 to 12 URLs also in that test.

fgalan:
> @fgalan, one more question about the HTTP header definitions.
>
> I thought one of these might be suitable to hold the definitions but I am not sure which: httpHeaders.h, globals.h and rest.h

I think the best one would be `httpHeaders.h`. @kzangeli what do you think?

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
```diff
@@ -55,11 +56,16 @@ std::string badVerbAllNotDelete
   OrionError   oe(SccBadVerb, ERROR_DESC_BAD_VERB);
 
   ciP->httpHeader.push_back("Allow");
-  ciP->httpHeaderValue.push_back("GET, PATCH, POST, PUT");
+  std::string headerValue = "GET, PATCH, POST, PUT";
```
kzangeli: Just for curiosity, is there any 'established' order for the verbs/methods ?

src/lib/rest/restReply.cpp
```diff
@@ -162,8 +162,8 @@ void restReply(ConnectionInfo* ciP, const std::string& _answer)
     }
   }
 
-  //Check if CORS is enabled
-  if (strlen(restAllowedOrigin) > 0)
+  //Check if CORS is enabled and the response is not bad verb response
+  if (strlen(restAllowedOrigin) > 0 && ciP->httpStatusCode != SccBadVerb)
```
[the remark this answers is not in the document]
fgalan: Suggested names:

* `allowedOrigin` for the string
* `hasAllowedOrigin` for the bool

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
fgalan: Interesting question... what does the CORS specification say about that?

kzangeli:
> I think the best one would be httpHeaders.h. @kzangeli what do you think?

Yeah, that sounds good

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
McMutton: CORS specification does not provide any rules for the order of the methods and the example scenarios provided do not seem to be following any pattern.

Please see the example scneario at the end of the subsection: [7.1.5 Cross-Origin Request with Preflight](https://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0).

src/lib/rest/restReply.cpp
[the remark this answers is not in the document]
McMutton: Agreed, this would be much better, thanks for the suggestion. 

One comment about the name of the boolean variable. In the code we basically do the check (the length check as of now) to see if CORS is enabled or not. I think it would make more sense to have the variable name something like `corsEnabled`. This would also make the code easier to understand. 

What do you think @kzangeli @fgalan ?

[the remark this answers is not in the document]
kzangeli: `corsEnabled` sounds great

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
kzangeli: So, NTC (alphabetic order seems ok to me)

I think I used an (by myself invented) "estimated use-frequency order", as: GOT. POST, PUT. PATCH ...
We should change the badVerb service routines to do alphabetic order instead . Can't go wrong ... (except perhaps for 100% backward compatibility :-( )
To be discussed, and nothing to be done in this PR. of course.

test/functionalTest/cases/0501_cors/allowed_headers.test
McMutton: And now... I see your point :) 

You are absolutely right, thanks for pointing this out. I'll fix this in the next commit.

src/lib/rest/restReply.cpp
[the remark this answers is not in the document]
fgalan: `corsEnabled` is fine for the booll

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
fgalan: > We should change the badVerb service routines to do alphabetic order instead . Can't go wrong ... (except perhaps for 100% backward compatibility :-( )
> To be discussed, and nothing to be done in this PR. of course.

Taking into account order doesn't matter from specification point of view, I don't see any value on changing that.

test/functionalTest/cases/0501_cors/allowed_headers.test
McMutton: Oh.. I just realized the name of this test file should have been allowed_methods. I will fixed that as well.

src/lib/serviceRoutinesV2/badVerbAllNotDelete.cpp
kzangeli: > Taking into account order doesn't matter from specification point of view, I don't see any value on changing that.

Uniformity seems like reason enough to me (surprised to hear that you don't care :-))
If it wasn't for the possible backward compatibility I would change it, it's easy enough.
But, let's not

test/functionalTest/cases/0501_cors/allowed_headers.test
McMutton: Fixed in b9a8b2f

src/lib/rest/restReply.cpp
```diff
@@ -162,8 +162,8 @@ void restReply(ConnectionInfo* ciP, const std::string& _answer)
     }
   }
 
-  //Check if CORS is enabled
-  if (strlen(restAllowedOrigin) > 0)
+  //Check if CORS is enabled and the response is not bad verb response
```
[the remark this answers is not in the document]
McMutton: Fixed in b9a8b2f

src/lib/rest/restReply.cpp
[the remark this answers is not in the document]
McMutton: Fixed in b9a8b2f

src/lib/serviceRoutines/badVerbGetDeleteOnly.cpp
```diff
@@ -54,7 +55,13 @@ std::string badVerbGetDeleteOnly
   OrionError   oe(SccBadVerb, ERROR_DESC_BAD_VERB);
 
   ciP->httpHeader.push_back("Allow");
-  ciP->httpHeaderValue.push_back("GET, DELETE");
+  std::string headerValue = "GET, DELETE";
+  //OPTIONS verb is only available for V2 API
+  if (strlen(restAllowedOrigin) > 0 && ciP->apiVersion == V2)
```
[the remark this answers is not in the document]
McMutton: Fixed in b9a8b2f

McMutton:
I have added a set of HTTP header definitions to `HttpHeaders.h` along with a `CORS_ALLOWED_HEADERS` macro. 

As you can see we only have the headers that are used in `CORS_ALLOWED_HEADERS` but this list of definitions can easily be extended to get rid of the hard-coded string occurrences in source files.

I know it is not related with this PR but this might be a good chance to talk if it would worth to create an issue focusing on replacing hard-coded HTTP header strings with constants to be defined in `HttpHeaders.h`

McMutton:
In `HttpHeaders.h`, we now have a list of definitions for HTTP headers that we allow in CORS mode.

@kzangeli , @fgalan I know it is not related to this PR but this might be a good opportunity to discuss if it is worth to create an issue for adding more header definitions to the list. 

There are many occurrences of hard-coded string values of HTTP headers in our source files and all these can be replaced by macros which can be added to HttpHeaders.h, extending the list of HTTP header definitions we have.

src/lib/rest/rest.cpp
```diff
@@ -1781,6 +1782,7 @@ void restInit
   mhdConnectionTimeout = _mhdTimeoutInSeconds;
 
   strncpy(restAllowedOrigin, _allowedOrigin, sizeof(restAllowedOrigin));
+  corsEnabled = (restAllowedOrigin[0] != 0);
```
fgalan: A cryptic way of checking if the char string is empty but it's ok... :)

Just to be sure... how `restAllowedOrigin` array is initialized? Can we asure that first element is zero? Or it is random and could be initilizalied with random stuff?

fgalan:
> @kzangeli , @fgalan I know it is not related to this PR but this might be a good opportunity to discuss if it is worth to create an issue for adding more header definitions to the list.
>
> There are many occurrences of hard-coded string values of HTTP headers in our source files and all these can be replaced by macros which can be added to HttpHeaders.h, extending the list of HTTP header definitions we have.

Makes fully sense. Issue has been created: https://github.com/telefonicaid/fiware-orion/issues/3035

src/lib/rest/rest.cpp
McMutton: Actually it was suggested by @kzangeli :) I think it is better than checking the length.

`restAllowedOrigin` gets the value of -corsOrigin switch or if the switch is not used, initialized with empty string, `""`. For empty string, the first char will always be zero.

By the way, since we are talking about the variable, we have re-used some of the code Jose Manuel wrote for the first CORS implementation and some of the variable names that made sense back then, now got a bit confusing. Especially restAllowedOrigin. So this is the flow we have now:

1. We take the `-corsOrigin` switch value and assign it to `allowedOrigin` in `contextBroker.cpp`
2. We call the restInit function, passing the value of `allowedOrigin` variable in `contextBroker.cpp`
3. We define a global variable in `rest.cpp` called `restAllowedOrigin`
4. In `restInit` function, the passed value is mapped to `_allowedOrigin` in the signature and the value is re-assigned to `restAllowedOrigin`. 

As you see, there are a lot of name changes which makes the code hard to read and harder to maintain. I propose using the same name corsOrigin except the method signature of `restInit` where we can have it as `_corsOrigin`.  Same goes for restCORSMaxAge.

fgalan: > if the switch is not used, initialized with empty string, ""

That's the piece of the puzzle I missed. I didn't know if the argument parsing library puts `""` in the case of missing swich or just left the holding variable with undefined content. I understand is the former.

> I propose using the same name corsOrigin except the method signature of restInit where we can have it as _corsOrigin. Same goes for restCORSMaxAge.

Sounds good.

fgalan: I guess this is related with

```
{ "-corsOrigin",       allowedOrigin,     "ALLOWED_ORIGIN",    PaString, PaOpt, _i "",          PaNL,  PaNL,     ALLOWED_ORIGIN_DESC    },
```
That `_i ""` ensures that in the case of not using `-corsOrigin` at the CLI, the value for allowedOrigin is univocally set to empty string, isn't it @kzangeli  ?

kzangeli: That's it

fgalan: Ok. So this comment (and with it the whole PR, as far as I remember) is only pending of a new commit with the refactor in the variable names, as @McMutton suggested above.

fgalan:
> Ok. So this comment (and with it the whole PR, as far as I remember) is only pending of a new commit with the refactor in the variable names, as @McMutton suggested above.

Anything else pending, @McMutton please?

McMutton:
> Anything else pending, @McMutton please?

No, there are no pending tasks, I plan to do the final commit today.

src/lib/rest/rest.cpp
McMutton: Fixed in e3a6471

fgalan:
I don't like `maxAge` too much as variable n ame, as it very generic and could be confused with some other "max ages" not related with CORS introduced in the future. However, we can keep it as it is now and change in the future if some conflict occur.

LGTM. Thasnk for the work! Passing the ball to @kzangeli for final LGTM.

kzangeli:
LGTM

PR #3037, opened 2017-11-13
Fix afiliations in NGSIv2 spec

fgalan:
Now that we are close to release a new RC (by the end of November) of NGSIv2 stable, it is a good time to do a little fixing in authors and editors affiliations.

@kzangeli @mrutid @jmcanterafonseca


kzangeli:
LGTM

Issue #2587, opened 2016-10-07
Update attribute value to null fails (it doesn't actually change the attribute value)


kzangeli:
Updating an attribute with a `null` value doesn't work:

```
payload='{
  "type": "T",
  "value": null
}'
orionCurl --url /v2/entities/E1/attrs/A1 --payload "$payload" -X PUT
```

The attribute type is changed, but the value stays the same.
The `null` value of A1 in the update is interpreted by the broker as "I don't want to change the value of the attribute".

We may need here a 'bool valueGiven' or something similar to fix this problem.

fgalan:
I've tried with current master (1.5.0-next, git version: f75b0e2acf2f62e132ae274608261ca9881b1819) and it seems to work ok.

This is the .test that I have used:

```
# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
#
# This file is part of Orion Context Broker.
#
# Orion Context Broker is free software: you can redistribute it and/or
# modify it under the terms of the GNU Affero General Public License as
# published by the Free Software Foundation, either version 3 of the
# License, or (at your option) any later version.
#
# Orion Context Broker is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
# General Public License for more details.
#
# You should have received a copy of the GNU Affero General Public License
# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
#
# For those usages not covered by this license please contact with
# iot_support at tid dot es

# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh

--NAME--
Update attr to null bugfix

--SHELL-INIT--
dbInit CB
brokerStart CB

--SHELL--

#
# 01. Create entity with attr value X and type T1
# 02. Get entity and see attr value X and type T1
# 03. Update entity with attr value null and type T2
# 04. Get entity and see attr value null and type T2
#

echo "01. Create entity with attr value X and type T1"
echo "==============================================="
payload='{
  "id": "E",
  "type": "T",
  "A": {
      "value": "X",
      "type": "T1"
   }
}'
orionCurl --url /v2/entities --payload "$payload"
echo
echo


echo "02. Get entity and see attr value X and type T1"
echo "==============================================="
orionCurl --url /v2/entities/E
echo
echo


echo "03. Update entity with attr value null and type T2"
[...]

PR #3042, opened 2017-11-16
ADD additional test for 2587 case

fgalan:
New .test for #2587. A copy of the existing .test but using no value instead of JSON `null`.


kzangeli:
LGTM

PR #3043, opened 2017-11-16
ISSUE: 3030. CORS Expose Headers.

McMutton:
Access-Control-Expose-Headers is now sent in the response, allowing the client to access certain response headers. 

With this PR, I took the liberty to add some improvements to our logic with respect to the CORS specification: We are no longer adding any CORS headers to the response if the client fails to add Origin to the request or the Origin value does not match the one allowed by Orion.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,6 +1,5 @@
-- Add: CORS Preflight Requests support for all NGSIv2 resources, -corsMaxAge switch (#501)
+- Add: CORS Preflight Requests support for all NGSIv2 resources, -corsMaxAge switch, CORS exposed headers (#501, #3030)
 - Fix: null not working in q/mq filter in subscriptions (#2998)
 - Fix: case-sensitive header duplication (e.g. "Content-Type" and "Content-type") in custom notifications (#2893)
 - Fix: bug in GTE and LTE operations in query filters (q/mq), both for GET operations and subscriptions (#2995)
 - Fix: Wrong "max one service-path allowed for subscriptions" in NGSIv2 subscription operation (#2948)
-- Fix: Add support for accessing response headers when making CORS requests (#3030)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -3,3 +3,4 @@
 - Fix: case-sensitive header duplication (e.g. "Content-Type" and "Content-type") in custom notifications (#2893)
 - Fix: bug in GTE and LTE operations in query filters (q/mq), both for GET operations and subscriptions (#2995)
 - Fix: Wrong "max one service-path allowed for subscriptions" in NGSIv2 subscription operation (#2948)
+- Fix: Add support for accessing response headers when making CORS requests (#3030)
--- a/src/lib/rest/HttpHeaders.h
+++ b/src/lib/rest/HttpHeaders.h
@@ -36,18 +36,21 @@
 *
 * HTTP Headers -
 */
-#define CONTENT_TYPE                 "Content-Type"
-#define FIWARE_SERVICE               "Fiware-Service"
-#define X_AUTH_TOKEN                 "X-Auth-Token"
-#define X_REAL_IP                    "X-Real-IP"
-#define X_FORWARDED_FOR              "X-Forwarded-For"
-#define FIWARE_CORRELATOR            "Fiware-Correlator"
-#define NGSIV2_ATTRSFORMAT           "Ngsiv2-AttrsFormat"
-#define FIWARE_SERVICEPATH           "Fiware-Servicepath"
-#define ACCESS_CONTROL_ALLOW_ORIGIN  "Access-Control-Allow-Origin"
-#define ACCESS_CONTROL_ALLOW_HEADERS "Access-Control-Allow-Headers"
-#define ACCESS_CONTROL_ALLOW_METHODS "Access-Control-Allow-Methods"
-#define ACCESS_CONTROL_MAX_AGE       "Access-Control-Max-Age"
+#define ACCESS_CONTROL_ALLOW_ORIGIN   "Access-Control-Allow-Origin"
+#define ACCESS_CONTROL_ALLOW_HEADERS  "Access-Control-Allow-Headers"
+#define ACCESS_CONTROL_ALLOW_METHODS  "Access-Control-Allow-Methods"
+#define ACCESS_CONTROL_MAX_AGE        "Access-Control-Max-Age"
+#define ACCESS_CONTROL_EXPOSE_HEADERS "Access-Control-Expose-Headers"
+#define CONTENT_TYPE                  "Content-Type"
+#define FIWARE_CORRELATOR             "Fiware-Correlator"
+#define FIWARE_SERVICE                "Fiware-Service"
+#define FIWARE_SERVICEPATH            "Fiware-Servicepath"
+#define FIWARE_TOTAL_COUNT            "Fiware-Total-Count"
+#define NGSIV2_ATTRSFORMAT            "Ngsiv2-AttrsFormat"
+#define RESOURCE_LOCATION             "Location"
+#define X_AUTH_TOKEN                  "X-Auth-Token"
+#define X_REAL_IP                     "X-Real-IP"
+#define X_FORWARDED_FOR               "X-Forwarded-For"
 
 
 
@@ -55,7 +58,15 @@
 *
 * CORS Allowed Headers -
 */
-#define CORS_ALLOWED_HEADERS CONTENT_TYPE", "FIWARE_SERVICE", "FIWARE_SERVICEPATH", "NGSIV2_ATTRSFORMAT", "FIWARE_CORRELATOR", "X_FORWARDED_FOR", " X_REAL_IP", " X_AUTH_TOKEN
+#define CORS_ALLOWED_HEADERS CONTENT_TYPE", "FIWARE_SERVICE", "FIWARE_SERVICEPATH", "NGSIV2_ATTRSFORMAT", "FIWARE_CORRELATOR", "X_FORWARDED_FOR", "X_REAL_IP", "X_AUTH_TOKEN
+
+
+
+/* ****************************************************************************
+*
+* CORS Exposed Headers -
+*/
+#define CORS_EXPOSED_HEADERS FIWARE_CORRELATOR", "FIWARE_TOTAL_COUNT", "RESOURCE_LOCATION
 
 
 
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -816,9 +816,9 @@ void firstServicePath(const char* servicePath, char* servicePath0, int servicePa
 * isOriginAllowedForCORS - checks the Origin header of the request and returns
 * true if that Origin is allowed to make a CORS request
 */
-bool isOriginAllowedForCORS(std::string requestOrigin)
+bool isOriginAllowedForCORS(const std::string& requestOrigin)
 {
-  return ( (requestOrigin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(requestOrigin.c_str(), corsOrigin) == 0)) );
+  return ((requestOrigin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(requestOrigin.c_str(), corsOrigin) == 0)));
 }
 
 
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -813,6 +813,18 @@ void firstServicePath(const char* servicePath, char* servicePath0, int servicePa
 
 /* ****************************************************************************
 *
+* isOriginAllowedForCORS - checks the Origin header of the request and returns
+* true if that Origin is allowed to make a CORS request
+*/
+bool isOriginAllowedForCORS(std::string requestOrigin)
+{
+  return ( (requestOrigin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(requestOrigin.c_str(), corsOrigin) == 0)) );
+}
+
+
+
+/* ****************************************************************************
+*
 * servicePathSplit -
 */
 int servicePathSplit(ConnectionInfo* ciP)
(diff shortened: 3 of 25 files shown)


test/functionalTest/cases/0501_cors/max_age.test
```diff
@@ -41,6 +41,7 @@ HTTP/1.1 200 OK
 Content-Length: 0
 Access-Control-Max-Age: 600
 Access-Control-Allow-Headers: REGEX(.*)
+Access-Control-Expose-Headers: REGEX(.*)
```
fgalan: As far as I see, the list is explicit (`define CORS_EXPOSED_HEADERS FIWARE_CORRELATOR", "FIWARE_TOTAL_COUNT", "RESOURCE_LOCATION`). Thus, I'd suggest to use the actual headers instead a generic regex, in order to make the test stronger.

src/lib/rest/restReply.cpp
```diff
@@ -179,16 +181,27 @@ void restReply(ConnectionInfo* ciP, const std::string& _answer)
       {
         MHD_add_response_header(response, ACCESS_CONTROL_ALLOW_ORIGIN, corsOrigin);
       }
-    }
+      // If there is no match, originAllowed flag is set to false
+      else {
```
fgalan: Style: `{` in next line.

src/lib/rest/HttpHeaders.h
```diff
@@ -36,26 +36,37 @@
...
-#define CORS_ALLOWED_HEADERS CONTENT_TYPE", "FIWARE_SERVICE", "FIWARE_SERVICEPATH", "NGSIV2_ATTRSFORMAT", "FIWARE_CORRELATOR", "X_FORWARDED_FOR", " X_REAL_IP", " X_AUTH_TOKEN
+#define CORS_ALLOWED_HEADERS CONTENT_TYPE", "FIWARE_SERVICE", "FIWARE_SERVICEPATH", "NGSIV2_ATTRSFORMAT", "FIWARE_CORRELATOR", "X_FORWARDED_FOR", "X_REAL_IP", "X_AUTH_TOKEN
+
+
+
+/* ****************************************************************************
+*
+* CORS Exposed Headers -
+*/
+#define CORS_EXPOSED_HEADERS FIWARE_CORRELATOR", "FIWARE_TOTAL_COUNT", "RESOURCE_LOCATION
```
fgalan: Should the three exposed headers be documented as part of Orion documentation? I mean, is this list something implementation-specific (so it should) or derived directly from CORS specification without ambiguity?

CHANGES_NEXT_RELEASE
```diff
@@ -3,3 +3,4 @@
 - Fix: case-sensitive header duplication (e.g. "Content-Type" and "Content-type") in custom notifications (#2893)
 - Fix: bug in GTE and LTE operations in query filters (q/mq), both for GET operations and subscriptions (#2995)
 - Fix: Wrong "max one service-path allowed for subscriptions" in NGSIv2 subscription operation (#2948)
+- Fix: Add support for accessing response headers when making CORS requests (#3030)
```
fgalan: Could this be considered part of "CORS Preflight Requests support for all NGSIv2 resources"? Or it is better to consider it a separate item of work?

kzangeli:
LGTM

test/functionalTest/cases/0501_cors/max_age.test
fgalan: I fact.. would it make sense to to the same for `Access-Control-Allow-Headers`, I mean, to pass from REGEX(.*) to the actual list of headers?

src/lib/serviceRoutinesV2/optionsAllNotDelete.cpp
```diff
@@ -45,8 +46,11 @@ std::string optionsAllNotDelete
   ParseData*                 parseDataP
 )
 {
-  ciP->httpHeader.push_back("Access-Control-Allow-Methods");
-  ciP->httpHeaderValue.push_back("GET, PATCH, POST, PUT, OPTIONS");
+  if ( (ciP->httpHeaders.origin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(ciP->httpHeaders.origin.c_str(), corsOrigin) == 0)) )
+  {
+    ciP->httpHeader.push_back("Access-Control-Allow-Methods");
```
fgalan: Should use ` ACCESS_CONTROL_ALLOW_METHODS`.

(Similar cases along others serviceRoutinesV2/ files in this PR)

src/lib/serviceRoutinesV2/optionsPostOnly.cpp
```diff
@@ -45,8 +46,11 @@ std::string optionsPostOnly
   ParseData*                 parseDataP
 )
 {
-  ciP->httpHeader.push_back("Access-Control-Allow-Methods");
-  ciP->httpHeaderValue.push_back("POST, OPTIONS");
+  if ( (ciP->httpHeaders.origin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(ciP->httpHeaders.origin.c_str(), corsOrigin) == 0)) )
```
fgalan: This check is done many times... Maybe is should be moved to a common function.

test/functionalTest/cases/0501_cors/max_age.test
McMutton: @fgalan, thank you for this comment. This made me realize I have skipped an important step in testing. Let me explain : )

My initial idea was to have a separate test file focusing on each CORS header and avoid failing of a specific test just because another CORS header has a wrong value.

For example, here in `max_age.test` we are focusing on the Access-Control-Max-Age header and this test should not fail in case there is something wrong with `Access-Control-Allow-Headers`. A specific test or a set of tests should be focusing on that header.

The reality is, there are no tests focusing on `Access-Control-Allow-Headers` or `Access-Control-Expose-Headers`. I'll go ahead and add them quickly to the set of CORS tests. Please check if the tests make sense as a whole then and we can discuss if we need any further improvements.

fgalan: Let me check if I'm undersetanding your point before providing feedback

You mean having a .test focusing on each headers, thus:

* In the first .test Access-Control-Max-Age: has particular values but Access-Control-Allow-Headers and Access-Control-Expose-Headers use REGEX(.*)
* In the second .test Access-Control-Allow-Headers: has particular values but Access-Control-Max-Age and Access-Control-Expose-Headers use REGEX(.*)
* In the third .test Access-Control-Expose-Headers: has particular values but Access-Control-Max-Age and Access-Control-Allow-Headers use REGEX(.*)

Is this correct?

McMutton: Yes that's the idea. We also have the Access-Control-Allow-Methods and Access-Control-Allow-Origin headers but they already have their own dedicated test files.

fgalan: The general rule we use in the .test is try to be as specific as possible, that is, to avoid the usage of `REGEX()` (and specially, `REGEX(.*)`) has much as possible.

Having said that, I think that in this particular case it makes sense to do an exception to the general rule and proceed as you suggest. However, I'd suggest to use the following in the preamble of the involved .test in order to be clear:

> This test is focused in header Access-Control-Max-Age. Thus, other CORS related headers use REGEX(.*) as value. However, note that those specific headers are tested with specific headers values in other .test files in this same directory

or something similar

src/lib/rest/HttpHeaders.h
McMutton: Well, these are the headers we put in Context Broker responses that the client could read. In the context of CORS, we need to have these headers listed in Access-Control-Expose-Headers in our responses so that the client can access and read them. 

Since we are not limiting or extending this set for CORS, I would say documenting might not be crucial but won't hurt either : ) 

By the way, if I have missed any headers that we put in the responses and expect it to be client accessible please let me know.

CHANGES_NEXT_RELEASE
McMutton: Well, it is not directly related with Preflight requests but since we are touching those parts in this PR as well, I can change the line below

> - Add: CORS Preflight Requests support for all NGSIv2 resources, -corsMaxAge switch (#501)

to 

- Add: CORS Preflight Requests support for all NGSIv2 resources, -corsMaxAge switch, CORS exposed headers (#501)

src/lib/rest/HttpHeaders.h
fgalan: > Since we are not limiting or extending this set for CORS, I would say documenting might not be crucial but won't hurt either : )

I agree.

Could you write a little cors.md piece of documentation about this (and any other CORS specific aspects to mention) in doc/manuals/user to be included in this PR, please? Don't worty too much about how the material fits in the overall manual, I'll deal with it once the PR gets merged.

CHANGES_NEXT_RELEASE
fgalan: Ok, with a small ammedment to include the number of the issue about exposed headers:

> Add: CORS Preflight Requests support for all NGSIv2 resources, -corsMaxAge switch, CORS exposed headers (#501, #3030)

src/lib/serviceRoutinesV2/optionsPostOnly.cpp
McMutton: Which file do you think should host the function? Is `rest.cpp` a good candidate?

fgalan: `rest.cpp` (along with extern declaration in `rest.h`) sounds good.

test/functionalTest/cases/0501_cors/max_age.test
McMutton: Fixed in 3e6e50a

src/lib/rest/restReply.cpp
McMutton: Fixed in 3e6e50a

src/lib/rest/HttpHeaders.h
McMutton: Fixed in 3e6e50a

CHANGES_NEXT_RELEASE
McMutton: Fixed in 3e6e50a

src/lib/serviceRoutinesV2/optionsAllNotDelete.cpp
McMutton: Fixed in 3e6e50a

src/lib/serviceRoutinesV2/optionsPostOnly.cpp
McMutton: Fixed in 3e6e50a

src/lib/serviceRoutinesV2/optionsAllNotDelete.cpp
```diff
@@ -45,8 +47,11 @@ std::string optionsAllNotDelete
   ParseData*                 parseDataP
 )
 {
-  ciP->httpHeader.push_back("Access-Control-Allow-Methods");
-  ciP->httpHeaderValue.push_back("GET, PATCH, POST, PUT, OPTIONS");
+  if ( isOriginAllowedForCORS(ciP->httpHeaders.origin) )
```
fgalan: I think is should be

`if (isOriginAllowedForCORS(ciP->httpHeaders.origin))`

per [M11 style rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#m11-spaces)

> Also, after `(` and before `)`, spaces MUST NOT be present.

fgalan: (Seen in other part of the PR also)

src/lib/rest/rest.cpp
```diff
@@ -811,6 +811,18 @@ void firstServicePath(const char* servicePath, char* servicePath0, int servicePa
...
 
 
+/* ****************************************************************************
+*
+* isOriginAllowedForCORS - checks the Origin header of the request and returns
+* true if that Origin is allowed to make a CORS request
+*/
+bool isOriginAllowedForCORS(std::string requestOrigin)
+{
+  return ( (requestOrigin != "") && ((strcmp(corsOrigin, "__ALL") == 0) || (strcmp(requestOrigin.c_str(), corsOrigin) == 0)) );
```
fgalan: If think it should change: `return ( ( .. ) );` -> `return ( )` per [M11 style rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#m11-spaces)

> Also, after `(` and before `)`, spaces MUST NOT be present.

src/lib/rest/rest.cpp
```diff
@@ -811,6 +811,18 @@ void firstServicePath(const char* servicePath, char* servicePath0, int servicePa
 
 
 
+/* ****************************************************************************
+*
+* isOriginAllowedForCORS - checks the Origin header of the request and returns
+* true if that Origin is allowed to make a CORS request
+*/
+bool isOriginAllowedForCORS(std::string requestOrigin)
```
fgalan: It should be `const std::string&`, see https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#programming-patterns:

> if the function should not be able to modify the object (i.e. "read only"), then const references should be used, e.g. `const BigFish& bf`

McMutton: Thanks for pointing out these mistakes @fgalan, I will fix them quickly.

fgalan: They aren't big issues, but I think is good to fix them so you get familiar with style aspects ;)

McMutton: Yep, we need mistakes to memorize style guidelines :)

McMutton: Fixed in da358b6

src/lib/rest/rest.cpp
McMutton: Fixed in da358b6

src/lib/serviceRoutinesV2/optionsAllNotDelete.cpp
McMutton: Fixed in da358b6

fgalan:
LGTM. Thanks!

src/lib/rest/rest.cpp
kzangeli: Just to add some clarification here ...
This is a bit more than just 'style guide' (understanding style guide as a "the way I prefer to see the code, my taste").
This is about avoiding to copy an entire object on the stack for every call to the function, and instead just passing the object by reference.

The style guide says something like:

"An entire object should never ever be sent over the stack."
"If the object is allowed to be altered by the function, use a C pointer"
"If not, use a C++ reference"

The reasoning behind this rule is that the ampersand in the call:
`funcCall(&v):`
**shows** that the variable `v` may be altered, while
`funcCall(v)`
does not.

PR #3048, opened 2017-11-21
FIX CORS integration into the main index and RTD

fgalan:
This PR integrates fixes some typos in cors.md and integrates with the RTD framework.

I'm not sure if RTD will keep links to the code but, even in the case it doesn't, the path to the file will be shown, so it is ok. Indepently from that, it is a big "hard" to include links to source code files in user documenation :) but I like how the document use these references, so I think we can leave it as it is now.

In addition, not related with CORS, I have take the opportunity to reractor index in README.md and mkdoc.yml in order to have have topipc in the same order.

@McMutton pls review


doc/manuals/user/cors.md
```diff
@@ -68,12 +71,12 @@ This header should be present in Orion's response for every `OPTIONS` request
 made to `/v2` resources. The user is free to set a value for the maximum time
```
McMutton: > The user is free to set a value for the maximum time (in seconds) a preflight request made to Orion can be cached by the client.

This is a bit confusing. Since you have already created a PR, may I ask you to change this line as well? Please see my suggestion below:

The user is free to set a value for the maximum time (in seconds) a client is allowed to cache a preflight request made to Orion.

fgalan: Fixed in 76ab66be28d95b9192b08e77a1120ffdacf1fee3

doc/manuals/user/cors.md
```diff
@@ -22,7 +25,7 @@ maximum preflight cache time
 
         contextBroker -corsOrigin specificdomain.com
 
-CORS is available for all /v2 resources but for /v1 resources, it is only
```
kzangeli: I see all in **bold** from this point ... A "bold-end" missing somewhere?

fgalan: It could be some "crazy diff effect" in github. The view buttom for that file in the branch (https://github.com/telefonicaid/fiware-orion/blob/hardening/cors_doc/doc/manuals/user/cors.md) shows it is fine.

kzangeli:
LGTM

PR #3057, opened 2017-12-07
FIX accumulator server failes if cert n key are not specified

ddmitrii:
[error description](https://ask.fiware.org/question/878/failed-to-execute-accumulator-server-in-orion/l)


fgalan:
LGTM

Thanks for the contribution!

Issue #3055, opened 2017-12-04
contextBroker shuts down when passing inappropriate parameter in query context payload


saurabhjangir:
```
[root@localhost scripts]# (curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json'  --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
 {
     "entities": [
         {
             "type": "Room",
             "isPattern": "false",
             "id": "Room1"
         }
     ],
     "attributes": [
         "temperature"
     ],
     "value": [
         "700"
     ]
 }
 EOF
**curl: (52) Empty reply from server**
**No JSON object could be decoded**
```

@fgalan : After this response, fiware orion stops. It is due to providing "value" parameter in payload.

fgalan:
Thanks for the report! I have labeles this issue as bug, while we can have a closer look to it.

Hint: it is better to use code style for curl fragments instead of quote style. I have edited your report to show what I mean :)

saurabhjangir:
Thanks @fgalan for your review and editing this report. For next time, I'll keep your suggestions i mind. :)

saurabhjangir:
In my investigation, I found that in this case Orion is trying to write inside a NULL pointer at line number 175 in file "src/lib/parse/compoundValue.cpp". Here _parseDataP->lastContextAttribute_  is NULL.
Hope this might help.

saurabhjangir:
PULL request #3064 is raised as the fix of this issue. 

Earlier we were getting a segmentation fault for the request mentioned in this issue.
REQUEST:
```
(curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
    --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
{
    "entities": [
        {
            "type": "Room",
            "isPattern": "false",
            "id": "Room1"
        }
    ],
    "attributes": [
        "temperature"
	],
	"value": [
		"23"
	]
} 
EOF
```
RESPONSE
```
curl: (52) Empty reply from server
No JSON object could be decoded
```

By making this change now we are getting appropriate error message shown below:

```
{
    "errorCode": {
        "code": "400",
        "details": "JSON Parse Error: unknown field: /value",
        "reasonPhrase": "Bad Request"
    }
}
```

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/3064

PR #3064, opened 2018-01-05
Fix of Issue 3055.

saurabhjangir:
By this change we can direct the code to proper flow and segmentation fault will not occur.

Fixes issue #3055


fgalan:
Thank you very much for the contribution!

However, in order to be ready to merge you have to include at least one functional test covering the fix. Please have a look to https://github.com/apinf/fiware-orion/blob/master/doc/manuals/devel/cookbook.md#adding-a-functional-test-case in the Development Manual regarding adding new functional test. In this case it should be easy: just one step, using the "crash request" you found in issue #3055

In addition, you should check that existing test are not broken by the fix. I mean, you have to do a full pass for unit test, function test and valgrind test.

```
make unit_test
make ft
make valgrind
```

General information on how to contribute can be found at https://github.com/apinf/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md. The Developers Manual is also a good source of information, you can found it at https://github.com/apinf/fiware-orion/tree/master/doc/manuals/devel

Don't hesistate to comment if you have doubts. I'll try to provide more feedback, once the functional test gets included.

saurabhjangir:
@fgalan I have added the required functional test case at appropriate location.
Also verified the existing test cases are working fine with these changes. Please provide your feedback on the same.
Could you please tell me about "crash request" and how it helps me to include test cases ?
Thank you

fgalan:
By "crash request" I refered to the request that caused the crash. I see the new .test uses that one, so it is ok.

Additional comment will be provided using line comments (as you can read at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#pull-request-protocol, line comments are preferred over general comments).

src/lib/jsonParse/jsonParse.cpp
```diff
@@ -378,7 +378,7 @@ static std::string jsonParse
```
fgalan: Please included an entry in CHANGES_NEXT_RELEASE file regarding the fix. For instance:

```
- Fix: bug that may contextBroker to crash when passing inappropriate parameter in query context payload (#3055)
```

The content of that file is eventually used to generate changelog between versions so it is good to track changes/issues fixed or implemented there.

saurabhjangir: @fgalan I've done that. Thanks for your valuable comment.
Please let me know if any further changes are required.

fgalan: According to https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#pull-request-protocol you should end this comment thread with a `"Fixed in <commit_hash>"` comment.

> * After discussion, the comment thread ends in one of the following ways:
>    * `Fixed in <commit hash>` in case the discussion involves a fix in the PR branch (which commit hash is included as reference)
>    *  `NTC`, if finally nothing needs to be done (NTC = Nothing To Change)

(It is a very minor thing but it worth it to get familiarity with the contribution process for future PRs ;)

saurabhjangir: Fixed in 767771823ba4f93291b77da32b72e1fa8764ab0b

src/lib/jsonParse/jsonParse.cpp
```diff
@@ -378,7 +378,7 @@ static std::string jsonParse
 
   boost::property_tree::ptree subtree = (boost::property_tree::ptree) v.second;
   int                         noOfChildren = subtree.size();
-  if ((isCompoundPath(path.c_str()) == true) && (nodeValue == "") && (noOfChildren != 0))
+  if ((isCompoundPath(path.c_str()) == true) && (nodeValue == "") && (noOfChildren != 0) && treated == true)
```
fgalan: According to [S3 code style rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#s3-parenthesis-in-part-conditions) you should use:

```
(treated == true)
```

saurabhjangir: Fixed in 54d6633b408a96d43d2efabd1ae5198d87f75bdc

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix: Bug that may contextBroker to crash when passing inappropriate parameter in query context payload (#3055)
```
fgalan: Typo: may contextBroker -> may cause contextBroker

Sorry, my fault, as author of the original suggested text :$

saurabhjangir: Fixed in 2182d329780232c2d7740582653132c4c5e8b6ec

PR #3065, opened 2018-01-08
Added the ability to start ContextBroker with core generation

vgarciag:
This PR add a config parameter to let the ContextBroker to start with the core generation in the service script (`/etc/init.d`)

This parameter is described and can be modified in the config file `/etc/sysconfig/contextBroker`.

**Tip:** the core generation could be monitorized using a script like:

```
#!/usr/bin/env bash

DIRECTORY_TO_CHECK="/var/cb_cores"
FILE_PATTERN_TO_CHECK="CB_core_*"

LAST_FILE_LIST="${DIRECTORY_TO_CHECK}/.last_file_list.txt"
NEW_FILE_LIST="${DIRECTORY_TO_CHECK}/.new_file_list.txt"
HOSTNAME_FROM="$(hostname)"

ls ${DIRECTORY_TO_CHECK}/${FILE_PATTERN_TO_CHECK} > ${NEW_FILE_LIST} 2> /dev/null

[[ ! -f ${LAST_FILE_LIST} ]] && touch ${LAST_FILE_LIST}

diff -q ${NEW_FILE_LIST} ${LAST_FILE_LIST} &> /dev/null 
if [[ ${?} -eq 1 ]]; then
	mailx  -S smtp={{ smtp_orion_server }} -s "{{ core_mail_subject }}" -r core_checker@${HOSTNAME_FROM} {{ core_mail_recipient_list }} <<< "{{ core_mail_message_body }}"
	rm -f ${LAST_FILE_LIST}
	mv ${NEW_FILE_LIST} ${LAST_FILE_LIST}
else
	echo "Content of ${DIRECTORY_TO_CHECK} has NOT changed"
fi
```
The previous script can be called in a cron and it will send an email when new core appears. Note: `{{ var }}` are Jinja2 template vars an can be replaced by your own messages.


etc/config/contextBroker
```diff
@@ -57,3 +57,10 @@ BROKER_DATABASE_NAME=orion
 # We need to start the CB with log append mode to not overwrite previous logs and logrotate work correctly
 BROKER_EXTRA_OPS="-multiservice -logAppend"
 
+# The next environment variable generates, if it is defined as 'true', an archive in the 
+# path /var/cb_cores with the core file and the logs when the process crash. When this var is 'true' 
+# the process is started with 'root' user. Use this parameter for debug purposes.
```
fgalan: Another important thing to mention is that the process is run in foreground (-fg) but back-grounded by the shell at the same time with `&`.

etc/init.d/contextBroker.centos
```diff
@@ -94,11 +96,56 @@ broker_start()
...
+broker_coredump_start()
+{
+    printf "%-50s" "Starting ${NAME}..."
+    if [[ -x ${EXECUTABLE} ]]; then
+        mkdir -p /var/cb_cores &> /dev/null
+        cd /var/cb_cores
+        rotate_cores &
+        ulimit -c unlimited
+        ulimit -n {{ orion_config.orion_root_max_open_files | mandatory }}
+        ${EXECUTABLE} ${BROKER_OPTS} -fg &> /dev/null &
```
fgalan: Which users starts Orion in coredump case? How do you ensure that user is root (as *"When this var is 'true' 
the process is started with 'root' user"* says).

etc/config/contextBroker
```diff
@@ -57,3 +57,10 @@ BROKER_DATABASE_NAME=orion
 # We need to start the CB with log append mode to not overwrite previous logs and logrotate work correctly
```
fgalan: An entry in the CHANGES_NEXT_RELEASE file should be added regarding this changes. Something like:

```
- Add: new GENERATE_COREDUMP variable at /etc/config/contextBroker to enable core generation and archiving (useful for debugging)
```

etc/config/contextBroker
```diff
@@ -57,3 +57,10 @@ BROKER_DATABASE_NAME=orion
...
 BROKER_EXTRA_OPS="-multiservice -logAppend"
 
+# The next environment variable generates, if it is defined as 'true', an archive in the 
+# path /var/cb_cores with the core file and the logs when the process crash. When this var is 'true' 
+# the process is started with 'root' user. Use this parameter for debug purposes.
+# The archive name has the next format: /var/cb_cores/CB_core_yyyymmdd_HHMMSS.tar.bz2. 
+# IE: /var/cb_cores/CB_core_220180108_151502.tar.bz2.
+# To avoid filling the disk only the las 8 cores and logs are maintained.
+GENERATE_COREDUMP=false
```
fgalan: The new variable should be documented at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/running.md#configuration-file

etc/config/contextBroker
```diff
@@ -57,3 +57,10 @@ BROKER_DATABASE_NAME=orion
 # We need to start the CB with log append mode to not overwrite previous logs and logrotate work correctly
 BROKER_EXTRA_OPS="-multiservice -logAppend"
 
+# The next environment variable generates, if it is defined as 'true', an archive in the 
+# path /var/cb_cores with the core file and the logs when the process crash. When this var is 'true'
```
fgalan: The directory has to exist? Or the core generation and archiving logic is able to create it? In the former case it should be mentioned in this piece of the documentation.

vgarciag: When the variable is set to `true` all the logic, creating, archiving and making the rotation in archives is fully functional...

May be indicate that a `bzip2` utility should be installed because in the `tar` is used the option `j` to compress.

etc/config/contextBroker
vgarciag: Correct, this is a difference between the to ways to start the process: with and without core generation

etc/config/contextBroker
vgarciag: 👍

etc/init.d/contextBroker.centos
```diff
@@ -94,11 +96,56 @@ broker_start()
...
+
+broker_coredump_start()
+{
+    printf "%-50s" "Starting ${NAME}..."
+    if [[ -x ${EXECUTABLE} ]]; then
+        mkdir -p /var/cb_cores &> /dev/null
+        cd /var/cb_cores
+        rotate_cores &
+        ulimit -c unlimited
+        ulimit -n {{ orion_config.orion_root_max_open_files | mandatory }}
```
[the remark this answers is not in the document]
vgarciag: This is a bug extracting the file from Ansible deployment code. I fix it with the value in the ansible config.

etc/init.d/contextBroker.centos
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
-#
+
+q#
```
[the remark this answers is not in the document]
vgarciag: yes!

etc/config/contextBroker
vgarciag: Ops, I always forget the CNR.

etc/init.d/contextBroker.centos
```diff
@@ -69,6 +70,7 @@ fi
 # Mandatory parameters
 BROKER_OPTS="-port ${BROKER_PORT} \
    -logDir ${BROKER_LOG_DIR} \
+   -logLevel ${BROKER_LOG_LEVEL} \
```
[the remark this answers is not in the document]
vgarciag: I add this variable to `/etc/config/contextBroker`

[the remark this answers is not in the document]
fgalan: Should be also taken into account in the CNR file (a new entry about this variable)

etc/init.d/contextBroker.centos
vgarciag: You are right I cannot ensure the `root`  user but probably the script fails in somewhere before the contextBroker starts. So, if the contextBroker starts it will be as root user with very high probability 😄

fgalan: Another option would be to override the BROKER_USER option in the case of coredump and use root to do the execution, i.e.:

```
su root -c "${EXECUTABLE} ${BROKER_OPTS} -fg &> /dev/null &"
```
What I'm not sure if it that is going to work... not sure of su + -fg + & work together...

etc/init.d/contextBroker.centos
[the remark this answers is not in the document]
fgalan: Fixed in 329a63aaa2f5df3dc1b8ee08ddfe908b1aa604fd

etc/config/contextBroker
fgalan: Fixed in 329a63aaa2f5df3dc1b8ee08ddfe908b1aa604fd and 1a7d031e866779bd1ee855b87375126f96b14851

etc/init.d/contextBroker.centos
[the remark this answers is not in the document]
fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/commit/329a63aaa2f5df3dc1b8ee08ddfe908b1aa604fd

etc/init.d/contextBroker.centos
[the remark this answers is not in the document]
fgalan: Fixed in 1a7d031e866779bd1ee855b87375126f96b14851

etc/config/contextBroker
fgalan: Documentation adjusted in 6d0e9e9

etc/config/contextBroker
fgalan: Fixed in 6d0e9e9

etc/config/contextBroker
fgalan: In a second though, this is very advanced configuration as it is related with debugging code. So it is better not include it in the .md (but improved the comment-based documentation in the sysconfig file itself).

NTC

doc/manuals/admin/running.md
```diff
@@ -35,7 +35,8 @@ The configuration used by the contextBroker service is stored in the
 /etc/sysconfig/contextBroker file, which typical content is:
 
 ```
-# BROKER_USER - What user to run orion-broker as
```
fgalan: I have take the opportunity to upgrade this part of the documentation. It was a bit obsolete, based in an old version of the sysconfig file.

NTC (just informative)

etc/init.d/contextBroker.centos
fgalan: After talking with @vgarciag , the best solution is keep the current way but checking BROKER_USER is 'root' previously (otherwise exit with an error message).

etc/init.d/contextBroker.centos
```diff
@@ -107,14 +154,15 @@ broker_start()
...
+
 }
 
 broker_stop()
 {
     printf "%-50s" "Stopping $NAME..."
     if [ -f "${BROKER_PID_FILE}" ]; then
-        kill -HUP $(cat ${BROKER_PID_FILE})
+        kill -HUP $(cat ${BROKER_PID_FILE}) 2> /dev/null
+        # kill $(cat ${BROKER_PID_FILE}) 2> /dev/null
```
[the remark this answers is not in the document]
kzangeli: SIGHUP (hangup) is a signal normally used in conjunction with controlling terminals. As the broker is started as a daemon ... this signal doesn't really fit.

Or, perhaps SIGHUP is used for more things now, not an expert.

SIGINT or SIGTERM are traditionally used for killing processes (in a controlled way).

Looking at the code of the broker, we've just seen (forgotten a long tima ago :-)) that just SIGINT and -TERM are captured and a 'graceful exit' is performed

etc/init.d/contextBroker.centos
```diff
@@ -94,11 +96,56 @@ broker_start()
...
+    if [[ -x ${EXECUTABLE} ]]; then
+        mkdir -p /var/cb_cores &> /dev/null
+        cd /var/cb_cores
+        rotate_cores &
+        ulimit -c unlimited
+        ulimit -n 10240
+        ${EXECUTABLE} ${BROKER_OPTS} -fg &> /dev/null &
         sleep 2 # wait some time to let the contextbroker start
-        PID=$(cat ${BROKER_PID_FILE} 2> /dev/null)
+        ps -ef | grep ${EXECUTABLE} | grep -v grep | awk '{print $2}' > ${BROKER_PID_FILE}
```
kzangeli: This here might give problems ...
If somebody is editing contextBroker.cpp, that line **emacs/vi contextBroker.cpp** would turn up in the 'grep' as well ...
And, overwriting the PID file, do we really want this?  It's the brokers responsibility to put its PID in the pid-file ...

etc/init.d/contextBroker.centos
```diff
@@ -94,11 +96,56 @@ broker_start()
         # logs are always managed by CB and written to file inside directory BROKER_LOG_DIR (see /etc/sysconfig/contextBroker) and rotated by logrotate
         BROKER_STDERR_LOG=${BROKER_LOG_DIR}/${NAME}-error.log
         su $BROKER_USER -c "${EXECUTABLE} ${BROKER_OPTS} 1>/dev/null 2>${BROKER_STDERR_LOG}"
+        sleep 2 # wait some time to leave contextbroker start
```
kzangeli: leave => let

etc/init.d/contextBroker.centos
fgalan: Makes sense to me

fgalan: In fact, in regular start (no core dumps) we are using cat. https://github.com/telefonicaid/fiware-orion/pull/3065/files#diff-705e27a30a796ccc4adca2c31d736a71R100

etc/init.d/contextBroker.centos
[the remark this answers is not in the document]
fgalan: I have done a quick test... Just start contextBroker with `-fg`, then get its PID and do `kill -HUP <pid>`. As result, the CB stops with a final `Colgar (hangup)` in the terminal (Spanish locale).

So, default behaviour for SIGHUP is to end the process? We should clarify this little mistery... :)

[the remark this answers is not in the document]
fgalan: Anyway, I'd suggest the following modifications:

* Use SIGTERM in the /etc/init.d/contextBroker script in the stop logic
* Modify signal handler in CB to include SIGHUP so at least we can exit in an ordered way

etc/init.d/contextBroker.centos
fgalan: Fixed in 48afdc3

etc/init.d/contextBroker.centos
```diff
@@ -42,6 +43,7 @@
 # For those usages not covered by this license please contact with
 # iot_support at tid dot es
 
+[[ $(id -u) -ne 0 ]] && echo "User root must execute this script" && exit 1
```
fgalan: @vgarciag , I have added here the root check we were discussion at skype.

etc/init.d/contextBroker.centos
fgalan: Fixed in 48afdc3740503daca08277b7df9ff9e3d69f6e57

etc/init.d/contextBroker.centos
fgalan: After discussing with @vgarciag , cat-based solution seems to be fine but, to be more robust, this issue should be solved before changing this PR: https://github.com/telefonicaid/fiware-orion/issues/3075

etc/init.d/contextBroker.centos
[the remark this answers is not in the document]
fgalan: Both implemented in acb784d5f26f4e7a251aee389c35d2932bb0140f

kzangeli:
LGTM

fgalan:
Thanks for the LGTM ;) but I think it would be premature to merge... #3075 should be solved first and the PR modified to use cat to get PID from file instead ps + grep.

etc/init.d/contextBroker.centos
```diff
@@ -141,8 +141,7 @@ broker_coredump_start()
         ulimit -c unlimited
         ulimit -n 10240
         ${EXECUTABLE} ${BROKER_OPTS} -fg &> /dev/null &
-        sleep 2 # wait some time to let the contextbroker start
-        ps -ef | grep ${EXECUTABLE} | grep -v grep | awk '{print $2}' > ${BROKER_PID_FILE}
+        sleep 2 # wait some time to let the contextbroker start
```
kzangeli: Why not do a smarter sleep?
A while with a timeout checking the broker port using nc, as we do in other places?

etc/init.d/contextBroker.centos
```diff
@@ -141,8 +141,7 @@ broker_coredump_start()
         ulimit -c unlimited
         ulimit -n 10240
         ${EXECUTABLE} ${BROKER_OPTS} -fg &> /dev/null &
-        sleep 2 # wait some time to let the contextbroker start
-        ps -ef | grep ${EXECUTABLE} | grep -v grep | awk '{print $2}' > ${BROKER_PID_FILE}
+        sleep 2 # wait some time to let the contextbroker start        
         PID=$(cat ${BROKER_PID_FILE})
         if [[ -z "${PID}" ]]; then
             printf "%s" "pidfile not found"
```
kzangeli: This error message is a bit cryptic.
If I understand it correctly, this is what is detected:

**Error starting Context Broker (no pidfile found)**

fgalan: I think this is concatenated with a previous message so what the user gets is:

"Starting context broker... pidfile not found"

which I think is descriptive enough, isn't it?

vgarciag:
LGTM

fgalan:
LGTM

PR #3082, opened 2018-01-19
Changing references distribution from CentOS 6 to CentOS 7

fgalan:
I have check that unit + ft + leak cheks works for:

* CentOS 7 (CI jenkins)
* CentOS 6 (CI jenkins, to keep working the code there for a while)
* Debian 8  (my personal developmeent environment)

Appart from Docker and documentation changes, some adaptation were needed in .test. However, as far as I understand no functional change on them. Please reviewers check.


CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Reference distribution changes from RHEL/CentOS 6 to RHEL/CentOS 7
```
kzangeli: changed?

doc/manuals/admin/build_source.md
```diff
@@ -145,7 +142,7 @@ sudo scons install --prefix=/usr/local --use-sasl-client --ssl  # This puts .h f
 
 ## Others
 
-If you have build orion in a system different from CentOS 6.x, don't hesitate to tell us and contribute to expand this section. Probably the best way if doing a pull request to modify this file with the new information. Thanks!
+If you have build orion in a system different from CentOS 7.x, don't hesitate to tell us and contribute to expand this section. Probably the best way if doing a pull request to modify this file with the new information. Thanks!
```
kzangeli: build => built
the best way if doing => the best way **is to issue** a pull request ...

doc/manuals/admin/install.md
```diff
@@ -12,20 +12,20 @@
...
+The recommended procedure is to install using RPM packages in CentOS 7.x. If you are interested in
 building from sources, check [this document](build_source.md).
 
 ## Requirements
 
 * System resources: see [these recommendations](diagnosis.md##resource-availability)
-* Operating system: CentOS/RedHat. The reference operating system is CentOS 6.3
-  but it should work also in any later CentOS/RedHat 6.x version.
+* Operating system: CentOS/RedHat. The reference operating system is CentOS 7.4.1708
+  but it should work also in any later CentOS/RedHat 7.x version.
```
kzangeli: and possible in earlier too ?
Or will the changes in this PR give problems in CentOS 6 ?

test/functionalTest/cases/0000_large_requests/large_discover_context_availability.test
```diff
@@ -29,8 +29,10 @@ brokerStart CB
 
 --SHELL--
 # Prepare the big file to use for the test
+# I have found that when the test in run in jenkins in CentOS 7, tr outputs a 'broken pipe' message which
```
kzangeli: in >= is  (plenty of times ...)

CHANGES_NEXT_RELEASE
fgalan: I think we tend to use present tense in CNR...

NTC

doc/manuals/admin/install.md
fgalan: It will not give problems with CentOS 6, but I think is cleaner to refer to the current version (and from no on current = CentOS 7) and the future.

NTC

test/functionalTest/cases/0000_large_requests/large_discover_context_availability.test
fgalan: Fixed in 952c2e2

doc/manuals/admin/build_source.md
fgalan: Fixed in 952c2e2

doc/manuals/admin/install.md
kzangeli: Sure, but ...

I think we should mention that the broker still works in CentOS 6. After all, most of our client base probably use that distro. They might get worried ...

Also, we will no longer make sure the broker works in CentOS 6, right?
If so, that we should mention as well

kzangeli: NTC

kzangeli:
LGTM

PR #3086, opened 2018-01-25
FIX recover 119 unit test previously disabled

fgalan:
This is part of #2760.

Orion 1.9.0 (the one introducing the changes in #2760) has been running for a while (more than a month) in production environments without problems, so test can be reeneabled.


src/lib/ngsi9/UpdateContextAvailabilitySubscriptionResponse.cpp
```diff
@@ -81,7 +81,7 @@ std::string UpdateContextAvailabilitySubscriptionResponse::render(void)
   out += startTag();
 
   out += subscriptionId.render(RtUpdateContextAvailabilitySubscriptionResponse, errorCodeRendered || durationRendered);
-  out += duration.render(false);
+  out += duration.render(errorCodeRendered);
```
fgalan: This could be a bugfixing... however, given it have nevered happened (in .test, etc.) and it is related with a very very very minor functionality (updating availability subscriptions... nobody uses that :) I think it doesn't worth the time further investigation.

NTC

kzangeli: Sure, NTC

PR #3087, opened 2018-01-25
FIX forwardingMode = false using Not Implemented error messages

fgalan:
Adjusting legacyForwardingMode error message, as discussed previosuly on another PR.

At the same time, I've taken the opportunity to unhardwire a couple of values from render logic (in fact the hardwire stills, but at mongoBackend level, which is better).


kzangeli:
LGTM

PR #3115, opened 2018-02-20
[ON HOLD] Doc/performance consideration

arigliano:
This pull request adds documentation about orion's performance


arigliano:
Thank you for the comments @fgalan. I will give you  a general answer. First of all, I agree with you on many of the issues you've raised, but I just summarized existing documentation  (I looked for the most updated material). Therefore, even understanding your point I can't do much for the issues such as: the choice of what to test, versions, environments and configurations or some missing information.
If you think the approach should be reviewed, I think you need to open a discussion with the foundation and qa-team members who deal with non-functional testing. For all the other comments I'm trying to fix them quickly.

fgalan:
Thanks for your answer and for the quick fixing. I think your proposal of having an open discussion with the foundation qa-team is a good idea. Let's have this PR on hold in the meanwhile.

arigliano:
@fgalan I tried to make the changes you requested with the information available. Do you think that (despite being based on old versions) they can be useful? Maybe, moving them to a different section of the documentation

fgalan:
We have a directory about load tests at https://github.com/telefonicaid/fiware-orion/tree/master/test/loadTest. Mabye that is a good place. However, before doing anything else I'd suggest to have 
having the discussion with the foundation qa-team we mentioned in previous comment. Until that moment, this PR should be on hold.

jmcanterafonseca:
@arigliano @fgalan we need to get consensus on this and land something

jmcanterafonseca:
after talking to @fgalan we have decided to close this PR as we would like to have a different approach:

+ Instead of copying the content of QA Reports on Orion Performance, we would just need a link to them, not only the one generated by the FIWARE QA team, but others that were generated by the Community

+ It would be good to provide some kind of summary in the form of tables so that readers can easily get an idea of the performance numbers 

+ We need a disclaimer that explicitly says that the numbers correspond to certain versions of Orion.

Issue #3119, opened 2018-03-05
[ISSUE] particular case of queryContext request.


saurabhjangir:
When a queryContext request which including utf-8 "\0" code is sent, a Orion process is downed.
A sample of sent queryContext is below.
```
{
  "entitites": [
    {
      "type": "Street",
      "isPattern": "true",
      "id": "\u0000"
    }
  ],
  "attributes": [
    "temperature"
  ]
}
```
Is this an expected behavior ?

fgalan:
What do you mean by "downed"? Orion crashes?

saurabhjangir:
Yes.
Please find the request and response below. Program is crashing after this request.
```
[root@localhost ~]# (curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
>     --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
> {
>     "entities": [
>         {
>             "type": "Room",
>             "isPattern": "true",
>             "id": "\u0000"
>         }
>     ]
> }
> EOF
curl: (52) Empty reply from server
No JSON object could be decoded
```
In case of isPattern is false in the same request we are getting appropriate error message as below,

```
[root@localhost fiware-orion]# (curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
>     --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
> {
>     "entities": [
>         {
>             "type": "Room",
>             "isPattern": "false",
>             "id": "\0000"
>         }
>     ]
> }
> EOF
{
    "errorCode": {
        "code": "400",
        "details": "JSON Parse Error",
        "reasonPhrase": "Bad Request"
    }
}
```

fgalan:
So, it seems to be a bug... :)

I would be great if you could point out the point of code in which Orion crashes. Set core genreation (`ulimit -c unlimited`), then make it crash to get the core, then use the core with `gbd` and do a backtrace with `bt` will provide that information.

kzangeli:
Seems clear we have a problem :-)

Now, a few doubts, The first payload contains the tag "entitites" ... is this a typo only in the issue?  (entitites ... an extra 't' there).

Second doubt, in the two payloads with id '0, the first one is `"id": "\u0000"` and the second one is without the 'u': `"\0000"`.  Is this how it should be? Typo in issue? Is it really the isPattern that makes one crash and the other one not? Or is the the difference in the value of "id" ?

[...]

PR #3124, opened 2018-03-07
Issue #3119 fix

saurabhjangir:
Fix and test cases added for Issue #3119


new
```diff
@@ -0,0 +1 @@
+Just checking new changes
```
fgalan: Seems you git-add more than expected ;) This files should be removed

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
```diff
@@ -0,0 +1,71 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: 2015 -> 2018 ;)

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
```diff
@@ -0,0 +1,71 @@
...
+HTTP/1.1 200 OK
+Content-Length: 128
+Content-Type: application/json
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+{
+    "errorCode": {
+        "code": "400",
+        "details": "empty entityId:id",
```
fgalan: Details is misleading... in this case what we have is actually something like "null characters are not allowed in isPattern entity id" or a similar wordking.

src/lib/ngsi/EntityId.cpp
```diff
@@ -140,7 +140,7 @@ std::string EntityId::toJson(void) const
 */
 std::string EntityId::check(RequestType requestType)
 {
-  if (id == "")
+  if (id == "" || ((id.find('\0')) != (std::string::npos)))
```
fgalan: The C++ check here is maybe too cryptic an potencially ineficient in a field that is going to be checked very ofter (a large set of operations in the API use entities id). Mabye @kzangeli have some C-based alternative that could be evaluated.

CHANGES_NEXT_RELEASE
```diff
@@ -3,3 +3,4 @@
 - Fix: case-sensitive header duplication (e.g. "Content-Type" and "Content-type") in custom notifications (#2893)
 - Fix: bug in GTE and LTE operations in query filters (q/mq), both for GET operations and subscriptions (#2995)
 - Fix: Wrong "max one service-path allowed for subscriptions" in NGSIv2 subscription operation (#2948)
+- Fix: Bug that may contextBroker to crash when passing inappropriate parameter in query context payload (#3055)
```
fgalan: I think correct issue number is `#3119`.

In addition, given you are going to edit this CNR line, change contextBroker -> ContextBroker, pls.

fgalan:
@saurabhjangir why you have closed this PR? Minor comments appart, it looks nice :)

saurabhjangir:
I am sorry this request is pulled by mistake.
Please track this issue fix at #3125

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
saurabhjangir: Entity ID with Null characters are actually an empty ID.

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
saurabhjangir: ohk

src/lib/ngsi/EntityId.cpp
saurabhjangir: Yes we can do it. I did it for an immediate fix. As we need it ASAP.
I will look for the workaround soon :P

CHANGES_NEXT_RELEASE
saurabhjangir: Done. :)

saurabhjangir: this was actually my last changed CNR ;)

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
fgalan: But you could have something like entity id = `\0afternull`. Strictly speaking that is not empty ID, but it has null... or not?

src/lib/ngsi/EntityId.cpp
kzangeli: Yeah, this is kind of strange ... '\0' is simply 0, there is no difference.
And the check is to make sure there is no ZERO inside the string. There cannot be any zero inside a string in C. The zero is the marker that the string has ended ... Very cryptid indeedd. I understand nothing of this.  The functest tells me it works though ... Might be something weird about std::string but ... beats me.

What happens if you do this...?:

```
std::string s = "abcde";
char*   cP = (char*) s.c_str();

cP[2] = 0;
```

In C ... something similar would set `s` to "ab". Position 2 is now the end of the string ...
Looking forward to hearing an explanation of all this ... :-)

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
kzangeli: As `"\0"` is 0, the string should end there and it is empty ...

src/lib/ngsi/EntityId.cpp
kzangeli: Ah!  Could it have something to do with UTF?  "Funny" charsets ... ?

test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
saurabhjangir: @fgalan Your concern was fixed now. Previously we were getting empty ID error messages. After few changes which I've just commited please find the request and response below:
```
[root@localhost fiware-orion]# (curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
>     --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
> {
>     "entities": [
>         {
>             "type": "Room",
>             "isPattern": "true",
>             "id": "\u0000"
>         }
>     ]
> }
> EOF
{
    "errorCode": {
        "code": "400",
        "details": "invalid regex for entity id pattern",
        "reasonPhrase": "Bad Request"
    }
}
```

saurabhjangir: Please track this at PR #3125

PR #3125, opened 2018-03-07
Fix of Issue #3119

saurabhjangir:
Changes related to Issue #3119


fgalan:
The change on this PR seems to be a subset of #3124...

saurabhjangir:
yeah right, Actually my previous changes also came in the same request. :)

saurabhjangir:
@fgalan Please provide your feedback and review comments on this PR.

src/lib/ngsi/EntityId.cpp
```diff
@@ -158,7 +158,7 @@ std::string EntityId::check(RequestType requestType)
   if (isTrue(isPattern))
   {
     regex_t re;
-    if (regcomp(&re, id.c_str(), REG_EXTENDED) != 0)
+    if ((regcomp(&re, id.c_str(), REG_EXTENDED) != 0) || ((id.find('\0')) != (std::string::npos)))
```
fgalan: The C++ check here is maybe too cryptic an potencially ineficient in a field that is going to be checked very often (a large set of operations in the API use entities id). Mabye @kzangeli have some C-based alternative that could be evaluated.

fgalan: In addition, maybe the check for `\0` should come first in the `||` expression, to avoid calling regcomp() if the check fails.

saurabhjangir: @fgalan I Think find method is using the simplest possible way of finding a substring in a given string and we don't need any separate algorithm for doing the same.
For Optimizing the performance here either we can use **MACRO** instead of a function call, which will avoid stack operations of a function call in program. We will implement a normal loop in that macro which will traverse through the string and returns TRUE if substring is found else FALSE.
Or the other approach could be, We can move the same check at some other place instead of EntityID::check which will reduces the operation on entityId.
Which'll you prefer among both ?

kzangeli: The cryptic part is that it is a search of ZERO in a string ... Strings can't contain zeroes.
What this really does, if I have understood correctly is to find the string `\u0000`, and in a slow way using std::string.

Would this do the trick?   `strstr(id.c_str(), "\u0000") != NULL`.

I have a few questions ...  The example functest is with `"id": "\u0000"`, i.e., an "empty" id.
Would  `"id": "E1\u0000"`also be an error?  
Both the "find implementation" and my strstr will flag an error for this.
Perhaps we should compare only the beginning of the string with `\u0000`?

One more doubt: do we support this kind of UTF chars? (`\uxxxx`)

saurabhjangir: @kzangeli 
\>Would "id": "E1\u0000"also be an error?
Yes, it's an error as well.
\>One more doubt: do we support this kind of UTF chars? (\uxxxx)
To the best of my knowledge, this is valid in JSON. So it needs to be handled in BSON which is used by mongodb. 
\>Would this do the trick? strstr(id.c_str(), "\u0000") != NULL.
No, This won't work. I didn't got into the details yet but after making these changes program is still crashing.

saurabhjangir: BY slightly changing your suggested command, it can be resolved.
```strstr(id.c_str(), "\0") == NULL```
This statement is working fine.

**Comment by kzangeli** [comment]
Still have my doubts about that zero inside a string ...
Need to see it in detail.
Fastest and easiest way for me to do that is to merge this PR and update my fork, so ...

LGTM

File: test/functionalTest/cases/3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+Date: REGEX(.*)
+
+{
+    "errorCode": {
+        "code": "400",
+        "details": "invalid regex for entity id pattern",
+        "reasonPhrase": "Bad Request"
+    }
+}
+
```
fgalan: I have run this .test in my environment I'm I get a different result:

```
01. Request context query with inappropriate value parameter in payload.
========================================================================
HTTP/1.1 200 OK
Content-Length: 105
Fiware-Correlator: 56455064-261d-11e8-8333-000c29173617
Date: Mon, 12 Mar 2018 17:46:47 GMT

{
    "errorCode": {
        "code": "400",
        "details": "invalid regex for entity id pattern",
        "reasonPhrase": "Bad Request"
    }
}
```

Most the same, but Content-Lenght differs.

Could you double-check, please? (it could be something wrong in my setup)

saurabhjangir: @fgalan Done. Updated it to 136.
Got successful result
```
make[1]: Leaving directory `/root/fiware-orion/BUILD_RELEASE'
./test/functionalTest/testHarness.sh
Mon Mar 12 13:27:12 EDT 2018
0001/1: 3119_qc_nullcharacter_payload/3119_qc_nullcharacter_payload.test .............................................  01 seconds
```

fgalan: I'm a bit surprissed...

In my case, it worked with Content-Lenght: 105. In you case it worked with Content-Lengh: 136. I don't know how to explain this fact...

But let's merge and let jenkins CI to check. In the worse case, the PR will be reverted and we could re-examine the case.

fgalan:
I'm afraid that jenkins CI has reported many fails related with this PR:

```
precated_checkings/geoquery_circle_deprecated.test
  o  0000_deprecated_checkings/legacy_geoquery_bad_coords.test
  o  0000_deprecated_checkings/legacy_geoquery_circle_json.test
  o  0000_deprecated_checkings/legacy_geoquery_polygon_json.test
  o  0000_deprecated_checkings/legacy_wgs84.test
  o  0000_ngsi10_geolocation_query/geoquery_bad_coords.test
  o  0000_ngsi10_geolocation_query/geoquery_circle_json.test
  o  0000_ngsi10_geolocation_query/geoquery_polygon_json.test
  o  0000_ngsi10_query/query_sequence.test
  o  0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_append-simpler.test
  o  0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_append.test
  o  0000_ngsi10_subscriptions/subscription_sequence_wildcards_onchange_update.test
  o  0000_ngsi9_subscriptions/subscription_availability_wildcards_sequence.test
  o  0559_new_url_paths/ngsi10_new_url_paths.test
  o  0559_new_url_paths/v1_paths.test
  o  0570_filters_for_entity_type/all_entities_with_emptyType.test
  o  0674_servicepath_in_notifications/service_path_in_onchange_notification.test
  o  0674_servicepath_in_notifications/service_path_in_single_onchange_notification.test
  o  0675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_both_cities_with_isPattern_and_deeper_service_path.test
  o  0675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_exact_city_with_isPattern_and_deeper_service_path.test
  o  0675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_with_isPattern_and_deeper_service_path.test
  o  0675_servicePath_in_ngsi10_subscriptions/simple_ngsi10_subscription_with_recursive_service_path_and_isPattern.test
  o  0690_geoquery_crash/invalid_polygon_geoscope.test
  o  0719_servicepath_for_registrations/forward_queries.test
  o  0719_servicepath_for_registrations/register_and_discovery_with_invalid_service_paths.test
  o  0719_servicepath_for_registrations/register_and_discovery_with_many_servicepaths.test
  o  0719_servicepath_for_registrations/register_and_discovery_with_servicepath.test
  o  0864_filter_string_scope/filter_string_scope.test
  o  0943_overnotification_bug/overnotification_bug.test
  o  1048_subscription_cache/cache_population_at_init.test
  o  1048_subscription_cache/three_subscriptions.test
  o  1103_order_by/order_by_attribute_ngsiv1.test
  o  1156_mq_as_uri_param_for_metadata/mq_as_scope_for_compound_values_in_v1_queries.test
  o  1156_mq_as_uri_param_for_metadata/mq_as_scope_in_v1_queries.test
  o  1308_subscription_cache/cache_reset_from_db.test
  o  1308_subscription_cache/subCacheMixedSubscriptionsWithUpdatesAndDeletes.test
  o  1308_subscription_cache/subCacheThrottlingAndExpiration.test
  o  1308_subscription_cache/subCacheThrottlingAndExpirationAfterRefresh.test
  o  1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdate.test
  o  1308_subscription_cache/subCacheThrottlingAndExpirationAfterSubUpdateWhenSubExpired.test
  o  1308_subscription_cache/sub_expired_and_updated_then_cache_refresh.test
  o  1449_notification_mode_cases/onchange_persisent.test
  o  1449_notification_mode_cases/onchange_simulated.test
  o  1449_notification_mode_cases/onchange_threadpool.test
  o  1677_geo_location_for_api_v2/ngsiv2_location_in_ngsv1.test
  o  1677_geo_location_for_api_v2/ngsiv2_location_in_ngsv1_near.test
  o  2723_entity_type_wrongly_as_pattern/enity_type_wrongly_as_pattern.test
  o  3119_qc_nullcharacter_payload/qc_nullcharacter_payload.test
```

I haven't looked to them in detail, but having a look to the first one I see several steps in which we are getting

```
{
    "errorCode": {
        "code": "400",
        "details": "invalid regex for entity id pattern",
        "reasonPhrase": "Bad Request"
    }
}
```

instead of the response defined in the .test.

Thus, I'm afraid I have to revert this PR and reopen the #3119

fgalan:
Reverted in https://github.com/telefonicaid/fiware-orion/pull/3130

saurabhjangir:
@fgalan Hello, I was testing those changes for some positive scenario and I found that it was failing.
```
[root@localhost fiware-orion]# (curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
>     --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
> {
>     "entities": [
>         {
>             "type": "Room",
>             "isPattern": "true",
>             "id": "Room*"
>         }
>     ]
> }
> EOF
{
    "errorCode": {
        "code": "400",
        "details": "invalid regex for entity id pattern",
        "reasonPhrase": "Bad Request"
    }
}

```
But the expected output is 

```
(curl localhost:1026/v1/queryContext -s -S --header 'Content-Type: application/json' \
    --header 'Accept: application/json' -d @- | python -mjson.tool) <<EOF
{
    "entities": [
        {
            "type": "Room",
            "isPattern": "true",
            "id": "Room*"
        }
    ]
} 
EOF
{
    "contextResponses": [
        {
            "contextElement": {
                "attributes": [
                    {
                        "name": "pressure",
                        "type": "Integer",
                        "value": 720
                    },
                    {
                        "name": "temperature",
                        "type": "Float",
                        "value": 23
                    }
                ],
                "id": "Room1",
                "isPattern": "false",
                "type": "Room"
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}

```

Previous changes in code suggested by me were working fine in all scenario, but as you said I am also looking for some C based alternative.

Also, Would like to know your reason for reverting this PR ?

fgalan:
> Would like to know your reason for reverting this PR ?

Not sure what the question is... but if you mean the resason from reverting the PR, that reason is that the PR broke the regression test (i.e. `make ft`) in the jenkins CI that triggers each time a PR is merged into master.

fgalan:
> Previous changes in code suggested by me were working fine in all scenario, but as you said I am also looking for some C based alternative.

If the alternative based in find() works for all cases, then let's go for it. If later @kzangeli is able to improve keeping all the .test regression, then it can be done in a next PR.

saurabhjangir:
@fgalan Ohk. Lets go for the previous solution to the problem check if it clears regression test. I am reverting it to the previous one.

saurabhjangir:
@fgalan  Do I need to create another PR?

fgalan:
>  Do I need to create another PR?

Yes, that's che cleanest way. However, you can use the same branch to cast the new PR.

PR #3131, opened 2018-03-13
Reverting C based alternative to the problem #3119

saurabhjangir:
C based alternative is failed during regression testing.


saurabhjangir:
@fgalan For my setup, It's working with full regression.

fgalan:
Here we go again with the merge... cross fingers ;)

fgalan:
This time it has gone better, although the following error was reported:

```
-----  Inappropriate parameter in query context payload.  -----
(3119_qc_nullcharacter_payload/3119_qc_nullcharacter_payload.test) output not as expected
VALIDATION ERROR: input line:
   Content-Length: 105
does not match ref line:
   Content-Length: 136
```

No need to revert the PR. A one-liner PR chaning the value to the expected one will fix the problem. However, although that would make CI happy, it could cause problems in your local enviornment if for you is working with 136 (that's weird... how can be different the result at CI and in your local environment? I cannot explain...)

saurabhjangir:
@fgalan For my environment content length is 136 so it is passing on my local system. Yes it could be some local environment issue. Just tell me if I change it to 105 or keep it to 136  ?
Also, Have you checked All ohter regresstion test on your environment, was it successful ?

fgalan:
I ran a test locally and, although I don't remember pretty well the result, I think is worked with 105.

Could you do a small PR to change to 105, pls? Let's see if that makes CI happy again :)

saurabhjangir:
@fgalan I did it PR #3133.  Is everything except this content length issue running successfully?

fgalan:
Everything is fine. CI has returned to ok status. Now I'll close #3119. Thanks for all! :)

saurabhjangir:
@fgalan  My pleassure :) Thanks for your inputs.

PR #3153, opened 2018-04-11
FIX style

fgalan:
Fix style "bugs" introduced during feature/ngsiv2_registration development fixes, excep for some changes that are properly marked with "NTC (informative)" comments.


src/lib/apiTypesV2/Registration.h
```diff
@@ -31,54 +31,15 @@
 #include "apiTypesV2/EntID.h"
 
 
-#if 0
```
fgalan: @kzangeli I understand this could be cleaned out, right?

src/lib/mongoBackend/mongoRegistrationDelete.cpp
```diff
@@ -67,7 +67,7 @@ void mongoRegistrationDelete
     return;
   }
 
-  reqSemTake(__FUNCTION__, "Mongo Delete Registration", SemReadOp, &reqSemTaken);
+  reqSemTake(__FUNCTION__, "Mongo Delete Registration", SemWriteOp, &reqSemTaken);
```
fgalan: This is not style fix, but a little bug. Delete operation must use write sem. @McMutton takes this into account in the future ;)

NTC (informative)

src/lib/apiTypesV2/Registration.h
kzangeli: Yes

fgalan: NTC

Issue #3162, opened 2018-04-24
Non-Primitive Values from a Registered Content Provider are not showing a value


jason-fox:
Orion /version:
```json
{
    "orion": {
        "version": "1.13.0-next",
        "uptime": "0 d, 16 h, 51 m, 32 s",
        "git_hash": "41276b3260b17fa2e20fc2be58767c786df9e984",
        "compile_time": "Tue Apr 17 17:08:42 UTC 2018",
        "compiled_by": "root",
        "compiled_in": "afaf7954cd8f",
        "release_date": "Tue Apr 17 17:08:42 UTC 2018",
        "doc": "https://fiware-orion.readthedocs.org/en/master/"
    }
}
```

I'm using the `/v2/registrations` endpoint to register a content provider with the `legacyForwarding` flag being set. Therefore my Content Provider is offering the `v1/queryContext` endpoint

When I am returning a simple value (Integer, String etc.) such as a temperature the data is added to the context correctly:

```json
{
    "contextResponses": [
        {
            "contextElement": {
                "attributes": [
                    {
                        "name": "temperature",
                        "type": "Number",
                        "value": 27
                    }
                ],
                "id": "urn:ngsi-ld:Store:001",
                "isPattern": "false",
                "type": "Store"
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}
```

However when trying to return an array of strings as shown from a Context Provider.

```json
{
    "contextResponses": [
        {
            "contextElement": {
                "attributes": [
                    {
                        "name": "tweets",
                        "type": "Array",
                        "value": [
                            "String 1",
                            "String 2"
                        ]
                    }
                ],
                "id": "urn:ngsi-ld:Store:002",
                "isPattern": "false",
                "type": "Store"
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}
```

I can see the request being sent in the log and I can retrieve the following entity:

```json
{
    "id": "urn:ngsi-ld:Store:002",
    "type": "Store",
    "address": {
        "type": "PostalAddress",
        "value": "",
        "metadata": {}
    },
    "location": {
        "type": "geo:json",
        "value": "",
        "metadata": {}
    },
    "name": {
        "type": "Text",
        "value": "Checkpoint Markt",
        "metadata": {}
    },
    "tweets": {
        "type": "Array",
        "value": "",
        "metadata": {}
    }
}
```

[...]

PR #3163, opened 2018-04-24
Additional Functional test for querying non-primitive data

jason-fox:
* 01. Start the contextBroker 'CB'
* 02. Start contextProvider 'CP' (contextBroker functioning as contextProvider)
* 03. Register an entity/attribute E1/A1 in broker, with CP as providing application
* 04. Update/APPEND E1/A1(=CP) in CP - this is a non-primitive element e.g. an Array
* 05. Query CP for E1/A1
* 06. Query broker for E1/A1 (will go to CP and it will work, A1=CP is returned)
* 07. Update/APPEND E1/A2(=broker) in broker - this is also a non-primitive element e.g. an Array
* 08. Query broker for E1/A2 (broker's local info is returned (A2=broker))
* 09. Query E1 in CB

This highlights the issue #3162 . It is expected that 06, 08, 09 will fail as all non-primitive data are being returned as blanks at the moment.

### DIFF
--- a/test/functionalTest/cases/3162_context_providers_legacy_non_primitive/query_for_arrays.test
+++ b/test/functionalTest/cases/3162_context_providers_legacy_non_primitive/query_for_arrays.test
@@ -1,0 +1,330 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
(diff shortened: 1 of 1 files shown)


test/functionalTest/cases/3162_context_providers_legacy_non_primitive/query_for_arrays.test
```diff
@@ -0,0 +1,330 @@
...
+brokerStart CP1 0
+
+
+--SHELL--
+echo "01. contextBroker running"
+echo "02. contextProvider running (contextBroker functioning as contextProvider)"
+echo
+echo
+
+echo "03. Register an entity/attribute E1/A1 in broker, with CP as providing application"
```
fgalan: Style: add a line above this one with:

echo "======== (as long as needed to cover the message in L52) ============"

That way is easier to identify the steps in test results.

test/functionalTest/cases/3162_context_providers_legacy_non_primitive/query_for_arrays.test
```diff
@@ -0,0 +1,330 @@
...
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+#
+# Uncached/Cached query
+#
+# 01. Start the contextBroker 'CB'
```
fgalan: Style: the steps here (and corresponding echo in SHELL part) assume that CB are any other needed system (e.g. CPs, accumulator server, etc.) are already started. Thus, 01 and 02 should be removed from the list.

fgalan:
Thank you for the contributition! Definitively, providing a .test associated to a new issue is the way to go :)

To be fully sure, this .test captures the current (supossely, buggy) behaviour? Or the expected behaviour? In the first case, it would be great to know which steps are the ones that are not showing the right behaviour (and the expected one),

fgalan:
> It is expected that 06, 08, 09 will fail as all non-primitive data are being returned as blanks at the moment.

That's solves my doubt (sorry, I haven't spotted it in a first read :)

PR #3173, opened 2018-05-09
#3000 issue - allow creation transient entities

### LINKED ISSUE #3000: Allow creation of transient entities
It will nice to have a functionality that allows to create transient entities. 

At creation time a new attribute `dateExpires` will state the date and time at which the entity will be expired. Expired entities will automatically removed by the system. 

`dateExpires` is an attribute that could be updated but probably not shown during a regular API query call as it happens with `dateCreated` and `dateModified`.

arigliano:
This PR fixes #3000 issue.
It was implemented the functionality that enables to define transient entities, by specifiyng the reserved attribute "dateExpiration".
In detail, has been implemented a set of dedicated methods (in dateExpiration file), for each event of the entity lifecicle (create, update and attribute deletion), which manage the specific expiration attribute, for which a TTL index is created at mongo layer (similar way as location attribute),
The attribute in the entity must be a string containing an ISO8601 date, with explictly defined type (DateTime).

Attribute example:

"dateExpiration":{
      	    "type": "DateTime",
      	    "value": "2018-05-05T14:25:05+00:00"
      }
	  
If the value represents a past date, the entity is considered expired and it will be removed in the next iteration of the Mongo TTL monitor (default 60 secs) 
The dateExpiration attribute will appear both in attrs array and at root entity level, as like as location attribute.
	  
Tests are being written...functionalTest (sh scripts way) or unittest (cpp way)??


arigliano:
Important Note. New tests are to be written and executed, as like as the existing ones. Please consider this PR not complete, but as a discussion starting point about the implementation done.

src/lib/mongoBackend/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ SET (SOURCES
     compoundResponses.cpp
```
fgalan: CNR entry missing. Something like this could be fine:

```
- Add: transient entities functionality (new virtual attribute: dateExpired) (#3000)
```

src/lib/mongoBackend/connectionOperations.cpp
```diff
@@ -575,7 +578,10 @@ bool collectionCreateIndex
 
   try
   {
-    connection->createIndex(col.c_str(), indexes);
+    if (isTTL)
```
fgalan: Style: better use:

```
if (...)
{
   ...
}
```

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -669,6 +671,23 @@ bool mongoLocationCapable(void)
...
+*/
+bool mongoExpirationCapable(void)
+{
+  int mayor;
+  int minor;
+
+  /* TTL (Time To Live) indexes was introduced in MongoDB 2.2 */
+  mongoVersionGet(&mayor, &minor);
+
+  return ((mayor == 2) && (minor >= 2)) || (mayor > 2);
```
fgalan: Not sure about this...

Although TTL index were introduced in MongoDB 2.2 according to this piece of documentation https://docs.mongodb.com/v2.2/tutorial/expire-data, the usage of 0 as value of expireAfterSeconds in order to expire at a specific date is not documented there. However https://docs.mongodb.com/v2.4/tutorial/expire-data uses that 0.

Maybe is more correct to check for 2.4. And tune the comment in the following way:

```
/* TTL (Time To Live) indexes was introduced in MongoDB 2.2, although the expireAfterSeconds: 0 usage is not shown in documentation until 2.4 */
```

I may be wrong... other pieces of documentation could say otherwise ;). In that case, please cite them.

src/lib/mongoBackend/dbConstants.h
```diff
@@ -80,6 +80,7 @@
 #define ENT_LOCATION_ATTRNAME        "attrName"
 #define ENT_LOCATION_COORDS          "coords"
 #define ENT_LAST_CORRELATOR          "lastCorrelator"
+#define ENT_EXPIRATION               "dateExpiration"
```
fgalan: To aling with other similar fields in databe, maybe this could be `expDate`.

Note this will not be the name used at API level (e.g. GET /v2/entities/foo?attrs=dateExpires`) which should use an style more aligned with NGSIv2 (e.g. `dateCreated` and `dateModified`).

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3328,11 +3387,28 @@ static void updateEntity
...
+
+    // This avoids strange behavior like as for the location, as reported in the #1142 issue
+    // In order to enable easy append management of fields (e.g. location, dateExpiration),
+    // it could be better to use a BSONObjBuilder instead the BSON stream macro below.
+    if (replaceDate)
+    {
+      updatedEntity.append("$set", BSON(ENT_ATTRS                   << toSetObj <<
+    	                                ENT_ATTRNAMES               << toPushArr <<
+    	                                ENT_MODIFICATION_DATE       << now <<
+										ENT_EXPIRATION              << currentDateExpiration <<
```
fgalan: Github diff shows overindent here. Mabye you are using tabs instead of spaces? Please check it and use the later.

(Also in other cases in this same PR)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2413,6 +2412,8 @@ static bool updateContextAttributeItem
   bool&                     entityModified,
   std::string*              currentLocAttrName,
   BSONObjBuilder*           geoJson,
+  mongo::Date_t*            dateExpiration,
+  bool*                     replaceDate,
```
fgalan: Maybe better: replaceExpirationDate

(There could be other date attributes, now or in the future, and replaceDate would be ambiguous)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2450,9 +2451,9 @@ static bool updateContextAttributeItem
       ca->found = false;
     }
   }
-
-  /* Check aspects related with location */
-  if (!processLocationAtUpdateAttribute(currentLocAttrName, targetAttr, geoJson, &err, apiVersion, oe))
+  /* Check aspects related with location and date expiration */
+  if (!processLocationAtUpdateAttribute(currentLocAttrName, targetAttr, geoJson, &err, apiVersion, oe)
```
fgalan: Doing this way, a false in `!processDateExpirationAtUpdateAttribute(targetAttr, dateExpiration, replaceDate, &err, oe)` would cause reporting a bug related with location.

Thus, it would be better to add the checking for !processLocationAtUpdateAttribute(currentLocAttrName, targetAttr, geoJson, &err, apiVersion, oe) in a separated if clause.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2520,7 +2524,8 @@ static bool appendContextAttributeItem
 
   /* Check aspects related with location */
   if (!processLocationAtAppendAttribute(currentLocAttrName, targetAttr, actualAppend, geoJson,
-                                        &err, apiVersion, oe))
+                                        &err, apiVersion, oe)
+      || !processDateExpirationAtAppendAttribute(dateExpiration, targetAttr, actualAppend, &err, oe))
```
fgalan: Same as previous comment https://github.com/telefonicaid/fiware-orion/pull/3173/files#r187559469

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: The code in the if clause, as far as i see, should be general in both cases. For this reason I kept a single if clause. In both cases it is printed the corresponding targetAttr name and error details.
On one hand, separating it in 2 clauses, ATM it would mean replicating the same code within both of them.
In the other hand, It would make sense to separate them if error handling will differ in the future for the two cases.

fgalan: Let me have a closer look in order to provide more precise feedback about this.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2593,6 +2599,13 @@ static bool deleteContextAttributeItem
       *currentLocAttrName = "";
     }
 
+    /* Check aspects related to date expiration.
+     * If the target attr is date expiration, delete the relative pointer
+     */
+    if (targetAttr->name == ENT_EXPIRATION){
+      *dateExpiration = 0;
```
fgalan: Not sure of the purpose of this... maybe is a way to signal the caller (processContextAttributeVector() function) that the dateExpiration field has to be removed from database?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -681,13 +700,29 @@ void ensureLocationIndex(const std::string& tenant)
...
+void ensureDateExpirationIndex(const std::string& tenant)
+{
+  /* Ensure index for entity expiration, in the case of using 2.4 */
+  if (mongoExpirationCapable())
+  {
+    std::string index = ENT_EXPIRATION;
+    std::string err;
+
+    collectionCreateIndex(getEntitiesCollectionName(tenant), BSON(index << 1), true, &err);
+    LM_T(LmtMongo, ("ensuring 2dsphere index on %s (tenant %s)", index.c_str(), tenant.c_str()));
```
fgalan: Wrong debug message... 2dsphere -> TTL (or something like that)

src/lib/mongoBackend/connectionOperations.cpp
fgalan: In addition, maybe you could elaborate (in a code comment) why ttl time (different from other indexes) needs the IndexSpec() stuff.

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,200 @@
...
+static bool getDateExpiration
+(
+  const ContextAttribute*  caP,
+  mongo::Date_t*           dateExpiration,
+  std::string*             errDetail
+)
+{
+  if ((caP->type == DATE_TYPE) || (caP->type == DATE_TYPE_ALT))
+  {
+    *dateExpiration = mongo::Date_t(caP->numberValue*1000);
```
fgalan: Why *1000 ?

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,200 @@
...
+  mongo::Date_t*                 dateExpiration,
+  std::string*                   errDetail,
+  OrionError*                    oe
+)
+{
+  for (unsigned ix = 0; ix < caV.size(); ++ix)
+  {
+    const ContextAttribute* caP = caV[ix];
+
+    if (caP->name != ENT_EXPIRATION)
```
fgalan: Check https://github.com/telefonicaid/fiware-orion/pull/3173/files#r187312857 regarding having a different name for the attribute at API level and the field name at DB level.

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+
+
+/* ****************************************************************************
+*
+* processDateExpirationAtEntityCreation -
+*
+* This function process the context attribute vector, searching for an attribute meaning
+* date expiration for the entity. In that case, it fills dateExpiration pointer. If a date expiration attribute is not found, then
+* dateExpiration is an empty pointer.
```
fgalan: In which point of the code dateExpiration is nullified in the case of error? I haven't find it either in this function or in getDateExpiration()... Maybe I'm missing it.

fgalan:
I have finished a first round of review and provide some feedback in comments. I'll try to go deeper in a next iteration (the update logic is tricky and I may have missed some detail ;)

Although there is yet missing functionality from #3000 (e.g. manage  `dateExpires` as virtual attribute in API rendering operations, etc. GET /v2/entities/E1?attrs=dateExpires) I think is a very promissing starting point. Good work @arigliano ! Keep on! :)

With regards to 

> Tests are being written...functionalTest (sh scripts way) or unittest (cpp way)??

I think is better to implement functional tests.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Looking in detail, what is done if the `if ()` detects an error is to build this error:
```
     std::string details = std::string("action: UPDATE") +
                           " - entity: [" + eP->toString() + "]" +
                           " - offending attribute: " + targetAttr->getName() +
" - " + err;
```

and raise a bad input alarm:

```
alarmMgr.badInput(clientIp, err);
```

So, it seems to be a generic processing that could apply to both location and data expiration errors.

Maybe it would be a good idea to improve the `details` error to distinguish between both causes (thus, in that case, we would need probably two if clauses). However, let's leave it as it is now and check once the PR gets closer to finish (i.e. when the .test get ready) to decide what's better.

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Because the value transformed by the upper parsing method (the one that fills numberValue starting  from the ISO8601 string in input), seems to be filled with seconds, when we create the mongo::Date_t object, we want milliseconds in the related constructor.

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,200 @@
...
+	    return false;
+	  }
+	  else
+	  {
+        *replaceDate = true;
+	  }
+    }
+    else
+    {
+      *dateExpiration = 0;
```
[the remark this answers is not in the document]
arigliano: It should be the same, as reported in
https://stackoverflow.com/questions/176989/do-you-use-null-or-0-zero-for-pointers-in-c
In any case, I will check if there are any difference

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: yes, in any case, I'm reviewing all the management of pointers and values, in cases of creation and cancellation, so as to better signal to the caller when a date is not present in a creation request or has been deleted as like in this case.

src/lib/mongoBackend/dateExpiration.cpp
fgalan: Ok. Clear. NTC.

src/lib/mongoBackend/dateExpiration.cpp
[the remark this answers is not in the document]
fgalan: Ok.

However, if you at the end needs to null-ify a point I think the right one would be:

```
dateExpiration = 0;
```

As far as I understand the `*dateExpiration = 0;` doesn't set the point to NULL, but updates the contain of the variable pointed by the pointer to 0.

Having said that, if you need it, use `dateExpiration = NULL` instead of `dateExpiration = 0`. Both are equivalente, but from a point of view of Orion code style, we typically use NULL.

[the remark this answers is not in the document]
arigliano: Sure, this is clear :) my intention actually was to set the value to 0 not the pointer. By setting 0 as dateExpiration value, mongo::Date_t is considered "false". As said previously, I'm reviewing all the logic of setting pointers and values for dateExpiration object in these utilities functions.

src/lib/mongoBackend/CMakeLists.txt
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/connectionOperations.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/MongoGlobal.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/dbConstants.h
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/MongoGlobal.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if (targetAttr->numberValue)
+    {
+	  if (!getDateExpiration(targetAttr, dateExpiration, errDetail))
+	  {
+	    oe->fill(SccBadRequest, *errDetail, "BadRequest");
+	    return false;
+	  }
+	  else
+	  {
+        *replaceDate = true;
```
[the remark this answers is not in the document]
arigliano: Fixed in 5a4703d

src/lib/mongoBackend/dateExpiration.cpp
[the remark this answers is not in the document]
arigliano: Fixed in 5a4703d

arigliano:
I'm implementing the virtual attribute functionality. I have some doubts. In the previous commits, i've implemented the attribute, by maintaining two fields in the Mongo document for the same attribute (similar to the location attribute). That is, when a new/updated entity contains the "dateExpires" attribute, according to the logic of a "normal attribute", this field goes into the "attrs" array. In particular, an attribute with DateTime or ISO8601 type is excepted and existing logic then converts the date string to seconds (numberValue field) representation. In addition, this seconds representation is used by the newly added logic (dateExpiration.cpp functions), in order to create the TTL indexed field, which must be a Mongo Date Object. This Date field (expDate) will be inserted at the root level of the document, as like as creDate and modDate. Now, as regards the virtual attribute behavior, the existing logic, by default, will query all the fields contained in the "attrs" array of the Mongo document, including the "dateExpires" in seconds, even when this attribute is not specified in the "attrs" or "options" URI parameters. The problem is that we would print either the "expDate" at root level or "dateExpires" in the attrs array, but only when expressly requested and with a human readable rendering. A better solution maybe could be to avoid the duplicate attribute, by avoiding, when creating or updating entities, the insertion of the "dateExpires" attribute into the attribute vector that will be stored in the document "attrs" array. This field, instead, could be directly translated to the "expDate" name and stored only at document root level. In this way, the attribute is only printed if I insert a dedicated logic, in order to retrieve also this root level field, if specified in the URI parameters.

fgalan:
Thinking on how to represent de dateExpires attribute, maybe  instead of considering it as a virtual attribute (as the original idea was), e.g. as dateCreated or dateModified, the best approach is use a similar one to location, as you mention.

I mean, in the case of location, we keep the "regular" representation of the location attribute in atrrsName and attrs DB field (as any other attribute) and a parallel location field at document root in the DB. The "regular" attribute in attrs use a string value, while the location field uses a GeoJSON (as the geo-index mandates to implement geo-queries correctl).

In a similar fashion, dataExpires could be a regular attribute in attrsName and attrsField, with a regular value (in this case, a DateTime string) and a parallel field ("expDate" if I'm remembering correctly) with the Data object that MongoDB expects to apply correctly the TTL index.

In this case, no virtual attribute special logic as the one that we have in place for dateModified/dataCreated (i.e. control the attrs= values, to include the virtual attribute upon user request) would be required. The implementation could be simpler and as far as I understand is almost done at the present moment (test pending of course, that typically raise new issues to fix).

Makes sense? :)

arigliano:
Yes, it does :) At this moment, the behavior is actually the one that you described for location case. Then, I think I can move to functional tests.

arigliano:
I'm having some problems with a test that fails. Is there a way to disable all tests except this, in order to isolate it and go back to what causes the error? I'm trying to put prints, but doing 600 tests is impossible to highlight those for that test.

kzangeli:
test/functionalTest/testHarness.sh <name of test file>

arigliano:
I was referring to unit tests (test/unittests folder), in particular the failing test is "mongoDbUpdateFail". As for functionalTests, is there a way to execute only that cpp test ?

fgalan:
Try with `make unit_test TEST_FILTER=mongoContextProvidersQueryRequest.noPatternAttrsAll` or `:` instead of `.` as separator

(I don't have an environment to test it at hand, but I think is that way)

arigliano:
Thanks, I was able to identify the problem with this last command (with . )

arigliano:
I'm facing an error in the functional tests. Every ShellInit fail. 
e.g. erroneous_input.shellInit.stdout
   - .stdout prints: 
     - 2018-06-01T09:51:09.656+0000 SyntaxError: Unexpected token ILLEGAL at (connect) 
   - .stderr prints:
     - exception: connect failed

The former should be caused by the following row in dbInit function:
  -   echo 'db.dropDatabase()' | mongo $host:$port/$db --quiet

In addition, I noticed that in localBrokerStart function, the CB_START_CMD is not set (prints void), because the script doesn't seem to enter in any if or elif case of **if [ "$role" == "CB" ]** block

Could be some problem with mongo init?

Any suggestion? I'm using a docker container created with the dockerfile provided in the project (with centos7.4.1708). I'm using the updated master branch, not the one belonging to this PR.

arigliano:
I reinstalled mongoDB 3.4. Now the stdout prints:
 `{ "ok" : 1 }`
`Unable to start contextBroker `

fgalan:
You mean runing the test suite (make ft or directly testHarness.sh) in one docker container and Orion and/or MongoDB in other?

Not sure if that setup is going to work... Test suite is designed to run CB, Mongo and make ft in the same machine. I think "localhot" is hardwired in several places in the test engine framework.

Anayway if there is a problem accessing MongoDB, the CB log (typically in /tmp/contextBroker.log in the case of ft) should provide some clue.

fgalan:
(Btw, PR https://github.com/telefonicaid/fiware-orion/pull/3192 is on the way, configuring Travis to pass the unit and functional tests automatically in the branch. It would take some time to merge it into master, but maybe it could be useful for you to check how a docker infraestructure for testing is set up)

arigliano:
Ok, i will take a look .  I run mongod locally to the Orion container, and it's accessible on localhost:27017 through mongo client. However, I've already used this docker setup for testing (in the previous issues that I fixed), but was from centos 6.6.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -724,7 +725,6 @@ static bool updateAttribute
     {
       attrType = caP->type;
     }
-
```
kzangeli: I would not remove these empty lines.
They help readability

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2593,6 +2599,13 @@ static bool deleteContextAttributeItem
       *currentLocAttrName = "";
     }
 
+    /* Check aspects related to date expiration.
+     * If the target attr is date expiration, nullifying dateExpiration ACTUAL value is the way
+     * of specifying that date expiration field is no longer used */
+    if (targetAttr->name == DATE_EXPIRES) {
```
kzangeli: style guide police:
`{` goes in a separate line

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2856,6 +2876,15 @@ static bool createEntity
     return false;
   }
 
+  /* Search for a potential date expiration attribute */
+  mongo::Date_t dateExpiration;
+
+  if (!processDateExpirationAtEntityCreation(attrsV, &dateExpiration, errDetail, oe))
```
kzangeli: If `oe` can be changed by `processDateExpirationAtEntityCreation`, it should be sent as a "C" pointer. That's the convention we use in the broker.
Read only objects are sent as C++ references (`const XXX& obj`) and objects that are to be modified are sent as C pointers.
This way, you see in the call that the called function may alter the object sent to it.

Read all about it in the style guide (contribution guide):
https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3180,7 +3215,6 @@ static void updateEntity
   std::string     locAttr = "";
   BSONObj         currentGeoJson;
   BSONObjBuilder  geoJson;
-
```
kzangeli: Again, the empty lines (well, most of them anyway), are there for a reason.

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -669,6 +671,24 @@ bool mongoLocationCapable(void)
...
+bool mongoExpirationCapable(void)
+{
+  int mayor;
+  int minor;
+
+  /* TTL (Time To Live) indexes was introduced in MongoDB 2.2,
+   *  although the expireAfterSeconds: 0 usage is not shown in documentation until 2.4 */
+  mongoVersionGet(&mayor, &minor);
+
+  return ((mayor == 2) && (minor >= 4)) || (mayor > 2);
```
kzangeli: `mayor >= 2`, perhaps ?

src/lib/mongoBackend/connectionOperations.cpp
```diff
@@ -575,10 +578,22 @@ bool collectionCreateIndex
...
+    {
+      connection->createIndex(col.c_str(), IndexSpec().addKeys(indexes).expireAfterSeconds(0));
+    }
+    else
+    {
+      connection->createIndex(col.c_str(), indexes);
+      releaseMongoConnection(connection);
+      TIME_STAT_MONGO_COMMAND_WAIT_STOP();
+      LM_I(("Database Operation Successful (createIndex: %s)", indexes.toString().c_str()));
+    }
```
kzangeli: Hmmm ...

Should probably be like this:
```
else
￼     {
￼       connection->createIndex(col.c_str(), indexes);
￼     }

￼     TIME_STAT_MONGO_COMMAND_WAIT_STOP();
￼     LM_I(("Database Operation Successful (createIndex: %s)", indexes.toString().c_str()));
￼ ```

File: src/lib/mongoBackend/dateExpiration.cpp [hunk 22]
```diff
@@ -0,0 +1,198 @@
...
+#include <vector>
+
+#include "common/string.h"
+#include "common/globals.h"
+#include "logMsg/logMsg.h"
+#include "ngsi/ContextAttribute.h"
+#include "parse/CompoundValueNode.h"
+#include "rest/OrionError.h"
+
+// FIXME P5: the following could be not necessary if we optimize the valueBson() thing. See
```
kzangeli: may not be necessary

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,198 @@
...
+  for (unsigned ix = 0; ix < caV.size(); ++ix)
+  {
+    const ContextAttribute* caP = caV[ix];
+
+    if (caP->name != DATE_EXPIRES)
+    {
+      continue;
+    }
+
+    if (!getDateExpiration(caP, dateExpiration, errDetail))
```
kzangeli: Just a little bit difficult to read ...
in one line it would be a lot easier (IMO):
```
if ((caP->name == DATE_EXPIRES) && !getDateExpiration(caP, dateExpiration, errDetail))
```
￼Also, once we have the attribute, and `getDateExpiration` worked, no real meaning to continue the loop, right?
=>
```
  if (caP->name == DATE_EXPIRES)
  ￼{
    if (!getDateExpiration(caP, dateExpiration, errDetail))
    {
      oe->fill(SccBadRequest, *errDetail, "BadRequest");
      return false;
￼    }

    return true;
  }

File: src/lib/mongoBackend/dateExpiration.cpp [hunk 24]
```diff
@@ -0,0 +1,198 @@
...
+)
+{
+  /*
+   * If the name of the target attribute is the date expiration,
+   * check for a number value that will be used in the mongo::Date_t constructor.
+   * If it is an empty value, is interpreted as a date expiration deletion and the ACTUAL date value
+   * is set to 0, in order to signal the caller function.
+   * If valid value, also replaceDate boolean is set to true, in order to manage the new date value
+   * in case the update is a replace operation.
+   */
```
kzangeli: This comment to the function header. that what it's for :)

test/functionalTest/cases/3000_allow_creation_transient_entities/allow_creation_transient_entities.test
```diff
@@ -0,0 +1,29 @@
...
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+Create a new transient entity, by specifying dateExpires attribute
+
+--SHELL-INIT--
+dbInit CB
+dbInit CP1
+brokerStart CB
```
kzangeli: Good start! :-D

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: `oe` is a pointer to `OrionError` (line 2848) so I  think is ok.

src/lib/mongoBackend/MongoGlobal.cpp
fgalan: I think the `= 2` case for mayor is already covered in the left part of the `||`

src/lib/mongoBackend/connectionOperations.cpp
fgalan: Good point! In fact, current implementation in the PR will leak connections, as releaseMongoConnection() is never called in the case of TTL.

test/functionalTest/cases/3000_allow_creation_transient_entities/allow_creation_transient_entities.test
```diff
@@ -0,0 +1,29 @@
...
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+Create a new transient entity, by specifying dateExpires attribute
+
+--SHELL-INIT--
+dbInit CB
+dbInit CP1
```
fgalan: Except if you use Context Providers (which I think is not the case for this functionality) CP1 initialization could be removed.

src/lib/mongoBackend/connectionOperations.cpp
arigliano: Right :)

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: Of course. I was too fast :)

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Didn't see that.  So, `oe` should be renamed to `oeP` and it will be clear.
Imagine we have something about this in the style guide

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,198 @@
...
+  mongo::Date_t*                 dateExpiration,
+  const ContextAttribute*        targetAttr,
+  bool                           actualAppend,
+  std::string*                   errDetail,
+  OrionError*                    oe
+)
+{
+  if (targetAttr->numberValue)
+  {
+  if (!getDateExpiration(targetAttr, dateExpiration, errDetail))
```
kzangeli: indentation of 2 spaces seems to be missing

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Ok, I noticed that we have the same situation in removeEntity, updateContextAttributeItem, appendContextAttributeItem, deleteContextAttributeItem, processContextAttributeVector, processContextAttributeVector. So I guess they should be fixed (maybe in a dedicated issue?)

fgalan: I think that by the moment use the rigth variable name in the part touched by this PR is enough. However, if you want to create an issue about the missing ones for the future, please go ahead :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in 0ee2585

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/connectionOperations.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in abdaa28

test/functionalTest/cases/3000_allow_creation_transient_entities/allow_creation_transient_entities.test
arigliano: Fixed in 2bd4701

test/functionalTest/cases/3000_allow_creation_transient_entities/allow_creation_transient_entities.test
arigliano: Fixed in 2bd4701

arigliano:
Both unit and functional tests are successfully performed.
In 2bd4701 are implemented the new functional tests for this PR.

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in abdaa28

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC by the moment (I'll re-review on the new version of the code, rasing a comment if needed)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC by the moment (I'll re-review on the new version of the code, rasing a comment if needed)

CHANGES_NEXT_RELEASE
```diff
@@ -2,3 +2,4 @@
 - Fix: bug having entities without attibutes AND matching registrations made the attr-less entity not be returned by GET /v2/entities (#3176)
 - Fix: bug when using "legacy" format for custom v2 notifications (#3154)
 - Fix: check for providers is done only in UPDATE or REPLACE actionType case in update entity logic (#2874)
+- Add: transient entities functionality (new virtual attribute: dateExpired) (#3000)
```
fgalan: Suggested change:

```
- Add: transient entities functionality (based on "dateExpired" attribute) (#3000)
```

The different may be subtle... but I understand it is more precise, as at the end dateExpired is similar to location, but not actually virtual attribute (in the sense e.g. dateModified or dateCreated are).

CHANGES_NEXT_RELEASE
```diff
@@ -2,3 +2,4 @@
 - Fix: bug having entities without attibutes AND matching registrations made the attr-less entity not be returned by GET /v2/entities (#3176)
```
fgalan: PR shows a renaming:

scripts/managedb/upgrade-0.14.1/swap_coords.js → scripts/managedb/swap_coords.js

which I understand is incorrect and should be rolled back

(I'm adding this comment in another file, as github doesn't allow me to provide a comment on the renamed file itself)

CHANGES_NEXT_RELEASE
arigliano: Right, that description was there before we decided not to use dateExpires as a virtual attribute.

src/lib/mongoBackend/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ SET (SOURCES
...
     location.cpp
     compoundValueBson.cpp
+    dateExpiration.cpp
```
fgalan: This new module in mongoBackend should be added to https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/mongoBackend.md#specific-purpose-modules. Maybe copy-pasting the same as `location` will be enougth, as the module seems to be quite similar.

In addition, it would be great if you could have a general look to the flows in https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/mongoBackend.md in order to see if some one would need adaptation. I don't think so, as TTL management index seems to be a secondary aspect and the flows doen't typically go into so deeper detail. But please check anyway  (hint: look for usages of the location.cpp/h module in the flows).

CHANGES_NEXT_RELEASE
arigliano: Perhaps caused by update with master, I will check it.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3189,6 +3225,19 @@ static void updateEntity
     currentGeoJson = getObjectFieldF(loc, ENT_LOCATION_COORDS);
   }
 
+  /* Is the entity using date expiration? In that case, we fill the currentdateExpiration attribute with that information.
+   * In any case, if the request contains a new date expiration, this will become the current one
+   * The replaceDate boolean is used in case of replace operation,
```
fgalan: "replaceDate boolean" -> "replaceDateExpiration boolean"

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2593,6 +2599,14 @@ static bool deleteContextAttributeItem
       *currentLocAttrName = "";
     }
 
+    /* Check aspects related to date expiration.
+     * If the target attr is date expiration, nullifying dateExpiration ACTUAL value is the way
+     * of specifying that date expiration field is no longer used */
+    if (targetAttr->name == DATE_EXPIRES)
+    {
+      *dateExpiration = 0;
```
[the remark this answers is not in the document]
fgalan: I have realized that is

```
 *dateExpiration = 0;
```

and not

```
 dateExpiration = NULL;
```

So 0 makes sense in this case. However, it is a bit confusing (as proved by these comments :) so I'd suggest using something like:

```
 *dateExpiration = NO_EXPIRATION_DATE;
```

using `#define NO_EXPIRATION_DATE 0` in the right place (maybe dateExpiration.h)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2946,12 +2976,18 @@ static bool createEntity
                                           ENT_LOCATION_COORDS   << geoJson.obj()));
   }
 
+  /* Add date expiration in the case it was found */
+  if (dateExpiration)
```
fgalan: In line with a previous comment, this could be change to

```
if (dateExpiration != NO_EXPIRATION_DATE)
```

which I think would be clearer.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3310,6 +3361,17 @@ static void updateEntity
     toUnset.append(ENT_LOCATION, 1);
   }
 
+  // We don't touch toSet in the replace case, due to
+  // the way in which BSON is composed in that case (see below)
+  if (currentDateExpiration && (strcasecmp(action.c_str(), "replace") != 0))
```
fgalan: actionTypes changed recently. We are now using enumeration instead of strings. This would need adaptation after merging with master (have a look to https://github.com/telefonicaid/fiware-orion/pull/3197, the modification is quite straighforward).

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2835,27 +2855,37 @@ static bool createEntity
...
 
-  if (!processLocationAtEntityCreation(attrsV, &locAttr, &geoJson, errDetail, apiVersion, oe))
+  if (!processLocationAtEntityCreation(attrsV, &locAttr, &geoJson, errDetail, apiVersion, oeP))
   {
     // oe->fill() already managed by processLocationAtEntityCreation()
     return false;
   }
 
+  /* Search for a potential date expiration attribute */
+  mongo::Date_t dateExpiration;
```
fgalan: I understand it would be safer to use:

```
mongo::Date_t dateExpiration = NO_EXPIRATION_DATE
```

(related with a previous comment about adding a NO_EXPIRATION_DATE constant)

src/lib/mongoBackend/MongoCommonUpdate.cpp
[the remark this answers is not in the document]
fgalan: By the way... it is a bit suprising that `0` gets accepted to set a variable of type `mongo::Date_t`. I understand that the implicit casting is solving the case, dangerous when the casting is from integer to a full fledged object. C/C++ is scaring some times :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Maybe the default constructor is using the same as NO_EXPIRATION_DATE by default... but I think it is better to set it explictely in our code.

fgalan: (These comment would apply not only to this place, but also in other places where `mongo::Date_t` objects are initalized)

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,198 @@
...
+  const ContextAttribute*  targetAttr,
+  mongo::Date_t*           dateExpiration,
+  bool*                    replaceDate,
+  std::string*             errDetail,
+  OrionError*              oe
+)
+{
+  if (targetAttr->name == DATE_EXPIRES)
+  {
+    if (targetAttr->numberValue)
```
fgalan: Could you elaborate on the purpose on this checking please? 

Not sure if this is the indended behaviour... maybe you should check if attribute type is correct (i.e. `tagetAttrs->type`)?

test/functionalTest/cases/3000_allow_creation_transient_entities/create_transient_entity.test
```diff
@@ -0,0 +1,115 @@
...
+
+#
+# 01. POST /v2/entities with dateExpires attribute (expired ISO8601 date value)
+# 02. GET /v2/entities the entity is present (expired but not removed yet)
+# 03. GET /v2/entities after 60 seconds (default removal interval of mongo TTL thread)
+# and check if the entity was removed.
+#
+
+echo "01. POST /v2/entities with dateExpires attribute (expired ISO8601 date value)"
+echo "====================================================================="
```
fgalan: Style: please align the number of "=" in the second echo with the length of the string in the first echo.

fgalan: (General comment in all .test, it seems it is happening in other places)

fgalan:
@arigliano I have done a new review today. Good work! The PR is progressing and the test cases are sound.

I have provided a new round of comments, which I think are more or less minor and straighforward.

Note that you would need to update with master as the PR is currently showing conflicts.

src/lib/mongoBackend/dateExpiration.cpp
arigliano: This checks if the target attribute has name DATE_EXPIRES and also has a numberValue different from 0. In this case, it calls getDateExpiration function, which checks if the type is a Date (```if ((caP->type == DATE_TYPE) || (caP->type == DATE_TYPE_ALT```). On the other hand, if the target attribute is DATE_EXPIRES  and numberValue is 0, it means that the dateExpiration has to be nullified, and thus we avoid calling getDateExpiration.

CHANGES_NEXT_RELEASE
arigliano: Fixed in ae70f62

CHANGES_NEXT_RELEASE
arigliano: Fixed in 77002fc

src/lib/mongoBackend/CMakeLists.txt
arigliano: Fixed in ae70f62

src/lib/mongoBackend/MongoCommonUpdate.cpp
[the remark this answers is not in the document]
arigliano: Fixed in faea551

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in faea551

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in faea551

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in faea551

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in faea551

test/functionalTest/cases/3000_allow_creation_transient_entities/create_transient_entity.test
arigliano: Fixed in faea551

test/functionalTest/cases/3000_allow_creation_transient_entities/create_transient_entity.test
```diff
@@ -0,0 +1,115 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: It would be great to include a .test checking error, i.e. trying to create an entity with invalidad dateExpires and check the requests fail with 400 Bad Request.

src/lib/mongoBackend/dateExpiration.cpp
fgalan: Thanks for the explanation. Maybe is a good idea to use

```
if (targetAttr->numberValue = NO_EXPIRATION_DATE
```

to make it more explicit.

test/functionalTest/cases/3000_allow_creation_transient_entities/create_transient_entity.test
fgalan: The .test covering error could also include the case of create a dateExpires atribute with a type different from DateTime (e.g. a number, a string, etc.) which I understand also result in 400 Bad Request.

arigliano: Fixed in 8ef8517

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in 8ef8517

test/functionalTest/cases/3000_allow_creation_transient_entities/create_transient_entity.test
arigliano: Second test fixed in fe0bab4

doc/manuals/devel/mongoBackend.md
```diff
@@ -186,7 +186,7 @@ Regarding the strategy used in `processContextAttributeVector()` to implement en
 * `toPush`: attributes that need to be added to the entity `attrsName` field in the database (list of attribute names), using the [`$addToSet`](https://docs.mongodb.com/manual/reference/operator/update/addToSet) and [`$each`](https://docs.mongodb.com/manual/reference/operator/update/each) operators.
 * `toPull`: attributes that need to be removed from the `attrsName` field in the database (list of attribute names), using the [`$pullAll` operator](https://docs.mongodb.com/manual/reference/operator/update/pullAll).
 * `locAttr` and `geoJson` are related to modifications in the geolocation information associated to the entity (entity `location` field in the database).
-
+* `dateExpiration` and `replaceDateExpiration` are related to modifications in the TTL expiration date information associated to a transient entity (entity `expDate` field in the database).
```
fgalan: Good addition to the devel manual, but a blank line is  needed just after this one. Otherwise, the text "The update is base [...]" appears in the same paragraph.

Just compare the render on [master](https://github.com/arigliano/fiware-orion/blob/master/doc/manuals/devel/mongoBackend.md#mongoupdatecontext-sr-and-mongonotifycontext-sr) and [your branch](https://github.com/arigliano/fiware-orion/blob/fe0bab47d90d0f9b7f84098348a93b8d1fe2bc3b/doc/manuals/devel/mongoBackend.md) to see what I mean.

fgalan:
New review round. Only 3 new comments this time, so probably this will be the last review on my side :)

In addition, it would be great if you could copy-paste the result of make test and make ft in the PR (only the last few lines) in order to check everything is correct. Thanks!

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3311,6 +3362,17 @@ static void updateEntity
     toUnset.append(ENT_LOCATION, 1);
   }
 
+  // We don't touch toSet in the replace case, due to
+  // the way in which BSON is composed in that case (see below)
+  if ((currentDateExpiration != NO_EXPIRATION_DATE) && (action != ActionTypeReplace))
```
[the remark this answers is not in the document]
arigliano: The effect should not be the same. In this way, if  the action is ActionTypeReplace and dateExpires attribute is not in the payload (then replaceDateExpiration is false) , it will not append this attribute to those to be unset, as the replace should do.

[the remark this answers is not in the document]
fgalan: Just to check if I'm understand correctly...

Do you mean that with my proposal if action type is replaced and dateExpires is not in the payload then unset ENT_EXPIRATION is not added to the unset array?

(Btw, reading your comment maybe a better name for replaceDateExpiration around the code is dateExpirationInPayload. Just a suggestion... replaceDateExpiration is also fine)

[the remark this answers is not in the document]
arigliano: Exactly, I mean this. Anyway, dateExpirationInPayload sounds better than replaceDateExpiration :)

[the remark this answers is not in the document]
fgalan: What about using

```
```
if (action != ActionTypeReplace) 
{
  if (currentDateExpiration != NO_EXPIRATION_DATE)
  {
  ..
  }
  else if (!replaceDateExpiration)
  {
  ...
  }
}
```

and, at the same time to use within the `if (action == ActionTypeReplace)` below:

```
    if (replaceDateExpiration)
    {
         ...
    else
    {
      updatedEntity.append("$set", BSON(ENT_ATTRS                   << toSetObj <<
                                        ENT_ATTRNAMES               << toPushArr <<
                                        ENT_MODIFICATION_DATE       << now <<
                                        ENT_LAST_CORRELATOR         << fiwareCorrelator));
      updatedEntity.append("$unset", BSON(ENT_EXPIRATION <<1));
    }

```

?

I mean, simplify the first part and be explicit with $unset in the second part.

(Again, just a suggestion :) Unfortunatelly, the updateEntity() is tricky and none of the solutions has a clear advantage in simplification terms over the other. Please, choose the one you prefer either keeping the existing one in the PR or what I propose, if it correct).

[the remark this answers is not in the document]
Comment (arigliano) [reply src/lib/mongoBackend/MongoCommonUpdate.cpp]: By using
```updatedEntity.append("$unset", BSON(ENT_EXPIRATION <<1));```
We lose other values appended in "toUnset" previously, such as:
```toUnset.append(ENT_LOCATION, 1);```

I would keep everything as is :)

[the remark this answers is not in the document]
Comment (fgalan) [reply src/lib/mongoBackend/MongoCommonUpdate.cpp]: Ok :)

Thus, only dateExpirationInPayload  is pending with regard this comment thread, isn't it?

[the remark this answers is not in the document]
Comment (arigliano) [reply src/lib/mongoBackend/MongoCommonUpdate.cpp]: Yes, already fixed. Once testing is finished, I will push it.

File: src/lib/mongoBackend/MongoCommonUpdate.cpp [hunk 41]
```diff
@@ -3329,11 +3391,28 @@ static void updateEntity
   if (action == ActionTypeReplace)
```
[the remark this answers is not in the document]
arigliano: Fixed in 9b3853d

doc/manuals/devel/mongoBackend.md
arigliano: Fixed in 9b3853d

arigliano:
[       OK ] rest.servicePathSplit (2 ms)
[----------] 2 tests from rest (2 ms total)

[----------] Global test environment tear-down
[==========] 862 tests from 131 test cases ran. (28898 ms total)
[  PASSED  ] 862 tests.

  YOU HAVE 26 DISABLED TESTS
  
  
  
Total test time: 3224.61 seconds

WARNING: 7 test cases disabled:
  o  2: 0000_bad_requests/exit.test
  o  27: 0000_ipv6_support/ipv4_only.test
  o  28: 0000_ipv6_support/ipv6_only.test
  o  406: 0917_queryContext_behaves_differently/query_with_and_without_forwarding.test
  o  575: 1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
  o  631: 1310_suspect_200OK/suspect_200OK.test
  o  851: 2207_not_spurious_decimals_in_custom_notifications/not_spurious_decimals_in_custom_notifications.test

src/lib/mongoBackend/MongoCommonUpdate.cpp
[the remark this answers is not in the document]
fgalan: Fixed in 9b3853d

doc/manuals/devel/mongoBackend.md
```diff
@@ -186,6 +186,7 @@ Regarding the strategy used in `processContextAttributeVector()` to implement en
 * `toPush`: attributes that need to be added to the entity `attrsName` field in the database (list of attribute names), using the [`$addToSet`](https://docs.mongodb.com/manual/reference/operator/update/addToSet) and [`$each`](https://docs.mongodb.com/manual/reference/operator/update/each) operators.
 * `toPull`: attributes that need to be removed from the `attrsName` field in the database (list of attribute names), using the [`$pullAll` operator](https://docs.mongodb.com/manual/reference/operator/update/pullAll).
 * `locAttr` and `geoJson` are related to modifications in the geolocation information associated to the entity (entity `location` field in the database).
+* `dateExpiration` and `replaceDateExpiration` are related to modifications in the TTL expiration date information associated to a transient entity (entity `expDate` field in the database).
```
fgalan: Should be "dateExpirationInPayload" (maybe the change in the name only touched code...)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3190,6 +3226,19 @@ static void updateEntity
     currentGeoJson = getObjectFieldF(loc, ENT_LOCATION_COORDS);
   }
 
+  /* Is the entity using date expiration? In that case, we fill the currentdateExpiration attribute with that information.
+   * In any case, if the request contains a new date expiration, this will become the current one
+   * The replaceDateExpiration boolean is used in case of replace operation,
```
fgalan: Should be "dateExpirationInPayload" (maybe the change in the name only touched code...)

src/lib/mongoBackend/dateExpiration.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+/* ****************************************************************************
+*
+* processDateExpirationAtUpdateAttribute -
+*
+* If the name of the target attribute is the date expiration,
+* check for a number value that will be used in the mongo::Date_t constructor.
+* If it is an empty value, is interpreted as a date expiration deletion and the ACTUAL date value
+* is set to 0, in order to signal the caller function.
+* If valid value, also replaceDate boolean is set to true, in order to manage the new date value
```
fgalan: Should be "dateExpirationInPayload" (maybe the change in the name only touched code...)

src/lib/mongoBackend/dateExpiration.h
```diff
@@ -0,0 +1,90 @@
...
+
+/* ****************************************************************************
+*
+* processDateExpirationAtUpdateAttribute -
+*/
+extern bool processDateExpirationAtUpdateAttribute
+(
+  const ContextAttribute*  targetAttr,
+  mongo::Date_t*           dateExpiration,
+  bool*                    replaceDate,
```
fgalan: Should be "dateExpirationInPayload" ?

arigliano: No, this is the actual date, not the boolean as in the processDateExpirationAtUpdateAttribute function.

fgalan: NTC

src/lib/mongoBackend/dateExpiration.cpp
arigliano: Fixed in 96cd352

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in 96cd352

doc/manuals/devel/mongoBackend.md
arigliano: Fixed in 96cd352

PR #3180, opened 2018-05-25
check_entities_consistency script

fgalan:


scripts/managedb/check_db_sanity.py
```diff
@@ -1,96 +0,0 @@
...
-            names[attr['name']] = 1
-    return at_least_one_dup
-
-def value_or_empty(d, k):
-    if k in d:
-       return d[k]
-    else:
-       return ''
-
-def check_dup_attrs():
```
fgalan: This check was obsolete. It made sense when attributes were stored in DB in a vector of name/type/value objects, but not with the new structure.

NTC

scripts/managedb/check_db_sanity.py
```diff
@@ -1,96 +0,0 @@
...
-
-    for doc in db[COL].find():
-        id = doc['_id']['id']
-        type = value_or_empty(doc['_id'], 'type')
-        sp = value_or_empty(doc['_id'], 'servicePath')
-
-        duplicated_attr_name(id, type, sp, doc['attrs'])
-
-def check_dup_id_subfields():
```
fgalan: This check has been recycled in the new check_entities_consistency.py script.

NTC

Issue #3001, opened 2017-10-06
MQTT notifications


fgalan:
This feature is about developing MQTT notifications in CB, in addition to the extisting ones in HTTP. The work will be based in the one done at branch https://github.com/telefonicaid/fiware-orion/tree/feature/mqtt_notifications_poc which already implemented a MQTT notification system (but very limited, as it is hardwired).

The payload used by MQTT notification will be the same one used for NGSIv2 notifications, i.e.: normalized, keyValues, values, legacy or customiced.

From an API point of view is is a matter of using `mqtt/mqttCustom` instead of existing `http/httpCustom` fields in subscriptions. E.g.:

```
"mqtt": {
  "topic": "/my/topic"
}
```

or 

```
"mqttCustom": {
  "topic": "/entity/{$id}/topic",
  "payload": "The temperature value is ${temperature}"
}
```

From a CLI point of view, a new parameter will be used to set MQTT broker host+port, e.g: `-mqttbroker` (similar to existing `-dbhost`). This parameter also acts as a switch to enable/disable MQTT notifications: if `-mqttbroker` is not defined then MQTT notifications are ignored.

Note that MQTT protocol doesn't have any mechanism similar to HTTP headers. Thus, Fiware-Service, Fiware-ServicePath, X-Auth-Token and Fiware-Correlator cannot be directly included in MQTT notifications. There are three approachs to cope with this:

1. Ignore them, assuming that the kind of receivers for MQTT notifications are not interested in that kind of inoformation.
2. Include that information in the MQTT payload as "virtual attributes" (in the same way dateCreated, dataModified, etc. works)
3. Modify the notification templating mechanisms defining ${fiware-service}, ${fiware-servicepath}, etc. tokens, so the client can use in the places of the notification (topic or payload fields) where to include.

McMutton:
Hi @fgalan, I have just started working on this issue. In case I have any questions, I'll post here, thank you.

McMutton:
Hi @fgalan, I have gathered enough information from the issue description, existing documents and the code in the task branch to start the implementation. 

Before doing so, I would like to share my work plan with you to check if that is OK with you. Instead of sending a huge pull request at the end, I would like to divide the task into several ones as below.

First PR: 
- MQTT CLI Switch
- Minor changes/improvements on the core MQTT handler
- Tests

Second PR:
- mqttCustom notifications
- Tests

Third PR:
- Handling of HTTP Headers 
- Tests

[...]

PR #3194, opened 2018-05-29
3001: MQTT inital module implementation

McMutton:
Issue #3001 

This PR implements the CLI and the core logic for MQTT client.

Orion can now be started with the mqtt command option: 
`contextBroker -mqttHost <host> -mqttPort <port> -mqttKeepAlive <keepalive>`

I'll leave review comments on the relevant code lines to start the discussions for the next PR.


src/lib/apiTypesV2/MqttInfo.cpp
```diff
@@ -0,0 +1,97 @@
+/*
```
McMutton: Right now, `MqttInfo` does not get compiled, I left it in the commit to start talking about what is the best way to handle the MQTT notification information.

As of now, `HttpInfo` is very commonly used throughout the Orion code. Creating `MqttInfo` which would hold the MQTT information and would be carried around as `HttpInfo`, sounds reasonable but this means we have to touch lots of lines of code.

Another option might be to carry the MQTT information in `HttpInfo` but this would make the code less understandable and maintainable. 

Creating something like `ConnectionInfo` might be nice but then again we have to touch maybe even more lines of code than the first option.

What do you think?

src/lib/mqtt/mqtt.cpp
```diff
@@ -0,0 +1,136 @@
...
+  int resultCode = mosquitto_connect(mosq, _mqttHost, _mqttPort, _mqttKeepAlive);
+  if (resultCode != MOSQ_ERR_SUCCESS)
+  {
+    LM_X(2, ("Fatal Error (Could not connect to MQTT Broker (%d): %s)", resultCode, mosquitto_strerror(resultCode)));
+  }
+  
+  LM_I(("Successfully connected to MQTT Broker."));
+
+  // Starts the client loop in its own thread. The client loop processes the network traffic
+  mosquitto_loop_start(mosq);
```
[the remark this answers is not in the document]
fgalan: Although I haven't deal with the details, all what you say makes sense to me.

My only "concern" is to we don't leak memory due to thread that are created but not destroyed and think likes that. I guess that the way you are using the handlers of the library is safe regarding this, but a double check would be great.

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
```diff
@@ -0,0 +1,129 @@
...
+    ]
+  },
+  "expires": "2040-01-01T14:00:00.00Z",
+  "throttling": 1
+}'
+orionCurl --url /v2/subscriptions --payload "$payload"
+echo
+echo
+
+echo "03. Update entity attribute A1 to trigger notification"
```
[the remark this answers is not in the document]
fgalan: Ideally we should have an `accumulator-mqtt.py` program similar to the one we already have now for HTTP (accumulator-server.py). I mean, a subscriber of the MQTT broker able to listen & store MQTT-based notification from Orion and able to print the stored notifications in standard output when the proper command (an special MQTT topic) is received.

Have you proficiency in Python and the related MATT libraries ([paho-mqtt](https://pypi.org/project/paho-mqtt/))? :) In positive case, you could develop this script. In negative case I could try to have a look and contribute with it to the feature/mqtt_notifications_poc (however, in this case we should find a way it doesn't block the ongoing work, as I'd do it in a best effort way and it could take some time).

(I'm suggesting Python as it is the one we typically use in our testing tooling and it is better not to diverge using a different lenguage)

src/lib/apiTypesV2/MqttInfo.cpp
fgalan: I think using `MqttInfo` is a good idea. Could you elaborate on the impact of such solution? I mean, why we have to touch lots of line of code in this case? As far as I understand, the main points to touch would be:

* Subscription parsing logic, to properly generated `MqttInfo` object based on JSON payload.
* Subscripton rendering logic, lo properly render JSON corresponding to `MqttInfo` logic.
* Notification logic, where the existence of `MqttInfo` instead of `HttpInfo` would divert to the part of the logic which actually deal with publication in the MQTT broker.

The logic sholud be quite parallel to the one already existing for `HttpInfo`. I mean, look for places in the code where `HttpInfo` is used probably would point out the points in which `MqttInfo` has to be introduced.

However, I could be missing something and be underestimating the difficulty... What do you think?

src/lib/mqtt/mqtt.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+/* ****************************************************************************
+*
+* mqttOnPublishCallback -
+* 
+*/
+void mqttOnPublishCallback(struct mosquitto *mosq, void *userdata, int mid)
+{
+  //FIXME: Do we have a usecase for the callback? To be resolved during second PR
+  LM_W(("PoC: in the publish callback"));
```
fgalan: When the callback is invoked? When the notification has been sucessfully published in MQTT broker?

src/lib/mqtt/mqtt.cpp
```diff
@@ -0,0 +1,136 @@
...
+/* ****************************************************************************
+*
+* sendMqttNotification -
+*
+*/
+int sendMqttNotification(const std::string& content)
+{
+  const char* msg = content.c_str();
+
+  // FIXME: unhardwire QoS, unhardwire retain - Planned for second MQTT PR
```
fgalan: Topic also has to be unhardwired. Although it would be improbable to forget about it :), maybe is a good idea to mention also in the FIXME line.

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
```diff
@@ -0,0 +1,129 @@
...
+    ],
+    "condition": {
+      "attrs": [
+        "A2"
+      ]
+    }
+  },
+  "notification": {
+    "http": {
+      "url": "mqtt://mqtttest.com"
```
fgalan: This doesn't follow the format described at https://github.com/telefonicaid/fiware-orion/issues/3001

```
"mqtt": {
  "topic": "/my/topic"
}
```

althought I guess is intentional, as the processing of that element is out of the scope of this PR (it is related with the `MqttInfo` discussion).

A bit off-topic, but... is `mqtt://` an standarized URL schema?

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
```diff
@@ -0,0 +1,129 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
```
fgalan: I'd suggest to include in this .test the following remark:

"FIXME: this .test has to be adjusted in a next PR as part of #3001 work"

or a similar wording.

makefile
```diff
@@ -177,6 +177,10 @@ post_install_libs:
 	cp src/lib/parse/*.h /usr/local/include/contextBroker/parse
 	cp $(CMAKE_BUILD_TYPE)/src/lib/parse/libparse.a  /usr/local/lib
 
+	cd /usr/local/include/contextBroker  && rm -rf mqtt && mkdir -p mqtt
```
fgalan: Maybe this first PR is a good point to include building documenation. I mean, to include in the https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/build_source.md document a bullet regarding libmqtt installation, just after libmicrohttp installation and in the same style (wget + make + etc.)

(Changes in the .spec to include mqtt dependencies would be another point... if you have proficiency in RPM .spec, please go ahead, if not don't worry, we can deal with it in the future).

fgalan: In addition to .spec, the doc/Dockerfile should be also adjusted...

fgalan:
@McMutton thanks for the contrution! I think is a very good starting point in the MQTT notification functionality. Please have a look to my coments in the PR.

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
McMutton: Yes it is intentional as you commented. 

Regarding the schema, good point, thanks for bringing it up. The documentation on [mqtt.org](http://mqtt.org/documentation) points to [this page](https://github.com/mqtt/mqtt.github.io/wiki/URI-Scheme) which suggests the following format:

    mqtt[s]://[username][:password]@host.domain[:port]

I think we can use this format unless you see any problems with it.

src/lib/mqtt/mqtt.cpp
McMutton: Exactly, it is invoked when a message is successfully sent to the MQTT broker.

Currently I see no use for the callback function other than hosting a proper log message. I'll move the log message below  from `sendMqttNotification` to here.

    LM_I(("MQTT notification sent to %s:%d", mqttHostname, mqttPortNumber));

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
fgalan: Not sure if that reference is yet a draft proposal or an official standard. Looking to IANA registry (https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml), it seems that "mqtt" is not there.

In addition, that schema seems to be aimed at defining the connection to the MQTT broker (which is solved in our case by `-mqttHost` and so on). It seems there isn't place for the topic in that URL.

In sum, I think the original idea of 

```
"mqtt": {
  "topic": "/my/topic"
}
```

is better (not for this PR, but for the next ones)

src/lib/mqtt/mqtt.cpp
fgalan: Yes, logs seem to be the only purpose (and very important when debugging, by the way :)

However, I'd suggest not moving the message about MQTT notification sents, but have two ones. In particular, for `sendMqttNotification` we could use the same we have now and for `mqttOnPublishCallback` we can use:

```
LM_I(("MQTT notification sucessfully published ...", ...);
```

including extra information (if any) about that (in the place where `...` are).

src/lib/mqtt/mqtt.cpp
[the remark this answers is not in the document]
McMutton: We now have the `mqttCleanup` function which takes care of releasing the resources claimed by the MQTT module. According to the mosquitto API documentation, we are doing what we should do.

What I am not sure is that if `mqttCleanup` is called everywhere it should be called. Right now it is only called from the `exitFunc` of `contextBroker.cpp`. Please let me know if I am missing anything.

In any case, I'll double-check to avoid the monster that is memory leak :smile:

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
McMutton: Yes, it is merely a suggestion which is based on community discussions. 

Having said that, I am sorry for the confusion, I totally agree with the original idea and I had already planned and implemented the necessary changes to remove that line. 

I just kept it to have the module working for the first PR until we finalize the discussions about MqttInfo. Then I forget what I am doing, I guess I need a break :man_facepalming:

src/lib/mqtt/mqtt.cpp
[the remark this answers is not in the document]
fgalan: If the behaviour of the library is to create a single thread (that is hold during the contextBroker process lifecycle) and removed when it shut downs (in exitFunc) it's ok. 

The problem would happen if we had "dynamicity" in thread lifecycle, for instance, if each new MQTT notification would involve a new thread being created, which removal should be control. However, I understand this is not the way the MQTT library works.

test/functionalTest/cases/3001_mqtt/subscribe_and_update.test
fgalan: Don't, worry, that's the point of having code review, detect this kind of things :D

src/lib/apiTypesV2/MqttInfo.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include "mongo/client/dbclient.h"
+#include "logMsg/logMsg.h"
+
+#include "common/JsonHelper.h"
+#include "mongoBackend/dbConstants.h"
+#include "mongoBackend/safeMongo.h"
+#include "apiTypesV2/MqttInfo.h"
+
+using namespace mongo;
```
kzangeli: This construct goes against the style guide for Orion, I'm afraid.
Check out this doc: https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md

The idea behind this rule is that `mongo::XXX` is clearer than just `XXX`, showing clearly where `XXX` comes from.

src/lib/apiTypesV2/MqttInfo.h
```diff
@@ -0,0 +1,61 @@
...
+  MqttInfo();
+  MqttInfo(const std::string& _topic);
+
+  std::string  toJson();
+  void         fill(const mongo::BSONObj& bo);
+};
+
+}
+
+#endif // SRC_LIB_APITYPESV2_MQTTINFO_H
```
kzangeli: Here's another style guide fix.
Two spaces before comment (`//`)
We have a script that checks the style guide for us, at least part of it:
```
kz@malika:apinf-orion> scripts/style_check.sh -u
Usage: style_check.sh [-u (usage)]
                      [-v (verbose)]
                      [-d <directory>]
```

Also, after fixing the style 'mistakes', add the directory `mqtt` to the list in `scripts/style_check_in_makefile.sh` (around line 65), if you want the directory to be checked for style by jenkins

src/lib/mqtt/mqtt.cpp
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  mosquitto_publish_callback_set(mosq, mqttOnPublishCallback);
+
+  LM_I(("Connecting to MQTT Broker at %s:%d", _mqttHost, _mqttPort));
+
+  int resultCode = mosquitto_connect(mosq, _mqttHost, _mqttPort, _mqttKeepAlive);
+  if (resultCode != MOSQ_ERR_SUCCESS)
+  {
+    LM_X(2, ("Fatal Error (Could not connect to MQTT Broker (%d): %s)", resultCode, mosquitto_strerror(resultCode)));
```
kzangeli: Not 100% sure the broker should die if the mqtt connection fails ...
It is true that we only use mqtt if explicitly requested and it's easier to catch the error if the broker exits but still ...

I'm not saying 'no'.
I just say we should discuss it.

src/lib/mqtt/mqtt.cpp
```diff
@@ -0,0 +1,136 @@
...
+*
+* sendMqttNotification -
+*
+*/
+int sendMqttNotification(const std::string& content)
+{
+  const char* msg = content.c_str();
+
+  // FIXME: unhardwire QoS, unhardwire retain - Planned for second MQTT PR
+  mosquitto_publish(mosq, NULL, "orion", (int) strlen(msg), msg, 0, false);
```
kzangeli: Can `mosquitto_publish` fail?
I imagine it can ...
If I'm right, then right after a failure we will se a log line saying that everything went well ...
That would be quite confusing :)

src/lib/mqtt/mqtt.h
```diff
@@ -0,0 +1,68 @@
...
+extern unsigned short   mqttPortNumber;
+extern unsigned short   mqttKeepAlivePeriod;
+
+
+
+/* ****************************************************************************
+*
+* mqttInit - 
+*/
+extern void mqttInit (
```
kzangeli: This could easilt fit on one single line.
If you prefer not to do that, for readability, then put the `(` on a separate line, and align `)` with it:
```
extern void mqttInit 
(
  const char*    _mqttHost,
  unsigned short _mqttPort,
  unsigned short _mqttKeepAlive
￼);

File: test/functionalTest/cases/3001_mqtt/subscribe_and_update.test (hunk 3 above)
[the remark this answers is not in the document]
Comment (kzangeli) [reply test/functionalTest/cases/3001_mqtt/subscribe_and_update.test]: Actually, the accumulator doesn't really speak HTTP, it just saves whatever text it received, and then spits it out on request.
I know absolutely nothing about MQTT ...
If it's a text based protocol, like HTTP, then the accumulator would work as is.
If not (which I hope), you'd need some python lib to add mqtt functionality (either inside the accumulator or write a new accumulator)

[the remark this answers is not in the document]
Comment (fgalan) [reply test/functionalTest/cases/3001_mqtt/subscribe_and_update.test]: Current accumulator is not going to work... The library we use there (Flask) is for HTTP and doesn't work for MQTT, which needs another libraries (the paho-mqtt I mention above). It is not rocket science, but it would required some knowledgue in Python to implement it. I can do it (best effort :) if @McMutton cannot.

I'd suggest do it in a separated .py file, for simplicity and cleaness.

File: src/lib/apiTypesV2/MqttInfo.cpp (hunk 9 above)
Comment (fgalan) [reply src/lib/apiTypesV2/MqttInfo.cpp]: Correct. Check rules M12 and M12bis in the documented cited by @kzangeli 

What is legal according to style rule is to use `using` which the specific function you want to use, e.g.:

```
using mongo::BSONObj;
using mongo::BSONElement;
```

Take it into account in the case it could help.

File: src/lib/mqtt/mqtt.cpp (hunk 11 above)
Comment (fgalan) [reply src/lib/mqtt/mqtt.cpp]: If the user uses `-mqttBroker` flag, then MQTT broker becomes a required external system, the same way the DB is. Thus, I think is safer to die if MQTT broker is not ready at startup time (in the same way we "die" when DB is not ready at statup time).

Comment (fgalan) [reply src/lib/mqtt/mqtt.cpp]: (I write "die" within quotes in the case of DB because as far as I remember, the process doesn't dies in that case, but it gets blocket after 100 ateemps or something like that... but this is off-topic for this PR ;)

File: src/lib/mqtt/mqtt.cpp (hunk 12 above)
Comment (fgalan) [reply src/lib/mqtt/mqtt.cpp]: The idea is to have a log just after sending and one if the sent was ok, as discussed in https://github.com/telefonicaid/fiware-orion/pull/3194/files/d7522bf5d5d2686d4355d44258349b07a2734c45#r191839450.

However, maybe the texts could be improved in order to make it clearer. If you have suggestion, please go ahead (better in the cited comment thread).

File: test/functionalTest/cases/3001_mqtt/subscribe_and_update.test (hunk 3 above)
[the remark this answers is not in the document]
Comment (McMutton) [reply test/functionalTest/cases/3001_mqtt/subscribe_and_update.test]: @fgalan I have Python experience, I'll give it a try. In any case, get ready for more questions :)

> I'd suggest do it in a separated .py file, for simplicity and cleaness. 

Sure :+1:

File: src/lib/mqtt/mqtt.cpp (hunk 2 above)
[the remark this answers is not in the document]
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: After a deeper look into the library, I don't see any reason to call mqttCleanup anywhere other than`exitFunc`. The library seems to manage itself within a single thread.

On the other hand, I found that I am missing a call to `mosquitto_lib_cleanup` just after `mosquitto_destroy`. I'll add this call to our `mosquittoCleanup` function.

File: src/lib/mqtt/mqtt.cpp (hunk 11 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Thanks for bringing this up @kzangeli . During the time of coding, my thoughts were more or less the same with what @fgalan commented. Since the user explicitly states their need for an external system through the CLI, it would be better to fail altogether. 

Also, considering the current design, even if we don't kill Orion, the user would have to restart it after a failed MQTT broker connection.

File: src/lib/mqtt/mqtt.cpp (hunk 12 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: @kzangeli Yes, it can and there are a number of different ways it can fail. Thanks for the pointing it out, I'll add the necessary code to handle it in a better way.

File: src/lib/mqtt/mqtt.cpp (hunk 5 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Agreed, thanks for the warning. It is improbable but it is better to approach impossible :)

File: src/lib/mqtt/mqtt.cpp (hunk 2 above)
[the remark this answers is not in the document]
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Fixed in 4896535

File: src/lib/mqtt/mqtt.cpp (hunk 4 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Fixed in 4896535

File: src/lib/mqtt/mqtt.cpp (hunk 5 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Fixed in 4896535

File: test/functionalTest/cases/3001_mqtt/subscribe_and_update.test (hunk 7 above)
Comment (McMutton) [reply test/functionalTest/cases/3001_mqtt/subscribe_and_update.test]: Fixed in 4896535

File: makefile (hunk 8 above)
Comment (McMutton) [reply makefile]: Fixed in 4896535

Mosquitto source build instructions are added to the build_source.md

The current list of dependencies cover what is needed by Mosquitto so I haven't changed the .spec.

Dockerfile is updated to handle Mosquitto installation

File: src/lib/mqtt/mqtt.cpp (hunk 12 above)
Comment (McMutton) [reply src/lib/mqtt/mqtt.cpp]: Fixed in 4896535

File: src/lib/mqtt/mqtt.cpp (hunk 11 above)
Comment (fgalan) [reply src/lib/mqtt/mqtt.cpp]: Thus... NTC?

File: doc/manuals/admin/build_source.md [hunk 14]
```diff
@@ -51,6 +51,16 @@ commands that require root privilege):
...
         sudo ldconfig      # just in case... it doesn't hurt :)
 
+* Install mosquitto from sources (config.mk file under mosquitto-1.5/ can be modified to fine tune the build)
+
+        wget -4 http://mosquitto.org/files/source/mosquitto-1.5.tar.gz
+        tar xvf mosquitto-1.5.tar.gz
+        cd mosquitto-1.5
+        make
+        sudo make install  # installation puts .h files in /usr/local/include and library in /usr/local/lib
+        echo "/usr/local/lib" | sudo tee /etc/ld.so.conf.d/99local.conf # tell runtime loader to look into /usr/local/lib
```
fgalan: Is this really needed? The other dependency libraries (e.g. libmicrohttpd above) also install in /usr/local/lib and no other needs this...

Maybe is something specific of your local system?

McMutton: > The other dependency libraries (e.g. libmicrohttpd above) also install in /usr/local/lib and no other needs this...

Exactly, I did not need to do this for the other libraries, so either my system or the mosquitto build has an issue. 

> Maybe is something specific of your local system?

To be sure, I double checked with a fresh CentOS 7 installation and the result was the same. 

Then I checked the makefiles of the dependencies and realized that, `mongo-cxx-driver`, `libmicrohttpd` and `gmock` are using [libtool](https://www.gnu.org/software/libtool/manual/libtool.html) but mosquitto is not. Libtool creates `.la` files alongside the shared objects which are used by libtool manage the library linking.

> Is this really needed?

I think yes, in order to get mosquitto working, the system should have a `.conf` file under `/etc/ld.so.conf.d`. There are other ways to tell the runtime loader to look in a specific path for libraries but the one we have right now is considered to be best practice (as far as I know).

Changing the /usr/local prefix in `config.mk` file of mosquitto to another location before running make would also be a solution but I prefer the one above.

Please let me know if I am missing something.

src/lib/apiTypesV2/MqttInfo.cpp
McMutton: > look for places in the code where HttpInfo is used

Yes, that's was what I did to estimate the impact. Please allow me to detail it further. Among `.h` and `.cpp` files, `HttpInfo` is referenced in 429 unique lines of code among 33 files.

This may not seem much but for example, although having only **8 unique lines of reference** to `HttpInfo`, the part of the code that is direcly related to parsing of `HttpInfo` in `parseSubscription.cpp` is **more than 150 lines**.

Consequently, I believe it is safe to assume that the impact would not be small in terms of number of lines and files we need to touch.

Nevertheless, I don't see any clean way to achieve this with a smaller impact. So I say, let's go with MqttInfo.

fgalan: > Nevertheless, I don't see any clean way to achieve this with a smaller impact. So I say, let's go with MqttInfo.

Yes, I agree. Thus, let's go with `HttpInfo` (I understand that in the next PR).

McMutton: Yes I plan it for the next PR, otherwise this one will be a bit hard to manage :)

fgalan: NTC? Or is there anything pending to decide?

McMutton: NTC for this PR.

scripts/accumulator-mqtt.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/python
```
fgalan: Good addition to the testing suite! :)

Place have a look to the `install_scripts` target in root `makefile`. This script should be also installed the same way accumulator-server.py is.

fgalan:
Any news on this @McMutton please?

Just to know :)

fgalan:
This PR cannot be considered stale. Although the MQTT notification implementation wasn't finalized, this PR contains a valuable contribution that could be used in the future to complete the task.

kzangeli:
Hi

The ETSI group defining NGSI-LD has recently decided to incorporate notifications using MQTT, and it has already been implemented in both the Scorpio broker and in Orion-LD.
It's all too new still to be in the official ETSI ISG CIM NGSI-LD spec but in a few words I can explain it right here:

First of all, only notifications are supported for MQTT (at least for now).

API details (not 100% relevant as we use different fields, but the info needs to be there):

The notification receiver of a subscription instead of defining an "http://host:port/path", for notifications in MQTT it is defined like this:
```
    "endpoint": "mqtt://host:port/topic"
```
Then we added an optional new field for Quality of Service and we're talking about another field for a password (MQTT brokers can be password protected as it seems).

So, the idea we have is to alongside the NGSI-LD broker start an MQTT broker, e,g, mosquitto.
A client wishing to receive notifications in MQTT will have to do two things:
* Create the subscription in the NGSi-LD broker - making the MQTT broker receptor of notifs
* Subscribe to the topic in the MQTT broker

This is a bit transparent to the broker so if you wish to skip the MQTT broker I guess that would work as well. I wouldn't recommend it though as I'm sure that mosquitto et al implement re-intents and stuff that you would have to implement in the broker otherwise ...

The payload of MQTT notifications has been decided to look like this;
```
{
  "metadata": {
    <HTTP header info>
  },
  "body" : {
    <payload data>
  }
}
```
As you see, we put *all* the info in the JSON object, as MQTT doesnb't have the concept of headers that HTTP has.

Needless to say, it would be nice to have Orion implementing the same as what the NGSI-LD specifications states (well, soon will state).

The next official version of the NGSI-LD specification is planned for the end of April and the MQTT notifications will be part of the spec by then.

If more information is needed (I assume it to be so), feel free to contact me or Fermin for details.

mapedraza:
FYI, latest cURL versions support mqtt. It is a very basic and experimental support but maybe could be easier to implement as  libcurl is the library used for http notification requests. More info [here](https://github.com/curl/curl/blob/master/docs/MQTT.md)

fgalan:
I'm going to merge this PR in telefonicaid:feature/mqtt_notifications_poc and continue from that point.

Issue #2721, opened 2016-11-25
Align actionType names with NGSIv2 API style (camelCase)


fgalan:
Currently NGSIv2 spec uses the following for `POST /v2/op/update` actionTypes:

> actionType, to specify the kind of update action to do: either APPEND, APPEND_STRICT, UPDATE, DELETE, REPLACE

(REPLACE is not currently included, by it will. See #2672)

This style is legacy from NGSIv1. However, NGSIv2 uses a camelCase style, e.g. "legacy", "keyValues", etc. Thus maybe we should change to:

* append, appendStrict, update, delete, replace

The old names (APPEND, etc.) will be marked as deprecated functionality but Orion will still supporting them in order to keep backward compatibility.

@jmcanterafonseca what do you think?

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/3197

PR #3197, opened 2018-05-31
camelCase actionTypes in POST /v2/op/update

fgalan:
Implements the new action types decribed in #2721, adjusting also documenation.

actionType has been changed in all POST /v2/op/update cases in all .test and a new one have been added in cases/0000_deprecated_checkings to ensure the uppdercase variants keep working.

Please @jmcanterafonseca have a look to modifications in .apib (you can ignore the rest of the PR). I think it would be ok, but I prefer if you could have a look and give LGTM, pls.


src/lib/jsonParse/jsonUpdateContextRequest.cpp
```diff
@@ -389,11 +389,13 @@ std::string jsonUpcrCheck(ParseData* reqData, ConnectionInfo* ciP)
 */
 void jsonUpcrPresent(ParseData* reqDataP)
 {
+  // FIXME PR: leftover?
   if (!lmTraceIsSet(LmtPresent))
```
fgalan: @kzangeli is this a leftover?

src/lib/orionTypes/UpdateActionType.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "common/globals.h"
+
+
+
+/* ****************************************************************************
+*
+* actionTypes enum (and rendering functions) -
+*/
+typedef enum ActionType
```
fgalan: This new enum is the cornerstone of all the PR. The idea is to use an enum internally instead of std::string, allowing  both action type dialects at parsing layer ("APPEND" and "append", etc.) but using the same enum-based at inner layers. The refactor in code could seem large, but almost all the changes are related with that.

NTC

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -2184,23 +2184,23 @@ Response:
 This operation allows to create, update and/or delete several entities in a single batch operation.
 The payload is an object with two properties:
 
-+ `actionType`, to specify the kind of update action to do: either APPEND, APPEND_STRICT, UPDATE,
-  DELETE, or REPLACE.
++ `actionType`, to specify the kind of update action to do: either `append`, `appendStrict`, `update`,
```
kzangeli: English lesson ...
Can 'either' be used if there is more than two options?

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -296,8 +296,14 @@ have been needed in the end. Thus, there is changed functionality that doesn't a
...
 (as [deprecated functionality](../deprecated.md)) in order to keep backward compability.
 
-In particular, the usage of `dateCreated` and `dateModified` in the `options` parameter (introduced
+In particular:
+
+* The usage of `dateCreated` and `dateModified` in the `options` parameter (introduced
 in stable RC-2016.05 and removed in RC-2016.10.) is still supported, e.g. `options=dateModified`. However,
 you are highly encouraged to use `attrs` instead (i.e. `attrs=dateModified,*`).
 
+* `POST /v2/op/update` accepts the same action types than NGSIv1, that is `APPEND`, `APPEND_STRICT`,
```
kzangeli: than => as

src/lib/jsonParse/jsonUpdateContextRequest.cpp
kzangeli: Seems like it.
However ... looking a little closer, the function is called jsonUpcr**Present**.
It does nothing but presenting info. And always using the Log trace `LmtPresent`, do, if that log level is not set, why spend any time inside this function?

So no, not a leftover.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2232,7 +2233,7 @@ static void updateAttrInNotifyCer
   ContextElementResponse* notifyCerP,
   ContextAttribute*       targetAttr,
   bool                    useDefaultType,
-  std::string             actionType
+  const std::string&      actionType
```
kzangeli: Good!!!

NTC

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -3150,7 +3151,7 @@ static void updateEntity
   cerP->contextElement.entityId.fill(entityId, entityType, "false");
 
   /* If the vector of Context Attributes is empty and the operation was DELETE, then delete the entity */
-  if (strcasecmp(action.c_str(), "delete") == 0 && ceP->contextAttributeVector.size() == 0)
+  if ((action == ActionTypeDelete) && (ceP->contextAttributeVector.size() == 0))
```
kzangeli: So, we had a `strcasecmp` here before ...
Is this a possible problem for backward compatibility? (depends on the action-type-parse function).

src/lib/ngsi10/UpdateContextRequest.h
```diff
@@ -52,7 +52,8 @@ struct UpdateContextAttributeRequest;
 typedef struct UpdateContextRequest
 {
   ContextElementVector    contextElementVector;  // Mandatory
-  UpdateActionType        updateActionType;      // Mandatory
+  //UpdateActionType        updateActionType;      // Mandatory
```
kzangeli: This here seems like a leftover ...

src/lib/orionTypes/UpdateActionType.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+
+
+/* ****************************************************************************
+*
+* parseActionTypeV1 -
+*/
+ActionType parseActionTypeV1(const std::string& action)
+{
+  if (strcasecmp(action.c_str(), "update") == 0)
```
kzangeli: Here it is.
No problem for V1

src/lib/orionTypes/UpdateActionType.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+
+
+/* ****************************************************************************
+*
+* parseActionTypeV2 -
+*/
+ActionType parseActionTypeV2(const std::string& actionType)
+{
+  if (actionType == "update")
```
kzangeli: In V2 we only accept lowercase.
It hasn't been processed earlier, right?
So, this here might be a problem - unless the spec says it must be all in lowercase. If so, it's ok.

[ Found the call to V1 parse function - NTC ]

src/lib/orionTypes/UpdateActionType.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include "common/globals.h"
+#include "orionTypes/UpdateActionType.h"
+
+
+
+/* ****************************************************************************
+*
+* actionTypeString -
+*/
+std::string actionTypeString(ApiVersion apiVersion, ActionType action)
```
kzangeli: If this function could return a `char*` instead, we wouldn't have to copy an object onto the stack ...

src/lib/orionTypes/UpdateActionType.cpp
```diff
@@ -0,0 +1,134 @@
...
+    return ActionTypeDelete;
+  }
+  else if (actionType == "replace")
+  {
+    return ActionTypeReplace;
+  }
+  else
+  {
+    // Last resort: try NGSIv1 based parsing. This could be removed in the future
+    return parseActionTypeV1(actionType);
```
kzangeli: Ah, I was too fast, strcasecmp comes in here ...

NTC (and adding an NTC to my former comment)

src/lib/orionTypes/UpdateActionType.h
kzangeli: Yes. It's a very good modification

src/lib/serviceRoutinesV2/postEntity.cpp
```diff
@@ -63,7 +63,8 @@ std::string postEntity
 )
 {
   Entity*        eP  = &parseDataP->ent.res;
-  std::string    op  = ciP->uriParam["op"];
+  //std::string    op  = ciP->uriParam["op"];
```
kzangeli: Leftover?
Wait, is the `op` uri param no longer used?

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I only change the literals, that part was that way already. However, no problem in changing the surronding text is something is wrong :)

src/lib/jsonParse/jsonUpdateContextRequest.cpp
fgalan: Thus, the following removed lines should be recovered also, right? https://github.com/telefonicaid/fiware-orion/pull/3197/files#diff-2635eb6d54466f77bca7735f68e47ee3L130

A bit off-topic but... how are we useing *present() methods at the present moment? I think we were using in the very early days of CB to debug, but I'm not pretty sure right now... In which points of the code are we invoking "root" *present() methods?

src/lib/orionTypes/UpdateActionType.cpp
fgalan: The NGSIv2 spec says that it has to be in lowercase (see modifications in this PR in fiware-ngsiv2-reference.apib).

NTC (I guess)

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Moved to new method `parseActionTypeV1()`, where `strcasecmp()` is used.

NTC (I guess)

src/lib/orionTypes/UpdateActionType.cpp
fgalan: NTC (I guess)

src/lib/orionTypes/UpdateActionType.cpp
fgalan: I tend to prefer returning std::string, for simplicity and readibility reasons (I think we have talked about this in the past... long time... many times :D)

Note that in the previous method equivalent to this (render() in the now deleted ngsi10/UpdateActionType.cpp file) was using the same pattern.

src/lib/serviceRoutinesV2/postEntity.cpp
fgalan: Yes, it's a leftover.

In fact, this was weird in the current code. `op` is set form a non-existing op= parameter (it was in very early NGSIv2 versions, but at the end we preferred options=append for the same)... to be overriden later. This PR fix that weird-ness.

fgalan: Fixed in 206e503

src/lib/ngsi10/UpdateContextRequest.h
fgalan: Fixed in 206e503

doc/manuals/user/ngsiv2_implementation_notes.md
fgalan: Fixed in 206e503

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: I don't know if it's wrong or right ... 
I was hoping Jason Fox would be with us and let us know ... :)

src/lib/jsonParse/jsonUpdateContextRequest.cpp
kzangeli: Not sure you read more than the first line of my comment ...
It is **not** a leftover.
But then again, as you say, the `present`  methods are no longer used (at least not by me, and I'm the one who invented them ... :))
I see no problem with removing all of them.

doc/apiary/v2/fiware-ngsiv2-reference.apib
kzangeli: https://english.stackexchange.com/questions/170409/whats-the-best-way-to-use-either-on-more-than-two-options

Seems it is correct English.
Still, I find it strange !!!

NTC

src/lib/jsonParse/jsonUpdateContextRequest.cpp
fgalan: > Not sure you read more than the first line of my comment ...
> It is not a leftover.

Yes, I dit read... that's why I'm saying that if it is **not** a leftover then the lines I removed at https://github.com/telefonicaid/fiware-orion/pull/3197/files#diff-2635eb6d54466f77bca7735f68e47ee3L130 thinking that part was also a leftover should be restored. 

I haven't done yet, but I will. I just what waiting to conclude the discussion on this comment.

> But then again, as you say, the present methods are no longer used (at least not by me, and I'm the one who invented them ... :))
> I see no problem with removing all of them.

I though that maybe at some point in parse library we have some invocation to present() under some conditions (e.g. `LmtPresent` activated). But if they are no longer called from any place and they are dead code, yest, better to remove.

However, not in this PR. This PR is about camelCase for actionTypes, so it is out my intention to touch too much other not related parts (as present() logic is).

kzangeli: Yes, not in thiCasa de aperosCasa de apeross PR.
Remove the "leftover" comment. It is incorrect.
And let's create an issue about removing all `present` methods

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Yeah, I saw.

NTC

src/lib/orionTypes/UpdateActionType.cpp
kzangeli: NTC

src/lib/orionTypes/UpdateActionType.cpp
kzangeli: NTC

src/lib/orionTypes/UpdateActionType.cpp
kzangeli: Yeah ... you prefer C++
In reality, the win in simplicity and readibility you talk about doesn't really exist.
Just changing std::string for char* doesn't change simplicity and readibility at all.

The problem here is that we return am object on the stack and put own style guide forbids that.
[ Which is a thing all style guides do, for good reasons ]
I do know we return std::string in a million places. I just don't see why we have to add more functions returning objects. What we should do is to not add more and fis the old ones.

What about:

`void actionTypeString(ApiVersion apiVersion, ActionType action, std::string* typeStringP)` ?

That option keeps std::string and removes the problem.

src/lib/jsonParse/jsonUpdateContextRequest.cpp
fgalan: Fixed in 2a3c07a3159e28aefe070e317edd6d10ac219f62. And issue created at https://github.com/telefonicaid/fiware-orion/issues/3199

src/lib/orionTypes/UpdateActionType.cpp
fgalan: Comparing:

```
std::string actionTypeString(ApiVersion apiVersion, ActionType action) {...}
...
  std::string details = std::string("action: ") + actionTypeString(apiVersion, action) + ...
...
```

with 

```
void actionTypeString(ApiVersion apiVersion, ActionType action, std::string* typeStringP) {...}
...
  std::string actionTypeS;
  void actionTypeString(apiVersion, action, &typeStringP);
  std::string details = std::string("action: ") + actionTypeS + ...
...
```

The first option is more compact (less code and more "direct"), readable (less lines of code, less extra variables at the caller) and mantenible (less points to touch if a change is needed there).

However, I think the discussion is out of the scope of this PR, so I'd suggest keep as it was before my PR (note that the render() methods where this comes from already uses std::string to return) and discuss in it in an independent (and more general) scope.

(With regards to style, the rule currently in place has to be reviewed, as originally  `std::string` was allowed to be returned, as reference (`std::string&`)... before we realize `const` cannot be used here. Note the rule mentions "inefficient copies", which has to do with large objects. Is std::string a large object? Which size has a std::string? We should think in this kind of questions...)

fgalan: > Yeah ... you prefer C++

I don't have a *general* preference of C++ among C (or the other way around ;). However, *particularly* you are right I like std::string, as they are a very useful abstraction to deal with text strings.

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -296,8 +296,14 @@ have been needed in the end. Thus, there is changed functionality that doesn't a
...
-In particular, the usage of `dateCreated` and `dateModified` in the `options` parameter (introduced
+In particular:
+
+* The usage of `dateCreated` and `dateModified` in the `options` parameter (introduced
 in stable RC-2016.05 and removed in RC-2016.10.) is still supported, e.g. `options=dateModified`. However,
 you are highly encouraged to use `attrs` instead (i.e. `attrs=dateModified,*`).
 
+* `POST /v2/op/update` accepts the same action types as NGSIv1, that is `APPEND`, `APPEND_STRICT`,
+`UPDATE`, `DELETE` and `REPLACE`. However, they shouldn't be used, preferring always the following counterparts:
+`append`, `appendStrict`, `update`, `delete and `replace`.
```
jmcanterafonseca: nit: check use of apostrophe

fgalan: Fixed in 5d9a45ae24481845f598068a1188708de1c58453

src/lib/orionTypes/UpdateActionType.cpp
kzangeli: So, shall we follow our style guide, and avoid sending objects on the stack or should we change the style guide and say it's OK to send objects on the stack?

Never mind for this PR. It is true that we return std::string in many places already.
Creating an issue about removing all of them.

kzangeli: NTC

kzangeli:
LGTM (but it seems a conflict needs to be resolved)

PR #3177, opened 2018-05-22
Fix for issue actionType is not checked before calling searchContextProviders() in updateEntity() function #2874

SourabhChourasiya-NEC:
Fix for issue **actionType is not checked before calling searchContextProviders() in updateEntity() function #2874**  [https://github.com/telefonicaid/fiware-orion/issues/2874]

PR #3198, opened 2018-06-04
FIX ASCII format in .test

fgalan:
The .test was introduced in PR https://github.com/telefonicaid/fiware-orion/pull/3177. However, it has broken CI due to a formatting problem. The .test was working in the original deverloper environment, so it is not an actual problem, but maybe something causes by diferent locales in develepor and CI machines.

It is not a big issue but please @SourabhChourasiya-NEC takes this into account for future .test ;)


test/functionalTest/cases/3177_updateContext_APPEND_CPr_Registration/updateContext_APPEND_CPr_Reg.test
```diff
@@ -1,230 +1,230 @@
-# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: No real change in this CI... just appliying `dos2unix` tool.

PR #3207, opened 2018-06-11
FIX mongo URL construction in harnessFunctions.sh

fgalan:
Maybe this error only happens with modern versions of the mongo shell. I have faced it just after upgrading my mongo package, when all test start to fail due to `dbInit` caused:

```
2018-06-11T15:40:08.988+0200 I NETWORK  [thread1] getaddrinfo("ftest") failed: Name or service not known
2018-06-11T15:40:08.989+0200 E QUERY    [thread1] Error: couldn't initialize connection to host ftest, address is invalid :
connect@src/mongo/shell/mongo.js:240:13
```

I guess that it will not hurt to people using old versions of the mongo shell but it would be a good idea to check with some other system before merging this. Any feedback?

### DIFF
--- a/test/functionalTest/harnessFunctions.sh
+++ b/test/functionalTest/harnessFunctions.sh
@@ -145,7 +145,7 @@ function dbInit()
   fi
 
   dMsg "database to drop: <$db>" 
-  echo 'db.dropDatabase()' | mongo $host:$port/$db --quiet
+  echo 'db.dropDatabase()' | mongo mongodb://$host:$port/$db --quiet
 }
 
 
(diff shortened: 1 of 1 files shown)


kzangeli:
Good that we have this all covered in aux-functions in harnessFunctions.sh.
However ... there might be uses still in *.test files ... 
Did you grep for it?

fgalan:
No, I didn't... good point! I'll check

fgalan:
> No, I didn't... good point! I'll check

Checked. No other place use Mongo connections string in test directly. A full pass of ft show it works.

kzangeli:
LGTM

PR #3211, opened 2018-06-13
Fix: NGSIv2 update passes command to IOTA but response to client is 404

### LINKED ISSUE #2871: NGSIv2 update passes command to IOTA but response to client is 404 (not ok)
Scenario: IOTA-JSON (using MQTT transport) and CB with some devices with commands provisiened at IOTA (in particular "CommandE" in entity "DevEnt1".

Updating using NGSIv1 results in 200 OK:

```
POST /v1/updateContext
{
    "contextElements": [
        {
            "type": "Device",
            "isPattern": "false",
            "id": "DevEnt1",
            "attributes": [
                {
                    "name": "CommandE",
                    "type": "int",
                    "value": "Fermin-test1"
                }
            ]
        }
    ],
    "updateAction": "UPDATE"
} 
```

But updating using NGSIv2 results in 404 Not Found:

```
PUT /v2/entities/DevEnt1/attrs/CommandE?type=Device

{ "value": "Fermin-test2", "type": "int" }
```

In both cases (NGSIv1 and NGSIv2) the command is correctly progressed to device, i.e. we get at MQTT subscribers:

```
/1234/DevID01/cmd {"CommandE":"Fermin-test1"}
/1234/DevID01/cmd {"CommandE":"Fermin-test2"}
```

My guess: some bug in the NGSIv1->NGSIv2 conversion logic for the updateContext responses at CB, in some part of updateForward() function.

fgalan:
Fixes #2871


scripts/accumulator-server.py
```diff
@@ -207,16 +207,18 @@ def bad_response():
     r.data = '{"name":"ENTITY_NOT_FOUND","message":"The entity with the requested id [qa_name_01] was not found."}'
```
fgalan: A lot of diff noise...

Basically what I have done is to refactor notification storage code in record_request() and add a new function record_2871() prepared to return the ouout that Orions needs to reproduce the case described in the issue. I don't like too much this kind of ad hoc code, as long as we don't have too much exceptional cases and they are easily "plugable" in record_XXXX() functions.

NTC (just informative)

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -711,15 +711,7 @@ std::string postUpdateContext
     // one used by the calling postBatchUpdate() function at serviceRoutineV2 library
     if (fails == response.contextElementResponseVector.size())
     {
-      // If all CER result in error, then it isn't a partial update, but a regular NotFound
-      if (ciP->apiVersion == V1)
```
fgalan: This block already runs in a `if (ciP->apiVersion == V2)` so we don't need to re-check here.

NTC (just informative)

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -729,6 +721,10 @@ std::string postUpdateContext
       // If some CER (but not all) fails, then it is a partial update
       parseDataP->upcrs.res.oe.fill(SccContextElementNotFound, "Attributes that were not updated: { " + failing + " }", "PartialUpdate");
     }
+    else {  // fails == 0
```
kzangeli: style guide ... ( `{` on a separate line)

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -729,6 +721,10 @@ std::string postUpdateContext
...
+      // No fail in this case, so invalidate any posible OrionError filled by mongoBackend on the mongoUpdateContext ste
```
kzangeli: No failure, so invalidate any possible OrionError ...

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -729,6 +721,10 @@ std::string postUpdateContext
...
+      parseDataP->upcrs.res.oe.fill(SccNone, "");
```
kzangeli: `SccNone` is a little dangerous (can give a `500` response if not modified).
Are we sure that the `SccNone` is always overwritten?
  
But, I imagine you have run the func tests ...

test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
```diff
@@ -0,0 +1,370 @@
...
+
+--SHELL--
+
+#
+# Note we don't typically use direct creation of entities/registrations in DB, but in this case
+# in order to reproduce precicesly the situation that causes issue #2871 we are doing that way
+# in steps 1 and 2
+#
+# 01. Create registration directly in DB
+# 02. Create entity direclty in DB
```
kzangeli: typo: direclty

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: SccNone is used by the NGSIv2 caller to check if an error has to be rendered. For instance, in postBatchUpdate(), just after calling postUpdateContext(), we have:

```
  // Check potential error
  if (parseDataP->upcrs.res.oe.code != SccNone )
  {
    TIMED_RENDER(answer = parseDataP->upcrs.res.oe.toJson());
    ciP->httpStatusCode = parseDataP->upcrs.res.oe.code;
  }
  else
  {
    ciP->httpStatusCode = SccNoContent;
  }
```

In addition, func test have passed, which validates the solution.

kzangeli: Yeah, I imagined you had the tests run already.

Nothing to do with this PR but ...
What if the oe.code is 200 ?
A bit risky that line checking only SccNone and not SccOk ...

NTC

test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
fgalan: Fixed in 5a3ae39

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: Fixed in 5a3ae39

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: Fixed in 5a3ae39

kzangeli:
LGTM

Issue #3215, opened 2018-06-15
Upsert operation


fgalan:
An upsert variant of the entity creation operation should be included using `?options=upsert`. For instance:

```
POST /v2/entities?options=upsert
{
  "id": "E",
  "type": "T",
  "A": {
    "value": 2,
    "type": "Number"
  }
}
```

will create the entity *or* update it if already exists.

Note that currently the same can be done using:

```
POST /v2/op/update
{
  "actionType": "append",
  "entities": [
    {
      "id": "E",
      "type": "T",
      "A": {
        "value": 2,
        "type": "Number"
      }
    }
  ]
}
```

However it involves more syntactical complexity (i.e. `actionType` and `entities` elements) as that operation is designed to a bulk update on entity (and a bulk of 1 is a bit missuse of the operation)

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/3217

PR #3217, opened 2018-06-15
ADD upsert operation

fgalan:
Implements #3215 

Requested LGTMs: @kzangeli (mainly for the code and test parts) @jmcanterafonseca (mainly for the .apib parts)


doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1106,8 +1107,10 @@ Response:
       + Members
           + keyValues - when used, the request payload uses the `keyValues` simplified entity
             representation. See "Simplified Entity Representation" section for details.
+          + upsert - when used, entity is updated if already exits (otherwise this situation causes
+            a 422 Unprocessable Entity error).
 
-+ Response 201
++ Response 201 204
```
fgalan: Not sure how this should be included in the .apib

* Enumerating the possible 2xx codes? (as it is now)
* Using two `+ Response` blocks, one for 201 the other for 204?
* Using only 201 as an example (as it was before)?

@jmcanterafonseca some advice on this would be great, please.

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1074,7 +1074,8 @@ the JSON entity representation format (described in a "JSON Entity Representatio
 
 Response:
 
-* Successful operation uses 201 Created. Reponse includes a `Location` header with the URL of the
+* Successful operation uses 201 Created (if upsert option is not used) or 204 No Content (if
+  upsert option is used). Reponse includes a `Location` header with the URL of the
```
jmcanterafonseca: Only Location header will be present when there was actually an entity creation

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1106,8 +1107,10 @@ Response:
       + Members
           + keyValues - when used, the request payload uses the `keyValues` simplified entity
             representation. See "Simplified Entity Representation" section for details.
+          + upsert - when used, entity is updated if already exits (otherwise this situation causes
+            a 422 Unprocessable Entity error).
```
jmcanterafonseca: clarify this better, perhaps putting that sentence separate

doc/apiary/v2/fiware-ngsiv2-reference.apib
jmcanterafonseca: Better to add a new response code

fgalan: Does .apib accept to have something like this?

```
+ Response 201
   ...
+ Response 204
  ...
```

(Maybe is a naïve question, but we haven't used that pattern in the .apib up to now :)

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: I think is better to include always the Location heaader. It will be redudant in the case of update, but at the same time it offerst a less variable response interface for clients. It doesn't hurt.

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: New block added in 6566b8aad31afa25fed44b9b8a5aec3661ef71f1... although I'm not sure if works this way (as asked in my previous comment).

doc/apiary/v2/fiware-ngsiv2-reference.apib
fgalan: Trying to clarify in 6566b8aad31afa25fed44b9b8a5aec3661ef71f1

doc/apiary/v2/fiware-ngsiv2-reference.apib
```diff
@@ -1106,13 +1107,21 @@ Response:
       + Members
           + keyValues - when used, the request payload uses the `keyValues` simplified entity
             representation. See "Simplified Entity Representation" section for details.
+          + upsert - when used, entity is updated if already exits. If upsert is not used and
```
jmcanterafonseca: use backtick for `422 Unprocessable Entity`

fgalan: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3217/commits/f01b3c3472c92bb5a2578e3892737a30ab2fbf5f

PR #3220, opened 2018-06-19
ADD documentation for transient entities

fgalan:
This corresponds to the documentation for #3000, aligned with the imlpementation done in PR #3173 . The ideas is to merge first #3173, then to merge this PR.

@arigliano please have a look to it, in order to ensure all what I describe is correct and that I'm not missing any important point. Thanks!


doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+no way of recovering it**. Ensure the information you set in a transient entity is actually "transient"
+(i.e. it is not important after the lifetime of the entity has passed and it gets expired).
+
+In addition, **have a look to the [backward compatibility considerations section](#backward-compatibility-considerations)
+in the case you are already using attributes with the exact name `dateExpires`**.
+
+## The `dateExpired` attribute
+
+Transient entities are based in the `dateExpires` attribute. This is an attribute of `DateTime` type,
+according to the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/stable/).
```
fgalan: Should transient entities functionality being included in NGSIv2? Or, in other words, should be this specification in the NGSIv2 .apib or in a Orion specific .md?

I think is better to keep it as an specific implementation aspect (i.e. in a Orion specific .md) in order to keep NGSIv2 spec simpler. However, I'm open to suggestions. @jmcanterafonseca what do you think?

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+## Deletion of expired entities
+
+Expiration relies on the MongoDB feture to [expire documents at a specific clock time](https://docs.mongodb.com/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time). This is based in a background thread thats wakes up every 60 seconds,
+so your transient entities may remain in the database up to 60 seconds (or a bit more, if the MongoDB load is high) after
+expiration date (see [MongoDB documentation](https://docs.mongodb.com/manual/core/index-ttl/#timing-of-the-delete-operation)
+for more details).
+
+**Once a transient entity is removed, it cannot be recovered.**
+
+## Backward compatibility considerations
```
fgalan: This is based in my "theoretical" interpretation of the implementation in #3173 . However, I may be wrong. This should be tested "on real" (once #3173) gets merged but on the menawhile let's try this description to be as precise as possible.

@arigliano what do you think? Is the description on this section correct?

arigliano: Yes, It's OK 👍

doc/manuals/user/transient_entities.md
fgalan: On a second thought, maybe `dataExpires` can be considered part of the "operational information" of the entity and, in that sense, should be a NGSIv2 system/builtin attribute, as `dateModified` or `dateCreated` (the original ideal... later discarded... now again under discussion :)

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
+# Transient Entities
+
+Orion Context Broker implements *transient entities*. A transient entity is like any other entity
```
jmcanterafonseca: A transient Entity is a regular Entity (i.e. it has id/type, a set of attributes, etc.) but with a expiration timestamp. When such point in time is reached the entity is automatically deleted from the Context managed by Orion.

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
+# Transient Entities
+
+Orion Context Broker implements *transient entities*. A transient entity is like any other entity 
+(i.e. it has id/type, a set of attributes, etc.) but with a limited lifetime, so when a given
+time has come, the entity is automatically deleted from the context managed by Orion.
+
+Thus, a first and very important piece of advice: **be careful if you use transient entities as once
+the expiration time has come, the entity will be automatically deleted from database and there is
+no way of recovering it**. Ensure the information you set in a transient entity is actually "transient"
```
jmcanterafonseca: it is not relevant once the Entity has expired (i.e. deleted)

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+(i.e. it is not important after the lifetime of the entity has passed and it gets expired).
+
+In addition, **have a look to the [backward compatibility considerations section](#backward-compatibility-considerations)
+in the case you are already using attributes with the exact name `dateExpires`**.
+
+## The `dateExpires` attribute
+
+Transient entities are based in the `dateExpires` attribute. This is an attribute of `DateTime` type,
```
jmcanterafonseca: The expiration timestamp of an Entity is defined by means of the `dateExpires` attribute

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+according to the [NGSIv2 specification](http://telefonicaid.github.io/fiware-orion/api/v2/stable/).
```
jmcanterafonseca: http://fiware.github.io/specifications/ngsiv2/stable/

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+  "type": "Ticket",
+  ...
+  "dateExpires": {
+    "value": "2028-07-07T21:35:00Z",
+    "type": "DateTime"
+  }
+}
+```
+
+creates an entity that will expire at July 7th, 2028 at 21:35.
```
jmcanterafonseca: 21:35 UTC

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+
+Additional considerations:
+
+* `dateExpires` has to have a valid `DateTime` value (check NGSIv2 specification for details).
+Otherwise a 400 Bad Request would be returned.
+
+* If `dateExpires` is set in the past, the entity is created expired (a bit weird, but functionally correct).
```
jmcanterafonseca: how the hell it can be created expired?? shouldn't this be a 400 Error?

doc/manuals/user/transient_entities.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+will make that entity to expire on October 12th, 2028 at 14:23.
+
+Additional considerations:
+
+* `dateExpires` has to have a valid `DateTime` syntax (check NGSIv2 specification for details).
+Otherwise a 400 Bad Request would be returned.
+
+* If `dateExpires` is set in the past, the entity gets automatically expired.
```
jmcanterafonseca: i.e. is not created ... or is it created and automatically deleted? a bit weird

fgalan: Note that MongoDB may take up to 60 seconds to delete an expired entity. Thus, we cannot talk about not creation or automatic deletion... to be precise, we talk about "expired entity" and, in another section (Deletion of expired entities) explain the behvaiour that ends with a expired entity to be deleted.

doc/manuals/user/transient_entities.md
fgalan: It can created expired, and MongoDB can take until 60 seconds to return it. Note we are mostly delegating on MongoDB for transient entities deletion, in order to keep Orion code simpler (otherwise, we would need to add more checkings in Orion).

doc/manuals/user/transient_entities.md
fgalan: Fixe din 78286a073f3b775bed26ebf74bb17da2bea33fe9... expect capitalization, given I Don't Want to Write in This Way :)

doc/manuals/user/transient_entities.md
fgalan: Fixed in Fixed in 78286a073f3b775bed26ebf74bb17da2bea33fe9

doc/manuals/user/transient_entities.md
fgalan: Fixed in 78286a073f3b775bed26ebf74bb17da2bea33fe9

doc/manuals/user/transient_entities.md
fgalan: It's the same, as they are identical mirrors... but anyway, chaned in 78286a073f3b775bed26ebf74bb17da2bea33fe9

doc/manuals/user/transient_entities.md
fgalan: Good point. Fixed in 78286a073f3b775bed26ebf74bb17da2bea33fe9

doc/manuals/user/transient_entities.md
fgalan: Thanks for checking it! NTC

jmcanterafonseca:
for me is ok that `dateExpires` is a built-in attribute. with that fixed and nailed down LGTM.

fgalan:
`dateExpires` as NGSIv2 builtin attribute in commit 935c7b7

PR #3221, opened 2018-06-20
ADD missing case in #2871 test case

fgalan:
We missed an attribute update variante in the .test include for #2871. This PR completes the .test with the missing case.

### DIFF
--- a/test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
+++ b/test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
@@ -40,15 +40,16 @@ accumulatorStart --pretty-print
 # 03. Update turn attribute using PATCH on entity
 # 04. Update turn attribute using PUT on attribute
 # 05. Update turn attribute using batch update
-# 06. Update turn attribute using NGSIv1 update
-# 07. Dump accumulator and see four forwarded updates
+# 06. Update turn attribute using PUT on attribute value
+# 07. Update turn attribute using NGSIv1 update
+# 08. Dump accumulator and see four forwarded updates
 #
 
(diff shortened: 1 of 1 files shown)


test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
```diff
@@ -40,15 +40,16 @@ accumulatorStart --pretty-print
...
+# 07. Update turn attribute using NGSIv1 update
+# 08. Dump accumulator and see four forwarded updates
 #
 
 echo "01. Create registration directly in DB"
 echo "======================================"
 doc='{
    "_id":ObjectId("5b1ff2ee49597334c9b5efc7"),
-   "expiration":NumberLong(1531412462),
+   "expiration":NumberLong(1845053934),
```
fgalan: Old expiration was on July 2018, so this .test would start to fail in less than a month. The new expiration is in June 2028, which I think is pretty long :)

NTC (just informative)

test/functionalTest/cases/2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test
```diff
@@ -78,28 +79,58 @@ echo
...
-        {
-          "name":"turn",
-          "type":"string",
-          "isDomain":"false"
-        }
-      ],
-      "providingApplication":"http://localhost:9997/bug2871"
-    }
-  ]
+        "_id": {
```
fgalan: Its funny... this was completely wrong (a reg doc inserted in the entities collection) but the .test was working anyway. I have recovered the entity example in the #2871 issue and check that the .test is failing with versions of CB previous to the fix.

NTC (just informative)

PR #3192, opened 2018-05-28
Travis integration

caa06d9c:

PR #3233, opened 2018-06-28
Fix CI retval for compliance job (and some additional polishing on the CI work)

fgalan:
This PR is based on the work done in https://github.com/telefonicaid/fiware-orion/pull/3192. It fix the build.sh CI script retval for the compliance tests. In addition, it includes some minor fixing and enhance README.md.

The results are:

imagen

It is ok test compliance file to fail, as compliance is broken in master branch (it will be fixes soon, the PR to achieve that already exists: https://github.com/telefonicaid/fiware-orion/pull/3230).

It is not ok that functional test to fail, but looking to the log, it seems to be a glitch and I don't pay too much importance. We'll fix in a separate PR if the problem stills.

Please @caa06d9c have a look and review.

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -1,12 +1,12 @@
 ## Overview
 Travis is enabled in this repository so each pull request is checked before being allowed to merge.
-The system is based on fiware/orion-ci:rpm7 which is built from master branch each time a new PR lands in master,
+The system is based on `fiware/orion-ci:rpm7` which is built from master branch each time a new PR lands in master,
 providing a clean environment with all build dependencies onboard. The Dockerfile used to build this docker is available
-in the ci/rpm7 directory.
+in the `ci/rpm7` directory.
 
-Note that fiware/orion-ci:rpm7 is *not* rebuilt due to changes in the PR branch under test. Thus, if you are developing
+Note that `fiware/orion-ci:rpm7` is *not* rebuilt due to changes in the PR branch under test. Thus, if you are developing
 a functionality that requires a new library or base system you need to do *first* a PR adding such library or base system
-to ci/rpm7/build-dep.sh and/or Dockerfile. Once that PR gets merged into master and fiware/orion-ci:rpm7 gets rebuild 
+to `ci/rpm7/build-dep.sh` and/or `Dockerfile`. Once that PR gets merged into master and `fiware/orion-ci:rpm7` gets rebuild 
 (checking progress in Dockerhub at: https://hub.docker.com/r/fiware/orion-ci/builds) your PR branch with the new 
 functionality is ready to be tested with travis. 
 
@@ -24,11 +24,11 @@ File compliance, payload and style checks are combined in one 'compliance' test.
 
 ## Changes in tests
 
-There is an special function in build.sh script named `_fix_tests()` which purpose is to do some on-the-fly adaptations
+There is an special function in `build.sh` script named `_fix_tests()` which purpose is to do some on-the-fly adaptations
 in order to make functional test to work under travis. In particular:
 
 * Test under `3000_allow_creation_transient_entities` reduce internal wait delay from 60 to 3 seconds. In addition, MongoDB
   TTL monitor thread sleep interval is changed to 3 seconds by configuration. This is needed to reduce the testing time, so
   it can fix in the 50 minutes hard limit used by travis.
 
-The work done by `_fix_tests` is rolled-back by the `_unfix_tests()` function.
+The work done by `_fix_tests()` is rolled-back by the `_unfix_tests()` function.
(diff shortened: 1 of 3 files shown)


ci/README.md
```diff
@@ -1,9 +1,15 @@
 ## Overview
 Travis is enabled in this repository so each pull request is checked before being allowed to merge.
-The system is based on fiware/orion-ci:rpm7  which is built from master branch each time a new PR lands in master,
+The system is based on fiware/orion-ci:rpm7 which is built from master branch each time a new PR lands in master,
 providing a clean environment with all build dependencies onboard. The Dockerfile used to build this docker is available
 in the ci/rpm7 directory.
 
+Note that fiware/orion-ci:rpm7 is *not* rebuilt due to changes in the PR branch under test. Thus, if you are developing
```
fgalan: Please  @caa06d9c ensure that what I say in this paragraph is fully correct (or if some modification needs to be done in order to make it 100% accurate).

ci/rpm7/build.sh
```diff
@@ -291,22 +290,24 @@ if [ -n "${test}" ] && [ "${stage}" = "compliance" ]; then
     if [ $? -ne 0 ]; then status=false; fi
 
     make style
-    if [ "$(cat LINT | grep 'Total errors found' | awk '{ print $4 }')" != "0" ]; then  status=false; fi
+    if [ $? -ne 0 ]; then status=false; else status=true; fi
```
caa06d9c: Not correct, make style always return 0, that is why makefile with targets (style) replaced the original one. Should be rollback or original makefile and backend function should be changed.

fgalan:
In fact, a second round of test (for the cdff0c7 commits) confirms that functional teste was a glitch:

imagen

ci/rpm7/build.sh
fgalan: You can do the following test:

```
$ git checkout hardening/just_playing_with_travis
$ make style
./scripts/style_check_in_makefile.sh
style_check.sh: [src/lib/serviceRoutinesV2]: found no mayor errors (4 minor errors) in 5796 lines of source code in 80 files (.27% style-guide-incompatibilities)
style_check.sh: [src/lib/logSummary]: found no mayor errors (0 minor errors) in 229 lines of source code in 2 files (0% style-guide-incompatibilities)
style_check.sh: [src/lib/jsonParseV2]: found no mayor errors (3 minor errors) in 6104 lines of source code in 46 files (.04% style-guide-incompatibilities)
style_check.sh: [src/lib/apiTypesV2]: found no mayor errors (11 minor errors) in 2571 lines of source code in 24 files (.42% style-guide-incompatibilities)
style_check.sh: found 136 errors (1 mayor, 134 minor errors) in 19497 lines of source code in 66 files (.69% style-guide-incompatibilities)
Lint Errors:

0043 errors of category 'Lines should be <= 120 characters long'
0001 errors of category 'Line ends in whitespace'
0091 errors of category 'Use int16/int64/etc'
....
makefile:262: recipe for target 'style' failed
make: *** [style] Error 1
fermin@neodeb:~/src/fiware-orion$ echo $?
2
```

As you can see `make style` return not always return 0. In this case (a fail) it return 2.

Thus, I'm afraid I'm not getting your point sorry... but the fact is that CI needs fixing with regards with compliance job. Previous to the change in this PR, travis marks as success the compliance job and this is not correct, as this job should be marked as fail (as there is an style fail in the code). If my fix is not correct, please propose an alternative fix.

(In fact, my original implementation wasn't fully correct, as a success style check may shadow possible fails in the previous stages, e.g. file compliance. The new commit db824f69cc68ad15519a02326fd68f40ef94e3cc should fix this).

fgalan:
> It wasn't an error. Exit function was turned off because compliance test was always failed, so to test other functionality the only message was on. And right now test fails, so I don't think that it is a good idea to exit on the error during the compliance test.

To check if I'm understanding you correctly... you mean that in the case of style fail in compliance test travis should not report a fail. Is my interpretation correect?

ci/README.md
fgalan: NTC

caa06d9c:
While the test fails in 100% I think yes, because unit and functional is more important. In case source code will be fixed, we can enable reporting.

fgalan:
You can assume the source code will be fixed once PR #3230 gets merged. So, I propose the following:

1. Merging the current PR, in which compliance step is failing (due to known reasons)
2. Upgrade PR #3230 merging from master, and check that compliance test goes back to success
3. Merge PR #3230 

Makes sense to you @caa06d9c ?

caa06d9c:
Ok) Let it be.

fgalan:
Ok. Great :)

To follow the "protocol" please provide LGTM on this PR (in no further feedback is pending)

caa06d9c:
this is my first
LGTM

PR #3246, opened 2018-07-16
add cyrus-sasl-devel to build dependencies

mathi123:


mathi123:
Travis tests where still running, but OK 👍

PR #3248, opened 2018-07-16
mongo 3.6 compatibility

mathi123:


mathi123:
Currently not possible without mongo driver update!

PR #3250, opened 2018-07-18
FIX MongoDB version in documentation

fgalan:
Orion development started time  ago, when 2.4 was the official MongoDB version. Along the time, we have modified the requirement to be inclusive with new versions, but this has lead long the year to a confusing requeriment, specially when many of such versions have reached now end of support or are close to it (see https://www.mongodb.com/support-policy).

This PR aims to simplify things, removing all versions prior to 3.4. Note also this is aligned with what we use in docker/docker-compose.yml.

Thanks to @mathi123 , whose feedback has raise this documentation flaw :)


doc/manuals/admin/install.md
```diff
@@ -71,16 +70,6 @@ Upgrade procedure depends on whether the *upgrade path* (i.e. from the installed
 * Upgrading MongoDB version
 * Migrating the data stored in DB (due to a change in the DB data model).
 
-### Upgrading MongoDB version
```
fgalan: This was here from the very beginning but actually is out the scope of Orion documentation... MongoDB official documentation described this information in a much more comprehensive a precise way.

NTC (only informative)

Issue #3070, opened 2018-01-10
MongoDB 3.6 support


fgalan:
Most of the functional test works with MongoDB 3.6, but the following ones are failling:

```
----------- Failing tests ------------------
  o  0117_convop_using_standard_ops/getAttributesForEntityType.test
  o  0117_convop_using_standard_ops/getEntityTypes.test
  o  0519_ngsi10_entity_types_listing/contextEntityTypeAttributes.test
  o  0519_ngsi10_entity_types_listing/contextEntityTypes.test
  o  0676_servicepath_context_types/servicepath_context_types.test
  o  0676_servicepath_context_types/servicepath_default_context_types.test
  o  0676_servicepath_context_types/servicepath_sametype.test
  o  0676_servicepath_context_types/servicepath_specific_sametype.test
  o  0686_contextTypes_with_no_attributes/contextTypes_with_no_attributes.test
  o  0740_crash_with_json_object_for_attributes/json_object_format.test
  o  1027_get_all_entity_types/get_all_entity_types.test
  o  1028_get_entity_type/get_entity_type.test
  o  1170_get_types_values_option/get_types_values_option.test
  o  1170_get_types_values_option/get_types_values_option_with_empty_type_entities.test
  o  1170_get_types_values_option/get_types_values_option_with_null_and_empty_types.test
  o  1170_get_types_values_option/get_types_values_option_with_null_type_entities.test
  o  1367_time_measures/time_measure_stats.test
  o  2073_uri_param_to_disable_type_detail/uri_param_to_disable_type_detail.test
  o  2115_wrong_count_no_entity_types/wrong_count_no_entity_types.test
```

My guess is that something has changed at DB level that makes some usages incopatible. Log traces may help to find the causes. For instance, the log trace for 0117_convop_using_standard_ops/getAttributesForEntityType.test shows the following message:

```
time=2018-01-09T16:42:19.317Z | lvl=ERROR | corr=0edb97fc-f55c-11e7-be7c-000c29173617 | trans=1515516138-264-00000000001 | 
from=0.0.0.0 | srv=<default> | subsrv=<default> | comp=Orion | op=safeMongo.cpp[315]:getField | 
msg=Runtime Error (field 'result' is missing in BSONObj <{ ok: 0.0, errmsg: "The 'cursor' option is required, except for 
aggregate with the explain argument", code: 9, codeName: "FailedToParse" }> from caller mongoAttributesForEntityType:624)
```

[...]

PR #1029, opened 2015-06-24
FIX Collapse standard operations. Tweak License statements

jmcanterafonseca:

PR #3251, opened 2018-07-19
Add cursor option to aggregation queries

fgalan:
Aggregation queries without cursor option are deprecated in MongoDB 3.4 and forbidden in MongoDB 3.6 (see #3070 for details). This PR fixes that.


mathi123:
:+1: looks good! next up: running tests with Mongo3.6 db.

fgalan:
@mathi123 I don't have a suitable testbed for MongoDB 3.6 right now... could you do the test of this PRs branch using MongoDB 3.6 as database, pls? (no worries if you cannot, the only thing is that we will take more time to ensure if it works also for MongoDB 3.6)

mathi123:
Yes I will set it up this weekend, but it will require some time!

mathi123:
@fgalan I had to merge the master into this branch before being able to build it. The --ssl configuration was causing issues. Can you merge master and push?

fgalan:
@mathi123 done!

mathi123:
@fgalan Could this be related to #3258? Could you merge master again?

fgalan:
@mathi123 of course. Done!

mathi123:
0876 tests are fixed. Other test fails now:

> Orion Functional Test Log File:
================================================================================
 Orion Functional tests starting
 Mon Aug 27 05:31:04 CEST 2018
 -----  Input and Output Format Test  -----
 (0000_content_related_headers/in_out_formats.test) output not as expected
 VALIDATION ERROR: input line:
    Content-Length: 120
 does not match ref line:
    Content-Length: 119
 -----  Missing Content-Type header  -----
 (0000_content_related_headers/missing_content_type_header.test) output not as expected
 VALIDATION ERROR: input line:
    Content-Length: 161
 does not match ref line:
    Content-Length: 160
 -----  failure stats in db and notifications  -----
 (1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications_stop_start_accumulator.test) output not as expected

mathi123:
More info about failing test:

some errors now return orionError instead of errorCode,
resulting in 1 less character and a regex mismatch.

2018-08-27 20_17_58- untitled-4 - digipolis-gip - visual studio code

fgalan:
Fails in the two above thest could be due to recent changes in PR https://github.com/telefonicaid/fiware-orion/pull/3261. However, it is a bit weird... as that PR also adjusted .test concecuently and, moreover, travis reports an ok status in this PR.

mathi123:
Would you consider this blocking for this PR?

fgalan:
> Would you consider this blocking for this PR?

No, it is not blocking.

However, before merging this PR I'd like to test a bit more. We have changed from a give-me-all-in-a-shoot way to a cursor-based way and I wonder what would happen with a large number of types to be retrieved. A .test should be developed based in that case so we can check if all the expected types are returned or if the result is "capped" and the implementation needs to be improved.

Looking to https://docs.mongodb.com/v3.6/reference/command/aggregate/ they mention `To indicate a cursor with the default batch size, specify cursor: {}.` but there is no clue about the value of the default batch size...

mathi123:
Hmm OK, 
Any idea on the timing of that? I have created my own docker image based of this branch, but of course we would prefer the official Orion image.

mathi123:
@fgalan I have created such a test here: #3273.
It creates 200 entity types, and then gets them back. All seems to work OK.

All tests work now on mongo 3.6, including the ones who failed before!

fgalan:
I'm looking this into detail after merging the .test in PR #3273.

It seems that the first batch is returning 101 results, not 200:

```
$cat test/functionalTest/cases/1029_get_many_entity_types/get_many_entity_types.test | grep 'type": "Room' | wc -l
101
```

Detail (same command, without `wc`):

```
$ cat test/functionalTest/cases/1029_get_many_entity_types/get_many_entity_types.test | grep 'type": "Room'
        "type": "Room1"
        "type": "Room10"
        "type": "Room100"
        "type": "Room101"
        "type": "Room102"
        "type": "Room103"
        "type": "Room104"
        "type": "Room105"
        "type": "Room106"
        "type": "Room107"
        "type": "Room108"
        "type": "Room109"
        "type": "Room11"
        "type": "Room110"
        "type": "Room111"
        "type": "Room112"
        "type": "Room113"
        "type": "Room114"
        "type": "Room115"
        "type": "Room116"
        "type": "Room117"
        "type": "Room118"
        "type": "Room119"
        "type": "Room12"
        "type": "Room120"
        "type": "Room121"
        "type": "Room122"
        "type": "Room123"
        "type": "Room124"
        "type": "Room125"
        "type": "Room126"
        "type": "Room127"
        "type": "Room128"
        "type": "Room129"
        "type": "Room13"
        "type": "Room130"
        "type": "Room131"
        "type": "Room132"
        "type": "Room133"
        "type": "Room134"
        "type": "Room135"
        "type": "Room136"
        "type": "Room137"
        "type": "Room138"
        "type": "Room139"
        "type": "Room14"
        "type": "Room140"
        "type": "Room141"
        "type": "Room142"
        "type": "Room143"
        "type": "Room144"
        "type": "Room145"
        "type": "Room146"
        "type": "Room147"
        "type": "Room148"
        "type": "Room149"
        "type": "Room15"
        "type": "Room150"
        "type": "Room151"
        "type": "Room152"
        "type": "Room153"
        "type": "Room154"
        "type": "Room155"
        "type": "Room156"
        "type": "Room157"
        "type": "Room158"
        "type": "Room159"
        "type": "Room16"
        "type": "Room160"
        "type": "Room161"
        "type": "Room162"
        "type": "Room163"
        "type": "Room164"
        "type": "Room165"
        "type": "Room166"
        "type": "Room167"
        "type": "Room168"
        "type": "Room169"
        "type": "Room17"
        "type": "Room170"
        "type": "Room171"
        "type": "Room172"
        "type": "Room173"
        "type": "Room174"
        "type": "Room175"
        "type": "Room176"
        "type": "Room177"
        "type": "Room178"
        "type": "Room179"
        "type": "Room18"
        "type": "Room180"
        "type": "Room181"
        "type": "Room182"
        "type": "Room183"
        "type": "Room184"
        "type": "Room185"
        "type": "Room186"
        "type": "Room187"
        "type": "Room188"
        "type": "Room189"
        "type": "Room19"
```

mathi123:
Indeed the test is not correct, the result check is wrong.

fgalan:
Maybe the following can be tested:

1. Set batch size to 1000 (the maximum hardwired limit in CB). Would it work? Or maybe mongodb hardwired limit for batch size is smaller than that?
2. Use $skip and $limit to implement pagination. Note that currently pagination was implemented in a different way:

```
   /* Another strategy to implement pagination is to use the $skip and $limit operators in the
   * aggregation framework. However, doing so, we don't know the total number of results, which can
   * be needed in the case of count request (using that approach, we need to do two queries: one to get
   * the count and other to get the actual results with $skip and $limit, in the same "transaction" to
   * avoid incoherence between both if some entity type is created or deleted in the process).
```

The downside of this is that (as described above) an additional aggregation operation is needed. However, maybe is a reasonable price to pay.

mathi123:
1. I don't think pagination with limit and skip will work 100% in his case, because you would indeed need a transaction to be sure. I haven't found any proof of transaction support in the legacy connector. 
2. I did some experiments with setting the batch size, and this does work for more than 1000 entity types aggregation.  The hard cap in Mongo seems to be on document size (16mb) rather than the document count. So that solution is also not 100% - but reasonably solid I believe.

mathi123:
I have added the batch size to #3275.

fgalan:
> I don't think pagination with limit and skip will work 100% in his case, because you would indeed need a transaction to be sure. I haven't found any proof of transaction support in the legacy connector.

MongoDB 3.6 doesn't support transactions. MongoDB 4.0 has some support, but not sure if only in regular operations(i.e. insert(), find(), etc.) or also in aggregations.

At this point, options are:

1. Not supporting MongDB 3.6, keeping the current mechanism (that gets all the results and them apply skip/limit to them, ensuring that the count is accurate)
2. Modify to code to support MongoDB 3.6, using $limit and $skip in the aggregation pipeline and a second aggregation command (when requested by user, i.e. `?options=count` in the request) to get the count, assuming that in some cases the count could not be accurate if some entity creation/deletion operation takes place in the middle.

I think option 2 is a good trade-off. We lose a bit (I think the probability of that operation in the middle is low), but gain a lot :)

> I did some experiments with setting the batch size, and this does work for more than 1000 entity types aggregation. The hard cap in Mongo seems to be on document size (16mb) rather than the document count. So that solution is also not 100% - but reasonably solid I believe.

I agree. Each result is small (a short list of attribute names, as far as I remember) so it is hardly improbable to reach the 16MB limit.

mathi123:
Fixed tests: #3309

mathi123:
Finally ready to merge :)

test/functionalTest/cases/1029_get_many_entity_types/get_many_entity_types.test
```diff
@@ -0,0 +1,2295 @@
+# Copyright 2015 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: Move this file to the proper place. The cases/1029 doesn't seem the right place (issue https://github.com/telefonicaid/fiware-orion/pull/1029 has nothing to do with this...)

(Self-note)

fgalan:
> Otherwise, with the current implementation, if we have for instance 1005 types and we try `GET /v2/types?offset=1000` we are not going to get types 1001 to 1005 (not fully sure but I think it will happen... a quick test to confirm/discard should be done).

Travis on 3aa5190 confirms the fail :(

```
1024/1047: 3055_qc_inappropriate_payload/qc_inappropriate_payload.test ..................................................  01 seconds
1025/1047: 3070_cursor_based_aggreation_queries/more_than_1000_types.test ...............................................  (ERROR 9 - .out and .regexpect differ) 
1025/1047: 3070_cursor_based_aggreation_queries/more_than_1000_types.test ............................................... (intent 2) (ERROR 9 - .out and .regexpect differ) 
1025/1047: 3070_cursor_based_aggreation_queries/more_than_1000_types.test ............................................... (intent 3) (FAIL 9 - .out and .regexpect differ) testHarness.sh/Get entity types passes the first 1000 types batch: (3070_cursor_based_aggreation_queries/more_than_1000_types.test) output not as expected
make: *** [functional] Error 9
```

Thus, the fix described in a comment above seems to be needed.

(Note that both more_than_1000_types.test and more_than_2000_types.test are working on master branch)

test/functionalTest/cases/1029_get_many_entity_types/get_many_entity_types.test
fgalan: Fixed in 663c53492c1c79a62818c2ba86e7dc6111e4cf97

fgalan:
After last modification in the aggregation logic I think the PR is completed and could be merged soon. 

Given that the aggregation commands has changed, it is wise to conduct some comparative performance testing between master and this implementation to check no penalization exists. I'll do that before merging, but in the meanwhile, it would be great if @mathi123 could have a look and provide LGMT on the PR.

Thanks @mathi123 for all your motivational feedback along the process along with you contributions as PRs! :)

test/functionalTest/cases/1170_get_types_values_option/get_types_values_option_with_null_and_empty_types.test
```diff
@@ -58,15 +58,15 @@ mongoCmd ${CB_DB_NAME} 'db.entities.insert({
                 "type": "",
                 "servicePath": "/"
         },
-        "attrNames": ["A1", "A3"],
```
fgalan: This ha no influence in the test outcome, but fix a typo (see echo line for this step)

NTC (informative)

src/lib/mongoBackend/mongoQueryTypes.cpp
```diff
@@ -174,6 +183,43 @@ static long long countEntities
...
+  std::vector<BSONElement> resultsArray = std::vector<BSONElement>();
+
+  if (result.hasField("cursor"))
+  {
+    resultsArray = getFieldF(getObjectFieldF(result, "cursor"), "firstBatch").Array();
+    return getIntFieldF(resultsArray[0].embeddedObject(), "count");
+  }
+  else
+  {
+    LM_E(("Runtime Error (ejecuting: %s, result hasn't cursor field"));
```
mathi123: Spanish typo? 😄

mathi123:
LGTM

mathi123:
Great to see this work done! Thx!

src/lib/mongoBackend/mongoQueryTypes.cpp
fgalan: Yes :)

Fixed in 97654564ec247e4e4c25e87e5e3f88a36f1690cd

fgalan:
> Given that the aggregation commands has changed, it is wise to conduct some comparative performance testing between master and this implementation to check no penalization exists. I'll do that before merging

This is the test result.

Three (service, subservices) tuples:

* (A, /sA) 13805 entities in 10 types with a small (~10) number of attributes per entity
* (B, /sA) 9968 entities in 65 types  with a small (~10) number of attributes per entity
* (C, /sA) 10303 entities in 4 types with a large (~100) number of attributes per entity

Orion 2.0.0 running in port 1026. Orion in this branch running in port 1027.

With Orion 2.0.0:

```
time curl -s -S -H 'fiware-service: A' -H 'fiware-servicepath: /sA' localhost:1026/v2/types
real	0m4.120s
user	0m0.012s
sys	0m0.000s

time curl -s -S -H 'fiware-service: A' -H 'fiware-servicepath: /sA' localhost:1026/v1/contextTypes
real	0m0.460s
user	0m0.008s
sys	0m0.000s

time curl -s -S -H 'fiware-service: B' -H 'fiware-servicepath: /sB' localhost:1026/v2/types
real	0m1.395s
user	0m0.016s
sys	0m0.000s

time curl -s -S -H 'fiware-service: B' -H 'fiware-servicepath: /sB' localhost:1026/v1/contextTypes
real	0m0.323s
user	0m0.008s
sys	0m0.004s

time curl -s -S -H 'fiware-service: C' -H 'fiware-servicepath: /sC' localhost:1026/v2/types
real	0m32.807s
user	0m0.008s
sys	0m0.008s

time curl -s -S -H 'fiware-service: C' -H 'fiware-servicepath: /sC' localhost:1026/v1/contextTypes
real	0m0.455s
user	0m0.008s
sys	0m0.008s
```

With Orion in this branch:

```
time curl -s -S -H 'fiware-service: A' -H 'fiware-servicepath: /sA' localhost:1027/v2/types
real	0m4.200s
user	0m0.012s
sys	0m0.008s

time curl -s -S -H 'fiware-service: A' -H 'fiware-servicepath: /sA' localhost:1027/v1/contextTypes
real	0m0.502s
user	0m0.008s
sys	0m0.004s

time curl -s -S -H 'fiware-service: B' -H 'fiware-servicepath: /sB' localhost:1027/v2/types
real	0m1.461s
user	0m0.012s
sys	0m0.004s


time curl -s -S -H 'fiware-service: B' -H 'fiware-servicepath: /sB' localhost:1027/v1/contextTypes
real	0m0.344s
user	0m0.016s
sys	0m0.000s


time curl -s -S -H 'fiware-service: C' -H 'fiware-servicepath: /sC' localhost:1027/v2/types
real	0m36.784s
user	0m0.008s
sys	0m0.004s

time curl -s -S -H 'fiware-service: C' -H 'fiware-servicepath: /sC' localhost:1027/v1/contextTypes
real	0m0.642s
user	0m0.012s
sys	0m0.000s
```

Times are more the same. A bit more slow in this branch Orion, but not so slow to preclude mergint this PR.

PR #3275, opened 2018-08-30
Feature/improve aggregate queries

mathi123:
@fgalan this cursor implementation is not working,

any ideas why?

### DIFF
--- a/src/lib/mongoBackend/mongoQueryTypes.cpp
+++ b/src/lib/mongoBackend/mongoQueryTypes.cpp
@@ -205,7 +205,7 @@ HttpStatusCode mongoEntityTypesValues
   /* Compose query based on this aggregation command:
    *
    * db.runCommand({aggregate: "entities",
-   *                cursor: {},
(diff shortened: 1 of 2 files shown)


mathi123:
To be more specific: the message 'getting with cursor' is always printed,

But the while loop is never entered. (so I always get 0 results).

Any suggestions?

fgalan:
The following is being used:

```
collectionQuery(connection, getEntitiesCollectionName(tenant), cmd, &cursor, &err)
```

Thus, you are using `cmd` as query on the entities collection, not as aggregation command. In other words is like you were running in the mongo shell a query like this:

```
db.entities.find({aggregate: ..., cursor:  ..., pipeline: ... })
```

so 0 documents is the expected result, as no entity document would match that (see destails of entities collection here: https://fiware-orion.readthedocs.io/en/master/admin/database_model/index.html#entities-collection).

I have some ideas that could be tested. I'll elaborate on this in the base PR, in order to have all the information there.

PR #3279, opened 2018-08-31
Clean NGSIv1 content from documentation (including deprecation statementa bout NGSIv1)

fgalan:
Aligned to the release of the final version of NGSIv2 (to be done soon), we declare NGSIv1 as deprecated API. According to "deprecation rules", NGSIv1 is being removed from user documentation.

It could seem that content is removed from Orion manual without a NGSIv2 counterpart (for instance, documentation related with filtering and geo-capabilities). However, note that in the case of NGSIv2 the specification itself (http://telefonicaid.github.io/fiware-orion/api/v2/stable) which is a much better approach (Orion manual shoudn't include content already in NGSIv2 spec).

To be merged after NGSIv2 final releasing (targeted for September 15th).


fgalan:
CC: @fisuda 

This PR modifieds .md files in the user and admin manuals.

jmcanterafonseca:
LGTM, thanks

Issue #1142, opened 2015-08-10
updateContext "replace" and geolocalization


fgalan:
It is not clear if the implementation of the "replace" functionality at mongoBackend (done in PR #1132) will work right in the case of using location attributes. Four situations needs to be tested:
- Existing location attribute L1, replace including L1
- Existing location attribute L1, replace no including any location attribute
- Existing location attribute L1, replace including location attribute but different (L2)
- Not existing locadtion attribute, replace including location attribute

The right strategy to test this (either unit test for mongoBackend, end2end test harness or a combination of both) will be decided in the moment of addressing this issue.

The best moment to implement this issue is once the new geolocation in NGSIv2 gets implemented (i.e. "geo:point" stuff).

**Effort: 3 man day**

arigliano:
Hello, I would like to work on this issue. Can you assign it to me? (due to possible relationship with #3167 )

fgalan:
Done. Thanks!

arigliano:
The problem is where building the final append query in case of replace action. As done for the expirationDate feature, the logic has to check if there is a new location, or if the old one must be deleted, and then append it either to set or unset BSONObjBuilders. In order to avoid code complication (too many if), in the updateEntity() function, a good approach could be using a BSONObjBuilder (replaceSet) instead BSON stream form, something like the following:

```
  if (action == ActionTypeReplace)
  {
    // toSet: { A1: { ... }, A2: { ... } }
    BSONObjBuilder replaceSet;
    int            now = getCurrentTime();

    // This avoids strange behavior like as for the location, as reported in the #1142 issue
    // In order to enable easy append management of fields (e.g. location, dateExpiration),
    // it could be better to use a BSONObjBuilder instead the BSON stream macro below.
//    if (dateExpirationInPayload)
//    {
//      updatedEntity.append("$set", BSON(ENT_ATTRS                   << toSetObj <<
//                                        ENT_ATTRNAMES               << toPushArr <<
//                                        ENT_MODIFICATION_DATE       << now <<
//                                        ENT_EXPIRATION              << currentDateExpiration <<
//                                        ENT_LAST_CORRELATOR         << fiwareCorrelator));
//    }
//    else
//    {
//      updatedEntity.append("$set", BSON(ENT_ATTRS                   << toSetObj <<
//                                        ENT_ATTRNAMES               << toPushArr <<
//                                        ENT_MODIFICATION_DATE       << now <<
[...]

Issue #3167, opened 2018-04-26
Inconsistent entity at DB: geo-located attribute but not location field


fgalan:
Sometimes CB reports error like this one:

```
"statusCode": {
                "code": "472",
                "reasonPhrase": "request parameter is invalid/not allowed",
                "details": "action: APPEND - entity: [E1, T] - offending attribute: position - attempt to define a geo location attribute [position] when another one has been previously defined []"
```

This is the reference request that were used to raise that error:

```
{
    "url": "http://localhost:1026/v1/updateContext",
    "method": "POST",
    "headers": {
        "fiware-service": "sc_smart_region_andalucia",
        "fiware-servicepath": "/and_sr_motril",
        "X-Auth-Token": null
    },
    "json": {
        "contextElements": [
            {
                "type": "T",
                "isPattern": "false",
                "id": "E1",
                "attributes": [
                    {
                        "name": "measure1",
                        "type": "string",
                        "value": "44",
                        "metadatas": [
                            {
                                "name": "TimeInstant",
                                "type": "ISO8601",
                                "value": "2018-03-20T16:58:31.421Z"
                            }
                        ]
                    },
                    {
                        "name": "measure2",
                        "type": "string",
                        "value": "21348",
                        "metadatas": [
                            {
                                "name": "TimeInstant",
                                "type": "ISO8601",
                                "value": "2018-03-20T16:58:31.421Z"
                            }
                        ]
                    },
                    {
                        "value": "32.7121, -2.39172",
                        "type": "geo:point",
                        "name": "position",
                        "metadatas": [
                            {
                                "name": "TimeInstant",
                                "type": "ISO8601",
                                "value": "2018-03-20T16:58:31.421Z"
                            }
                        ]
                    },
                    {
                        "name": "TimeInstant",
                        "type": "ISO8601",
                        "value": "2018-03-20T16:58:31.421Z"
                    }
                ]
            }
        ],
        "updateAction": "APPEND"
    }
}
```

The error message corresponds to a checking that CB does and it is correct in normal circustance. But what is abnormal (and buggy) in this case is the `...defined []` part, without specifying an actual attribute.

Looking into DB we have the following:

```
﻿{
    "_id" : {
        "id" : "E1",
        "type" : "T",
        "servicePath" : "/subserv1"
    },
    "attrNames" : [
        "TimeInstant",
        "position",
        "measure1",
        "measure2",
        "measure3"
    ],
    "attrs" : {
[...]

PR #3283, opened 2018-09-03
1142 update context replace and geolocalization

arigliano:
This PR fixes issue #1142 
The problem was in the correct building of Mongo update query, in particular when appending toSet and ToUnset Builders. In the replace case, now it also uses a cleaner BSONObjectBuilder instead of the BSON Stream syntax. 

Unit tests for the following cases was added, (as stated in #1142):
 - Existing location attribute L1, replace including L1
 - Existing location attribute L1, replace no including any location attribute
 - Not existing locadtion attribute, replace including location attribute

The following one should be discussed:

- Existing location attribute L1, replace including location attribute but different (L2)
It should be first discussed/explained, as I wrote in the last part of previous comment.


fgalan:
Thank you very much for you contribution!

I'm currently a bit busy but when I have a change I'll have a look and provide feedback in the PR and associated issue.

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
```diff
@@ -493,6 +493,313 @@ TEST(mongoUpdateContextGeoRequest, updateLocAttribute)
...
+
+    ent = connection->findOne(ENTITIES_COLL, BSON("_id.id" << "E1" << "_id.type" << "T1"));
+    EXPECT_STREQ("E1", C_STR_FIELD(ent.getObjectField("_id"), "id"));
+    EXPECT_STREQ("T1", C_STR_FIELD(ent.getObjectField("_id"), "type"));
+    EXPECT_FALSE(ent.hasField("creDate"));
+    EXPECT_FALSE(ent.hasField("creDate"));
+    attrs = ent.getField("attrs").embeddedObject();
+    attrNames = ent.getField("attrNames").Array();
+    ASSERT_EQ(2, attrs.nFields());
+    ASSERT_EQ(2, attrNames.size());
```
fgalan: Given the request used to replace contains only an attribute ("A2", "geo:point", "2, -4") both attrs.nField() and attrNames.size() should be 1, isn't it?

fgalan: I have just seen ("E2", "T2", "false") in the request... I'm looking to the wrong place. Sorry the noise :)

NTC

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
```diff
@@ -493,6 +493,313 @@ TEST(mongoUpdateContextGeoRequest, updateLocAttribute)
     utExit();
 }
 
+/* ****************************************************************************
```
fgalan: Unit test are fine but it would be great if you could also add a .test covering the same cases :) What do you think?

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1,4 +1,4 @@
-﻿
```
fgalan: Should this PR add an entry to CHANGES_NEXT_RELEASE file? Has this PR fixes something previously broken or it is only a refactor of things that already were working? (in order to check, just run the new unit test with the old code in master). In the former case, an entry about the fix should be added.

emastrosimone: Yes, the fix should be added to the file.

test/unittests/mongoBackend/mongoUpdateContextGeo_test.cpp
arigliano: Fixed in 40da980

src/lib/mongoBackend/MongoCommonUpdate.cpp
arigliano: Fixed in 40da980

arigliano:
Unit and functional test (including new ones), performed:

- Unit tests **OK**:
```
[----------] Global test environment tear-down
[==========] 866 tests from 131 test cases ran. (28303 ms total)
[  PASSED  ] 866 tests.
```
- **2** Functional tests **FAILED**:

```
-----  failure stats in db and notifications  -----
(1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications.test) output not as expected

-----  Test of the fix for the problem of 404 Not Found responses when updating attribute using NGSIv2 and CB forwards it  -----
(2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test) output not as expected
```
I'm investigating.

arigliano:
@fgalan , It seems that also in the master these 2 tests are failing, can you confirm it? 

- 1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications.test
- 2871_fwd_update_ok_but_not_found_response/fwd_update_ok_but_not_found_response.test

fgalan:
I haven't found problems with ftest in master recently, although some of them glitch some times depending on the system.

Are these tests uploaded to the branch of the PR? In that caase, I think is find, given that travis shows ok status at commit 9ca6f79

arigliano:
So should be fine, as these are old tests and now travis shows ok.

arigliano:
- FIXME comment should be removed, I will fix it.

- For CNR line: #3167 is the issue related to the #1142 one. So both numbers in parentheses are issues. I think that this PR fixes both, I don't know if further tests are needed to consider also the issue #3167 closed. In that case, we could for now include in the CNR line only the #1142.

fgalan:
> For CNR line: #3167 is the issue related to the #1142 one. So both numbers in parentheses are issues. I think that this PR fixes both, I don't know if further tests are needed to consider also the issue #3167 closed. In that case, we could for now include in the CNR line only the #1142.

Ah... ok. I missed that, sorry. Thus, the CNR line is ok. I'll have a look to #3167 after it merge and try to model the case with a .test. Thanks!

arigliano:
> > For CNR line: #3167 is the issue related to the #1142 one. So both numbers in parentheses are issues. I think that this PR fixes both, I don't know if further tests are needed to consider also the issue #3167 closed. In that case, we could for now include in the CNR line only the #1142.
> 
> Ah... ok. I missed that, sorry. Thus, the CNR line is ok. I'll have a look to #3167 after it merge and try to model the case with a .test. Thanks!

FIXME comment removed in 09763cc

fgalan:
LGTM. Thanks!

(Not waiting for travis this time, as the last commit includes only changes in commented code)

PR #3320, opened 2018-09-27
Adjust documentation note related with issue #3162

fgalan:
Based on [new findings on the issue](https://github.com/telefonicaid/fiware-orion/issues/3162#issuecomment-425160902) it seems documentation was wrong an need to be changed.

Please, @mainakae and @versatildefuy have a look and tell if you find it ok. Thx! 

CC: @fisuda in the case this PR gets merged at the end, I think it would be need to be synced in Japanese translation.


fisuda:
Noted with thanks

versatildefuy:
From my side, I would precise that the issue with forwarded requests appears only in v2 queries. 
I haven't read the entire documentation, so if it's specified above the documentation is perfectly well.
I hope that helps :)
Thank you very much for your efforts.

fgalan:
Issue #3162 was fixed at the end, so this PR is no longer valid. Closing.

Thus @fisuda no adaptation in the Japanese translation is needed due to this. Thanks!

PR #3321, opened 2018-09-28
FIX version in devel manual (+ some others post 2.0.0 releasing adjusts)

fgalan:
CC: @fisuda I understand this merge needs to be synced in Japanese translation

PR #3322, opened 2018-09-28
(JP) FIX version in devel manual (#3321)

fisuda:
@fgalan 
Please let me know changes in the .pptx documents on slideshare.


fisuda:
Thanks.

PR #3324, opened 2018-10-01
NGSIv1 regs isDomain/metadta removal and render() to toJsonv1() renaming

fgalan:
This PR is related with #1298 and includes two pieces of work:

* NGSIv1 regs isDomain/metadta removal
* render() to toJsonv1() renaming

I'd recommend to review commit per commit. Only the first two ones are interesing, the others are rutinary adjustment in ftest and utest.


fgalan:
CC: @fisuda 

There are some changes in .md files in this PR that I think would need to be synced in the Japanese translation.

fisuda:
Noted with thanks.

src/lib/ngsi/ContextAttribute.cpp
```diff
@@ -601,14 +601,14 @@ std::string ContextAttribute::getLocation(ApiVersion apiVersion) const
 
 /* ****************************************************************************
 *
-* renderAsJsonObject - 
+* toJsonV1AsJsonObject -
 */
-std::string ContextAttribute::renderAsJsonObject
+std::string ContextAttribute::toJsonV1AsJsonObject
```
kzangeli: As json is now part of the prefix, it could be dropped in the end of the name:

`toJsonV1AsObject`

Just a suggestion

fgalan: Fixed in 7e25b65c1f5bf757b5e8dda2b346dabb1a657d53

PR #3329, opened 2018-10-03
Add JsonVectorHelper class and some related refactors

fgalan:
This PR is part of issue #1298 refactors and contains the following items:

1. Add a new class JsonVectorHelper, similar to current JsonHelper for objects
2. Ensure all NGSIv2 rendering logic (i.e. `toJson()` methods) are using the helpers. Removal of dead code and some fixes. After this change, no string concatenation is implemented in toJson() methods directly, all that is encapusulated in Json*Helper classes. Detail of the fixed methods (as far as I remember):
   * NotifyContextRequeset::toJson()
   * ContextElementResponseVector::toJson()
   * EntityID::toJson()
   * StatusCode::toJson()
   * SubscribeError::toJson()
   * SubscribeContextResponse::toJson()
   * OrionError::toJson()
3. Avoid code duplication unifying toJsonString() and jsonIvalidCharsTransformation() code
4. Avoid + for std::string in Json*Helper clases (and some other placers), according to the recommendations of https://www.thecodingdelight.com/string-cplusplus/

I'd try to coduct a test with Apache Benchmark in order to see if item 4 provides some improvement, but PR review can be done in the meanwhile.

### DIFF
--- a/src/lib/apiTypesV2/Registration.cpp
+++ b/src/lib/apiTypesV2/Registration.cpp
@@ -67,7 +67,7 @@ Registration::~Registration()
 */
 std::string Registration::toJson(void)
 {
-  JsonHelper jh;
+  JsonObjectHelper jh;
 
   jh.addString("id", id);
 
@@ -101,7 +101,7 @@ std::string Registration::toJson(void)
 */
 std::string DataProvided::toJson(void)
 {
-  JsonHelper jh;
+  JsonObjectHelper jh;
 
   jh.addRaw("entities", vectorToJson(entities));
   jh.addRaw("attrs", vectorToJson(attributes));
@@ -117,10 +117,10 @@ std::string DataProvided::toJson(void)
 */
 std::string Provider::toJson(void)
 {
-  JsonHelper jhUrl;
+  JsonObjectHelper jhUrl;
   jhUrl.addString("url", http.url);
 
-  JsonHelper   jh;
+  JsonObjectHelper   jh;
   jh.addRaw("http", jhUrl.str());
   jh.addString("supportedForwardingMode", forwardingModeToString(supportedForwardingMode));
   jh.addBool("legacyForwarding", legacyForwardingMode? "true" : "false");
@@ -136,7 +136,7 @@ std::string Provider::toJson(void)
 */
 std::string ForwardingInformation::toJson()
 {
-  JsonHelper  jh;
+  JsonObjectHelper  jh;
 
   jh.addNumber("timesSent", timesSent);
 
--- a/src/lib/apiTypesV2/Registration.cpp
+++ b/src/lib/apiTypesV2/Registration.cpp
@@ -117,10 +117,11 @@ std::string DataProvided::toJson(void)
 */
 std::string Provider::toJson(void)
 {
-  JsonHelper   jh;
-  std::string  urlAsJson = "{\"url\": \"" + http.url + "\"}";
+  JsonHelper jhUrl;
+  jhUrl.addString("url", http.url);
 
-  jh.addRaw("http", urlAsJson);
+  JsonHelper   jh;
+  jh.addRaw("http", jhUrl.str());
   jh.addString("supportedForwardingMode", forwardingModeToString(supportedForwardingMode));
   jh.addBool("legacyForwarding", legacyForwardingMode? "true" : "false");
 
--- a/src/lib/common/JsonHelper.h
+++ b/src/lib/common/JsonHelper.h
@@ -32,11 +32,10 @@
 #include <map>
 
 
-// FIXME P10: should be renamed to JsonObjectHelper
-class JsonHelper
+class JsonObjectHelper
 {
 public:
-  JsonHelper();
+  JsonObjectHelper();
 
   void        addString(const std::string& key, const std::string& value);
   void        addRaw(const std::string& key, const std::string& value);
--- a/src/lib/common/JsonHelper.h
+++ b/src/lib/common/JsonHelper.h
@@ -31,6 +31,8 @@
 #include <vector>
 #include <map>
 
+
+// FIXME P10: should be renamed to JsonObjectHelper
 class JsonHelper
 {
 public:
@@ -42,6 +44,26 @@ public:
   void        addNumber(const std::string& key, double value);
   void        addDate(const std::string& key, long long timestamp);
   void        addBool(const std::string& key, bool b);
+
+  std::string str();
+
+private:
+ std::string  ss;
+ bool         empty;
+};
+
+
+class JsonVectorHelper
+{
+public:
+  JsonVectorHelper();
+
+  void        addString(const std::string& value);
+  void        addRaw(const std::string& value);
+  void        addNumber(long long value);
+  void        addNumber(double value);
+  void        addDate(long long timestamp);
+  void        addBool(bool b);
 
   std::string str();
 
--- a/src/lib/common/JsonHelper.h
+++ b/src/lib/common/JsonHelper.h
@@ -98,10 +98,12 @@ std::string vectorToJson(std::vector<T> &list)
 
   std::string ss;
 
-  ss += '[' + list[0].toJson();
+  ss += '[';
+  ss += list[0].toJson();
   for (size_type i = 1; i != list.size(); ++i)
   {
-    ss += ',' + list[i].toJson();
+    ss += ',';
+    ss += list[i].toJson();
   }
   ss += ']';
   return ss;
--- a/src/lib/ngsi/StatusCode.cpp
+++ b/src/lib/ngsi/StatusCode.cpp
@@ -33,6 +33,7 @@
 #include "common/globals.h"
 #include "common/string.h"
 #include "common/tag.h"
+#include "common/JsonHelper.h"
 #include "common/limits.h"
 #include "ngsi/Request.h"
 #include "ngsi/StatusCode.h"
@@ -85,7 +86,7 @@ StatusCode::StatusCode(HttpStatusCode _code, const std::string& _details, const 
 
 /* ****************************************************************************
 *
-* StatusCode::render -
+* StatusCode::toJsonV1 -
 */
 std::string StatusCode::toJsonV1(bool comma, bool showKey)
 {
@@ -127,43 +128,12 @@ std::string StatusCode::toJsonV1(bool comma, bool showKey)
 *
 * StatusCode::toJson -
 *
-* For version 2 of the API, the unnecessary 'reasonPhrase' is removed.
+* For version 2 of the API, based on OrionError.
 */
-std::string StatusCode::toJson(bool isLastElement)
+std::string StatusCode::toJson(void)
 {
-  std::string  out  = "";
-
-  if (strstr(details.c_str(), "\"") != NULL)
-  {
-    int    len = details.length() * 2;
-    char*  s    = (char*) calloc(1, len + 1);
-
-    strReplace(s, len, details.c_str(), "\"", "\\\"");
-    details = s;
-    free(s);
-  }
-
-  char codeV[STRING_SIZE_FOR_INT];
-
-  snprintf(codeV, sizeof(codeV), "%d", code);
-
-  out += "{";
-
-  out += std::string("\"code\":\"") + codeV + "\"";
-  
-  if (details != "")
-  {
-    out += ",\"details\":\"" + details + "\"";
-  }
-
-  out += "}";
-
-  if (!isLastElement)
-  {
-    out += ",";
-  }
-
-  return out;
+  OrionError oe(code, details, reasonPhrase);
+  return oe.smartRender(V2);
 }
 
 
--- a/src/lib/ngsi/SubscribeError.cpp
+++ b/src/lib/ngsi/SubscribeError.cpp
@@ -25,6 +25,7 @@
 #include <string>
 
 #include "common/tag.h"
+#include "common/JsonHelper.h"
 #include "ngsi/StatusCode.h"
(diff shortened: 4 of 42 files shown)


src/lib/ngsi/StatusCode.cpp
```diff
@@ -127,43 +128,12 @@ std::string StatusCode::toJsonV1(bool comma, bool showKey)
 *
 * StatusCode::toJson -
 *
-* For version 2 of the API, the unnecessary 'reasonPhrase' is removed.
+* For version 2 of the API, based on OrionError.
 */
-std::string StatusCode::toJson(bool isLastElement)
+std::string StatusCode::toJson(void)
 {
-  std::string  out  = "";
```
fgalan: Lot of complexity removed here...

NTC

src/lib/ngsi/SubscribeError.cpp
```diff
@@ -46,37 +47,14 @@ SubscribeError::SubscribeError()
...
-  std::string  out;
+  JsonHelper jh;
 
-  out += JSON_VALUE("error", errorCode.reasonPhrase.c_str());
-  out += ",";
-  out += JSON_VALUE("description", errorCode.details.c_str());
+  jh.addString("error", errorCode.reasonPhrase);
+  jh.addString("description", errorCode.details);
 
-
```
fgalan: Lot of complexity removed here...

NTC

src/lib/ngsi/SubscriptionId.cpp
```diff
@@ -119,40 +119,6 @@ const char* SubscriptionId::c_str(void) const
 
 
 
-/* ****************************************************************************
-*
-* SubscriptionId::toJson -
-*/
-std::string SubscriptionId::toJson(RequestType container, bool comma)
```
fgalan: Dead code. Mabye sometime is was used, but not at the present moment.

NTC

test/functionalTest/cases/2763_postUpdateContext_error_return/postEntities_with_errors_in_service_path.test
```diff
@@ -53,14 +53,14 @@ echo
 01. Try to create an entity with more than one service path - see error
 =======================================================================
 HTTP/1.1 400 Bad Request
-Content-Length: 79
+Content-Length: 91
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
```
fgalan: It seem we were wrongly using NGSIv1-like error response for POST /v2/entity. This PR has raised this and fixed (used to happen during cleanup refactors :)

A bit surprising not to found this in more places. I think it deservers a CNR line like this:

```
- Fix: NGSIv1-like error responses were used in some NGSIv2 operations
```

test/functionalTest/cases/3004_v2_reg_create/v2_reg_create.test
```diff
@@ -97,7 +97,7 @@ Date: REGEX(.*)
 02. GET the registration using APIv2
 ====================================
 HTTP/1.1 200 OK
-Content-Length: 312
```
fgalan: Content length changes in this PR are due to httpInfo rendering has moved from `http":{"url": "http://foo.bar"}` to `http":{"url":"http://foo.bar"}` (1 less whitespace). Thus, the change in .test is fine.

NTC

src/lib/common/JsonHelper.h
```diff
@@ -31,6 +31,8 @@
 #include <vector>
 #include <map>
 
+
+// FIXME P10: should be renamed to JsonObjectHelper
```
fgalan: Two additional actions could be done:

1. Rename JsonHelper class to JsonObjectHelper class (I think this could be done in this same PR)
2. Split JsonHelper.h.cpp into JsonObjectHelper.h/cpp and JsonVectorHelper.h/cpp. I think this case is better to have them in only , as both classes are highly related and use some common logic such as toJsonString() function (which would force to have an additional couple jsonHelperCommon.h/cpp in the second case) and we have used this pattern in other places (e.g. Subscription related classes are all them in the same cpp/h module). However, I don't have an strong opinion on it so if you prefer the other way for me is ok and I can also implement it in this PR.

src/lib/ngsi/SubscribeError.cpp
```diff
@@ -46,37 +47,14 @@ SubscribeError::SubscribeError()
...
-  else if ((requestType          == SubscribeContext)           &&
-           (subscriptionId.get() != "000000000000000000000000") &&
-           (subscriptionId.get() != ""))
-  {
-    out += ",";
-    out += JSON_PROP("affectedItems") + "[" + JSON_STR(subscriptionId.toJson(requestType, true)) + "]";
-  }
-
-  return out;
+  return jh.str();
```
kzangeli: This we can do as we have deprecated API V1, I imagine ...
But, even deprecated, the broker still supports the requests, right?

src/lib/parse/CompoundValueNode.cpp
```diff
@@ -691,32 +694,28 @@ std::string CompoundValueNode::toJson(bool toplevel)
...
-      }
-      out += "]";
+      jh.addRaw(childV[ix]->toJson(false));
     }
+    out = jh.str();
     break;
 
   case orion::ValueTypeObject:
+    // In thic case we cannot use JsonHelper to build the object, as we don't have a
+    // key-value sequence to invoke addXX() method
```
kzangeli: An empty JsonHelper would render as "{}", no?

src/lib/parse/CompoundValueNode.cpp
```diff
@@ -742,7 +741,10 @@ std::string CompoundValueNode::toJson(bool toplevel)
 
   if (parentIsObject)
   {
-    return toJsonString(name) + ":" + out;
+    std::string preOut = "\"";
+    preOut += toJsonString(name);
+    preOut += "\":";
+    return preOut + out;
```
kzangeli: Or:
```
preOut += out:
return preOut;
```
If you really want to avoid ALL std::string operator+

src/lib/common/JsonHelper.h
fgalan: I can try with something as you propose in order to see how looks like. Probably the level of reutilizanton will not be too high (as the way to add a key-value, for object, is different for the way of adding just the value, for vector) but let's see. Let's also see how it looks like from the point of view of the "user code".

If at the end we go with just one class them, two comments:

1. Maybe the name JsonConteiner first better than JsonHelper, as the later is less precise ("helper" may mean many things :), "container" is more precise).
2. I think is more clear to use a meanifull type (a enum) instead of a bool to select the container type:

```
JsonConatiner jco(jc::object);
...
JsonContainer jcv(jc::vector);
```

(alternative usggestion for jh:: namespace are welcome)

src/lib/apiTypesV2/Registration.cpp
```diff
@@ -117,10 +117,11 @@ std::string DataProvided::toJson(void)
...
 std::string Provider::toJson(void)
 {
-  JsonHelper   jh;
-  std::string  urlAsJson = "{\"url\": \"" + http.url + "\"}";
+  JsonHelper jhUrl;
+  jhUrl.addString("url", http.url);
 
-  jh.addRaw("http", urlAsJson);
+  JsonHelper   jh;
+  jh.addRaw("http", jhUrl.str());
```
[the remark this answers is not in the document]
fgalan: Yes. Basically JsonHelper is an absraction for JSON objects (the same way new JsonHelper is an abstraction for JSON vectors) so the user doesn't have to deal with commas, ending items, brachets and so on and use add*() method to add things to the container.

A more powerfull approach would be to have only one constructor. We have discussed that in the past as a possible enhacement, but not for this PR :)

src/lib/common/JsonHelper.h
```diff
@@ -76,10 +98,12 @@ std::string vectorToJson(std::vector<T> &list)
 
   std::string ss;
 
-  ss += '[' + list[0].toJson();
+  ss += '[';
```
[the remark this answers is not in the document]
fgalan: I'll take into account in the context of https://github.com/telefonicaid/fiware-orion/pull/3329#discussion_r222380841

src/lib/ngsi/SubscribeError.cpp
fgalan: I think so. What I think is happening is that NGSIv2 never renders this object. NGSIv1 still rendering it using the toJsonV1() method.

NTC (I guess)

src/lib/parse/CompoundValueNode.cpp
fgalan: Yes, but I cannot use the helper here. The way the helper works it needs both a key and value to invoke addXX():

```
addXX(key, value)
```

However, in this method I don't have a key because the child toJson() method I'm using returns the key along the value.

Maybe it can be solved with a deeper refactor of CompoundValue::toJson() but that's complex so I left out of this PR.

src/lib/parse/CompoundValueNode.cpp
fgalan: Good catch!

src/lib/common/JsonHelper.h
fgalan: At the end we will keep with two classes as it is now (discussed at skype).

The JsonHelper -> JsonObjectHelper renaming is pending.

src/lib/common/JsonHelper.h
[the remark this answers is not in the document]
fgalan: At the end we will keep with two classes as it is now (discussed at skype).

So, NTC.

src/lib/parse/CompoundValueNode.cpp
fgalan: Fixed in edd01ffba0453591e991b12d5f62687890108b09

src/lib/common/JsonHelper.h
fgalan: Rename done in 33b9f6de10f03a2f775bae250b4a3bcc66670eb6

test/functionalTest/cases/2763_postUpdateContext_error_return/postEntities_with_errors_in_service_path.test
fgalan: Fixed in edd01ffba0453591e991b12d5f62687890108b09

fgalan:
> I'd try to coduct a test with Apache Benchmark in order to see if item 4 provides some improvement, but PR review can be done in the meanwhile.

Comparing results [from master](https://gist.github.com/fgalan/0ef83b52c8ac4d007448600dc5d99cb5) and from the this [PR branch](https://gist.github.com/fgalan/5248dc4e4697a8241c78bad072fd860d) there is a slight throughput increase of 2.14% or 7.41% depending the case. Not bad.

src/lib/common/JsonHelper.h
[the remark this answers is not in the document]
kzangeli: Yes, NTC

src/lib/ngsi/SubscribeError.cpp
kzangeli: If still used in NGSIv1, with this change, the NGSIv1 protocol is no longer followed.
That was my question, actually.
Not that I am very fond of NGSIv1 but, it is defined as it is ... with sub-id as mandatory field.

Or, am I wrong?

src/lib/parse/CompoundValueNode.cpp
kzangeli: My point was that with an empty object, you wouldn't call any addXX, the loop would be empty and you would end up with '{}' ...  Or not?

fgalan: Not sure what you mean... you mean to use JsonObjectHelper in the case of no child? I mean, something like:

```
    if (childV.size() == 0)
    {
      JsonObjectHelper jh;
      out = jh.str();
    }
```
    
That could be use, but I think is better the current approach. However, I don't have a strong opinion on it, if you prefer to be changed to the above.

src/lib/ngsi/SubscribeError.cpp
fgalan: NGSIv1 protocol is no longer followed... in a place where doesn't need to follow. Note that here we are inside toJson() method that is only used by NGSIv2 logic.

NGSIv1 render logic is done in toJsonV1() that, as far as I understand, it still using this stuff.

(Maybe I haven't understood the question and my answer doesn't make sense :)

src/lib/parse/CompoundValueNode.cpp
kzangeli: Getting confused ... Sorry about that.
I wan't really interested in JsonHelper here, only the special treatment of empty objects. My bad. 
Using code instead of words. More exact.

Something like this:

```
out = "{";
unsigned int ix;
for (ix = 1; ix < childV.size(); ix++)
{
  out += childV[0].toJson(false);
  if (ix != childV.size() - 1)  out += ",";
}

out += "}";
```
No special treatment for 0 children.

fgalan: Ok. I'll fix.

fgalan: Fixed in 03e44cf

src/lib/ngsi/SubscribeError.cpp
kzangeli: I think I understand now ... Finally!
It's a bad diff. The content that was for NGSIv1 seems removed but it has probably simply been moved to a toJsonV1 method.
NTC
NTC

PR #3339, opened 2018-10-10
Fix devel manual after changes implemented as part of #1298

fgalan:
Related with #1298 

Basically, remove references to ContextElement and ContextElementVector (now removed) and include information about Json*Helper classes.

CC: @fisuda , as I understand that the Japanese version of devel manual should be updated after merging this PR.


fisuda:
Noted with thanks!

Issue #2359, opened 2016-07-14
String filters and Null values


kzangeli:
The string filters don't consider attribute/metadata values of type `Null`.
For some reason this case wasn't taken into consideration when implementing the string filters.

**Effort: 1 man day**

fgalan:
To be precisse, this issue refers to the JSON datatype `null`.

fgalan:
Both for `q` and `mq`.

kzangeli:
We might have just forgotten about NULL when implementing string filters.
However, I did a quick test and it failed:

    StringFilter::mongoFilterPopulate()
    {
      ...
      case SfvtNull:
        bb.append("$exists", true).append("$in", BSON_ARRAY("null"));
        break;

That try gives the string "null" in the query: 

    attrs.A1.value.N: { $exists: true, $in: [ "null" ] } }

And trying with NULL:

      case SfvtNull:
        bb.append("$exists", true).append("$in", BSON_ARRAY(NULL));
        break;

Well, that doesn't even compile.
Not very important, leaving it for later.
To fix this, look at servicePath and copy how it's done there:

    servicePath: { $in: [ /^/.*/, null ] }

fgalan:
Not sure, but I remember something like this in a similar case... it may work.

```
    BSONObj empty;
    bb.append("$exists", true).append("$in", BSON_ARRAY(empty));
```

fgalan:
Fixed by PR #2806

PR #3340, opened 2018-10-10
FIX some issues in case/1156 .test

fgalan:
I recencly spoted some issues in cases/1156, which this PR solves.

More detail in inline comments.


test/functionalTest/cases/1156_qfilters_and_compounds/qfilters_and_compounds_equals_null.test
```diff
@@ -29,8 +29,6 @@ brokerStart CB 0-255
 
 --SHELL--
 
-#
-# FIXME P2, once 'null' values are supported (issue #2359), this test should be enabled
```
fgalan: It seems we fixed #2359 but forget to enable this .test... :)

NTC (informative)

fgalan: In addition, a pararell test for mq= has been added as new .test file

PR #3342, opened 2018-10-15
Simplification of CompoundValueNode class

fgalan:
CompoundValuesNode class is simplified removing a bunch of extra fields not actually used (maybe there were useful when this class was developed, but now it is mature and there is no need for them).

The simplification has allowed to address easyly a couple of pending FIXME P7 issues:

```
// FIXME P7: next line is counterintuitive. If the object is a vector, why [...]
// FIXME P7: why don't have a case for Object (valueType = orion::ValueTypeObject) [...]
```

### DIFF
--- a/src/lib/jsonParse/jsonParse.cpp
+++ b/src/lib/jsonParse/jsonParse.cpp
@@ -286,10 +286,9 @@ void eatCompound
       }
 
       containerP->add(orion::ValueTypeString, nodeName, nodeValue);
-      LM_T(LmtCompoundValue, ("Added string '%s' (value: '%s') under '%s'",
+      LM_T(LmtCompoundValue, ("Added string '%s' (value: '%s')",
                               nodeName.c_str(),
-                              nodeValue.c_str(),
-                              containerP->cpath()));
+                              nodeValue.c_str()));
     }
     else if ((nodeName == "") && (nodeValue == "") && (noOfChildren == 0))  // Unnamed String with EMPTY VALUE
     {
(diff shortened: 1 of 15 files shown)


test/functionalTest/cases/2457_json_tags_with_same_name/json_tags_with_same_name.test
```diff
@@ -296,13 +296,13 @@ Date: REGEX(.*)
...
 HTTP/1.1 400 Bad Request
-Content-Length: 76
+Content-Length: 75
 Content-Type: application/json
 Fiware-Correlator: REGEX([0-9a-f\-]{36})
 Date: REGEX(.*)
 
 {
-    "description": "duplicated tag-name: /a/ in path: /b/",
+    "description": "duplicated tag-name: /a/ in path: /b",
```
fgalan: The change in this .test is in a error description text, so I shouldn't be problematic from backward compatibiltiy point of view. Moreover, it can be considered a kind of bugfix, solving the unhomogenity reporting this kind of problem (step 7 and 8 where using different ways of reporting before this PR, one with final slash, the other without it).

NTC (just informative)

kzangeli: Can't say I agree ...
The '/' is used instead of quotes, as quotes would need to be URI-escaped.
The slashes mark the start and end of the strings.
`/a/ in path: /b`  ... the last string is not ended. Very confusing. Seems like a bug

src/lib/jsonParse/jsonParse.cpp
```diff
@@ -286,10 +286,9 @@ void eatCompound
       }
 
       containerP->add(orion::ValueTypeString, nodeName, nodeValue);
-      LM_T(LmtCompoundValue, ("Added string '%s' (value: '%s') under '%s'",
+      LM_T(LmtCompoundValue, ("Added string '%s' (value: '%s')",
                               nodeName.c_str(),
-                              nodeValue.c_str(),
-                              containerP->cpath()));
+                              nodeValue.c_str()));
```
kzangeli: I'm assuming the 'container' has been removed.
And yes, that was indeed useful during debugging. Probably not needed anymore.

test/functionalTest/cases/2457_json_tags_with_same_name/json_tags_with_same_name.test
fgalan: So you mean chaging here (step 7) and in step 8:

/b -> /b/

?

kzangeli:
Yes, that would solve the problem.
Also, understanding why the ending slash disappeared ... (that is what
seems like a bug to me)

test/functionalTest/cases/2457_json_tags_with_same_name/json_tags_with_same_name.test
fgalan: (Moved from general comment, @kzangeli comment)

> Yes, that would solve the problem. Also, understanding why the ending slash disappeared ... (that is what seems like a bug to me)

fgalan: Ending slash disappeared during the refactor of check(). Easy to solve.

src/lib/parse/CompoundValueNode.cpp
```diff
@@ -67,16 +61,10 @@ CompoundValueNode::CompoundValueNode():
 * CompoundValueNode - constructor for toplevel 'node'
 */
 CompoundValueNode::CompoundValueNode(orion::ValueType _type):
-  name        ("toplevel"),
+  name        (""),
```
kzangeli: As far as I know, a std::string is already initialized to the empty string, so this line can probably be removed

src/lib/parse/CompoundValueNode.cpp
```diff
@@ -270,7 +211,7 @@ std::string CompoundValueNode::finish(void)
     show("");
   }
 
-  check();  // sets 'error' for toplevel node
+  error = check("");  // sets 'error' for toplevel node
 
   return error;
```
kzangeli: Unless the variable 'error' is used elsewhere, perhaps we can remove it completely and simply `return check("");` ?
[ Even better would be to make the function return bool ... but that's probably a lot of work ]

src/lib/jsonParse/jsonParse.cpp
fgalan: Correct. 'container' has been removed.

NTC  (I guess)

kzangeli: NTC

Issue #3122, opened 2018-03-07
NGSIv2-NGSIv1 location metadata issues


fgalan:
Create an entity E1 using NGSIv2 API with metadata location (old way of marking location in NGSIv1):

```
POST /v2/entities

{
	"id": "E1",
	"type": "T",
	"A": {
		"value": "40, -3",
		"type": "whatever",
		"metadata": {
			"location": {
				"type": "string",
				"value": "WGS84"
			}
		}
	}
}
```

Create an entity E2 using NGSIv1 API with metadata location (old way of marking location in NGSIv1):

```
POST /v1/contextEntities/type/T/id/E2

{
    "attributes": [
        {
            "name": "A",
            "type": "whatever",
            "value": "39,-2",
            "metadatas": [
               {
                  "name": "location",
                  "type": "string",
                  "value": "WGS84"
               }
            ]
        }
    ]
}
```

Getting entities using NGSIv2. Note that metadata location is not rendered in any entity.

```
GET /v2/entities

[
    {
        "id": "E1",
        "type": "T",
        "A": {
            "type": "whatever",
            "value": "40, -3",
            "metadata": {}
        }
    },
    {
        "id": "E2",
        "type": "T",
        "A": {
            "type": "whatever",
            "value": "39,-2",
            "metadata": {}
        }
    }
]
```

Getting entities using NGSIv1. The location metadata is rendered only for E2 (the entity created using NGSIv1):

```
GET /v1/contextEntities

{
    "contextResponses": [
        {
            "contextElement": {
                "type": "T",
                "isPattern": "false",
                "id": "E1",
                "attributes": [
                    {
                        "name": "A",
                        "type": "whatever",
                        "value": "40, -3"
                    }
                ]
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        },
        {
            "contextElement": {
                "type": "T",
                "isPattern": "false",
                "id": "E2",
                "attributes": [
                    {
                        "name": "A",
                        "type": "whatever",
                        "value": "39,-2",
                        "metadatas": [
                            {
                                "name": "location",
                                "type": "string",
                                "value": "WGS84"
                            }
                        ]
                    }
                ]
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}
```

Conclusions:

* The location metadata is ignored in NGSIv2 creation operation at two levels:
[...]

Issue #3045, opened 2017-11-17
Wrong inclusion of "location" metadata in geo:json notifications


fgalan:
Entity created like this:

```
curl -vvvv localhost:1026/v2/entities -s -S --header 'Content-Type: application/json' -d @- <<EOF
{
  "id": "E",
  "type": "T",
  "name":{
    "value": "SD3",
    "type": "Text"
  },
  "perimeter": {
    "type":"geo:json",
    "value":{
    "type": "Polygon",
    "coordinates": [
                    [
                        [-13.806454999999991,33.46727800000001,0],
                        [-13.806707999999981,33.46723400000001,0],
                        [-13.806863999999999,33.46720599999997,0],
                        [-13.807017000000008,33.46719100000001,0],
                        [-13.806454999999991,33.46727800000001,0]
                    ]
                ]
        }
   }
}
EOF
```

A subscription on that entity is created like this:

```
curl -vvvvv localhost:1026/v2/subscriptions -s -S --header 'Content-Type: application/json' -d @- <<EOF
{
  "description": "A subscription to get info about E",
  "subject": {
    "entities": [
      {
        "id": "E",
        "type": "T"
      }
    ],
    "condition": {
      "attrs": [
      ]
    }
  },
  "notification": {
    "http": {
      "url": "http://localhost:1028/notify"
    },
    "attrs": [     
    ]
  } 
}
EOF
```

Notifications are triggered with something like this:

```
curl -vvvv localhost:1026/v2/entities/E/attrs/name -X PUT -s -S --header 'Content-Type: application/json' -d @- <<EOF
{
  "value": "$1"
}
EOF
```

Notification wrongly include a `location` metadata with `"WGS84"` value for the `perimeter` atribute, i.e. mixing with the old NGSIv1-base mechanism. Note that GET on the same entity is not including it. Thus, GET is ok but notifications isn't.

Some additional remarks:

* Tested with geo:json but it may happen with geo:point, geo:line, geo:box and geo:polygon. All them should be covered by the associated .test.
* The problem seems to happen both in NGSIv2 and legacy NGSIv1 notifications. The associated .test should cover both.

arigliano:
Hello, I would like to work on this issue. Can you assign it to me?

fgalan:
Of course. Done!

arigliano:
[...]

PR #3346, opened 2018-10-24
#3122 location metadata issues and #3045 wrong location metadata inclusion

arigliano:
This PR fixes issues #3122  and #3045 

Functional tests added.
It was removed the part where we "manually" add the "location" metadata in the NGSIV1 case.
The metadata now is correctly handled and shown as a regular "custom metadata" (when creating/getting both in V1 and V2").
In the NGSIV1 case, we use additionally this metadata to create the geolocation attribute, according to that "old way".


test/functionalTest/cases/3122_location_metadata_issues/create_and_get_location_metadata_v1_v2.test
```diff
@@ -0,0 +1,256 @@
...
+echo
+echo
+
+
+echo "04. Get E1 and E2 with NGSIV1"
+echo "============================="
+orionCurl --url /v1/contextEntities
+echo
+echo
+
```
fgalan: I'd suggest to add a step 5 with a geo query covering both (40, -3) and (39,-2), The result, as far as I understand, should be E2 but not E1 as metadata-based location is not supported in NGSIv2 so "location" there is a regular metadata.

Let's add the step and see what's happen :)

test/functionalTest/cases/0000_deprecated_checkings/legacy_wgs84.test
```diff
@@ -46,7 +46,7 @@ payload='{
         {
           "name": "location",
           "type": "string",
-          "value": "WSG84"
```
fgalan: This .test case is designed to test the support for "WGS84" literal. It was used in old versions of Orion before we note it was wrong, but we needed to keep support for not breaking backward compatibility.

What would happen if WGS84 is used in this .test? It breaks?

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
```diff
@@ -88,21 +88,24 @@ static void prepareDatabase(void)
   BSONObj A = BSON("_id" << BSON("id" << "A" << "type" << "Point") <<
                    "attrNames" << BSON_ARRAY("pos" << "foo") <<
                    "attrs" << BSON(
-                     "pos" << BSON("type" << "location" << "value" << "2, 3") <<
+                     "pos" << BSON("type" << "location" << "value" << "2, 3" << "md" <<
+                              BSON("location" << BSON("type" << "string" << "value" << "WGS84"))) <<
```
fgalan: Looking this modification, I understand the entities currently created in DB using NGSIv1-based location will not work if Orion is update to a version after this change.

We should think a way of avoiding it... otherwise it would need a data migration process, which use to be painfull.

test/functionalTest/cases/0000_deprecated_checkings/legacy_wgs84.test
arigliano: Yes, because the Context Broker now only holds location metadata as "custom", so it is rendered as it was stored during POST /v2/entities. No hardcoded location metadata additions, which caused duplication, wrongly inclusions and inconsistency among v1/v2 GET/POSTs.

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
arigliano: Yes, because now Orion handles metadata in a consistent way. A solution should be reintroducing some "If v1" in creation and/or in the rendering logic (as in the case of WSG84 test above), which leads this PR to be not very useful, if the purpose was to close these issues.

arigliano: In any case, for this specific unit test, that modification is needed to match the assertion about metadata vector size, the way in which the geolocation attribute is stored, and then the geoqueries are preserved, as other unit/functional tests are ok.

fgalan:
Let's consider the location functionality in general and from the point of view of backward compatibility. This PR should fix the addressed issues but in a way it doesn't break backward compatibility and I have some concern about this could be happening...

I'm trying to summarize my understanding on how location functionality should work in a couple of diagrams, one capturing the creation/update case, the other capturing the retrieve case.

imagen

imagen

I think the creation/update case is pretty clear. The retrieval case is a bit more tricky because it includes a check (and "corrective" action associated to the check) in the case we use NGSIv1 and we have a "location" GeoJSON field in the DB but without the corresponding "location" metadata. This will be inconsistent (i.e. it couldn't happen) for entities created according to the creation/update diagram but note that DB could hold old entities created with old Orion versions in which this situation has occurred. In fact, the DB populated in unit test mongoQueryContextGeo_test.cpp is representing this situation.

So, let's consider next actions. First of all, ensure that the behaviour we want is the one I have expressed in the diagram (the issue of backward compatibility is tricky for location and I may have misses something, don't hesitate to tell me in that case :). What do you think @arigliano ?

If we agree on this, next action would be trying to model the behaviour in the diagrams with .test. In particular:

* For the creation/update case, let's have one .test per API version. The .test should be pretty simple: create an entity with a location attribute, then check in DB that the result is as expected using mongoCmd (there are many examples of usage of mongoCmd in the existing .test if you are not familiar with it)
* For the retrievall case, I'd suggest to have three .test files:
  * NGSIv2: creating an entity with location in a preparation step (using NGSIv2 API) before the step corresponding to the test itself
  * NGSIv1-sane: creating an entity with location in a preparation step (using NGSIv1 API) before the step corresponding to the test itself
  * NGSIv2-degenerate: creating an "inconsisetent" entity (i.e. "location" GeoJSON field but without "location" metadata) with direct insertion in the DB (using mongoCmd) before the step corresponding to the test itself

Sorry for the long comment, but I hope I have provided a detailed explanation. @arigliano don't hesitate to provide feedback or ask if you have questions/doubts.

arigliano:
I agree with the logic, now it seems to me more clear ;) Only one question. For each retrieval test, should we do a GET in both NGSIv1 and v2, right?

fgalan:
> For each retrieval test, should we do a GET in both NGSIv1 and v2, right?

I haven't considered in my first approach, but I think is a good idea :)

test/functionalTest/cases/0000_deprecated_checkings/legacy_wgs84.test
fgalan: Let's pospone this particular .test until the ones described at https://github.com/telefonicaid/fiware-orion/pull/3346#issuecomment-433969358 get completed.

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
fgalan: See general comment: https://github.com/telefonicaid/fiware-orion/pull/3346#issuecomment-433969358

fgalan:
Not fully sure... but I think that using more than one REGEX() per line could be problematic.

A solution would be to add `| python -mjson.tool` after mongoCmd command

```
mongoCmd ${CB_DB_NAME} ... | python -mjson.tool
```

so now you get a multi-line JSON as output and you can use each REGEX() in a separate line.

Please, have a look and confirm if that works.

arigliano:
Thank you very much. It worked very well ;)

arigliano:
> Let's consider the location functionality in general and from the point of view of backward compatibility. This PR should fix the addressed issues but in a way it doesn't break backward compatibility and I have some concern about this could be happening...
> 
> I'm trying to summarize my understanding on how location functionality should work in a couple of diagrams, one capturing the creation/update case, the other capturing the retrieve case.
> 
> imagen
> 
> imagen
> 
> I think the creation/update case is pretty clear. The retrieval case is a bit more tricky because it includes a check (and "corrective" action associated to the check) in the case we use NGSIv1 and we have a "location" GeoJSON field in the DB but without the corresponding "location" metadata. This will be inconsistent (i.e. it couldn't happen) for entities created according to the creation/update diagram but note that DB could hold old entities created with old Orion versions in which this situation has occurred. In fact, the DB populated in unit test mongoQueryContextGeo_test.cpp is representing this situation.
> 
> So, let's consider next actions. First of all, ensure that the behaviour we want is the one I have expressed in the diagram (the issue of backward compatibility is tricky for location and I may have misses something, don't hesitate to tell me in that case :). What do you think @arigliano ?
> 
> If we agree on this, next action would be trying to model the behaviour in the diagrams with .test. In particular:
> 
> * For the creation/update case, let's have one .test per API version. The .test should be pretty simple: create an entity with a location attribute, then check in DB that the result is as expected using mongoCmd (there are many examples of usage of mongoCmd in the existing .test if you are not familiar with it)
> * For the retrievall case, I'd suggest to have three .test files:
>   
>   * NGSIv2: creating an entity with location in a preparation step (using NGSIv2 API) before the step corresponding to the test itself
>   * NGSIv1-sane: creating an entity with location in a preparation step (using NGSIv1 API) before the step corresponding to the test itself
>   * NGSIv2-degenerate: creating an "inconsisetent" entity (i.e. "location" GeoJSON field but without "location" metadata) with direct insertion in the DB (using mongoCmd) before the step corresponding to the test itself
> 
> Sorry for the long comment, but I hope I have provided a detailed explanation. @arigliano don't hesitate to provide feedback or ask if you have questions/doubts.

Fixed in 68020b7

fgalan:
Looking to travis report on last commit:

imagen

I see:

```
1024/1055: 3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test ...................  (ERROR 9 - .out and .regexpect differ) 
1024/1055: 3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test ................... (intent 2) (ERROR 9 - .out and .regexpect differ) 
1024/1055: 3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test ................... (intent 3) (FAIL 9 - .out and .regexpect differ) testHarness.sh/Wrong inclusion of "location" metadata in geo:json notifications: (3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test) output not as expected
```

Maybe something need to be adjusted yet in .test @arigliano ?

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2493,7 +2496,7 @@ static BSONArray processConditionVector
       {
         conds.append(nc->condValueList[jx]);
       }
-
+std::cout << "\nprocessConditionVector apiVersion:\n " << apiVersion << std::endl;
```
fgalan: Seems to be code for debugging. I understand it should be removed.

arigliano: Fixed in a0cb1aa

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test
```diff
@@ -0,0 +1,275 @@
...
+
+--SHELL--
+
+#
+# 01. Create E with type T 
+# 02. Create subscription for E
+# 03. Trigger notification
+# 04. See notification and make sure the metadata is correct
+#
+
```
fgalan: I would suggest to use more precise name and steps (including echo messages) in this test. For instance

```
--NAME--
Check "location" metadata is not wrongly included in NGSIv2 notifications

--SHELL--

# 01. Create E with type T 
# 02. Create NGSIv2 subscription for E
# 03. Trigger notification
# 04. See NGSIv2 notification and make sure "location" metadata is not included
```

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test
```diff
@@ -0,0 +1,275 @@
...
+}'
+orionCurl --url /v2/subscriptions --payload "$payload"
+echo
+echo
+
+
+echo "03. Trigger notification"
+echo "========================"
+payload='{
+  "value": "$1"
```
fgalan: Why to use a variable here? It would be more clear to use a fixed value (as long as you use something different from "SD3", the notification will be triggered).

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V1.test
```diff
@@ -0,0 +1,215 @@
...
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create E1 with NGSIV1
+# 02. Create E2 with NGSIV1
+# 03. Check E1 entity in DB
+# 04. Check E2 entity in DB
+
```
fgalan: I would suggest to use more precise steps (including echo messages) in this test. For instance

```
# 01. Create E1 with NGSIV1 with location metadata
# 02. Create E2 with NGSIV1 without location metadata
# 03. Check E1 entity in DB has location metadata and location GeoJSON field
# 04. Check E2 entity in DB has no location metadata nor location GeoJSON field
```

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V2.test
```diff
@@ -0,0 +1,166 @@
...
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create E1 with NGSIV2
+# 02. Create E2 with NGSIV2
+# 03. Check E1 entity in DB
+# 04. Check E2 entity in DB
+
```
fgalan: I would suggest to use more precise steps (including echo messages) in this test. For instance

```
# 01. Create E1 with NGSIV2 with location metadata but not geo:xxx type
# 02. Create E2 with NGSIV2 without location metadata but with geo:xxx type
# 03. Check E1 entity in DB has location metadata but no location GeoJSON field
# 04. Check E2 entity in DB has no location metadata but location GeoJSON field
```

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V1_creation.test
```diff
@@ -0,0 +1,176 @@
...
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create E1 with NGSIV1
+# 02. Get E1 with NGSIV1
+# 03. Get E1 with NGSIV2
+
```
fgalan: I would suggest to use more precise steps (including echo messages) in this test. For instance

```
# 01. Create E1 with NGSIV1 with location metadata
# 02. Get E1 with NGSIV1 and check location metadata is included
# 03. Get E1 with NGSIV2 and check location metadata is included
```

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_db_creation.test
```diff
@@ -0,0 +1,152 @@
...
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create E1 directly in DB
+# 02. Get E1 with NGSIV1
+# 03. Get E1 with NGSIV2
+
```
fgalan: I would suggest to use more precise steps (including echo messages) in this test. For instance:

```
# 01. Create E1 directly in DB with location GeoJSON field but without location metadata nor geo:xxx type
# 02. Get E1 with NGSIV1 with location metadata
# 03. Get E1 with NGSIV2 without location metadata
```

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test
arigliano: Fixed in c76622d

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata.test
arigliano: Fixed in c76622d

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V1.test
arigliano: Fixed in c76622d

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V2.test
arigliano: Fixed in c76622d

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V1_creation.test
arigliano: Fixed in c76622d

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_db_creation.test
arigliano: Fixed in c76622d

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -341,7 +341,8 @@ static void setCondsAndInitialNotifyNgsiv1
                                             status,
                                             fiwareCorrelator,
                                             sub.notification.attributes,
-                                            sub.notification.blacklist);
+                                            sub.notification.blacklist,
+                                            V2);
```
fgalan: The function where this call is located is setCondsAndInitialNotifyNgsiv1(). Thus, maybe it should be `V1`?

In that case, this is a bit weird... it should fail with `V2` as it is now or maybe the condition is not properly captured by a test (it shoulld).

src/lib/mongoBackend/MongoCommonSubscription.cpp
```diff
@@ -333,7 +333,8 @@ void setCondsAndInitialNotify
                                             status,
                                             fiwareCorrelator,
                                             notifAttributesV,
-                                            blacklist);
+                                            blacklist,
+                                            V2);
```
fgalan: This call is in setCondsAndInitialNotify() function, which is called from mongoCreateSubscription() and (indirectly) from mongoUpdateSubscription().

Both mongoCreateDusbsciprion() and mongoUpdateSubscription() are common methods from NGSIv1 (called from mongoSubscribeContext and mongoUpdateContextSubscription, check [doc1](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/mongoBackend.md#mongosubscribecontext-sr) and [doc2](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/mongoBackend.md#mongoupdatecontextsubscription-sr)) and NGSIv2 (called directly from  V2 service routines) logic.

Thus, maybe `V2` should not be hardwired here and come from the API version corresponding to the request?

In that case, as with https://github.com/telefonicaid/fiware-orion/pull/3346/files#r233028287, this should be failing now so probably we are not properly capturing this in a test.

src/lib/mongoBackend/mongoUpdateSubscription.cpp
arigliano: The setCondsAndInitialNotifyNgsiv1 seems to be called only if in the subscription update request (v1/updateContextSubscription ) was present the subject. But that field shouldn't be only for NGSIv2? So even if we wrongly pass V2 above, this function is not called at all. This function shouldn't be used when we have a subscription coming from NGSIv1 (subUp.fromNgsiv1) ? 
So, to summarize:
 * mongoUpdateSubscription -> 
* setCondsAndInitialNotify (wrapper)
*  ->  if (subUp.fromNgsiv1) ->
* either setCondsAndInitialNotifyNgsiv1 or setCondsAndInitialNotify. 

But **if (subUp.fromNgsiv1)** is in the  **if (subUp.subjectProvided)**, which is excepted for NGSIV2 subscription. I am missing something?

src/lib/mongoBackend/MongoCommonSubscription.cpp
arigliano: See previous comment, should be related. (Considering that passing not hardwired values for apiVersion could be better for sure).

src/lib/mongoBackend/mongoUpdateSubscription.cpp
arigliano: ` subUp->subjectProvided      = (notifyConditionVector.size() > 0); `

I figured out, the setCondsAndInitialNotifyNgsiv1 is triggered when updating Context Subscription (v1) with a "notifyConditions" array.

arigliano: I'm writing specific functional tesst to trigger notifications when updating a subscription both on NGSIv1 and NGSV2. One question. How to get the subscriptionId from the response payload in the creation step, in order to use it in the update step (to create the /v2/subscription/subscriptionId url)?

arigliano: > NOTE: maybe this function should be renamed setCondsAndInitialNotifyNgsiv2() to make things more clear.

I would keep the name of the function as it is, because also the mongoSubscribeContext (V1) uses it (through mongoCreateSubscription)

arigliano: > Have a look for instance to the test named Subscription with duplicated items in attrs filter. The trick it to get SUB_ID just after creation request this way:

Thanks. How to get the subscriptionId from the response payload in the creation step, in the NGSIV1 case instead (v1/subscribeContext) ? 
I need to extract the subscriptionId field from a JSON like the following:
```
{
    "subscribeResponse": {
        "subscriptionId": "5bf28aad456d31fc96bd52dc",
        "duration": "PT24H"
    }
}
```

fgalan: Ah ok... you need it for NGSIv1 :)

In that case, you can have a look to test named "X-Auth-Token From Subscribe Request".

```
SUB_ID=$(echo "$_response" | grep subscriptionId | awk -F\" '{ print $4 }')
```

arigliano: I need for both :) thanks

arigliano: Fixed in 5845ebd

src/lib/mongoBackend/MongoCommonSubscription.cpp
arigliano: Fixed in 5845ebd.
 I added two new functional tests for 3045 that should cover this and previous thread.

arigliano:
Fixed style error causing compliance test to fail, in 03d8b80

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata_V1.test
```diff
@@ -0,0 +1,458 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This test is covering this case

imagen

Great!

NTC

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata_V2.test
```diff
@@ -0,0 +1,370 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This test is covering the case

imagen

Great! 

NTC

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V2.test
```diff
@@ -0,0 +1,166 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This test covers the following cases (both in the left branch but with different results in the checkings)

imagen

Great!

NTC

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V1_creation.test
```diff
@@ -0,0 +1,176 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This .test covers the following cases:

imagen

Great!

NTC

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V2_creation.test
```diff
@@ -0,0 +1,129 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This covers:

imagen

Great!

NTC

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_db_creation.test
```diff
@@ -0,0 +1,152 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: This test covers:

imagen

Great!

NTC

test/functionalTest/cases/0000_deprecated_checkings/legacy_wgs84.test
fgalan: This test was postponed while the other .test gets completed. Now is the moment to resume discussing :)

As far as I understand, this .test doesn't needs a change. I'll try to explin graphically. At creation stage:

imagen

What to you think?

fgalan:
After last commits the PR is almos done. Impressive work!

Next steps would be:

1. Please check that my intepretation of the .test based in the use case diagram is correct (seven diagrams/comments). Please refresh to the last version of the diagrams at the present moment (as I have edited my early versions) and confirm they are ok.
2. Check that no use case left to be covered (basically, all the possible flow branches and posivilities in the diagram has been covered). I think all they are covered, but I would like of course to know also your opinion.
3. Have a look to https://github.com/telefonicaid/fiware-orion/pull/3346#discussion_r235328543. I think that .test could be reverted to its original version.

fgalan:
This other is pending:

4. https://github.com/telefonicaid/fiware-orion/pull/3346#discussion_r228126091

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
fgalan: Looking to the last version of the PR, I understand this changes were finally reverted (as I suggested). Could you confirm?

fgalan:
...and finally confirming at this one (sorry going drop by drop :)

5. https://github.com/telefonicaid/fiware-orion/pull/3346#discussion_r228127279

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
arigliano: Yes, I confirm.

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata_V1.test
arigliano: It's ok.

test/functionalTest/cases/3045_wrong_inclusion_location_metadata_notification/wrong_inclusion_location_metadata_V2.test
arigliano: It's ok

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V1.test
```diff
@@ -0,0 +1,215 @@
+# Copyright 2018 Telefonica Investigacion y Desarrollo, S.A.U
```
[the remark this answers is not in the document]
arigliano: It's ok

test/functionalTest/cases/3122_location_metadata_issues/create_location_attr_V2.test
arigliano: It's ok

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V1_creation.test
arigliano: It's ok

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V2_creation.test
arigliano: It's ok

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_db_creation.test
arigliano: It's ok

test/functionalTest/cases/0000_deprecated_checkings/legacy_wgs84.test
arigliano: The creation stage matches with the current behavior, thus it's ok. It creates the location field because it finds a location metadata in NGSIv1, which is stored as is ("WSG84").

In the DB we have the following:
```
{
    "_id" : {
        "id" : "Madrid",
        "type" : "City",
        "servicePath" : "/"
    },
    "attrNames" : [ 
        "location"
    ],
    "attrs" : {
        "location" : {
            "type" : "coords",
            "creDate" : 1542813795,
            "modDate" : 1542813795,
            "value" : "40.418889, -3.691944",
            "md" : {
                "location" : {
                    "type" : "string",
                    "value" : "WSG84"
                }
            },
            "mdNames" : [ 
                "location"
            ]
        }
    },
    "creDate" : 1542813795,
    "modDate" : 1542813795,
    "location" : {
        "attrName" : "location",
        "coords" : {
            "type" : "Point",
            "coordinates" : [ 
                -3.691944, 
                40.418889
            ]
        }
    },
    "lastCorrelator" : "5e1fd5c4-eda1-11e8-8e61-0242ac110002"
}
```

The test fails in the retrieve operation. As a location metadata (even if WSG84) is present, it is render as it is (WSG84) and not as WGS84. We should add a dedicated check.
 So,  I modified the following:
```
/* The flag below indicates that a location metadata with WGS84 was found during iteration.
        * It needs to the NGSIV1 check below, in order to add it if the flag is false */
        if (md->name == NGSI_MD_LOCATION && md->stringValue == LOCATION_WGS84)
        {
          noLocationMetadata = false;
        }
```
In this way, either if the location was not present at all or present but with any wrong value (including WSG84), the right WGS84 metadata is added. The problem is that now it returns both:
```
{
    "contextResponses": [
        {
            "contextElement": {
                "type": "City",
                "isPattern": "false",
                "id": "Madrid",
                "attributes": [
                    {
                        "name": "location",
                        "type": "coords",
                        "value": "40.418889, -3.691944",
                        "metadatas": [
                            {
                                "name": "location",
                                "type": "string",
                                "value": "WSG84"
                            },
                            {
                                "name": "location",
                                "type": "string",
                                "value": "WGS84"
                            }
                        ]
                    }
                ]
            },
            "statusCode": {
                "code": "200",
                "reasonPhrase": "OK"
            }
        }
    ]
}
```
I guess we should add some logic to check explicitly that a WSG84 metadata but not a WGS84 was added or can we keep the response reported above?

arigliano: We should remove WSG84 from metadataVector and add the right WGS84 ? Maybe this logic should be implemented in the metadata value rendering logic (by replacing there directly the WSG84 value with WGS84), in the latter case, we would avoid to deal with metadataVector pop (WSG84) and then push (WGS84)

arigliano: I solved with:
```
  /* The flag below indicates that a location metadata with WGS84 was found during iteration.
        *  It needs to the NGSIV1 check below, in order to add it if the flag is false
        *  In addition, adjust old wrong WSG84 metadata value with WGS84 */
        if (md->name == NGSI_MD_LOCATION)
        {
          noLocationMetadata = false;

          if (md->valueType == orion::ValueTypeString && md->stringValue == LOCATION_WGS84_LEGACY)
          {
            md->stringValue = LOCATION_WGS84;
          }
        }

        caP->metadataVector.push_back(md);
      }
    }

    if (apiVersion == V1)
    {
      /* Setting location metadata (if location attr found
       *  and the location metadata was not present or was present but with old wrong WSG84 value) */
      if ((locAttr == ca.name) && (ca.type != GEO_POINT) && noLocationMetadata)
      {
        /* Note that if attribute type is geo:point then the user is using the "new way"
         * of locating entities in NGSIv1, thus location metadata is not rendered */
        Metadata* md = new Metadata(NGSI_MD_LOCATION, "string", LOCATION_WGS84);
        caP->metadataVector.push_back(md);
      }
    }
```

Fixed in 6bc0cc4

test/functionalTest/cases/3122_location_metadata_issues/create_and_get_location_metadata_v1_v2.test
arigliano: Fixed in 1fa60ea

test/unittests/mongoBackend/mongoQueryContextGeo_test.cpp
arigliano: It was fixed in 68020b7

test/functionalTest/cases/3122_location_metadata_issues/get_location_attr_with_V2_creation.test
fgalan: Looking this into detail, entity has been created with geo:xxx in NGSIv1, so it has a "location" GeoJson field. Thus, in fact I think the second condition on the right is "Yes" in this case. Actually it is:

imagen

fgalan:
In spite the little fail in my initial analysis I found (see https://github.com/telefonicaid/fiware-orion/pull/3346#discussion_r235473891, please check) I think that the only pending thing is:

> Check that no use case left to be covered (basically, all the possible flow branches and posivilities in the diagram has been covered). I think all they are covered, but I would like of course to know also your opinion.

If you confirm that, I think it would be merged (once travis ends).

arigliano:
> In spite the little fail in my initial analysis I found (see [#3346 (comment)](https://github.com/telefonicaid/fiware-orion/pull/3346#discussion_r235473891), please check) I think that the only pending thing is:
> 
> > Check that no use case left to be covered (basically, all the possible flow branches and posivilities in the diagram has been covered). I think all they are covered, but I would like of course to know also your opinion.
> 
> If you confirm that, I think it would be merged (once travis ends).

Yes, it seems to me that now all cases have been covered.

PR #3355, opened 2018-10-31
Improve perf tunning documentation

fgalan:
Related with https://stackoverflow.com/questions/52814894/erron-on-quering-fiware-orion-with-orderby-custom-attribute/52834026#52834026

@fisuda please take this PR into account in order to update Japanese translation.


fisuda:
Noted with thanks!

PR #3365, opened 2018-12-04
FIX user manual ToC in markdown

fgalan:
CC: @fisuda just in case the modification in this PR needs some sync in the Japanese translation


AlvaroVega:
LGTM

PR #3371, opened 2018-12-06
New token tool

caa06d9c:


doc/manuals/quick_start_guide.md
```diff
@@ -9,8 +9,8 @@ A good learning resource for such API is the
 
 First of all, you need an account in FIWARE Lab, so register for one in [the following link](https://account.lab.fiware.org/sign_up) if you don't have one (it's free, all you need is a valid email adress :). With that account you can obtain a valid authentication token to use in the REST API calls to Orion. To get that token, get and run the `token_script.sh` script. Introduce your FIWARE Lab user and password when the scripts ask for it (**note you have to use the complete username, including email domain**, e.g. if you email were "foo@gmail.com" you have to use "foo@gmail.com", not just "foo"):
 
-    # wget --no-check-certificate https://raw.githubusercontent.com/fgalan/oauth2-example-orion-client/master/token_script.sh
-    # bash token_script.sh
+    # wget --no-check-certificate https://raw.githubusercontent.com/FIWARE-Ops/Tools/master/GetToken/get_token.sh
+    # bash get_token.sh orion-gi
```
fgalan: What "gi" in "orion-gi" means? Global Instance?

caa06d9c: Yes. I can fix name very easy, if you want

fgalan: The name is fine but it should be better to include a little explanation. For instance, you could replace above:

> To get that token, get and run the `token_script.sh` script

by

> To get that token, get and run the `get_token.sh` script using `orion-gi` as parameter (`orion-gi` means the Orion Global Instance at FIWARE Lab)

or something like that

caa06d9c: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3371/commits/5edc1ccca25ec1a8e18da757ef1a567957d174dc

fgalan: I'm testing with the same credentials I used to access to https://cloud.lab.fiware.org/auth/login/ but I'm getting

```
Token: Invalid username or password
```

Are these credentials ok or should I use others?

caa06d9c: Yes, it's a bug, cloud.lab.fiware.org and account.lab.fiware.org, are separated, so you should use account.lab.fiware.org credentials, but... 

I think to make orion global instance public, how do you think, is it a good idea?

fgalan: In the past I think Orion public instance needed authentication in order to avoid abuse (so users not in FIWARE Lab cannot modify existing data used by another users). If this security concern has changed and this is no longer a problem, then I guess it can be fully public.

However, I'm not the right person to take that decission :) Maybe somebody rom FIWARE Lab staff (maybe @flopezag) should provide better feedback on this than mine.

fw-ops: Ok, we decided to keep security, so you should use account.lab.fiware.org credentials.

caa06d9c: Ok, we decided to keep security, so you should use account.lab.fiware.org credentials.

fisuda:
LGTM. Thank you!

I will sync the Japanese translation after merging.

Issue #2882, opened 2017-05-04
cache sem is not taken before subCacheItemLookup() call


fgalan:
In `mongoUpdateSubscription()` function we have:

```
  if (!noCache)
  {
    subCacheP = subCacheItemLookup(tenant.c_str(), subUp.id.c_str());
  }
```

This access to cache is not protected by sem taken / give. Potentially, it may cause a segfault.

fgalan:
Flow diagram MB-12 should be changed once this issue gets completed.

SourabhChourasiya-NEC:
Fixed in #3373

fgalan:
Doc fixes in PR #3376

PR #3373, opened 2018-12-11
fix for cache sem is not taken before subCacheItemLookup() call #2882

SourabhChourasiya-NEC:


src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -888,7 +888,11 @@ std::string mongoUpdateSubscription
```
fgalan: Please add an entry in CHANGES_NEXT_RELEASE bout this fix.

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -888,7 +888,11 @@ std::string mongoUpdateSubscription
 
   if (!noCache)
   {
+    cacheSemTake(__FUNCTION__, "Looking for subscription in cache subscription");
```
kzangeli: Perhaps "subscription cache" would have been better wording ...
Too late now, but it could be included in a future PR.

PR #3376, opened 2018-12-12
FIX devel documentation MB-12 diagram and description

fgalan:
The doc part of issue #2882 

Please @fisuda takes this PR into account, in the case it needs to be synched in the Japanese translation.

CC: @SourabhChourasiya-NEC


fisuda:
Noted with thanks.

PR #3384, opened 2018-12-14
DEPRECATE Rush support

fgalan:
First step toward the final removal of the support in the code. Rush was useful in the past, when HTTPS was not natively supported in Orion. However, now that HTTPS is nativiely supported, it doesn't have to much sense to keep supporting it.


fgalan:
Please @fisuda take into account the changes in documentation involved by this PR, to be synchronized with the Japanese transalation.

PR #3386, opened 2018-12-17
Update context_providers.md

juliozinga:
Fix example

### DIFF
--- a/doc/manuals/user/context_providers.md
+++ b/doc/manuals/user/context_providers.md
@@ -62,10 +62,7 @@ EOF
 ```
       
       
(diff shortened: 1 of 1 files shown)


doc/manuals/user/context_providers.md
```diff
@@ -62,7 +62,9 @@ EOF
 ```
       
       
-* Next, consider that a client queries the Street4 temperature
```
fgalan: I understand that this line should not be modified. Or maybe I'm missing something?

juliozinga: Yes, I may have made a mistake

PR #3395, opened 2019-01-08
FIX perf tunning documentation

fgalan:
A little fix in the performance tunning documentation.

CC: @fisuda in order to sync in Japanese translation once this PR gets merged.


fisuda:
Noted with thanks.

Issue #3400, opened 2019-01-18
Documentation: context information - what is the definition


ilarimikkonen:
the documentation refers at many points to "context information" but it's not defined anywhere. Could you please add a short section to Readme explaining what "context information" means?

fgalan:
Thanks for the feedback. Pull request https://github.com/telefonicaid/fiware-orion/pull/3406 addresses it. Please tell us what do you think, so the PR can be merged to solve this issue.

PR #3406, opened 2019-01-21
FIX define context information in main README.md

fgalan:
Addresses issue #3400 

CC: @ilarimikkonen


fgalan:
I like your suggestions. Thanks!

Please have a look again to check if you find it ok.

PR #3426, opened 2019-02-06
Change to mongo 3.6 in docker-compose examples

fgalan:
It seems we misses updating this when Orion 2.1.0 was released.

@fisuda please take this into account to upgrade also the Japanese transaltion, pls.


fisuda:
Noted with thanks.

PR #3437, opened 2019-02-14
Prepare introductory presentations for Orion 2.2.0

fgalan:
This PR change the links for introductory presentations for version 2.2.0. In addition, I have updated the reference version in devel manual.

@fisuda please have a look as I have touched one of the documents in the Japanese translation and provide LGTM if it is ok.


fisuda:
LGTM. Thanks!

PR #3441, opened 2019-02-18
Add AGPL disclaimer

jason-fox:
* As requested by TSC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,25 @@ tag `fiware-orion`
 
 ## License
 
-Orion Context Broker is licensed under Affero General Public License (GPL)
-version 3.
+Orion Context Broker is licensed under [Affero General Public License (GPL)
+version 3](./LICENSE).
 
 © 2019 Telefonica Investigación y Desarrollo, S.A.U
+
+### Are there any legal issues with AGPL 3.0? Is it safe for me to use?
+
+There is absolutely no problem in using a product licensed under AGPL 3.0. Issues with GPL 
+(or AGPL) licenses are mostly related with the fact that different people assign different 
+interpretations on the meaning of the term “derivate work” used in these licenses. Due to this,
+some people believe that there is a risk in just _using_ software under GPL or AGPL licenses
+(even without _modifying_ it).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -269,7 +269,25 @@ tag `fiware-orion`
...
+There is absolutely no problem in using a product licensed under AGPL 3.0. Issues with GPL 
+(or AGPL) licenses are mostly related with the fact that different people assign different 
+interpretations on the meaning of the term “derivate work” used in these licenses. Due to this,
+some people believe that there is a risk in just _using_ software under GPL or AGPL licenses
+(even without _modifying_ it).
+
+For the avoidance of doubt, the owners of this software licensed under an AGPL-3.0 license  
+wish to make a clarifying public statement as follows:
+
+> Please note that software derived as a result of modifying the source code of this
```
fgalan: This is the same text we have in the old catalogue, isn't it?

jason-fox: Yes - more specifically this the text from the TSC minutes from this morning.

fgalan: Ok. However, I think the text should be moved up, so the copyright mark "© 2019 Telefonica Investigación y Desarrollo, S.A.U" ends at the end of the README.md file.

jason-fox: Satisfying all conflicting requirements within the limitations of a `README` is tricky. 

screenshot 2019-02-19 at 10 12 51

The idea of keeping the copyright with the License section  it is obvious that  the copyright refers to the license - i.e. _Telefonica holds the copyright and licenses under APGL_. In an ideal world we wouldn't need the final disclaimer.

If you move the copyright below the disclaimer it appears to imply the following  _This software is licensed under AGPL, the  copyright of the **disclaimer** held by Telefonica_

Yes, ideally common practice states license at the bottom, copyright last line. It is typically just a couple of lines and doesn't have any extra text. However adding the disclaimer is an old requirement and since this week's TSC agreement was that the  disclaimer must be placed prominently in the `README` I see  leaving copyright above the disclaimer as less likely to cause confusion than moving it below.

fgalan: Makes sense.

NTC

PR #3454, opened 2019-03-09
Hardening/necessary fixes for ubuntu 17 10

kzangeli:
Made orion compile and tests work in Ubuntu 17.10


src/lib/logMsg/logMsg.h
```diff
@@ -1838,9 +1838,9 @@ inline void lmTransactionStart(
   const char* __service    = strlen(_service)    == 0? "<none>" : _service;
   const char* __subService = strlen(_subService) == 0? "<none>" : _subService;
 
-  snprintf(service,    sizeof(service),    __service);
-  snprintf(subService, sizeof(subService), __subService);
-  snprintf(fromIp,     sizeof(fromIp),     _fromIp);
+  strncpy(service,    __service,    sizeof(service));
+  strncpy(subService, __subService, sizeof(subService));
+  strncpy(fromIp,     _fromIp,      sizeof(fromIp));
```
fgalan: Why this change is needed?

(The change is fine... it's just curiosity)

kzangeli: The compiler complains about "no format string" or something similar.
Not sure why we used snprintf here, doesn't make much sense.
snprintf is very good to insert variables into a format string but here there are no variables, only the format string ...
strncpy is simpler, faster and enough for what we need here.

PR #3455, opened 2019-03-11
Task/3287 toix as new cli for functest script

kzangeli:
New CLI parameter for testHarness.sh: --toIx.
This parameter sets the upper limit ofthe functests to be executed.


fgalan:
This PR has change some .test to .DISABLED. I understand this is not an intentional change and should be reverted before merging this PR.

kzangeli:
Creating a new fork. My old fork is dirty ...

Issue #3463, opened 2019-03-22
Basic query and update forwarding cases sin registration '.*' pattern


fgalan:
The basic cases are:

1. regR = `.*`, query = `.*`, attrs = [ A, ... ]
2. regR = `.*`, query = 'E', attrs = {null}
3. regR = `.*`, query = 'E', attrs = [ A ... ]
4. regR = `.*`, query = 'E', attrs = {null}
4. regR = `.*`, update = 'E', attrs = [ A ... ]

This first one was covered in issue #3458. The present issue is about covering the remaining cases.

kzangeli:
I have done some experimentation and found that query cases with entity id (cases 2-4) don't work.
I haven't tested but I strongly doubt that the update case with entity id works.
Doesn't really matter.

The implementation of forwarding is made with something else in mind.
Instead of looking at the query/update, we look at the registrations. Modifying the existing code is far more difficult than rewriting the forwarding part.

My suggestion would be to leave NGSIv1 as is (for "backward compatibility") and completely rewrite the forwarding algorithm for NGSIv2.

kzangeli:
See step 08 in the following functional test:
[legacy-query-over-entity-id-for-regs-with-idpattern.txt](https://github.com/telefonicaid/fiware-orion/files/3003836/legacy-query-over-entity-id-for-regs-with-idpattern.txt)

kzangeli:
To troubleshoot, use the branch (pushed to origin):
`bug/3463_forwarded_query_with_entity_in_for_reg_with_idpattern`

In that branch, run the functest `legacy-query-over-entity-id-for-regs-with-idpattern.test`


Search for the second occurrence of **Serving request GET /v2/entities/E1** in the logfile

What I have seen, `registrationsQuery` (I modified it) seems to do its job, it finds the three matching regs.
But this trace line:
`MongoGlobal.cpp[1901]:processEntity : Not Included ...`
is present for all three hits from registrationsQuery and the output from mongoQueryContext is an empty vector and no forwards are done by the service routine, when really, all three hits should be output in the vector and three forwards should be performed.

fgalan:
I have had a look to the branch and test. And made it to work.

The "Not Found" problem in step 8 is solved just taking into account the ".*" pattern into account for matching. In particular, this modification in `processEntity()`:

```
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
[...]

Issue #3068, opened 2018-01-09
NGSIv2 based forwarding for CPrs


fgalan:
NGSIv2 based forwarding (i.e. forward NGSIv2 queries/updates from CB to CPr, based in existing registrations) has not been yet defined.

First task with regard to this issue is specificaiton of the mechanism. In order to do so, I have recovered some previous discussion on PR https://github.com/telefonicaid/fiware-orion/pull/2990 comments that may be useful.

@jmcanterafonseca comment:

> First of all a context provider in theory should be just as any other context broker i.e. supporting the same set of APIs. Having said that I have to acknowledge the fact that Orion Broker main use case for context providers is the integration of external datasources and in that case we can not force developers to implement all the stack, and as a result my proposal is: 
>
>The operations that would have to be supported by an NGSIv2 context provider are:
>
>```
>/v2/entities?type=XX,JJ&id=YY,ZZ&attrs=at1,atn
>```
>
>I assume that at this stage we don't want to forward full query requests i.e. `q` parameter, 
>
>We need essentially to support the attribute PATCH method so that we can forward update requests >properly i.e. currently we have many many flavours for operations in charge of changing an attribute >but the essential one is PATCH, so that's why this is proposed. 
>
>```
>/v2/entities/<entityId>/attrs
>```
>
>What do you think? 
  
@fgalan comment:

> Regarding the operations I guess you mean (in your previous comment verb is missing):
>
> ```
>GET /v2/entities?type=XX,JJ&id=YY,ZZ&attrs=at1,atn
> PATCH /v2/entities/<entityId>/attrs
> ```
>
> I think it may sense. Next step in this line would be to include that "CPr requirement" in the 
specification in this PR along with a functional explanation on how forwarding works (paying special attention to "multi forwarding" case, e.g. a query on entity E ends in a forwared query for E-A1, E-A2 and E-A3 as E-A{1|2|3} belong to different CPrs).
>
>This specification could be either in the .apib (if we agree forwarding is part of the NGSIv2 API) or specific .md in the Orion manual (if we agree on the opposite), which I see by your last comment https://github.com/telefonicaid/fiware-orion/pull/2990#issuecomment-334728474 is yet a matter of discussion.
>
>PD. I agree `?q` shouldn't be a requirement for CPrs.

Second task regarding this issue is to *implement* the mechanism in CB.

[...]

PR #3467, opened 2019-03-27
100% NGSIv2 forwarding now works for queries, for simple cases

kzangeli:
First of all, the Forwarding Mode is stored in mongo.
For registrations created with legacyFowarding set to false, queries are now forwarded in NGSIv2 format.
NOTE: Only queries.
NOTE: Complex cases don't work, due to old bugs in the forwarding mechanism


src/lib/apiTypesV2/Entity.h
```diff
@@ -64,6 +64,7 @@ class Entity
   double                  modDate;          // used by dateModified functionality in NGSIv2
 
   std::vector<ProvidingApplication> providingApplicationList;    // Not part of NGSI, used internally for CPr forwarding functionality
+  std::string                       forwardingMode;              // Not part of NGSI, used internally for CPr forwarding functionality
```
fgalan: This assumes that all providing applications in the list are using the same forwarding mode, but this is not necesarily true as each providing application may come from a different registration, using different modes. It would be better to have `std:.vector<string> forwardingModeList`, so the i-th element in `providingApplicationList` corresponds to the i-th element in `fowardingModeList`.

In addition, I understand this refers to the nature of the CPr, either a legacy one implementing NGSIv1 or not legacy and implementing NGSIv2. In that case, I'd suggest a different name, e.g. `providerTypeList`. Note that the "forwarding mode" is a different thing in the NGSIv2 spec (in particular a way of specifying if CPr is able to process queries, updates of both) and out of the scope of this PR.

Finally, maybe it should a enum instead of string, Currently only two elements in the enum ("v1" and "v2") but in the future it could be expanded if new forwarding mechanism are defined.

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -504,7 +504,14 @@ static std::string parseNotification(ConnectionInfo* ciP, SubscriptionUpdate* su
 
   if (notification.HasMember("attrs"))
   {
-    if (parseStringVector(&subsP->notification.attributes, notification["attrs"], "attrs", true, true, &errorString) == false)
```
fgalan: Please try to minimize changes not related with the PR, e.g. refactoring in a file which nothing has to do with the PR. A single chande doesn't hurt, but the accumulation of many of them makes review process harder.

(General comment, no only for this PR. In fact, in this PR probably doesn't worth the time to rollbacking the changes, so NTC)

src/lib/apiTypesV2/Entity.h
fgalan: Looking to how this apperas in the DB:

> **format**: the format to use to send forwarded requests. The only accepted value for now is **JSON** (meaning NGSIv1 format), although this may change in the future.

I change my suggestion to: `providerFormatList`.

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: I got 4 errors from "make style" from travis.
3 about too long lines and one about space before newline.
I fixed the four of them.

NTC

src/lib/apiTypesV2/Entity.h
fgalan: And enum: `JSON` and `V2`.

Why V2 instead of "normalized"? This is not the same than NGSIv2 representation modes (normalized, keyValues, values) and I think `V2` is enough.

kzangeli: Yeah, what I would have wanted would be a 'apiVersion'.
Instead we have to use "JSON" for APIv1 and "normalized"/"keyValues"/"values" or something similar for APIv2. Not perfect.
For now, just to move on I am using ONLY "normalized".

Perhaps I haven't fully understood the providingApplicationList ...
Didn't really stop and consider it ...
If it is as you say, I believe the forwardingMode should be part of the providingApplication

src/lib/mongoBackend/mongoRegistrationCreate.cpp
```diff
@@ -211,7 +211,10 @@ void mongoRegistrationCreate
   setServicePath(servicePath, &bob);
   setContextRegistrationVector(regP, &bob);
   setStatus(regP->status, &bob);
-  setFormat("JSON", &bob);   // FIXME #3068: this would be unhardired when we implement NGSIv2-based forwarding
+
+  // FIXME #3068: this must be unhardwired once we fully implement NGSIv2-based forwarding (not only "normalized")
```
fgalan: I understand this comment should be removed. This PR has done the unhardwiring.

src/lib/ngsi/ContextRegistrationResponse.h
```diff
@@ -39,8 +39,11 @@
 */
 typedef struct ContextRegistrationResponse
 {
-  ContextRegistration   contextRegistration;    // Mandatory
-  StatusCode            errorCode;              // Optional
+  ContextRegistration   contextRegistration;        // Mandatory
+  StatusCode            errorCode;                  // Optional
+
+  std::string           forwardingMode;             // During forwarding, this field is used to record the 'format' from the registration that provoked
```
fgalan: Better to use "format" or "providerFormat".

src/lib/ngsi10/QueryContextRequest.h
```diff
@@ -56,10 +56,11 @@ typedef struct QueryContextRequest
   int               restrictions;
   StringList        metadataList;     // From URI param 'metadata'
   std::string       contextProvider;  // Not part of the payload - used internally only
+  std::string       forwardingMode;   // Not part of the payload - used internally only
 
   QueryContextRequest();
-  QueryContextRequest(const std::string& _contextProvider, EntityId* eP, const std::string& attributeName);
-  QueryContextRequest(const std::string& _contextProvider, EntityId* eP, const StringList& attributeList);
+  QueryContextRequest(const std::string& _contextProvider, EntityId* eP, const std::string& attributeName, const std::string& _forwardingMode);
+  QueryContextRequest(const std::string& _contextProvider, EntityId* eP, const StringList& attributeList, const std::string& forwardingMode);
```
fgalan: Why query context (a class that models the query context request itself) needs include aspects related with CPr format?

fgalan: I see the existing contextProvider there. So I think there should be a good reason but I don't remember it :)

src/lib/rest/ConnectionInfo.h
```diff
@@ -168,6 +168,7 @@ class ConnectionInfo
   RequestType                requestType;  // FIXME P2: To Be Removed (found inside restServiceP->request (restServiceP->type))
   std::string                acceptHeaderError;
   struct timeval             transactionStart;  // For metrics
+  std::string                entityIdFromUrlPath;
```
fgalan: Why is this needed?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -55,7 +57,25 @@
 */
 static char* jsonPayloadClean(const char* payload)
 {
-  return (char*) strstr(payload, "{");
```
fgalan: I understand this has changed due to the old way doesn't work. Why?=

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -101,26 +123,70 @@ static bool queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, QueryCo
...
-  std::string     resource     = prefix + "/queryContext";
+  std::string     payload;
+  std::string     verb;
+  std::string     resource;
   std::string     tenant       = ciP->tenant;
   std::string     servicePath  = (ciP->httpHeaders.servicePathReceived == true)? ciP->httpHeaders.servicePath : "";
-  std::string     mimeType     = "application/json";
+  std::string     mimeType;
+
+  LM_TMP(("qcrP->forwardingMode == '%s'", qcrP->forwardingMode.c_str()));
```
fgalan: Please check that no LM_TMP reamins in the PR before merging.

src/lib/mongoBackend/mongoRegistrationCreate.cpp
kzangeli: Not really, now it is either "JSON" or "normalized"

src/lib/ngsi/ContextRegistrationResponse.h
kzangeli: Not sure where I got that name from ...
So, "providerFormat"

src/lib/ngsi10/QueryContextRequest.h
kzangeli: Wherever the provider is I needed to add a "providerFormat".
But, as I stated before, perhaps the providerFormat should be a part of the Provider?

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -101,26 +123,70 @@ static bool queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, QueryCo
...
+    verb      = "POST";
+    resource  = prefix + "/queryContext";
+    mimeType  = "application/json";
+
+    LM_T(LmtForward, ("Rendering payload for the forward"));
+    TIMED_RENDER(payload = qcrP->toJsonV1());
+  }
+  else
+  {
+    std::string  extraHeaders;
```
fgalan: extraHeaders could be confusing here, as it could be missinterpreted as HTTP header.

I think what you actually are building are query params, so queryParams or extraParams are better names.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: The query response of GET /v2/entities is a JSON Vector (the initial '[' was eaten).
Before, in v1 the response was always an Object and the function worked just fine

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -101,26 +123,70 @@ static bool queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, QueryCo
 
 
   //
-  // 2. Render the string of the request we want to forward
+  // 2. Send the request to the Context Provider (and await the reply)
+  //    Render the payload if V1 forward
```
fgalan: Not fully sure (the diff is complex) but I think you are jumping from 2 to 4 in the sequence of comments.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Ah yes, of course. extraParams is what I meant

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Yeah, the old "3" is "Render the payload", which is only done for V1 forwards.
The 4 should be a 3, the 5 a 4 etc. Will decrement the following numbers

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
```diff
@@ -198,13 +198,13 @@ Date: REGEX(.*)
                     "serviceTime": REGEX(.*)
                 },
                 "root-subserv": {
-                    "incomingTransactionResponseSize": 320,
```
fgalan: I understand you are relalxing this becuse the current 320 doesn't work. What do you get in your system? I'd like to have a look also to mine.

The number itself doesn't care, but I'd like to ensure that is not the sympton of a more importante error in some point.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
```diff
@@ -0,0 +1,113 @@
...
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 38,186-187,235 IPV4
+accumulatorStart --pretty-print
+
+--SHELL--
+
+#
+# 01. Register entities of type T, with ID .*, and attribute A1, for the accumulator
```
fgalan: It would be also great to check in a similar fashion (i.e. with accumulator)

* 0x. Register entities of type T, with ID .*, and no attributes, for the accumulator
* 0x. Register entities of type T, with ID E1, and attribute A1, for the accumulator
* 0x. Register entities of type T, with ID E1, and no attributes, for the accumulator

(Maybe already done... I haven't finished to review the .test, but I don't want to forget this comment :)

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
```diff
@@ -0,0 +1,113 @@
...
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+[]
+
+
+03. Dump the accumulator, see the forwarded query
+=================================================
+GET http://localhost:REGEX(\d+)/v2/entities?type=T
+Fiware-Servicepath: /
```
fgalan: Registration use

```
 01. Register entities of type T, with ID .*, and attribute A1, for the accumulator
```

so I wonder why we are not seeing `attrs=A1` here... Maybe a but in the implementation?

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
```diff
@@ -0,0 +1,113 @@
...
+        "idPattern": ".*"
+      }
+    ],
+    "attrs": [ "A1" ]
+  },
+  "provider": {
+    "http": {
+      "url": "http://localhost:'${LISTENER_PORT}'/v2"
+    },
+    "legacyForwarding": false
```
fgalan: Do we have some test case testing the ommision of the legacyForwarding field (as equivalente to "V2") ?

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: Another one:

0x. Register entities of type T, with ID E1, and attribute A1+A2, for the accumulator

(To check that we see `attrs=A1,A2` in the accumulator)

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
```diff
@@ -0,0 +1,395 @@
...
+brokerStart CB 38,186-187,235 IPV4
+brokerStart CP1
+brokerStart CP2
+brokerStart CP3
+brokerStart CP4
+
+--SHELL--
+
+#
+# 01. Register entities of type T, with ID .*, and attribute A1, for CP1
```
fgalan: NGSIv2 based forwarding should cover also not-patter ID registrations.

I'd recomment to have a look to .test in cases/0787_cprs_full_functional_v2/. We did a quite comprehensive testing when we implemented query/update forwarding first time so "porting" these cases (or the relevant sub-sets) to NGSIv2 forwarding would be great.

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
kzangeli: I get 316, but only once every 2-3 tries

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: No not done, those I intended to try with real providers.
This functest I only used in the beginning.
But sure, I can add these new cases, doesn't hurt.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: In addition, it would be great to include some basic interoperability case. For instance:

1. Register E1-A1 in NGSIv1-based CPr
3. Register E1-A2 in NGSIv2-based CPr
4. Query GET /v2/entities/E1 and get E-(A1,A2) seamlessly
5. Query POST /v1/queryContext (for E1) and get E-(A1,A2) seamlessly

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: No test, and I believe I implemented that as equivalent to V1 ...
No wait, that was the famous "JSON" that in some cases is empty. If empty =>V1.
V2 only for "normalized" etc.
 
Adding a test to make sure how it works. It's a bit confusing ...

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: ok. Will copy the query part of that functest and "translate" it into V2

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
```diff
@@ -0,0 +1,395 @@
...
+# 02. Register entities of type T, with ID .*, and attribute A2, for CP2
+# 03. Register entities of type T, with ID .*, and attribute A3, for CP3
+# 04. Register entities of type U, with ID .*, and attribute A4, for CP4
+# 05. Create E1/T in CP1
+# 06. Create E2/T in CP2
+# 07. Create E3/T in CP3
+# 08. Create E4/U in CP4
+# 09. Query entities of type T - see E1, E2 and E3
+# 10. Query entities of type U - see E4
+# 11. Query entities of type T and with attr A1 - see E1 with A1, and E2+E3 without attrs
```
fgalan: This is a bit weird. I'd say that the right one should be:

* Query entities of type T and with attr A1 - see E1 with A1

However, if the behaviour is the same for NGSIv1 based registration, it would be out of this PR so ok (the NGSIv2 behaviour shoul "mimic" the NGSIv1, issues included :). Could you check and tell, pls?

CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 - Fix: idPattern '.*' is now allowed in NGSIv2 registrations (#3458)
```
fgalan: With regards to documentation I think the following docs should be touched (I'm using RTD references for convenicence, please do the modificaitons in their .md counterparts):

* https://fiware-orion.readthedocs.io/en/master/user/ngsiv2_implementation_notes/index.html#registrations. 
    * From: "Although NGSIv1 is deprecated, for the time being, NGSIv2-based forwarding has not been defined (see this issue about it) so the only valid option is to always use "legacyForwarding": true (otherwise a 501 Not Implemented error response will be the result)."
   * To: "Although NGSIv1 is deprecated, some Context Provider could not has been migrated yet to NGSIv2, so this mode could be useful".
* https://fiware-orion.readthedocs.io/en/master/user/context_providers/index.html. Several actions:
   * Remove "legacyForwardingMode: true" in the example and modify the steps describing the NGSIv1-based forwarding for NGSIv2-based ones.
   * This sentence "First, Orion doesn't include the filter in the POST /v1/queryContext operation forwarded to CP" should use GET /v2/entities instead of POST /v1/queryContext
   * The picture needs to be modified. I can take care of this (I'll pass you the new png file, to be used in the PR)
* https://fiware-orion.readthedocs.io/en/master/admin/database_model/index.html#registrations-collection.
   * From: "format: the format to use to send forwarded requests. The only accepted value for now is JSON (meaning NGSIv1 format), although this may change in the future (see issue about NGSIv2-based forwarding)."
   * To: "format: the format to use to send forwarded requests: JSON (meaning NGSIv1 format, used by legacy CPrs), or "V2""

fgalan: Maybe I have missed some one... A grep -i search in *.md for key words (cpr, registration, provider, forwarding, etc.) could be raise additional places to fix. Please, check it.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: All matching entities (not looking at attributes) are returned.
If they have any attributes among the ones in the URI param `attrs`, then those attributes are included.
If no attributes among the ones in the URI param `attrs`, then the entities enter "naked" - only entity ID and Type.
Seems like V1 works like this.

src/lib/apiTypesV2/Entity.h
fgalan: Mabye I'm not make me clear enough. I'll try again :)

* "forwaringMode" term is about the kind of operations that de CPr accept (query/updates/both). Out of the scope of this PR.
* "format" (or whichever other name we want to use) is about the version of NGSI that the CPr understands. JSON for NGSIv1 (for legacy reasons) or V2 for NGSIv2 (normalized is also fine).

Please, use the proper name for the varible. You have all the code plentiy of "fowardingMode" but what you really mean is "format" or whichever other name we want to use).

fgalan: With regards to:

> If it is as you say, I believe the ~~forwardingMode~~ format should be part of the providingApplication

It makes sense. Do we have already some struct/class modeling Cprs in which we could base? Or would it be a new one?

src/lib/jsonParseV2/parseSubscription.cpp
fgalan: That's weird... as that case should be already in master and travis didn't complain about it.

In this PR we are not going to do anything about it (thus, I agree, NTC) but next time it happens, please tell me (by skype). I'd like to have a closer look.

src/lib/mongoBackend/mongoRegistrationCreate.cpp
fgalan: I wonder if it makes sense to implement other NGSIv2 formats appart from normalized for forwarding...

src/lib/ngsi10/QueryContextRequest.h
fgalan: I understand the rationale ("given there is a provider-thing there, I'm addint this new one") but my question is more basic... why query context request class included a context provider to begin with (I mean, as it is now in master)?

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: Not confusing :), the rule is:

legacyForwarding ommission == legacyForwarding: false

But we should have a test to cover it.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: Seems like a commit is missing. Pretty sure I implemented that (not in the first run, but afterwards).
Checking ...

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: Let's transform that "it seems like" in "it works that way" to be fully sure :)

src/lib/apiTypesV2/Entity.h
kzangeli: We have a class called ProvidingApplication.
Haven't looked at it yet, but I hope to be able to put this field inside that class

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: (Travis is currently failing... maybe due the missing commit?)

src/lib/apiTypesV2/Entity.h
fgalan: I think that class is intended to model the provider in the "scope" of a context registration request/response. However, it seems pretty appropiate also for this case (vs. the alternative of inventing a new class for so close concept).

src/lib/jsonParseV2/parseSubscription.cpp
kzangeli: I think I know what happened.
The tree errors about "line too long" have "always" been there, but they're minor.
Now I added a MAYOR error, "space before newline" and travis shows all errors (minor and mayor) from the file. Thus, I got four errors from travis (really from 'make style').
I fixed not only the MAYOR error but also the three MINor errors.

src/lib/ngsi10/QueryContextRequest.h
kzangeli: That I do not know.
We could look into it if you think it smells ...
Also, when making this change work, I "blindly added" providerFormat wherever I found a provider ...
This one might not be necessary. Let's look into it

src/lib/rest/ConnectionInfo.h
kzangeli: This is meant for requests of the type GET /v2/entities/<entityId>

src/lib/jsonParseV2/parseSubscription.cpp
fgalan: Ok. Mistery solved ;)

src/lib/mongoBackend/mongoRegistrationCreate.cpp
kzangeli: Oh yes, it makes sense, but let's do that in another PR

fgalan: I don't think so. Extra formats beyond normalized (keyValues, etc.) are a way of givin flexibility to the final user. So, they make sense in operations done by the client.

However, CB-CPr is an internal communication in the platform so I don't see the need of adding extra complexity.

Whatever the case, the FIXME can be removed. The FIXME was about implementing any kind of NGSIv2 based format. Once we have one (normalized), it can be removed.

src/lib/apiTypesV2/Entity.h
kzangeli: Fixed in c522e4eaf6d7bb140f7f12c7815049bcd23ff200

src/lib/mongoBackend/mongoRegistrationCreate.cpp
kzangeli: fixed in c522e4eaf6d7bb140f7f12c7815049bcd23ff200

src/lib/ngsi/ContextRegistrationResponse.h
kzangeli: Fixed in c522e4e

src/lib/ngsi10/QueryContextRequest.h
kzangeli: Still not 100% sure (it's a jungle ... :)) but I believe it is necessary.

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: NTC

src/lib/rest/ConnectionInfo.h
kzangeli: NTC

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Sorry about that.
Fixed in c522e4e

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in c522e4e

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Fixed in 9db6f1005

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: So, I looked into this test.
All seems just OK.
The query in step 2 is like this:
```
orionCurl --url /v2/entities?type=T
```
So, that is what is forwarded.

What you propose, _to check the registration data and extract the "smallest" part of the query and the registration_ is part of the final implementation for fully working forwarding.
For that we need a complete rewrite of the forwarding mechanism. Not worth the trouble to do it in this "attempt to make forwarding a little better".

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: Bad choice of words  (_seems like V1 works like this_) ...
What is done is that all entities that match (using entity id and type) are included.
After that, the attributes to be included are filtered according to the attr-list.
Entities that have no attributes of that list are returned without attributes.

NTC

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: No need of a complete rewrite of the forwarding mechanism. As I explain at https://github.com/telefonicaid/fiware-orion/issues/3463 is just a matter of improving the logic in a couple of points. But that's another history to be addresses in another issue.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: This is another case of wanting forwarding to be perfect in this PR.
That is not what this PR is about.
For perfect forwarding we need to rewrite the mechanism.
Not a good idea to waste time trying to convert the current mechanism into something else.
Would take me months.
Rewriting it should take a week.

CHANGES_NEXT_RELEASE
kzangeli: Docu fixed in 7001535f8.
Found a few places more. Hopefully this is all.
To be completely sure we'd need to read through the entire documentation.
Perhaps we can ask for help from our Japanese co-corker?
I don't mean to read through all of it, but for him, as he started not so long ago, all of this is a lot mor e"fresh" than for us ...

kzangeli:
Some info:

The test I just added in commit XXX: `legacy_forwarding_true_false_default_with_accumulator.test` fails due to two providers are set in the list, instead of just one. There are four calls to registrationsQuery (conditionally) in mongoQueryContext(). The case that fails calls registrationsQuery in "case" 1 and 4.
I reied to add one extra condition for the fourth call, but that only made 20 other functests fail (it was a desperate attempt ... :-)).

This is the diff of the attempt (in mongoQueryContext.cpp):
-   if (requestP->attributeList.size() == 0)
+  if ((requestP->attributeList.size() == 0) && (rawCerV.size() == 0))

The attempt failed and has not been pushed to the branch.

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: To keep the original implementation (which is simpler) we could use the following:

```
static char* jsonPayloadClean(const char* payload, const &std::string init)
{
    return (char*) strstr(payload, init);
}
```

So you call `jsonPayloadClean(out.c_str(), "[");` or `jsonPayloadClean(out.c_str(), "{");` depending the case.

In fact, I guess it should even better to use as signature

```
static char* jsonPayloadClean(const char* payload, char init)
```

using the strstr() equivalent for char internally.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: It is not a problem to keep it as it is now if the behaviour in NGSIv1 is the same.

Could you check and confirm that in the equivalent NGSIv1 registration case (client doing `GET /v2/entities?id=E1` but registrtion done in NGSIv1 for ID E1 and attribute A1+A2) the accumulator gets `POST /v1/queryContext` on E1 without any attribute?

fgalan: In addition, it seems that the cases I commented:

* 0x. Register entities of type T, with ID .*, and no attributes, for the accumulator
* 0x. Register entities of type T, with ID E1, and attribute A1, for the accumulator
* 0x. Register entities of type T, with ID E1, and no attributes, for the accumulator

has not been yet added (or maybe I have missed them in the new version of the .test?)

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: This idea works if we are 100% sure that there is no '[' in the HTTP header information that comes before the payload.
The old implementation was unsafe. The new implementation is not.
Not a good idea, so:

NTC

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: (It seems I do a missplaced comment. In fact https://github.com/telefonicaid/fiware-orion/pull/3467#discussion_r271605574 actually applies here :)

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: Did this .test finally added?

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: Where cases/0787_cprs_full_functional_v2/ finally examined for relevant .test cases to migrate to V2?

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: To check if NGSIv1 behaves the same.

doc/manuals/admin/database_model.md
```diff
@@ -194,9 +194,10 @@ Fields:
     we ensure that registration IDs are unique and that queries by
     registration IDs will be very fast (as there is an automatic default
     index in \_id).
--   **format**: the format to use to send forwarded requests. The only accepted value for now
-    is **JSON** (meaning NGSIv1 format), although this
-    may change in the future (see [issue about NGSIv2-based forwarding](https://github.com/telefonicaid/fiware-orion/issues/3068)).
+-   **format**: the format to use to send forwarded requests.
+    For NGSIv1 format, use **JSON** as value for `format`.
+    For NGSIv2, as of today, only **normalized** format is supported.
+    In the near future **keyValues** and **values** will be implemented as well (only for NGSIv2 forwarding).
```
fgalan: It's not sure (see https://github.com/telefonicaid/fiware-orion/pull/3467#discussion_r270308979). So, better not include this sentence in the documentation.

doc/manuals/user/context_providers.md
```diff
@@ -70,17 +69,49 @@ curl localhost:1026/v2/entities/Street4/attrs/temperature?type=Street -s -S \
...
-  [this issue about it](https://github.com/telefonicaid/fiware-orion/issues/3068)).
+  <http://sensor48.mycity.com/v1/queryContext> (i.e., the URL used in
+  the Providing Application field at registration time, and adding
+  "/queryContext" to the URL PATH).
+  The query is forwarded using NGSIv2 format by default.
+  It doesn't matter whether NGSIv1 or NGSIv2 was used in the query, what decides the format of the
+  forward is what was demanded by the registration.
+  To have the NGSIv1 format used for the forwarding, the field "legacyForwarding" must be set to `true` in the registration:
+  ```
+  "legacyForwarding": true
```
fgalan: context_providers.md should provide a description based in NGSIv2, as if NGSIv1 didn't exist. The legacyForwarding explanation is provided in "implementation notes".

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -1060,8 +1060,8 @@ context availability management is about the *source of the entities and attribu
 in context availability management is the *registration* resource. A registration contains information about a source of
 information (named a "context provider") and whose entities and attributes are provided by that source.
 
-Let's illustrate with an example, creating a simple registration. We are stating that the attributes temperature and
-pressure of Room5 are provided by a context provider in the URL http://mysensors.com/Rooms:
+Let's illustrate with an example, with forwards in NGSIv1 format (legacyForwarding set to true), creating a simple registration.
```
fgalan: I missed walkthrough_apiv2.md in my original analyis... It should be modified to be fully based in NGSIv2 (before this PR we cannot, but now it's possible).

src/lib/ngsi/ProvidingApplication.h
```diff
@@ -34,16 +34,31 @@
...
+*
+* ProvidingApplication -
 */
 typedef struct ProvidingApplication
 {
-  std::string   string;
-  MimeType      mimeType;     // Not part of NGSI itself, used by the CB to specify the preferred Mime-Type for CPr interaction
+  std::string     string;
+  MimeType        mimeType;        // Not part of NGSI itself, used by the CB to specify the preferred Mime-Type for CPr interaction
+  ProviderFormat  providerFormat;  // PfJson ("JSON" in mongo): NGSIv1, PfV2: NGSIv2
```
fgalan: `providerFormat` and `mimeType` should be unified (in fact, `mimeType` seems to be a legacy from the XML days) but in a separate PR (in this PR with a `// FIXME PR: ...` mark is enough).

src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -2123,10 +2126,13 @@ bool registrationsQuery
 
     MimeType                  mimeType = JSON;
     std::vector<BSONElement>  queryContextRegistrationV = getFieldF(r, REG_CONTEXT_REGISTRATION).Array();
+    std::string               format                    = getStringFieldF(r, REG_FORMAT);
+    ProviderFormat            providerFormat            = (format == "")? PfJson : (format == "JSON")? PfJson : PfV2;  // FIXME PR: Make V2 default format
```
fgalan: Default format should be V1 to keep backward compatibility behevaiour in registrations created before providerFormat was invended. I understand current code is correct, the removal of the FIXME PR mark is pending.

kzangeli: Hmm, did you not say default format should be V2 just a while ago?
Misunderstanding, let's try to find out what has happened

doc/manuals/user/context_providers.md
kzangeli: That we should do once NGSIv2 is fully implemented.
For now, we only have queries in V2.
So, postponed?

doc/manuals/user/walkthrough_apiv2.md
kzangeli: Same comment here, better to postpone until NGSIv2 forwarding is ready.

src/lib/mongoBackend/MongoGlobal.cpp
kzangeli: I'm probably thinking of a missing `legacyForwarding` in the creation of a registration.
There we said that the default should be V2.
 It's a bit strange but as the same time it makes sense. Regs created before legv1acyForwarding should probably be NGSIv1.

kzangeli: Fixed in 2ca5e661c

doc/manuals/user/context_providers.md
fgalan: The example in this .md is covering only queries so it could be done in this PR and, at the same time, minimize the times we touch the documentation (which makes also life easier for people that does translations on the documentation).

doc/manuals/user/walkthrough_apiv2.md
fgalan: Same https://github.com/telefonicaid/fiware-orion/pull/3467#discussion_r271797488

src/lib/ngsi/ProvidingApplication.h
kzangeli: Fixed in 2ca5e66 (only a FIXME)

doc/manuals/admin/database_model.md
kzangeli: Only thinking of Orion as context provider, it makes sense what you say, to not implement keyValues/values.  But, perhaps somebody wants to implement a context provider of their own and would like this feature ...
I wouldn't rule it out. The idea is good, in my opinion.
"Near future" that's perhaps a little bit optimistic.

That said, I removed the line in:
Fixed in 2ca5e66

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: Here: 3068_non_legacy_notifications/legacy_forwarding_true_false_default_with_accumulator.test

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: The first of the three has been added,
As none of them work right now, no meaning to add any more test cases until the problem is fixed.

TBI

doc/manuals/user/context_providers.md
kzangeli: Attempt made in 5404560b1

doc/manuals/user/walkthrough_apiv2.md
kzangeli: Complete proof-read and corrections in 45e7eedd0.
There wasn't really much about NGSIv1

src/lib/ngsi10/QueryContextRequest.h
fgalan: It seems it is used in queryForward() to associate the query request to the provider to which the request is forwareded, e.g:

```
  if (parseUrl(qcrP->contextProvider, ip, port, prefix, protocol) == false)
```

This fact means 100% security for me :)

NTC

src/lib/rest/ConnectionInfo.h
fgalan: I understand this parameter is used in queryForward() in

```
    if (ciP->entityIdFromUrlPath != "")
    {
      extraParams = "&id=" + ciP->entityIdFromUrlPath;
    }
```

However, the entity id should be taken from the QueryContextRequest which is passed as argument to the function. Note that the service routing logic would have transformed the NGSIv2 request in the equivalent QueryContextRequest and the id should be there.

Moreover, maybe this should be prepared for multiple IDs (i.e. `id=E1,E2`). The QueryContextRequest could have several entity ids, e.g. `POST /v2/op/query` or `POST /v1/queryContext` with several entities in the payload.

fgalan: Also related with the usage of QueryContextRequest parameter in queryForward() as source of entity ids, the following should be removed:

```
    if (ciP->uriParam["id"] != "")
    {
      extraParams = "&id=" + ciP->uriParam["id"];
    }
```

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: I'm still trying to understand the logic... I guess that that magic number of 13 means `\r`.

I understand you are looking for either a `\n\n` or `\n\r\n` pattern. In that case, why don't just do somethink like this?

```
char* p;
if ( (p = (char*) strstr(payload, "\n\n")) != NULL)
{
   return p;
}
else
{
   return (char*) strstr(payload, "\n\r\n");
}
```

Clearer from my point of view and also 100% safe.

test/functionalTest/cases/3068_non_legacy_notifications/legacy_forwarding_true_false_default_with_accumulator.test
```diff
@@ -0,0 +1,112 @@
+# Copyright 2019 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan: We should provide in positive, I mean, instead of "non legacy notifications" (notifications here is wrong, by the way as this is not related with subscriptinos) we should use something like "ngsiv2 forwarding".

Not only for directory and .test file names, but also for step descriptions and echo messages inside (if applicable).



(In front of me from the very begingin, but I have just realized about this, sorry)

kzangeli: Yeah, I saw it too.
A forwarding request is so similar to a notification ...
I make this mistake all the time :(

kzangeli: Fixed in ad0ca86a2

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: Just took a look at this ...
There are 24 functest files.
We seems to use V1 registrations in all 24 test files. I assume this (078will 7) is long before we implemented regs in NGSIv2.
What probably should be done here is to remove all usage of V1 regs and only use V2 regs.
V1 is deprecated and will (hopefully) soon be removed.

However, this PR has gone on for quite some time. I'd prefer to postpone this to some other time.
Perhaps to when we finally decide to remove V1. There will be an important clean of functest that day ...

src/lib/rest/ConnectionInfo.h
kzangeli: Found another code smell with extraParams (`extraParams =` instead of `extraParams +=`).
Fixed in 6486136a8

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Sure, there are a hundred ways to solve a thing like this.
However ... Your solution is no good.
If `strstr(payload, "\n\r\n")` returns NULL, then 3 will be returned ...
Might be that my solution is bad in a similar way.
I'll take a look at it

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: Maybe the task about translation case/0787 to NGSIv2 could be done in the future. Probably some of the .test there mix query and update forwards, so it is better to do it at the end. I mean in this order:

* NGSIv2 query forwarding with basic .test (this PR)
* NGSIv2 update forwarding with basic .test (the next PR)
* "Translate" the 0787 cases to NGSIv1 (the next of the next PR)

Makes sense?

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: Nope, no problem there. NULL returned if no payload is found.
Added a few comments to make the function even easier to understand, in bf50da234

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: Sure. Another option is to not modify all those tests but to copy them and convert into pure V2.
Like we've done a few times already.
When we remove all V1, we'll remove the V1 tests as well (they will fail, so no issue/FIXME needed).

fgalan: Right, "copy & translate" instead of "just translate" (case/0787 will remain as it is now, a new case/3068 will be created based on that).

NTC

kzangeli: For this, we may need an issue ...
Or, will we remember a few weeks from now, when V2 Update forwarding is ready?

fgalan: It has been noted in https://github.com/telefonicaid/fiware-orion/issues/3068#issuecomment-479933791 so we don't forget about it.

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: Fixed in 19efa15ad

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
kzangeli: Fixed in 19efa15

src/lib/rest/ConnectionInfo.h
fgalan: Good for the changes in 6486136, but in addition I understand that `ciP->entityIdFromUrlPath = entityId;` should be removed in getEntity.cpp

CHANGES_NEXT_RELEASE
fgalan: More things I have spotted (in the devel manual):

* https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/sourceCode.md#srclibserviceroutines. Remove the "IMPORTANT" paragraph (and join together the line just after it and the line just before it).
* https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/cprs.md (a document in the devels manual devoted to CPr and fowarding!).
    * Remove references to NGSI9 (obsolete term from NGSIv1 days)
    * Changes in the FW-04 description at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/cprs.md#flow-fw-04. Currently the description only covers NGSIv1 but this method has been expanded to cover also NGSIv2. Hint: try to do it in a way in which changes in the diagram get minimized. If after that modifications in the png are needed I can help with that.
        * Take note also off FW-02. Not for this PR but in the next one for fowared updated it would need also to be modified.

fgalan: (If you need adjustments in the diagramas just tell me so I can provide the new .png files for that)

kzangeli: Fixed in 69093c2b8

src/lib/rest/ConnectionInfo.h
kzangeli: Fixed in 7d1cb1056

test/functionalTest/cases/3068_ngsi_v2_based_forwarding/legacy_forwarding_true_false_default_with_accumulator.test
```diff
@@ -0,0 +1,133 @@
...
+#
+#
+# FIXME: It seems like older versions of libcurl (CentOS 7 used version 7.29 of libcurl) contain a bug that makes the Content-Type
+#        always be added to the sent message, also when the Content-Type is EMPTY (which should inhibit this behaviour).
+#        The result of this bug is that the line
+#          Content-Type:
+#        is a part of forwarded messages.
+#        In newer versions of libcurl this haqs been fixed (Ubuntu 18.04 uses libcurl 7.58)
+#        For now, a "grep -v 'Content-Type'" is added to calls of accumulatorDump as a work-around.
+#        The real fix would be to use a newer version of libcurl in CentOS 7 (Travis, Jenkins)
```
fgalan: To use grep -v to hide this may make jenkins and travis happy, but the problem is still there and could potencially broke with real CPrs. Not a real fix of the problem.

Thus, I propose the following:

1. Review the changelog between 7.29 and 7.58 to confirm the bug and the version in which it has been solved. Let's name it version X. Some additional comments:
    * In my system the problem also happens. I'm using Debian with 7.38.0-4+deb8u9. Thus, I'd suggest to explore the delta from 7.38 to 7.58 (shorter than the delta from 7.29)
    * If the fix is not evident after exploring the changelog, then ask in the libcurl devel list or in SOF
2. Do a quick test in local with version X in a system wtich previously failed (mine could be a valid place)
3. Create a fresh branch from master and implement the fix there. The CB code will not be touched but fixes in documentation, dockerfile and travis will be needed.
4. Merge that branch in master, ensuring that CI e2e regression keep working
5. Upgrade this PR branch master (removing the grep -v hack)

While that comes, let's keep this PR on hold.

kzangeli: The `grep -v` ensures that the only error in the test is the unwanted occurrence of "Content-Type:", everything else is as it should. That was the reason for the modification.

Question raised in SOF:
https://stackoverflow.com/questions/55588447/empty-content-type-in-curl-easy-perform-get-request

doc/manuals/user/context_providers.md
```diff
@@ -159,7 +131,6 @@ Some additional comments:
     the CPr results before responding them back to client. An issue corresponding to this limitation has been created:
```
fgalan: On above line I have found a NGSIv1 mention that should be fixed:

```
First, Orion doesn't include the filter in the `POST /v1/queryContext` operation forwarded to CPr
```

should be changed to:

```
First, Orion doesn't include the filter in the `GET /v2/entities` operation forwarded to CPr
```

kzangeli: Fixed in 138fbfa943a2aec7992c0c84422d254a2c2f319f

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_accumulator.test
fgalan: I found it:

```
 # 01. Registration of E1 without Legacy Forwarding, for the accumulator - forward to be in NGSIv2 format
```

NTC

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: I missed this in the long thread of comments above:

> In addition, it would be great to include some basic interoperability case. For instance:
> 
> 1. Register E1-A1 in NGSIv1-based CPr
> 3. Register E1-A2 in NGSIv2-based CPr
> 4. Query GET /v2/entities/E1 and get E-(A1,A2) seamlessly
> 5. Query POST /v1/queryContext (for E1) and get E-(A1,A2) seamlessly

Has been finally included?

If not, I suggest to adding them to this PR, but I'll write them down as pending work (i.e. an extra PR) in #3068

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: I suggest to add a FIXME in the .test preamble with this (or a similar) wording:

> FIXME: step 11 should be just "Query entities of type T and with attr A1 - see E1 with A1", I mean without "and E2+E3 without attrs". To do so, CB should be more precise in the way it invokes CPrs or in the way it filters the CPr results before responding to the client. Note this is not a NGSIv2-specific flaw, NGSIv1 forwarding behaves the same way.

CHANGES_NEXT_RELEASE
fgalan: > (If you need adjustments in the diagramas just tell me so I can provide the new .png files for that)

At the end you didn't need a replacing .png? (just to confirm)

doc/manuals/user/context_providers.md
```diff
@@ -53,79 +53,51 @@ curl localhost:1026/v2/registrations -s -S -H 'Content-Type: application/json' -
...
+  <http://sensor48.mycity.com/v2> does know about the Street4 temperature, so it forwards the query
   (message number 3) to the URL
-  <http://sensor48.mycity.com/v1/queryContext> (i.e. the URL used in
-  the Providing Application field at registration time, plus the
-  "/queryContext" operation). Note that the query is forwarded using
-  NGSIv1 format although the original request from the client used NGSIv2
-  (NGSIv1 is deprecated, but we need to do so as there is yet lack of support for NGSIv2 based forwarding, see
-  [this issue about it](https://github.com/telefonicaid/fiware-orion/issues/3068)).
+  <http://sensor48.mycity.com/v2/entities> (i.e., the URL used in
+  the Providing Application field at registration time, and adding "/entities" to the URL PATH).
```
fgalan: ...the URL used in the `url` field

(Providing Application is the old one in NGSIv1)

CHANGES_NEXT_RELEASE
fgalan: I have looked this in detail. I think the png should be replaced with this one:

imagen

And FW-04 step 2 description changed to (for instance):

> * Fowarded request has to be built (step 2). In the case of NGSIv1, we need to extract information of the binary object to text to be able to send the REST request (plain text) to the Context Provider using `POST /v1/queryContext`. In the case of NGSIv2, the use `GET /v2/entities` an no payload is required.

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -181,15 +181,15 @@ curl localhost:1026/<operation_url> -s -S [headers] -X DELETE
 
 Regarding \[headers\] you have to include the following ones:
 
--   Accept header to specify which payload format
-    you want to receive in the response. You should explicitly specify JSON.
+-   Accept header to specify the payload formatin which
```
fgalan: Typo? formatin -> format in

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -738,9 +734,8 @@ Write it down because you will need it later in this tutorial.
 < Date: Wed, 25 May 2016 11:05:35 GMT
 ```
 
-Let's have a look now at accumulator-server.py. We will see one (and
-just one by the moment, no matter how much you wait)
-notifyContextRequest, similar to this one:
+Let's have a look now at accumulator-server.py. We will see one notifyContextRequest (and
```
fgalan: We will see one notification (and ...

(notifyContextRequest is an old term from NGSIv1. This comment is not really related with fowareding functionality, but given you have touched this part of the walkthrough document, let's ensure is as correct as possible :)

src/lib/ngsi/ProvidingApplication.h
fgalan: I think the moment is (actually, was) "when XML is removed" (because NGSIv1 forwarding also is based in application/json as MIME type), but it's fine in the case you don't want to bother with a little adjust in the comment text.

test/functionalTest/cases/3068_ngsi_v2_based_forwarding/legacy_forwarding_true_false_default_with_accumulator.test
fgalan: So at the end, after research, this wasn't a problem with libcurl but something related with Flask and the libraries in which it depends.

Thus, NTC (happily :)

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: I'm stil thinking that the version above based in two strstr() statements in simpler and shorter (specially if you add +4 lines to the PR implementation, due to style). However, not importar enough to keep this thread opened. I'll have a look and eventally modify it when fowarding works ends.

NTC

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
fgalan: That's randomness scares me a bit if we cannot find the cause (too many "misteries" recently, i.e. libcurl vs. Flask, etc.)... and when I'm scared I prefer to leave the things as they are :)

Thus, please revert the changes in this .test. Travis will tell if things keep stable. Sorry for the fails this may cause in your local environment, but travis/jenkins rules.

kzangeli: Are you aware that this functest has 12 'incomingTransactionResponseSize', of which 9 have already a REGEX?
This will fail in my system and failing functests is a bad idea. Too strict testing gives a lot of problems ...

src/lib/ngsi/ProvidingApplication.h
kzangeli: Not sure we ever did forwarding in both JSON and XML.
Changing the text for something broader:
```
// FIXME: Remove 'mimeType'
```
Because saying that is should be removed "when" something happens that has already happened and many years ago ...

src/lib/serviceRoutines/postQueryContext.cpp
kzangeli: If you do two calls to `strstr`, you start looking from the beginning again the second time. Bad idea.

NTC

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
fgalan: What is better? Not to fail locally with relatex tests + not to fail in reference CI (travis/jenkins) with relaxed tests? Or to fail locally with strict tests + not to fail in reference CI (travis/jenkins) with strict tests?

An answer for that would require discusion, out of the scope of this PR (maybe even we have already an issue about this).

And yes... I know we have currently a mixture of REGEX(...), both relaxed and strict, in the test related with metrics. We have ended in an undersired middle position that should go either to one end or the other. But not in this PR, which should leave the things as they are now.

src/lib/ngsi/ProvidingApplication.h
fgalan: What about?

```
// FIXME: Remove mimeType as it is useless (we always use application/json to forward, no matter NGSIv1 or NGSIv2)
```

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: Not included. Add to issue

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
kzangeli: Never mind.
320 is back in aef7cb845

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
kzangeli: Fixed in aef7cb8

CHANGES_NEXT_RELEASE
kzangeli: Fixed in aef7cb8

doc/manuals/user/context_providers.md
kzangeli: Fixed in aef7cb8

doc/manuals/user/walkthrough_apiv2.md
kzangeli: Fixed in aef7cb8

doc/manuals/user/walkthrough_apiv2.md
kzangeli: Fixed in aef7cb8

test/functionalTest/cases/3068_non_legacy_notifications/non_legacy_notifications_with_providers.test
fgalan: Added to the pending-things comment https://github.com/telefonicaid/fiware-orion/issues/3068#issuecomment-479933791

NTC in this PR

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
fgalan: Maybe github is fooling me but I'm still seeing changes in 2750_reset_metrics.test in the PR (it should disappear from the PR diff view).

src/lib/mongoBackend/mongoRegistrationGet.cpp
```diff
@@ -97,6 +97,7 @@ static void setProvider(ngsiv2::Registration* regP, const mongo::BSONObj& r)
   else
   {
     // FIXME #3068: to be implemented once we define NGSIv2 based forwarding
```
fgalan: Once this has been implemented, this FIXME could be removed.

test/functionalTest/cases/2750_common_metrics/2750_reset_metrics.test
kzangeli: My fault. Thought there were only one. There were four. 
Now fixed in 3c29a94ca

src/lib/ngsi/ProvidingApplication.h
kzangeli: Done

src/lib/serviceRoutines/postQueryContext.cpp
```diff
@@ -101,26 +124,120 @@ static bool queryForward(ConnectionInfo* ciP, QueryContextRequest* qcrP, QueryCo
...
+  {
+    //
+    // NGSIv2 forward: instead of payload, URI params are used
+    //
+    std::string  extraParams;
+
+    verb      = "GET";
+    resource  = prefix + "/entities";
+
+    if (ciP->uriParam["type"] != "")
```
fgalan: I think this strategy to set the type is not going to work in the case the type comes in the payload, i.e. client request is `POST /v2/op/query` or `POST /v1/queryContext`. All data should come from `qcrP` which supposelly has been "translated" from the original request and have all information.

However, better to include in a next PR, after having some .test confirming and raising the problem. In this PR just marking this with a FIXME mark would be enough.

fgalan: I think the solution could be something similar to what is done for attrs a bit bellow:

```
    if (ciP->uriParam["attrs"] != "")
    {
    ...
    }
    else if (qcrP->attributeList.size() != 0)
```
I mean, first trying to get from URI params. If fails, try with `qcrP` info.

kzangeli: Makes sense what you say.
This PR was about `GET /v2/entities` where the type will only come as uri param.
Adding FIXME for the future

src/lib/mongoBackend/mongoRegistrationGet.cpp
fgalan: Fixed by 3c29a94

src/lib/serviceRoutines/postQueryContext.cpp
fgalan: Fixed by 3c29a94

fgalan:
CC: @fisuda 

This PR touches several .md files. Many of the changes are related with English language typo fixing, so maybe they doesn't have impact in the Japanese translation, but other changes are relevant. I'd recommend to have a look and sync any needed change, pls. Thanks!

fisuda:
@fgalan ,
Noted with thanks. I will check this PR.

PR #3475, opened 2019-04-23
hardening/service_routine_doc

kzangeli:
Fix in service routine document and a bug fix about the `Allow:` HTTP header


src/app/contextBroker/orionRestServices.cpp
```diff
@@ -401,7 +401,7 @@ static RestService badVerbV[] =
   { BatchQueryRequest,                             3, { "v2", "op", "query"                                                            }, badVerbPostOnly           },
   { BatchUpdateRequest,                            3, { "v2", "op", "update"                                                           }, badVerbPostOnly           },
   { RegistrationRequest,                           3, { "v2", "registrations", "*"                                                     }, badVerbGetDeleteOnly      },
-  { RegistrationsRequest,                          2, { "v2", "registrations"                                                          }, badVerbPostOnly           },
+  { RegistrationsRequest,                          2, { "v2", "registrations"                                                          }, badVerbGetPostOnly        },
```
fgalan: Should the bugfix be mentioned in CNR (without referrencing issue number)? Or is too minor?

kzangeli: I'd say it's very minor, but I have no problem with adding it to the change log.

fgalan: Ok. Add it, pls.

kzangeli: Fixed in cb6054704

PR #3480, opened 2019-04-24
Task/3068 ngsiv2 update forwarding

kzangeli:
Implementation of NGSIv2 forwarding of updates for the request
```
PATCH /v2/entities/{EntityID}/attrs
```


CHANGES_NEXT_RELEASE
```diff
@@ -1,4 +1,5 @@
 - Fix: idPattern '.*' is now allowed in NGSIv2 registrations (#3458)
 - Add: Forwarded queries work for registrations using idPattern == '.*', but only for simple cases (#3458)
+- Add: Forwarded updates work for registrations using idPattern == '.*', but only for simple cases (#3458)
```
fgalan: This line seems to be duplicated.

doc/manuals/devel/cprs.md
```diff
@@ -16,7 +16,7 @@ In NGSIv1 (deprecated), the request `POST /v1/updateContext` has a field called
 * APPEND_STRICT
 * REPLACE
 
-> Side-node: The first three are "standard NGSIv1" while the second two were added for NGSIv2.
```
fgalan: This file modificaiton seems to be related with fixes in the query forward documentation, but the update forwared documetnation about (FW-02 if I'm remembering correctly) has not be modified. Probably an update of the png for FW-02 is needed also, but I haven't seen yet the changes in the code so I'm not fully sure.

doc/manuals/user/ngsiv2_implementation_notes.md
```diff
@@ -335,8 +335,9 @@ for the following aspects:
   directly. I.e., updates must be done deleting and re-creating the registration. Please
   see [this issue](https://github.com/telefonicaid/fiware-orion/issues/3007) about this.
 * `idPattern` is supported but only for the exact regular expression `.*`
-  And, right now (2019-03-21), only forwarding of queries is working for registrations with idPatterns.
-  forwarded updates do not work for registrations with idPattern (the plan is to fix this asap). 
+  And, right now (2019-04-24), forwarding for registrations with idPatterns works for queries using `GET /v2/entities`
+  as well as updates using `PATCH /v2/entities/{Entity-ID}/attrs`.
+  More requests will be added to this list.
```
fgalan: I think is better the old wording, without providing the detail of the particular operations. At least until the translation of cases/0787 test so we can have more insight in which operations are supported and which operations doesn't.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2703,17 +2703,24 @@ static void searchContextProviders
   std::string                        err;
```
fgalan: Is there any change in this file appart from LM_T LmtForward adddition? It seems so but I would like to have a confirmation on your side, pls.

src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -137,13 +138,39 @@ static void updateForward(ConnectionInfo* ciP, UpdateContextRequest* upcrP, Upda
...
+  {
+    std::vector<std::string>  nullFilter;
+    Entity*                   eP = upcrP->entityVector[0];
+
+    eP->renderId = false;
+
+    TIMED_RENDER(payload = eP->toJson(NGSI_V2_NORMALIZED, nullFilter, false, nullFilter));
+
+    resource = prefix + "/entities/" + eP->id + "/attrs";
+    verb     = "PATCH";
```
fgalan: I'm afraid this is contrary to specification https://github.com/telefonicaid/fiware-orion/issues/3068#issuecomment-470005604:

> * `POST /v2/op/update` will use `"update"` as `actionType` (to be consistent with NGSIv1)

That's the operation has to be used to implemente NGSIv2 forwarding.

src/lib/serviceRoutinesV2/patchEntity.cpp
```diff
@@ -25,6 +25,9 @@
 #include <string>
 #include <vector>
 
+#include "logMsg/logMsg.h"
```
fgalan: Really needed in this file? I don't see any change in this file using the things added by these includes...

test/functionalTest/harnessFunctions.sh
```diff
@@ -715,6 +715,14 @@ function accumulatorStart()
     shift
   fi
 
+  url="/notify"
```
fgalan: This is afix in harness suite? I mean, until this fix the `--url` parameter was being ignored?

CHANGES_NEXT_RELEASE
kzangeli: One line is about queries, the other about updates.
They **are** very similar but also completely different

fgalan: I misses the "queries" vs "udpates" difference. I's ok. Sorry the noise.

NTC

doc/manuals/user/ngsiv2_implementation_notes.md
kzangeli: Until we have a clear path forward, let's wait with these kind of changes. No need to spend time before we are sure what we want exactly

src/lib/mongoBackend/MongoCommonUpdate.cpp
kzangeli: Seems to be only LM_T in this file, yes

doc/manuals/user/ngsiv2_implementation_notes.md
fgalan: ... and if there is no need of spending time, why this file was modified? ;)

I'd suggest to go back for its previous version.

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: NTC

doc/manuals/devel/cprs.md
kzangeli: As you probably can see, the modifications in this file are just corrections.

About documenting _Update Forward_, part of it is added, but most probably not **all** that is necessary

src/lib/serviceRoutines/postUpdateContext.cpp
kzangeli: No really contrary to the "specification" though.
You are talking about batch ops (POST /v2/op/update). This PR only deals with the PATCH operation.

As before, once we have agreed on a way to attack the forwarding, this will either stay the same or be modified.
Before we have that agreement, nothing can be done really.

src/lib/serviceRoutinesV2/patchEntity.cpp
kzangeli: I probably had an LM_T and then removed it.
These things happen during PRs.
The inclusion of logMsg.h doesn't hurt (except for a millisecond extra during compilation)

doc/manuals/user/ngsiv2_implementation_notes.md
kzangeli: I was following a perhaps unclear spec and this was my interpretation, a logical one by the way.
I did not see this as spending time for nothing.

src/lib/serviceRoutinesV2/patchEntity.cpp
fgalan: I agree it doesn't hurt, but let's try to keep files tidy. If an include is doesn't use by a file, the it should be removed (if I'm remembering correctly, I think we have some rules about that in style guide).

test/functionalTest/harnessFunctions.sh
fgalan: NTC

PR #3481, opened 2019-04-24
FIX direct_https_notifications_no_accept_selfsigned.test to cover both CB_THREADPOOL posibilities

fgalan:
Current .test works ok with normal run of test harness (i.e. just `testHarness.sh`), which is the one done by travis for instance.

However, one of our jenkins instances does two passes, one with just `testHarness.sh` and another adding CB_NO_CACHE=ON CB_THREADPOOL=OFF. The latter (CB_THREADPOOL=OFF) makes this .test to fail.

I don't like the fix too much (it's a kind of hack) so any other suggestion fixing the problem is welcome.


kzangeli:
LGTM

PR #3498, opened 2019-05-29
NGSIv2 forwarding regression (based in old NGSIv1 tests)

fgalan:
This PR is a translation of the reference tests for forwarding (cases/0787_cprs_full_functional and cases/0787_cprs_full_functional_v2) to a version using NGSIv2 based forwarding (*). The resulting tests are in parallel directories cases/3068_cprs_full_functional and cases/3068_cprs_full_functional_v2

_How to review this PR?_

This PR is pretty large I’d suggest to review individual commits:

* 8e6a6ce , contains the modifications to the code to make the .test to work. In can be reviewed in the usual way (i.e. through github diff).
* a4c7648 , contains all the new .test files. The best way to review the new .test is comparing to their original counter parts. [This script](https://gist.github.com/fgalan/d60921b445b99bf397ef987abab440ef) (to be run in test/functionTest directory) may help.

If you run the script, you will see changes only in the preamble comments and in the registrations request/responses (from NGSIv1 registration to NGSIv2 registrations format). **This ensures that from an end-to-end perspective (i.e. the original client sending the query/update to CB) NGSIv2 forwarding is at least functionally equivalent to the old NGSIv1 forwarding implementation. Guarantees are high, given the large set of tests that have been translated (49).**

_There is one exception_. You will see some differences in update cases with partial updates (i.e. the update is forwarded containing several attributes, some in the CPr and some not in the CPr). In this case, NGSIv1 response from the CPr is able to “separate” success and missed attribute, so CB can use this information to response consequently to client (using a similar response to the used by the CPr in NGSIv1 case, using “PartialUpdate” and a description with the detail in NGSIv2 case).

However, in NGSIv2 when the update is forwarded containing several attributes, some in the CPr and some not in the CPr, the response is always “Not Found” without additional detail in the “description”. Thus, CB does not know which one failed and which one did not fail and he that all attribute failed (and responses consequently to the client).

You can see this effect in the following test:

* cases/3068_cprs_full_functional/fwd_v2_update_forward_depends_on_attr_field.test
* cases/3068_cprs_full_functional/fwd_v2_update_found_and_not_found_in_local_and_three_cprs.test
* cases/3068_cprs_full_functional/fwd_v2_update_three_found_on_cps_two_local_four_not_found.test
* cases/3068_cprs_full_functional_v2/fwd_v2_ngsiv2_update_forward_depends_on_attr_field.test
* cases/3068_cprs_full_functional_v2/fwd_v2_ngsiv2_update_found_and_not_found_in_local_and_three_cprs.test
* cases/3068_cprs_full_functional_v2/fwd_v2_ngsiv2_update_three_found_on_cps_two_local_four_not_found.test

In this a real problem? I do not think so given that:

* The effect in NGSIv1 is really corner case and the API is deprecated. Response to update requests never were not too important and in addition in only occurs for a subset of CPrs (NGISv2 based ones) in a subset of forwarding cases (mixing successful and missing attributes in the same request to a CPr). If NGSIv1 were not deprecated this could be more important, but it is, so as long as we don’t break NGSIv1 too much (and this is not the case :) it’s fine.

* In NGSIv2 we can relax the “description” of “PartialUpdate” in this case to "Some of the following attributes were not updated: …". Note that change “description” text is legal as they are intended for humans and may vary between implementations (as NGSIv2 specification describes in “Error Responses” section). There is not breakage is backward compatibility as “error” field doesn’t change. This behavior has to be documented anyway.

Pending:

- [x] The aforementioned changes in code and documentation regarding “PartialUpdate”
- [x]  Create a follow up issue about possible modifications to “Not Found” in cases of partial updates, so the “description” could provide more information about it. Issue is: https://github.com/telefonicaid/fiware-orion/issues/3499

(*) Note that some .test files within cases/0787_cprs_full_functional have not been translated, as they do not contain any registration. Probably they are there due to miss-classification.


src/lib/serviceRoutines/postUpdateContext.cpp
```diff
@@ -312,11 +312,20 @@ static bool updateForward(ConnectionInfo* ciP, UpdateContextRequest* upcrP, Upda
   else  // NGSIv2
   {
     LM_T(LmtForward, ("upcrP->providerFormat == V2. out: '%s'", out.c_str()));
-    // NGSIv2 forward - no payload to be received, just a 204 No Content HTTP Header
-    if (strstr(out.c_str(), "204 No Content") != NULL)
+    // NGSIv2 forward - no payload to be received
+
+    //if (strstr(out.c_str(), "204 No Content") != NULL)
```
fgalan: We cannot use string matching for this. Note that reason phrases such as "No Content" or "Not Found", although used 
by most implementations, are not mandatory from a formal point of view. From the HTTP spec at https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html: _"The reason phrases listed here are only recommendations -- they MAY be replaced by local equivalents without affecting the protocol"_.

It seems I left a leftover comment. Pending to be removed.

fgalan:
I'm afraid that travis has reached its limit with the new tests added in this PR :(

"The job exceeded the maximum time limit for jobs, and has been terminated."

Thus, ignore travis in this PR (we should rely in jenkins after merging)

src/lib/serviceRoutines/postUpdateContext.cpp
fgalan: Fixed in 2164edf

fgalan:
@fisuda this PR does a small modification to context_providers.md. I think it should be synched in the Japanese translation, pls.

You can do the PR using master as base branch or hardening/3068_cpr_regression_for_ngsiv2_based_in_ngsiv1 (the branch of this PR), as you prefer.

Thanks!

fisuda:
Noted with thanks.

Issue #3364, opened 2018-12-03
Deb based image for Orion


caa06d9c:
I've prepared image based on Debian (~112mb). I can refactor CI too. Is is a good idea?
[Link](https://github.com/caa06d9c/fiware-orion/tree/docker/docker)

`docker build -t fiware/orion:deb --build-arg REV=master`

fgalan:
Thanks for the link! I'll have a look as soon as I can.

For the records: current Orion image (https://hub.docker.com/r/fiware/orion/tags/) is around 105-108 MB

caa06d9c:
This size is compressed, real size at local storage ~280MB, so dev version in compressed size ~ 45mb, in real - 112

Issue #3287, opened 2018-09-07
Travis: parallelize functional test in sub-jobs


fgalan:
As far as I understand, travis runs sub-job in parallel:

![imagen](https://user-images.githubusercontent.com/1534240/45217063-03533c00-b2a3-11e8-94a8-1053a8000d3a.png)

Could we exploit this in order to paralelize functional testing? I mean, use the `--fromIx` and `--toIx` in order to run diferent blocks in different sub-jobs.

For instance, if we chose 100 tests as block sice we can have:

* functional1 sub-job with test from 1 to 100 (`./testHarnes --fromIx 1 --toIx 100`)
* functional2 sub-job with test from 101 to 200 (`./testHarnes --fromIx 101 --toIx 200`)
...
* functional11 sub-job with test from 1000 to 1100 (`./testHarnes --fromIx 1000 --toIx 1100`). Actually, we currently have 1030 but this will work. 

When we reach teset 1101 a functional12 sub-job will be created.

If this is feasible, it will speed up **a lot** the time travis takes to finish its job.

@caa06d9c what do you think?

fgalan:
I'm realizing that testHarness doesn't have a `--toIx` CLI...

```
fermin@neodeb:~/src/fiware-orion/test/functionalTest$ ./testHarness.sh -u
vie sep  7 13:42:08 CEST 2018
Usage: testHarness.sh [-u (usage)]
                      [-v (verbose)]
                      [--filter <test filter>]
                      [--match <string for test to match>]
                      [--keep (don't remove output files)]
                      [--dryrun (don't execute any tests)]
                      [--dir <directory>]
                      [--fromIx <index of test where to start>]
                      [--ixList <list of test indexes>]
                      [--skipList <list of indexes of test cases to be skipped>]
                      [--stopOnError (stop at first error encountered)]
                      [--no-duration (removes duration mark on successful tests)]
                      [--noCache (force broker to be started with the option --noCache)]
                      [--cache (force broker to be started without the option --noCache)]
                      [--noThreadpool (do not use a threadpool, unless specified by a test case. If not set, a thread pool of 200:20 is used by default in test cases which do not set notificationMode options)]
                      [ <directory or file> ]*
```

However, if the division of jobs described above is possible, it think it should worth the effort to implement `--toIx` in testHarness.sh. It shouldn't be too dificuult.

kzangeli:
--toIx could be implemented in 30 min, testing included

[...]

PR #3500, opened 2019-05-29
changes in ci

caa06d9c:
Changes:
- testHarness can be executed in several flows
- Debian platform added

Resolving: 
- https://github.com/telefonicaid/fiware-orion/issues/3287
- https://github.com/telefonicaid/fiware-orion/issues/3364

how to use:
`docker build -t fiware/orion .` - to build centos based image
`docker build --build-arg REV=${BRANCH} -t fiware/orion .` - to build centos based image, not from the master
`docker build --build-arg STAGE=deps -t fiware/orion:deps-centos .` - to build centos based support image with dependencies
`docker build -t fiware/orion -f Dockerfile-debian .` - to build debian based image


caa06d9c:
So it took 17 min 45 sec (but Travis was free, in other time it will take more time)

fgalan:
Thanks for your contribution!

However, I think it would be better to split this into two ones. I mean, it seems this PR mixes the split of functional tests in several stages on one hand (which is pretty straightforward to review) with changes in the docker artifacts under docker/ in the other hand (which is not so straightforward and need to be analyzed with more detail because is a bigger change).

Could you do a PR just with the splitting of tests? I guess it would be a matter of adjusting .travis.yml and maybe some existing files under ci/. Does it make sense?

fgalan:
I have found in a build of a different PR (https://travis-ci.org/telefonicaid/fiware-orion/jobs/539315869) the following error:

 ```
Error response from daemon: manifest for fiware/orion-ci:rpm7 not found
```

From https://hub.docker.com/r/fiware/orion-ci/tags, it seems that fiware/orion-ci:rpm7 has been removed.

Could you restore it, please? Other existing PR should not be affected while we are discussing this one (if at the end of this PR fiware/orion-ci:rpm7 is not needed any longer it could be removed, but please let's wait for that)

**Comment by caa06d9c** [comment]
Ops, sorry, restored

**Comment by fgalan** [comment]
Thanks for your prompt answer! :)

**Comment by caa06d9c** [comment]
@fgalan, could you review last changes?

File: ci/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,14 @@
+FROM centos:centos7
```
fgalan: Is this the same file previously in ci/rpm7/Dockerfile? In that case, I'd suggest to use the old place (to minimize changes in the PR that are more costly to review)

ci/build.sh
```diff
@@ -0,0 +1,373 @@
+#!/bin/bash
```
fgalan: Is this the same file previously in ci/rpm7/build.sh? In that case, I'd suggest to use the old place (to minimize changes in the PR that are more costly to review)

ci/Dockerfile
caa06d9c: Not the same,  rpm7 removed, because scenarios adopted to work with any platform and version

ci/build-centos.sh
```diff
@@ -0,0 +1,200 @@
+#!/bin/bash
```
fgalan: What's the difference between build-centos.sh and build.sh files?

caa06d9c: As it was before. build.sh provide an interface to control testing and other activities, build-${platform} - specific scenarios to work with platforms. Original idea was to unite all build scenarios into 1 recipe, to avoid duplication.

caa06d9c:
Don't merge it, I need to add 1 more commit

ci/build.sh
caa06d9c: Not the same, rpm7 removed, because scenarios adapted to work with any platform and version

fgalan:
As a general comment, I think we are mixing two things in this PR:

1. The splitting of the functional test in several travis stages
2. Improvements in dockerfiles, build scripts, etc. to make it more general and support more platforms.

Maybe I'm wrong, but they seem to be independent things.

Could we focus this PR only in item 1 applying only the minimal changes to make it possible from master, pls? I'd suggest to move improvements in dockerfiles, build scripts, etc. to a next PR, for the sace of clarify and reviewability.

fgalan:
@caa06d9c maybe we can do the following:

In other no bother you too much with this :), leave me some time to analyze this PR in deep and I'll try to do the minimum set of changes to parallelize in travis in a separate PR. Don't do any additional change on this until I provide more feedback.

Thanks for your help!

caa06d9c:
ok, thanks!

Issue #3505, opened 2019-06-11
Suboptimal DB queries


fgalan:
Queries to DB done by Orion could be suboptimal. In particular:

```
curl -H 'fiware-servicepath: /waste' localhost:1026/v2/entities/E

{ query: { $or: [ { _id.id: "E" } ], _id.servicePath: { $in: [ /^/waste$/ ] } }, orderby: { creDate: 1 } }
```

```
curl -H 'fiware-servicepath: /waste' localhost:1026/v2/entities/E?type=T

{ query: { $or: [ { _id.id: "E", _id.type: "T" } ], _id.servicePath: { $in: [ /^/waste$/ ] } }, orderby: { creDate: 1 } }
```

```
curl -H 'fiware-servicepath: /waste' localhost:1026/v2/entities

{ query: { $or: [ { _id.id: /.*/ } ], _id.servicePath: { $in: [ /^/waste$/ ] } }, orderby: { creDate: 1 } }
```

```
curl ' -H 'fiware-servicepath: /waste' localhost:1026/v2/entities?type=T

{ query: { $or: [ { _id.id: /.*/, _id.type: "T" } ], _id.servicePath: { $in: [ /^/waste$/ ] }, _id.type: { $exists: true } }, orderby: { creDate: 1 } }
```

```
curl -H 'fiware-service: sc_santander' -H 'fiware-servicepath: /waste' localhost:1026/v2/entities?q=A

{ query: { $or: [ { _id.id: /.*/ } ], _id.servicePath: { $in: [ /^/waste$/ ] }, attrs.A.value: { $exists: true } }, orderby: { creDate: 1 } }
```

```
curl -H 'fiware-servicepath: /waste' 'localhost:1026/v2/types?options=noAttrDetail'


pipeline = [ { $match: { "_id.servicePath": { $in: [ /^\/waste$/ ] } } }, { $project: { "_id.type": { $ifNull: [ "$_id.type", null ] }, attrNames: 1 } }, { $project: { attrNames: { $cond: [ { $eq: [ "$attrNames", [] ] }, [ null ], "$attrNames" ] } } }, { $unwind: "$attrNames" }, { $group: { _id: { $cond: [ { $in: [ "$_id.type", [ null, "" ] ] }, "", "$_id.type" ] }, attrs: { $addToSet: "$attrNames" } } }, { $sort: { _id: 1 } }, { $skip: 0 }, { $limit: 20 } ]
```

Sumary of issues that can be improved:

[...]

PR #3510, opened 2019-06-17
Optimice servicePath queries

fgalan:
Part of issue #3505:

> When we only have an element in servicePath, _id.servicePath: { $in: [ /^/waste$/ ] } can be simplified (not use $in and probably event not use a regex)


AlvaroVega:
LGTM

PR #3528, opened 2019-07-29
FIX documentation about Notify only attributes that change

fgalan:
Extra documentation related with #3190

@SourabhChourasiya-NEC if would be great if you could have a look and provide feedback/LGTM

CC: @fisuda the changes in this PR should be synched in the Japanese translation. Of course, any feedback from your side is also welcome. Thanks!

PR #3534, opened 2019-08-02
(JP) FIX documentation about Notify only attributes that change (#3528)

fisuda:


fgalan:
Maybe the entry for **onlyChanged** in https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals.jp/admin/database_model.md#csubs-collection should be included in this PR? 

You can find it in English at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/database_model.md#csubs-collection

fisuda:
Thank you for your comment. I updated the document.

PR #3572, opened 2019-11-06
FIX accumulator-server.py python requirements

fgalan:
Feedback from @WyRe is welcome.


WyRe:
I think that would be enough. Thank you :+1:

fgalan:
@fisuda this PR modifies a .md file. Could you sync the change in the Japanese translation, pls? Thanks!

Issue #3568, opened 2019-10-31
Notification flow control on update


fgalan:
Sometimes (especially with batch updates, i.e. `POST /v2/op/update`) a single update may trigger a large set of notifications which take some time to be processed (e.g. due to the receiver doesn't responses immediately). If the sender of the updates goes faster than notification processing (due to Orion responses immediately to the update sender, without waiting for notification finalization) this can cause saturation problems (typically, the notification queue gets full and new notifications are discarded). Thus, some flow control mechanism is needed.

Let's consider a flow control algorithm with depends on three parameters: notifications coefficient (C) from 0 to 1, evaluation step (S) in milliseconds, maximum time (M) in milliseconds.

The algorith is as follows:

1. When start to processing the update, get the current value of the notification queue. Let's name that value Q_0
2. During the processing of the update, count how many subscriptions are triggered (i.e. how many subscriptions are added to the notification queue). Let's name that value N.
3. Now that the processing as ended:
    1. Gets the new value of the queue: Q_i
    2. If Q_i <= Q_0 +(1 - C) * N then go to 4. Else wait for form S milliseconds and continue to 3.iii.
        * Note that if C = 1 corresponds to _highest flow control_. In this case: Q_i <= Q_0, that is: the size of the queue as large as we found it at the start of the update. 
       * Note that If C = 0 corresponds to _lowest flow control_. In this case: Q_i <= Q_0 + N, that is: we won't want to wait for any of the notifications to be processed (regarding the initial size of the queue at the start of the update).
   3. If the cumulative wait time is more than M, then go to 4. Otherwise, go back to 3.i (a value of 0 for M could mean always go to 3.1).
4. Orion response to the client of the update

Additional notes:

* The client must explicitly require flow control. This could be done with an NGSIv2 option, e.g. `?options=flowControl`.
* The flow control option could be applied to any NGSIv2 update operation. Note that although the one that would take more advantage of this new functionality is batch update even a single-entity update could trigger several notification in that entity has several matching subscriptions. From an implementation point of view, it doesn't involve an extra cost (as all updates are using the same core logic at mongoBackend at the end).
* A CLI will be used to enable flow control and define its parameters. For instance: `-flowControl C:S:M`. If this CLI is not active, `?options=flowControl` doesn't have any effect.
* Flow control needs the notification queue so it will be allowed only when threadpool notification mode is used.
[...]

PR #3577, opened 2019-11-14
Notification flow control

fgalan:
Implements #3568

Doc pending:

- [x] CLI doc
- [x] NGSIv2 notes doc
- [x] Perf tunning doc
- [ ] ~Devel manual?~ at the end, I think this is not necessary as the additions to mongoUpdateContext() are somehow a "side functionality", outside of the main flow described in the devel document


src/lib/mongoBackend/mongoUpdateContext.cpp
```diff
@@ -38,12 +38,60 @@
...
+  // get current size of the queue
+  unsigned int qi = ((QueueNotifier*) getNotifier())->queueSize();
+#endif
+  LM_T(LmtNotifier, ("flow control pass %d, delay %d, notification queue size is: %d",
+                     pass, accumulatedDelay, qi));
+
+
+  while (qi > target)
+  {
+    pass++;
```
fgalan: We should check that the connection with the client is still active or break in negative case. Not sure how to do this... we have a ciP->connection object (a MDH_Connection) which hopefully could be used to check this. However libmicrohttpd library documentation is not very clear on this. Maybe the MHD_get_connection_info() function can be used for this.

fgalan: It is not clear how to implement this at the presesnt moment (suggestions/hints welcome ;) so probably I'll move the text to a code comment with a FIXME mark. It is not a big issue, after all (only involves some inefficiency in the case of client cutting the connection, which is not usual).

fgalan: Comment added in commit 612b80d12a88236b1927641547d83491d95fb3ba

doc/manuals/admin/perf_tuning.md
```diff
@@ -122,6 +123,60 @@ on [the `notifQueue` block](statistics.md#notifqueue-block) may help you to tune
...
+
+Orion implements a flow control mechanism that allow to slow-down the updates flow sent by a client.
+Using flow control, Orion does not respond immediately to update request, but when the notifications
+triggered by the update has been sent (totally or partially, depending on the flow control
+configuration). This way, we avoid Orion saturation due to too much accumulated notifications
+in the notification queue (which, at the end, causes notifications being discarded when the queue
+is full).
+
+Flow control mechanism is configured using `-notifFlowControl` which takes three parameters:
+`-notifFlowControl gauge:stepDelay:maxInterval`. It requires threadpool notification mode
```
manucarrace: Three parameters. It is not clear the meaning of each parameter and its possible values.

fgalan: Actually is explained below. However, at this point the reader don't know it. I have slightly modified the text in 9a8663688bda2800ec8dc92b917cc2cb4010f447 to make it clearer.

fgalan:
@fisuda it would be great if you could modify cli.md, perf_tunnin.md and ngsiv2_implementation_nodes.md to sync the Japanese translation, please. Thanks!

fisuda:
Noted with thanks.

PR #3584, opened 2019-12-23
Bug/hyperlinks

kzangeli:
Hopefully fixed hyperlinks to installation guides for different distros


fgalan:
It seems this PR is not as intended... I understand it should be about adding/fixing some links in the corresponding .md files (given the body: *Hopefully fixed hyperlinks to installation guides for different distros*) but the PR includes much more (+89,681 −2,757 changes, more than 250+, a lot of conflicts, etc.).

Probably this PR should be closed and the right one with the hyperlinks stuff should be created from a fresh branch from master.

Issue #3550, opened 2019-09-13
Special attribute type to skip forbidden chars


fgalan:
Context Broker doesn't allow forbidden charts (https://fiware-orion.readthedocs.io/en/master/user/forbidden_characters/index.html) in attribute values. This can be restrictive in some scenarios.

We could define an special attributeType (for instance, "TextUnrestricted") so ContextBroker broker will not do checks on it. Eg:

```
"description": {
   "type": "TextUnrestricted",
   "value": "I'm a description (and I'm using forbidden chars)"
}
```

Of course, the user should know the potential security implications of this and assume them in his client code.

This would be backward compatible and enhance Orion.

fgalan:
Implemented by PR https://github.com/telefonicaid/fiware-orion/pull/3592

fgalan:
@pcanoesp could you provide the request you are sending to the CB, pls? verb + url + payload

pcanoesp:
In our real case of use, payload filed is the one including forbiden characters (base64 coding)

PR #3592, opened 2020-01-17
ADD TextUnrestricted attribute type to avoid forbidden chars checking

fgalan:
Implements #3550


fgalan:
@fisuda it would be great if you could include the modifications in the couple of files in this PR also in the Japanese translation, please. Thanks!

fisuda:
Noted with thanks!

PR #3600, opened 2020-01-23
FIX improve documentation

fgalan:


fgalan:
@fisuda it would be great if you could sync this change in the Japanese translation, please. Thanks!

fisuda:
Noted with thanks.

Issue #3460, opened 2019-03-19
GET /v2/entities doesnt return 500 Internal Sever error when DB is failing


fgalan:
Orion returns 200 OK and `[]` for GET /v2/entities when DB fails. It should return 500 Internal Server Error.

Note that GET /v2/subscriptions is behaving correctly in this aspect (have a look and copy the same behaviour).

ArqumFarooqui-NEC:
Hi @fgalan  @kzangeli sir

As per my investigation, `500 Internal Server Error` is returning as response when the queries `GET /v2/subscriptions`, `GET /v2/subscriptions/<subID>` and `GET /v2/entities/<entID>` passed to Orion where DB is failing.

As the behavior of `GET /v2/entities/<entID>` is also correct so I'm following current implementation of the same and will implement the similar behavior to `GET /v2/entities`

Please confirm my understanding and approach and please assign this issue to me.

fgalan:
I think your understanding it correct. Thanks!

ArqamFarooqui110719:
Hi @fgalan sir,
As I have commented on [PR](https://github.com/telefonicaid/fiware-orion/pull/3616#issuecomment-920617623) of this issue, that this issue has been fixed by other PR and currently `GET v2/entities` API is returning `500 Internal Server Error` when DB fails.
So this issue should be closed.

fgalan:
> Hi @fgalan sir,
> As I have commented on [PR](https://github.com/telefonicaid/fiware-orion/pull/3616#issuecomment-920617623) of this issue, that this issue has been fixed by other PR and currently `GET v2/entities` API is returning `500 Internal Server Error` when DB fails.
> So this issue should be closed.

I agree. Closing.

Thanks!

PR #3616, opened 2020-01-30
Fix status code for v2/getEntities when DB fails

Chandradeep-NEC:
Fix for Issue #3460


src/lib/serviceRoutinesV2/getEntities.cpp
```diff
@@ -314,31 +314,23 @@ std::string getEntities
 
   // 03. Render Entities response
 
-  if (parseDataP->qcrs.res.contextElementResponseVector.size() == 0)
```
fgalan: Not sure if the proposed change is correct...

It seems you have removed some checks (the one for `parseDataP->qcrs.res.contextElementResponseVector.size()` in the original external if clause and the one for `oe.code != SccNone` in the original internal if clause), I understand that with the aim of simpliying the logic.

However, although I don't remember why, I understand these checks are here for some reason :). In fact, if you look to the travis run of the functional tests at https://travis-ci.org/telefonicaid/fiware-orion/jobs/643774476?utm_medium=notification&utm_source=github_status you will see that some of the them, probably for this reason (did you run the test locally before creating the PR?)

Not sure, but maybe the issue should be solved in a lower part (in mongoBackend)

fgalan:
Any update in the progress of this PR, please?

Thanks!

fgalan:
This PR has opened comments that would need an answer, please. In addition, they have conflict that needs to be solved.

ArqamFarooqui110719:
> This PR has opened comments that would need an answer, please. In addition, they have conflict that needs to be solved.

Hi @fgalan sir,
Sorry for late reply, I was busy in some other priority tasks.
I have looked into the issue, and currently `GET v2/entities` API is returning 500 Internal Server Error when DB fails. So I think this issue has been resolved. For reference I'm attaching screenshots:
**When DB is active:**

Screenshot (360)

**When DB is exited(fail):**

Screenshot (362)

If your understanding is also same then Please close this PR and issue #3460

fgalan:
I agree with your analysis: the issue no longer occurs. Not sure when it happened, but probably some other issue fix has indirectly fixed this one.

Thus, I understand this PR is no longer needed so it will be closed. Thanks!

PR #3627, opened 2020-02-13
FIX CMakeLists.txt compiler flags bases in gcc veresion instead of distro version

fgalan:
This way the build system will be more resilient to the evolution of distributions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,15 +78,14 @@ add_definitions(-fPIC)
 # This workaround consists of putting back the -Werror flag and instead solving the problem by
 # adding another flag: -Wno-deprecated-declarations
 #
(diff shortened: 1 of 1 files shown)


fgalan:
_Testing the "version less than 6.3" if-branch:_

In travis run (https://travis-ci.org/telefonicaid/fiware-orion/builds/650105041?utm_source=github_status&utm_medium=notification), based in a CentOS 7 image (gcc version 4.8.5)

For unit test:

imagen

For functional test:

imagen

_Testing the "version equal or hight than 6.3" if-branch:_

Building in Debian 9 (gcc version 6.3.0) I get (both in make debug and make unit_test):

```
gcc version 6.3 or higher, compiling with -Wno-deprecated-declarations
```
-----------

Thus, under my understanding, the change works in both cases.

Issue #3434, opened 2019-02-12
"Special Metadata Types" in NGSIv2 specification


fgalan:
Looking to the [NGSIv2 specification](http://fiware.github.io/specifications/ngsiv2/stable/) we have a section for "Special Attribute Types". 

However, we don't have a section for "Special Metadata Types". The DateTime attribute type is the only one to be included there (note that in "Builtin Metadata" section we mention DateTime).

NGSI v2.0 was already released so in theory no new version can be done without changing the spec. But maybe a v2.0.1 with errata-like modifications (like this one) is aceptable.

ArqumFarooqui-NEC:
Hi @fgalan sir,
I’ve started working on this issue. Please assign this issue to me.
As per my understanding: I’ll have to add a doc section for `Special Metadata Types` i.e. only `DateTime` attribute.
Please confirm the above understanding??
As per my investigation, this new section can be added in `fiware-orion/blob/master/doc/manuals/user/metadata.md` file?

fgalan:
Of course, I can assign this issue to you if you want to work on it... however, maybe it is not a very interesting issue. Let me explain myself.

Note we don't plan to release NGSIv2 errata version soon, so maybe your changes in the .apib file (the one that should be modified) will not appear in places like this one http://telefonicaid.github.io/fiware-orion/api/v2/stable in the next future.

Thus, I'd recommend to leave this issue on hold and choose another more profitable one. What do you think?

ArqumFarooqui-NEC:
I've done some investigation on this and I think I'm bit closure. Once I'll raise the PR and then I'll pick another one. 
What do you think like this, please let me know your opinion?

fgalan:
If you have started the work and are close to the finish, then I think the best idea is to raise the PR (my comment arrived late... sorry for that ;)

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/3654

PR #3652, opened 2020-05-19
Issue/3434

ArqumFarooqui-NEC:
Included the section for `Special Metadata Types`


fgalan:
Thank for your contribution!

I have had a look to the new section and I think it would be better to copy the same style than for the attributes built-in section. I'd suggest to copy-paste them and change only the differences (note that even in some casess the name of the built-in attribute matches the name of similar metadata, pe. dateCreated).

In addition, try to work on a cleaner way with regards the number of commits. Over-commiting in a PR is in general not a good idea, as it adds noise. This PR has now 15 commits and I wonder if it could be has been done in just one. Please, work on your working copy of the repository (not in github.com) and do all the changes derived of a review round all them in the same committ. Thanks!

doc/apiary/v2/fiware-ngsiv2-2.0-2018_09_15.apib
```diff
@@ -354,6 +354,35 @@ The list of builtin attributes is as follows:
 Like regular attributes, they can be  used in `q` filters and in `orderBy`.
 However, they cannot be used in resource URLs.
 
+## Special Metadata Types
+
+Similar to attributes, each piece of metadata has a metadata name, a metadata type, a metadata value.
+The metadata field can also be used to explicitly include some special metadata e.g. dateCreated,
+dateModified and dateExpires i.e. for attribute creation date, last modification date and expiration date
+respectively (These are not included by default).
```
fgalan, line 362:
```suggestion

Generally speaking, user-defined metadata types are informative; they are processed by the NGSIv2
server in an opaque way. Nonetheless, the types described below are used to convey a special
meaning:
```

doc/apiary/v2/fiware-ngsiv2-2.0-2018_09_15.apib
```diff
@@ -354,6 +354,35 @@ The list of builtin attributes is as follows:
...
+```
+
+Datetimes are composed of date, time and timezone designator, in one of the following patterns:
+<date>
+<date>T<time>
+<date>T<time><timezone>
+
+Orion always provides datetime metadata using the format YYYY-MM-DDThh:mm:ss.ssZ. It uses UTC/Zulu timezone 
+(which is the best default option, as clients/receivers may be running in any timezone). 
+For more details about DateTime follow [DateTime section](https://fiware-orion.readthedocs.io/en/master/user/ngsiv2_implementation_notes/index.html)
```
fgalan, line 384: By similarity with "Special Attribute Types" section, I'd suggest to delete this block of lines.

fgalan:
> I have had a look to the new section and I think it would be better to copy the same style than for the attributes built-in section. I'd suggest to copy-paste them and change only the differences (note that even in some casess the name of the built-in attribute matches the name of similar metadata, pe. dateCreated).

I have provided more precise comments about this in the PR.

In addition, please add the following entry:

> Add "Special Metadata Types" section

just above "Changes since RC-2018.07:" line (for a future changelog, if needed).

Thanks!

fgalan:
Something weird has happened with this PR... it shows no content in files changes...

I see a "force-pushed" was done. Be careful with that operation... in general it is not recomended. It can be dangerous.

Chandradeep-NEC:
> Something weird has happened with this PR... it shows no content in files changes...
> 
> I see a "force-pushed" was done. Be careful with that operation... in general it is not recomended. It can be dangerous.

Dear @fgalan 

We are sorry for such commits in the PR. I'm assuring you as same will not be happen in future.

Thanks!!

fgalan:
> We are sorry for such commits in the PR. I'm assuring you as same will not be happen in future.

Don't worry. Something it happens ;)

Thus, what's next? You will retry the same modifications again? Maybe it's a good idea to do it in a fresh new PR (in that case, please try to use only a small number of commits for the whole contribution)?

doc/apiary/v2/fiware-ngsiv2-2.0-2018_09_15.apib
ArqumFarooqui-NEC: Fixed in 0a39e5b in PR #3654

doc/apiary/v2/fiware-ngsiv2-2.0-2018_09_15.apib
ArqumFarooqui-NEC: Fixed in 0a39e5b in PR #3654

ArqumFarooqui-NEC:
> In addition, please add the following entry:
> 
> > Add "Special Metadata Types" section
> 
> just above "Changes since RC-2018.07:" line (for a future changelog, if needed).
> 
Fixed in 0a39e5b in PR #3654 

Hi @fgalan sir, I have created a new PR #3654 , for issue #3434

fgalan:
Ok, thanks. Thus, let's close this one and continue in PR #3654

PR #3668, opened 2020-07-02
feature/miliseconds_in_timestamps

kzangeli:
Creation date and modification date now have milliseconds

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,2 +1,3 @@
 - Fix: wrong 204 code was responded (instead of 501) when using idPattern in POST /v2/op/update operation (#3427)
 - Fix: changed the name of pretty much all the environment variables for Orion. E.g. adding prefix 'ORION_'
+- Add: Creation date and modification date now have milliseconds
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,3 +1,3 @@
 - Fix: wrong 204 code was responded (instead of 501) when using idPattern in POST /v2/op/update operation (#3427)
 - Fix: changed the name of pretty much all the environment variables for Orion. E.g. adding prefix 'ORION_'
-- Add: milliseconds support in DateTime attributes and metadata, included dateCreated and dateModified built-ins (#432, #2670, #3412, #3666, )
+- Add: milliseconds support in DateTime attributes and metadata, included dateCreated and dateModified built-ins (#432, #2670, #3412, #3666)
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -1022,8 +1022,8 @@ std::string isodate2str(double timestamp)
   // We've seen that e.g. the number 12.60 is stored as 12.599999999 and as here we cut
   // the number it ends up as 12.59, which is no good ...
   //
-  // Two choices here, enther call some rounding function, or use a trick.
-  // The trick is to add a microsecond (we're only interested in milliseconds) and then cut
+  // Two choices here, either call some rounding function, or use a trick.
+  // The trick is to add a microsecond (we're only interested in milliseconds) and then cut.
   //
   // Assuming we have nine decimals: 12.599999999, adding a microsecond to this gives 12.600000999:
   //
@@ -1034,12 +1034,12 @@ std::string isodate2str(double timestamp)
   // If instead the rounding error would go on the upper side, i.e. 12.60 would be 12.6000000001, adding that microsecond
   // doesn't change anything. All still work just fine.
   //
-  // Better example: what if we wish to store 12.5999?
-  // A typical rounding error on the "upper side" would then be 12.599900001, and again, adding that microsecond after
+  // Better example: what if we wish to store 12.599?
+  // A typical rounding error on the "upper side" would then be 12.599000001, and again, adding that microsecond after
   // cutting doesn't change a thing.
   //
   //
-  // So, the algorithm is as follows:
+  // So, the "trick algorithm" is as follows:
   //
   // 1. Get the integer part of "double timestamp" - this cuts all decimals, no rounding is done,
   //    and store it in the variable 'seconds'
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -1011,10 +1011,13 @@ std::string isodate2str(double timestamp)
 {
   // 80 bytes is enough to store any ISO8601 string safely
   // We use gmtime() to get UTC strings, otherwise we would use localtime()
-  // Date pattern: 1970-04-26T17:46:40.00Z
+  // Date pattern: 1970-04-26T17:46:40.000Z
+
   char    buffer[80];
   time_t  seconds   = (time_t) timestamp;
-  int     millis    = (timestamp * 1000) - (seconds * 1000);   // (timestamp - seconds) * 1000 gives rounding errors ...
+  double  ms        = timestamp - (double) seconds;
+  int     micros    = ms * 1000000;
+  int     millis    = (micros + 1) / 1000;   // (timestamp - seconds) * 1000 gives rounding errors ...
 
   strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%S", gmtime(&seconds));
 
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -999,6 +999,7 @@ std::string double2string(double f)
 }
 
 
+
 /*****************************************************************************
 *
 * isodate2str -
@@ -1006,14 +1007,19 @@ std::string double2string(double f)
 * FIXME P6: change implementation to use gmtime_r
 *
 */
-std::string isodate2str(long long timestamp)
+std::string isodate2str(double timestamp)
 {
   // 80 bytes is enough to store any ISO8601 string safely
   // We use gmtime() to get UTC strings, otherwise we would use localtime()
   // Date pattern: 1970-04-26T17:46:40.00Z
-  char   buffer[80];
-  time_t rawtime = (time_t) timestamp;
-  strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%S.00Z", gmtime(&rawtime));
+  char    buffer[80];
+  time_t  seconds   = (time_t) timestamp;
+  int     millis    = (timestamp * 1000) - (seconds * 1000);   // (timestamp - seconds) * 1000 gives rounding errors ...
+
+  strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%S", gmtime(&seconds));
+
+  char* eob = &buffer[strlen(buffer)];
+  sprintf(eob, ".%03dZ", millis);
   return std::string(buffer);
 }
 
--- a/src/lib/common/string.cpp
+++ b/src/lib/common/string.cpp
@@ -1014,6 +1014,39 @@ std::string isodate2str(double timestamp)
   // Date pattern: 1970-04-26T17:46:40.000Z
 
   char    buffer[80];
+
+  //
+  // About the added microsecond in  **int millis = (micros + 1) / 1000;**
+  //
+  // Floating point numbers are often a problem for a CPU, giving funny rounding errors.
+  // We've seen that e.g. the number 12.60 is stored as 12.599999999 and as here we cut
+  // the number it ends up as 12.59, which is no good ...
+  //
+  // Two choices here, enther call some rounding function, or use a trick.
+  // The trick is to add a microsecond (we're only interested in milliseconds) and then cut
+  //
+  // Assuming we have nine decimals: 12.599999999, adding a microsecond to this gives 12.600000999:
+  //
+  //  12.599999999 + 0.000001 == 12.600000999
+  //
+  // After that we cut it at three decimals and end up with 12.600.
+  //
+  // If instead the rounding error would go on the upper side, i.e. 12.60 would be 12.6000000001, adding that microsecond
+  // doesn't change anything. All still work just fine.
+  //
+  // Better example: what if we wish to store 12.5999?
+  // A typical rounding error on the "upper side" would then be 12.599900001, and again, adding that microsecond after
+  // cutting doesn't change a thing.
+  //
+  //
+  // So, the algorithm is as follows:
+  //
+  // 1. Get the integer part of "double timestamp" - this cuts all decimals, no rounding is done,
+  //    and store it in the variable 'seconds'
+  // 2. Store the decimals in 'ms', by extracting the integer part ('seconds') from 'timestamp'
+  // 3. Convert the decimals into microseconds, by multiplicating with one million (1000000) - in the variable 'micros'.
+  // 4. Add a microsecond and then divide by one thousand - cutting all decimals
+  //
   time_t  seconds   = (time_t) timestamp;
   double  ms        = timestamp - (double) seconds;
   int     micros    = ms * 1000000;
--- a/src/lib/mongoBackend/safeMongo.cpp
+++ b/src/lib/mongoBackend/safeMongo.cpp
@@ -108,8 +108,8 @@ BSONObj getObjectField(const BSONObj& b, const std::string& field, const std::st
   }
   else
   {
-    LM_E(("Runtime Error (field '%s' was supposed to be an object but the type is '%s' in BSONObj <%s> from caller %s:%d)",
-          field.c_str(), mongoTypeName(b.getField(field).type()), b.toString().c_str(), caller.c_str(), line));
+    LM_E(("Runtime Error (field '%s' was supposed to be an object but the type is '%s' (type as integer: %d) in BSONObj <%s> from caller %s:%d)",
+          field.c_str(), mongoTypeName(b.getField(field).type()), b.getField(field).type(), b.toString().c_str(), caller.c_str(), line));
   }
   return BSONObj();
 }
@@ -139,8 +139,8 @@ BSONArray getArrayField(const BSONObj& b, const std::string& field, const std::s
   }
   else
   {
-    LM_E(("Runtime Error (field '%s' was supposed to be an array but the type is '%s' in BSONObj <%s> from caller %s:%d)",
-          field.c_str(), mongoTypeName(b.getField(field).type()), b.toString().c_str(), caller.c_str(), line));
+    LM_E(("Runtime Error (field '%s' was supposed to be an array but the type is '%s' (type as integer: %d) in BSONObj <%s> from caller %s:%d)",
+          field.c_str(), mongoTypeName(b.getField(field).type()), b.getField(field).type(), b.toString().c_str(), caller.c_str(), line));
   }
   return BSONArray();
 }
@@ -169,8 +169,8 @@ std::string getStringField(const BSONObj& b, const std::string& field, const std
   }
   else
   {
-    LM_E(("Runtime Error (field '%s' was supposed to be a string but the type is '%s' in BSONObj <%s> from caller %s:%d)",
-          field.c_str(), mongoTypeName(b.getField(field).type()), b.toString().c_str(), caller.c_str(), line));
+    LM_E(("Runtime Error (field '%s' was supposed to be a string but the type is '%s' (type as integer: %d) in BSONObj <%s> from caller %s:%d)",
+          field.c_str(), mongoTypeName(b.getField(field).type()), b.getField(field).type(), b.toString().c_str(), caller.c_str(), line));
   }
 
   return "";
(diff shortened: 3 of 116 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,2 +1,3 @@
 - Fix: wrong 204 code was responded (instead of 501) when using idPattern in POST /v2/op/update operation (#3427)
 - Fix: changed the name of pretty much all the environment variables for Orion. E.g. adding prefix 'ORION_'
+- Add: Creation date and modification date now have milliseconds
```
fgalan, line 3:
```suggestion
- Add: Creation date and modification date now have milliseconds (#432, #2670, #3412, #3666, )
```

I think these issues are about miliseconds support, so should be mentioned.

src/lib/mongoBackend/safeMongo.cpp
```diff
@@ -69,8 +108,8 @@ BSONObj getObjectField(const BSONObj& b, const std::string& field, const std::st
   }
   else
   {
-    LM_E(("Runtime Error (field '%s' was supposed to be an object but type=%d in BSONObj <%s> from caller %s:%d)",
-          field.c_str(), b.getField(field).type(), b.toString().c_str(), caller.c_str(), line));
+    LM_E(("Runtime Error (field '%s' was supposed to be an object but the type is '%s' in BSONObj <%s> from caller %s:%d)",
```
fgalan, line 111: Maybe including both number and string could be useful (general comment for all similar LM_E in this PR).

src/lib/mongoBackend/safeMongo.cpp
```diff
@@ -288,15 +327,52 @@ bool getBoolField(const BSONObj& b, const std::string& field, const std::string&
...
   return false;
 }
 
 
 
+/* ****************************************************************************
+*
+* getNumberFieldAsDouble -
+*/
+double getNumberFieldAsDouble(const BSONObj& b, const std::string& field, const std::string& caller, int line)
```
fgalan, line 343: What the different between this new function and existing `double getNumberField()` ?

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
```diff
@@ -49,7 +49,7 @@ payload='{
   "type": "T",
   "A": "foo"
 }'
-T_0=$(TZ=GMT date +"%Y-%m-%dT%H:%M:%S.00Z")
+T_0=$(TZ=GMT date +"%Y-%m-%dT%H:%M:%S")
```
fgalan, line 52: I understand `date` command is not able to provide date with miliseconds precission, isn't it?

test/functionalTest/cases/1156_mq_as_uri_param_for_metadata/mq_and_spaces_in_metadata_value.test
```diff
@@ -0,0 +1,101 @@
+# Copyright 2016 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan, line 1: This new .test doesn't seem related with DateTime support... maybe I'm missing something?

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
```diff
@@ -0,0 +1,90 @@
...
+# 02. GET E1 and see its 'ts' attribute with its milliseconds
+#
+
+echo "01. Create an Entity E1, with an attribute 'ts' that is of DateTime type and with a value with milliseconds"
+echo "==========================================================================================================="
+payload='{
+  "id": "E1",
+  "ts": {
+    "type": "DateTime",
+    "value": "2020-07-03T09:24:04.456"
```
fgalan, line 43: Although tis is a format perfectly valid in ISO8601, for completness, I'd suggest to add another cases using:

```
"2020-07-03T09:24:04.456Z" (to test Z timezone)
"2020-07-03T09:24:04.456+0100" (to test a not Z timezone)
```

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
```diff
@@ -0,0 +1,90 @@
...
+===========================================================================================================
+HTTP/1.1 201 Created
+Content-Length: 0
+Location: /v2/entities/E1?type=Thing
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+
+
+02. GET E1 and see its 'ts' attribute with its milliseconds
```
fgalan, line 69: Maybe we should test also the value is correctly rendered in notifications?

test/functionalTest/cases/3162_context_providers_legacy_non_primitive/blanked_out_attr.test
```diff
@@ -199,9 +199,9 @@ Date: REGEX(.*)
     ],
     "attrs": {
         "address": {
-            "creDate": REGEX([0-9]+),
+            "creDate": REGEX([0-9]+\.[0-9]+),
```
fgalan, line 202: I understand this time we cannot use `[0-9][0-9][0-9]` or `\d\d\d` instead of `[0-9]+` or `\d+` because in the DB due to rounding effect more than 3 decimal digits may appear, isn't it?

test/functionalTest/cases/3442_ngsiv1_interop_location_fix/ngsiv1_geobox_create_does_not_break_location.test
```diff
@@ -39,18 +39,18 @@ echo "01. Create entity with geo:box with NGSIv1 (APPEND)"
...
+      "type": "E",
+      "isPattern": "false",
+      "id": "E1",
+      "attributes": [
+        {
+          "name": "position",
+          "type": "geo:box",
+          "value": ["7.7, 8.8", "9.9, -10.10"]
+        }
+      ]
```
fgalan, line 52: Why this change? Indent only?

test/functionalTest/testHarness.sh
```diff
@@ -146,8 +146,15 @@ function usage()
...
-  echo "CB_SKIP_FUNC_TESTS:  list of names of func tests to skip"
+  echo "CB_MAX_TRIES:             The number of tries before giving up on a failing test case"
+  echo "CB_SKIP_LIST:             Default value for option --skipList"
+  echo "CB_SKIP_FUNC_TESTS:       List of names of func tests to skip"
+  echo "CB_NO_CACHE:              Start the broker without subscription cache (if set to 'ON')"
+  echo "CB_THREADPOOL:            Start the broker without thread pool (if set to 'OFF')"
+  echo "CB_DIFF_TOOL:             To view diff of failing tests with diff/tkdiff/meld/... (e.g. export CB_DIFF_TOOL=tkdiff)"
+  echo "CB_WITH_EXTERNAL_BROKER:  The broker is started externally - not 'automatically' by the test harness (if set to 'ON')"
+  echo
+  echo "Please note that, if using CB_WITH_EXTERNAL_BROKER (or --xbroker, which is the same), only a single test case should be run."
```
fgalan, line 157: Fixes in the -u implementation?

src/lib/mongoBackend/safeMongo.cpp
kzangeli: This new function accepts int32, int64 or double in mongo and returns a double.
Does getNumberField do that?
If it does, then no difference whatsoever.
I will check - didn't imagine a function like that would exist already :)

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
kzangeli: I don't know. Didn't check that - didn't even think about it to tell you the truth . ..

test/functionalTest/cases/1156_mq_as_uri_param_for_metadata/mq_and_spaces_in_metadata_value.test
kzangeli: Ah, true, this is an old functest I found in my cloned repo, from some older work.
It got added by mistake (I had many many FTs modified so I did: `git add test/func*/cases/*/*.test`).

Why don't you take a look at it and tell me if you're interested in keeping it. If not, I'll just remove it.
I think it was forgotten in some old PR

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
kzangeli: There are many many functests already with rendered milliseconds.

test/functionalTest/cases/3162_context_providers_legacy_non_primitive/blanked_out_attr.test
kzangeli: I guess we could do something better, I just copied the same approach that was already in the functest - there are many different approaches in the functests and I tried to always use the same one

test/functionalTest/cases/3442_ngsiv1_interop_location_fix/ngsiv1_geobox_create_does_not_break_location.test
kzangeli: Yeah, it hurt my eyes. Sorry

test/functionalTest/testHarness.sh
kzangeli: Yeah, I needed my '-eb (external broker)' to run the broker under gdb.
So, I stole it from Orion-LD, and while I was at it I copied the entire list of env vars, as Orion's list was incomplete.
Then I found out that my '-ld' already existed, called '-xbroker' ...

I left the added env vars as I think it's an improvement.

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
fgalan: What I mean, if at the end a new .test (2670_milliseconds_for_DateTime_attrs.test) has been added to check milliseconds rendering (which I think is a good idea) let's do it checking the two main ways of rendering context information: synchronous GET and asynchronous notifications.

kzangeli: Sure, no probs

src/lib/mongoBackend/safeMongo.cpp
kzangeli: So, looked it up.
getNumberField only accepts double. Any other type in mongo will result in an error.
I'd suggest a change in the name of the function to "getDoubleField".
Not really for this PR though

NTC

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
kzangeli: `man date`:
%N     nanoseconds (000000000..999999999)

However, it doesn't help.
If we get nanoseconds we still can't compare with the milliseconds we now get from the broker.

NTC

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
kzangeli: This is a little bit out of scope for this PR.
I'll test it - if it works, then all good, I'll include it. If it does not work, then I'd propose to create an issue for it.
Has this ever been tested in Orion?

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
kzangeli: Fixed in 6d18bc1

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
fgalan: Yes. Orion supports timezones and we have .test covering it (as far as I remember). See https://fiware-orion.readthedocs.io/en/master/user/ngsiv2_implementation_notes/index.html#datetime-support

src/lib/mongoBackend/safeMongo.cpp
kzangeli: Fixed in a794057

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
kzangeli: Yes, you are right - it works.
Fixed in e8a770d

CHANGES_NEXT_RELEASE
fgalan: Fixed in 75e4f2b

test/functionalTest/cases/1156_mq_as_uri_param_for_metadata/mq_and_spaces_in_metadata_value.test
fgalan: I'll have a closer look but I think it can be kept. I undestand it would make stronger the regression on old issue #1156 

NTC

test/functionalTest/cases/3162_context_providers_legacy_non_primitive/blanked_out_attr.test
fgalan: Good enough. NTC

test/functionalTest/cases/3442_ngsiv1_interop_location_fix/ngsiv1_geobox_create_does_not_break_location.test
fgalan: It makes review a little harder ;) but it's fine. NTC.

test/functionalTest/testHarness.sh
fgalan: Yes, I agree it's an improvement. NTC.

src/lib/common/string.cpp
```diff
@@ -999,21 +999,30 @@ std::string double2string(double f)
...
-  char   buffer[80];
-  time_t rawtime = (time_t) timestamp;
-  strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%S.00Z", gmtime(&rawtime));
+  // Date pattern: 1970-04-26T17:46:40.000Z
+
+  char    buffer[80];
+  time_t  seconds   = (time_t) timestamp;
+  double  ms        = timestamp - (double) seconds;
+  int     micros    = ms * 1000000;
+  int     millis    = (micros + 1) / 1000;   // (timestamp - seconds) * 1000 gives rounding errors ...
```
fgalan, line 1020: Maybe a comment should be added explaining why the "+1" is needed? It can be a little tricky...

src/lib/mongoBackend/safeMongo.cpp
fgalan: I'll have a look to both functions once this PR gets merged. I have the "feeling" that they are so close they could be unified, but not sure of it.

As you said, not in this PR.

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
```diff
@@ -135,13 +135,13 @@ echo
 
 echo "06. Check dateCreate == t_0 and dateModified == t_1"
 echo "==================================================="
-if [ "$CRE_DATE" == "$T_0" ]
+if [[ "$CRE_DATE" == "$T_0"* ]]
```
fgalan, line 138: I think it would be clearer to keep this (and other) ifs as they were, removing the milis part at CRE_DATE setting date. Something like this:

```
# We remove miliseconds from variables as the reference T_0 and T_1 dates doesn't include them
CRE_DATE=$(echo "$_response" | python -mjson.tool | grep dateCreated | awk -F '"' '{print $4}' | awk -F '.' '{print$2}')
MOD_DATE=$(echo "$_response" | python -mjson.tool | grep dateModified | awk -F '"' '{print $4}' | awk -F '.' '{print$2}')
```

Similar comment applicable to other .test (test/functionalTest/cases/0876_attribute_dates/attribute_dates.test and not use if others9

test/functionalTest/cases/2670_milliseconds_for_DateTime_attrs/2670_milliseconds_for_DateTime_attrs.test
```diff
@@ -0,0 +1,178 @@
+# Copyright 2020 Telefonica Investigacion y Desarrollo, S.A.U
```
fgalan, line 1: I would suggest to add a new .test to test miliseconds filtering capabilties. Something like this

```
01. Create E1 with attribute t=2020-05-10T01:02:03.123Z
02. Create E2 with attribute t=2020-05-10T01:02:03.456Z
03. Create E3 with attribute t=2020-05-10T01:02:03.789Z
04. Query for entities t>2020-05-10T01:02:03Z, get E1, E2, E3
05. Query for entities t<2020-05-10T01:02:03.500Z, get E1
06. Query for entities t=2020-05-10T01:02:03.200Z...2020-05-10T01:02:03.600Z, get E2
```

and a simila test based in sub filtering instead of GET filtering.

fgalan: Discussed with @kzangeli at skype, this will be done in a next PR. @fgalan will do it. NTC

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
fgalan: Discussed with @kzangeli at skype, this will be done in a next PR. @fgalan will do it. NTC

kzangeli: Perhaps ... perhaps this works:
```
if [[ "$CRE_DATE" == "$T_0.???Z" ]]
```

Worth a try ...

src/lib/common/string.cpp
kzangeli: Fixed in 4d8fe4857

fgalan: Fixed in 4d8fe48

test/functionalTest/cases/0000_deprecated_checkings/get_entity_dates_with_options.test
fgalan: I'll try that.

fgalan: Tested. I'm afraid it doesn't work.

kzangeli: Pity.
Frankly, I believe the current solution is pretty good, only that is doesn't make sure there are exactly three decimals and also that the trailing 'Z' is present.

If you want it "even better", I'm sure this can be solved using regex in bash though.
It seems the "=~" operator is used.
Found a SOF question about something similar: https://stackoverflow.com/questions/18709962/regex-matching-in-a-bash-if-statement/18710850.

kzangeli: I found a way:

```
dt="2020-07-03T09:24:04"
dtWithMillis="$dt.456Z"
echo dt: $dt
echo dtWithMillis: $dtWithMillis


if [[ "$dtWithMillis" =~ $dt.[0-9]{3}Z ]]
then
  echo MATCH
else
  echo NO MATCH
fi
```

fgalan: Even in the case comparison condition can be changed, the statement in case of fail

```
   echo "dateCreated: not t_0 ($CRE_DATE != $T_0)"
```

would be a bit weird, as CRE_DATE and T_0 are not sharing a common reference format. So, I think that using the "mínimo común denominador" (I don't know right now how is in English :) of both formats is better. I mean, in this case, to take the one used by `date` as a minimum and "tail" part of the other.

Test modified according to this idea in PR https://github.com/telefonicaid/fiware-orion/pull/3672

Issue #3694, opened 2020-09-11
Improvements in Orion logs system


fgalan:
The following modifications has to be done in the Orion Context Broker logs system:

* Move all the current LM_I traces to LM_T, choosing a suitable (new) tracelevel in DEBUG level **except** the trace about Orion startup, that must remain in INFO level
* Introduce the following new traces at INFO level
    * At the end of each GET request on the CB API endpoint, log the URL used and the response status code to that request. No payload is loged.
    * At the end of each non-GET request on the CB API endpoint, log the URL used, the response status code to that request and the request payload
    * At the end of each notification send, log the URL of the endpoind and subId of the associated subcription. No payload is logged
* For notifications, add the following token to the correlator ID: `; notif=n` where "n" is an autoincremental counter (1, 2, 3, etc.) local to the e2e flow that caused the nofiication. I mean, if a given update has `corr=10405986-e4e7-4ee4-95c7-bfec5282e3a4` and three notifications are generated as consecuence of that update, then notifications will have the following correlators:
    * `corr=10405986-e4e7-4ee4-95c7-bfec5282e3a4; notif=1`
    * `corr=10405986-e4e7-4ee4-95c7-bfec5282e3a4; notif=2`
    * `corr=10405986-e4e7-4ee4-95c7-bfec5282e3a4; notif=3`
* Modify documentation in logs.md

fgalan:
> Move all the current LM_I traces to LM_T, choosing a suitable (new) tracelevel in DEBUG level except the trace about Orion startup, that must remain in INFO level

The following old-INFO traces will remain at the end:

```
src/app/contextBroker/contextBroker.cpp:    LM_I(("Orion context broker exiting due to receiving a signal"));
src/app/contextBroker/contextBroker.cpp:    LM_I(("Orion context broker exits in an ordered manner (%s)", reason.c_str()));
src/app/contextBroker/contextBroker.cpp:  LM_I(("Orion Context Broker is running"));
src/app/contextBroker/contextBroker.cpp:  LM_I(("Startup completed"));
```

which I think is good enough.

fgalan:
Regarding

> For notifications, add the following token to the correlator ID: `; notif=n` where "n" is an autoincremental counter (1, 2, 3, etc.) local to the e2e flow that caused the nofiication.

[...]

PR #3696, opened 2020-09-17
Log improvements

fgalan:
Issue #3694 

- [x] Move (almost) all the current LM_I traces to LM_T, choosing a suitable (new) tracelevel in DEBUG level
- [x] Introduce the following new traces at INFO level
    - [x] At the end of each GET request on the CB API endpoint, log the URL used and the response status code to that request. No payload is loged.
    - [x] At the end of each non-GET request on the CB API endpoint, log the URL used, the response status code to that request and the request payload
    - [x] At the end of each notification send, log the URL of the endpoind and subId of the associated subcription. No payload is logged
- [x] For notifications, add the following token to the correlator ID: `; notif=n` where "n" is an autoincremental counter (1, 2, 3, etc.) local to the e2e flow that caused the nofiication.
- [x] Take into account feedback in https://github.com/telefonicaid/fiware-orion/issues/3694
- [x] Add new .tests and adjusts existing ones
- [x] Modify documentation in logs.md (and maybe others .md)
- [x] CNR entry


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -81,6 +81,39 @@ using orion::CompoundValueNode;
...
+*
+* For instance if full correlator is
+*
+*   f320136c-0192-11eb-a893-000c29df7908; cbnotif=32
+*
+* then the correlator root is:
+*
+*   f320136c-0192-11eb-a893-000c29df7908
+*
+* This is needed as the self-notification loop protection logic is based on the correlator root
```
fgalan, line 98: This modification (and other similar ones) make me to realize that self-notification could not be properly documented. It seems that only a small mention in data model document with regards to `lastCorrelator` field in DB mention it.

I should be described in some place of the documentation. Maybe in the one(s) related with custom notifications.

To be addressed within this PR.

src/lib/common/logTracing.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+/* ****************************************************************************
+*
+* truncatePayload -
+*
+* NOTE: this function allocated dynamic memory, be careful with memory leaks!
+*/
+static char* truncatePayload(const char* payload)
```
fgalan, line 89: There should be simpler ways of doing this in C/C++... but probably aren't :)

fgalan: NTC

AlvaroVega (approved):
LGTM

fgalan:
@fisuda this PR is modifying some .md files. It would be great if you could update the Japanese translation based on that, please.

Thanks!

fisuda:
Noted with thanks.

Issue #3716, opened 2020-11-03
notificationMode different in the documentation


PedroVictorB:
In the documentation (https://fiware-orion.readthedocs.io/en/2.5.0/admin/cli/index.html), the notificationMode parameter can be transient, permanent or threadpool:q:n .

After trying to use permanent, I got this message:

`
orion_1  | time=Tuesday 03 Nov 14:21:43 2020.326Z | lvl=FATAL | corr=N/A | trans=N/A | from=N/A | srv=N/A | subsrv=N/A | comp=Orion | op=contextBroker.cpp[739]:notificationModeParse | msg=Fatal Error parsing notification mode: invalid mode (permanent)
fiware_orion_1 exited with code 1
`

After taking a look at the source code it seems that the mode is actually "persistent".

contextBroker.cpp - Line 745: (strcmp(mode, "transient") == 0 || strcmp(mode, "persistent") == 0)

fgalan:
You are right, the right keyword is "persistent". Usage message also confirms it:

```
$ contextBroker -u | grep notificationMode
                      [option '-notificationMode' <notification mode (persistent|transient|threadpool:q:n)>]
```

The following sections in the documentation need to be fixed:

* https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/cli.md#command-line-options
* https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/perf_tuning.md#notification-modes-and-performance

Do you want to do the typo fixing yourself (in a pull request on this repo)? Of course we can also do it, but maybe you want (the one that discover the bug should be the one to get the glory of solving it ;). Just tell us what do you prefer.

Thanks!

PedroVictorB:
@fgalan The pull request is there (#3718 ). I will try to contribute more in the future!

Thanks!

fgalan:
Once PR https://github.com/telefonicaid/fiware-orion/pull/3718 has been merged, this issue can be closed.

Thanks for the contribution!

PR #3718, opened 2020-11-04
Fix for the issue #3716

PedroVictorB:
Fix for the documentation showing a wrong mode for the notificationMode parameter. #3716


fgalan (approved):
LGTM. Thanks for the typo fixing! :)

We will merge this PR even if travis has not ended the pass yet. The PR only touch documentation so regression is not needed.

fgalan:
@fisuda this PR solves a kind of typo in the English documentation. Not sure if the typo is also in the Japanese translation but, in that case, it would be great if you could modify it. Thanks!

fisuda:
Noted with thanks.

PR #3720, opened 2020-11-05
add disableFileLog option, fix usage of fg and logForHumans as ENV vars

wistefan:
When running a service in kubernetes, its better to not log into a file, but instead logging to stdout and handle them with another log tool to prevent write to the nodes file-system from within a pod. Therefore I added an option to disable file logging(e.g. -disableFileLog / ORION_DISABLE_FILE_LOG).
Additionally, I figured out that the env vars ORION_FOREGROUND and ORION_LOG_FOR_HUMANS are ignored. To fix that, I extracted the logSetup from the options parsing and call it explicitly after the options where parsed and set.


wistefan:
Will fix tests.

doc/manuals/admin/cli.md
```diff
@@ -166,6 +166,7 @@ The list of available options is the following:
 -   **-disableCustomNotifications**. Disabled NGSIv2 custom notifications. In particular:
     * `httpCustom` is interpreted as `http`, i.e. all sub-fields except `url` are ignored
     * No `${...}` macro substitution is performed.
+-   **-disableFileLog**. To prevent orion from logging into a file. This option might be usefull if you are running on kubernetes.
```
fgalan, line 169:
```suggestion
-   **-disableFileLog**. To prevent Orion from logging into a file (default behaviour is to use a log file). This option might be useful if you are running on kubernetes.
```

src/app/contextBroker/contextBroker.cpp
```diff
@@ -245,6 +247,7 @@ unsigned long   fcMaxInterval;
 #define RELOGALARMS_DESC       "log messages for existing alarms beyond the raising alarm log message itself"
 #define CHECK_v1_ID_DESC       "additional checks for id fields in the NGSIv1 API"
 #define DISABLE_CUSTOM_NOTIF   "disable NGSIv2 custom notifications"
+#define DISABLE_FILE_LOG       "disable logging into files"
```
fgalan, line 250:
```suggestion
#define DISABLE_FILE_LOG       "disable logging into file"
```

fgalan: Impact in .test files. The 's' has to be deleted also there.

src/lib/parseArgs/paParse.cpp
```diff
@@ -565,22 +565,11 @@ int paParse
...
   {
     RETURN_ERROR("paLimitCheck");
   }
 
-  if ((s != -2) && (paConfigActions(false) == -1))
-  {
-    RETURN_ERROR("paConfigActions");
-  }
-
-
```
fgalan, line 583: We this removal is needed? Could you elaborate?

wistefan: I moved both methods to the contextBroker main method, because in the previous order the log-system was setup before the env variables where evaluated. Therefor the -fg, -logForHumans and the new -disableFileLog options where only available as commandline options. With extracting paLogSetup and paConfigActions from the parse method, its now possible to parse the options and configure the logging based on the results of the parsing. See contextBroker.pp l953 - l988

src/app/contextBroker/contextBroker.cpp
wistefan: Fixed in ce77dfe455713e7b52992d067daebfad21136745

doc/manuals/admin/cli.md
wistefan: Fixed in ce77dfe455713e7b52992d067daebfad21136745

kzangeli:
I might have found a simpler approach to get the same functionality ...
simplerApproach

wistefan:
> I might have found a simpler approach to get the same functionality ...
> simplerApproach

I originally had the same idea, but this would only make the option available via command line, and will continue to ignore the environment variables.

kzangeli:
So, the problem is that paIsSet only looks at command line args, not at env vars?
If so, let's fix that

kzangeli:
I checked, and yes, paIsSet only looks at the command line args.
```
bool paIsSet(int argC, char* argV[], const char* option)
{
  int i;

  for (i = 1; i < argC; i++)
  {
    if (strcmp(argV[i], option) == 0)
    {
      return true;
    }
  }

  return false;
}
```

We'd need to add a second loop, going over all items of the PaArgument array, checking for set env-vars.

wistefan:
Ok, I will do that. I thought the second loop is avoided on purpose, to not loop twice.

kzangeli:
The thing is, we haven't made a heavy use of env vars until just recently.
This bug would have been found earlier otherwise.
Thank you very much for finding it and for making the broker able to run inside kubernetes!

src/app/contextBroker/contextBroker.cpp
```diff
@@ -64,6 +63,7 @@
 #include <string>
 #include <vector>
 #include <limits.h>
+#include <stdio.h>             /* snprintf                                    */
```
kzangeli, line 66: Why not put it back where it was (11 lines before), and use C++ comments, indented like for `unistd.h` and `fcntl.h` ?
Just for "prettiness" :)

src/app/contextBroker/contextBroker.cpp
```diff
@@ -888,8 +888,6 @@ static void logEnvVars(void)
   }
 }
 
-
-
```
kzangeli, line 888: Here you seem to have removed two empty lines.
I'd like them back.
Throughout Orion source code, we try to maintain three empty line between functions.
It makes the file easier to read.

src/app/contextBroker/contextBroker.cpp
```diff
@@ -971,22 +968,16 @@ int main(int argC, char* argV[])
...
   {
-    paConfig("log to file",                   (void*) true);
+    paConfig("log to file", (void*) true);
   }
-  
-  // setup logging after the env vars evaluated and log config is set.
-  if(paLogSetup() == -1) {
-    printf("Was not able to set up the logging system.");
-    exit(1);
-  }
```
kzangeli: This is a bit weird ...
Looks like new code (as it doesn't follow the style guide) but ... it's removed ...
Would be nice to understand what happened here.

src/lib/parseArgs/paParse.cpp
```diff
@@ -570,6 +570,17 @@ int paParse
     RETURN_ERROR("paLimitCheck");
   }
 
+  if ((s != -2) && (paConfigActions(false) == -1))
+  {
+    RETURN_ERROR("paConfigActions");
+  }
+
+  if ((s != -2) && (paConfigActions(false) == -1))
```
kzangeli, line 578: This seems like a duplication of code.
Cut 'n paste error?   (`paConfigActions` is called twice)

kzangeli (changes requested):
Very good as a first PR.
Hope to see many more in the future :)
Thanks!

src/app/contextBroker/contextBroker.cpp
wistefan: Fixed in 7ea8ab6

src/app/contextBroker/contextBroker.cpp
wistefan: Fixed in 17364b1

src/lib/parseArgs/paParse.cpp
wistefan: Fixed in e00dc42

src/app/contextBroker/contextBroker.cpp
wistefan: This was from my first approach, when I extracted the log setup from the paParse method and called it explicitly after the env vars where evaluated. Since paIsSet now takes env-vars into account, it not longer required.

kzangeli (approved):
LGTM

src/app/contextBroker/contextBroker.cpp
```diff
@@ -63,7 +63,7 @@
 #include <openssl/ssl.h>
 #include <string>
 #include <vector>
-#include <limits.h>
+#include <limits.h>
```
fgalan, line 66:
```suggestion
#include <limits.h>
```

To avoid extra trailing whitespace

src/app/contextBroker/contextBroker.cpp
```diff
@@ -52,7 +52,7 @@
 * To override the security added with the certificate, curl can always be called using the
 * CLI option '--insecure'.
 */
-#include <stdio.h>
+#include <stdio.h>                              // snprintf
```
fgalan, line 55:
```suggestion
#include <stdio.h>                              // snprintf
```

To avoid extra trailing whitespace.

doc/manuals/admin/cli.md
```diff
@@ -166,6 +166,7 @@ The list of available options is the following:
 -   **-disableCustomNotifications**. Disabled NGSIv2 custom notifications. In particular:
     * `httpCustom` is interpreted as `http`, i.e. all sub-fields except `url` are ignored
     * No `${...}` macro substitution is performed.
+-   **-disableFileLog**. To prevent Orion from logging into a file (default behaviour is use a log file). This option might be useful if you are running on kubernetes.
```
fgalan, line 169: An entry in CHANGES_NEXT_RELEASE about the new flag should be added. For instance:

```
- Add: -disableFileLog CLI parameter (ORION_DISABLE_FILE_LOG env var) to prevent Orion from logggin into a file
```

wistefan: Fixed in 521246b5f94f5366050a06501199621e8abd6046

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Add: -disableFileLog CLI parameter (ORION_DISABLE_FILE_LOG env var) to prevent Orion from logggin into a file
```
fgalan, line 1: Small typo:

```suggestion
- Add: -disableFileLog CLI parameter (ORION_DISABLE_FILE_LOG env var) to prevent Orion from logging into a file
```

(My fault... the typo was already in my original comment)

fgalan (approved):
LGTM. Thanks for your contribution! :)

fgalan:
@fisuda this PR does an small modification to cli.md. It would be great if you could do a PR to sync the Japanese traslation. Thanks!

fisuda:
Noted with thanks.

PR #3722, opened 2020-11-11
(JP) Add document about disableFileLog option (#3720)

fisuda:
This PR adds document about disableFileLog option.


fgalan (approved):
LGTM. Thanks!

Issue #3347, opened 2018-10-25
Return 400 instead of 500 when GeoJSON syntax error occur


fgalan:
Orion returns 500 error when client send a wrongly formed GeoJSON:

Example 1:

```
curl -i -X POST http://localhost:10026/v2/entities -H 'Content-Type:
application/json' -d '{ "id": "J",
"type":"T","location":{"type":"geo:json","value":{"type":"Point","coord
inates":[[2,1]]  }}}' -H 'fiware-service: smartcity' -H 'fiware-
servicepath: /'
```

```
HTTP/1.1 500 Internal Server Error
Connection: Keep-Alive
Content-Length: 1444
Content-Type: application/json
Fiware-Correlator: 51e178d8-d81d-11e8-996d-0242ac110011
Date: Thu, 25 Oct 2018 06:15:07 GMT

{"error":"InternalError","description":"Database Error &#40;collection:
orion-smartcity.entities - insert&#40;&#41;: { _id: { id:
&quot;J&quot;, type: &quot;T&quot;, servicePath: &quot;/&quot; },
attrNames: [ &quot;location&quot; ], attrs: { location: { type:
&quot;geo:json&quot;, creDate: 1540448107, modDate: 1540448107, value:
{ type: &quot;Point&quot;, coordinates: [ [ 2.0, 1.0 ] ] }, mdNames: []
} }, creDate: 1540448107, modDate: 1540448107, location: { attrName:
&quot;location&quot;, coords: { type: &quot;Point&quot;, coordinates: [
[ 2.0, 1.0 ] ] } }, lastCorrelator: &quot;51e178d8-d81d-11e8-996d-
0242ac110011&quot; } - exception: OperationException: { index: 0, code:
16755, errmsg: &quot;Can&#39;t extract geo keys: { _id: { id:
&quot;J&quot;, type: &quot;T&quot;, servicePath: &quot;/&quot; },
attrNames: [ &quot;location&quot; ], attrs: { location: { type:
&quot;geo:json&quot;, creDate: 15...&quot;, op: { _id: { id:
&quot;J&quot;, type: &quot;T&quot;, servicePath: &quot;/&quot; },
attrNames: [ &quot;location&quot; ], attrs: { location: { type:
&quot;geo:json&quot;, creDate: 1540448107, modDate: 1540448107, value:
{ type: &quot;Point&quot;, coordinates: [ [ 2.0, 1.0 ] ] }, mdNames: []
[...]

PR #3733, opened 2020-11-26
fix for issue#3347 regarding wrong GeoJson syntax

ArqumFarooqui-NEC:
Fix for Issue #3347 when wrong GeoJson syntax

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2689,7 +2689,7 @@ static bool createEntity
 
   if (!collectionInsert(getEntitiesCollectionName(tenant), insertedDoc.obj(), errDetail))
   {
-    oeP->fill(SccReceiverInternalError, *errDetail, "InternalError");
+    oeP->fill(SccBadRequest, *errDetail, "BadRequest");
     return false;
   }
 
--- a/src/lib/mongoBackend/connectionOperations.cpp
+++ b/src/lib/mongoBackend/connectionOperations.cpp
@@ -384,12 +384,7 @@ bool collectionInsert
     releaseMongoConnection(connection);
     TIME_STAT_MONGO_WRITE_WAIT_STOP();
 
-    std::string msg = std::string("collection: ") + col.c_str() +
-      " - insert(): " + doc.toString() +
-      " - exception: " + e.what();
-
-    *err = "Database Error (" + msg + ")";
-    alarmMgr.dbError(msg);
+    *err = "Bad Request";
 
     return false;
   }
--- a/test/functionalTest/cases/3347_createEntity_with_wrong_GeoJSON/createEntity_with_wrong_GeoJSON.test
+++ b/test/functionalTest/cases/3347_createEntity_with_wrong_GeoJSON/createEntity_with_wrong_GeoJSON.test
@@ -1,0 +1,172 @@
+# Copyright 2020 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+Create entity with WRONG GeoJSON
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB 0
+
+--SHELL--
+
+#
+# 01. Create entity with correct GeoJSON
+# 02. GET /v2/entities/K
+# 03. Create entity with WRONG GeoJSON1
+# 04. Create entity with WRONG GeoJSON2
+#
+
+echo "01. Create entity with correct GeoJSON"
+echo "======================================"
+payload='{
+   "id": "K",
+   "type":"T",
+   "location":
+   {
+      "type":"geo:json",
+      "value":
+      {
+        "type":"Point",
+        "coordinates":[2,1]
+      }
+   }
+}'
+orionCurl --url '/v2/entities' --payload "${payload}"
+echo
+echo
+
+
+echo "02. GET /v2/entities/K"
+echo "======================"
+orionCurl --url /v2/entities/K
+echo
+echo
+
+
+echo "03. Create entity with WRONG GeoJSON1"
+echo "====================================="
+payload='{
+   "id": "M",
+   "type":"N",
+   "location":
+   {
+      "type":"geo:json",
+      "value":
+      {
+        "type":"PointKK3",
+        "coordinates":[2,1]
+      }
+   }
+}'
+orionCurl --url '/v2/entities' --payload "${payload}"
+echo
+echo
+
+
+echo "04. Create entity with WRONG GeoJSON2"
+echo "====================================="
+payload='{
+   "id": "M",
+   "type":"N",
+   "location":
+   {
+      "type":"geo:json",
+      "value":
+      {
+        "type":"Point",
+        "coordinates":[[2,1]]
+      }
+   }
+}'
+orionCurl --url '/v2/entities' --payload "${payload}"
+echo
+echo
+
+
+--REGEXPECT--
+01. Create entity with correct GeoJSON
+======================================
+HTTP/1.1 201 Created
+Content-Length: 0
+Location: /v2/entities/K?type=T
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+
+
+02. GET /v2/entities/K 
+======================
+HTTP/1.1 200 OK
+Content-Length: 111
+Content-Type: application/json
+Fiware-Correlator: REGEX([0-9a-f\-]{36})
+Date: REGEX(.*)
+
+{
+    "id": "K",
+    "location": {
+        "metadata": {},
+        "type": "geo:json",
+        "value": {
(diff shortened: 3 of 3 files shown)


ArqumFarooqui-NEC:
Hi @fgalan sir,
Please review this PR, if it is OK then Please merge it.
Thanks:)

kzangeli:
I'm afraid this is not the way to solve this problem.
Instead, the incoming payload should be inspected (I'm surprised it isn't!) and the error detected long before mongo is called.

Right after parsing the incoming payload, the "type" and "coordinates" of the geo-prop value should be checked and if the "type" is not "Point", "Polygon", etc, a clear error response about this should be returned.
Same same for invalid coordinates, or "type" missing, "coords" missing, any field duplicated or empty or ...

The current proposal assumes every error in mogno is a bad request, which it is of course not.

So, there is a much better way to fix this issue.

ArqumFarooqui-NEC:
Thank you for your guidance:)
Sorry for granting the solution in wrong way, I'll follow the suggestions.

kzangeli:
And thank You! for contributing :)

I haven't looked too closely on this but I imagine ContextAttribute::check() is where this check should be implemented ...
Right, @fgalan ?

fgalan:
Not sure if checking at the parsing stage (ContextAttribute::check() or similar) is the best way to do.

You could do some syntax checks at parsing check (as the ones suggested by @kzangeli above about missing fields in the JSON). However, some other problems are not easy to detect at the parsing stage. For instance, if a polygon has the vertices in a wrong order, such as this one:

```
{
  type: "Polygon",
  coordinates: [ [
    [2, 2],
    [-2, -2],
    [2, -2],
    [-2, 2],
    [2, 2]
  ] ]
}
```

Thus, I think a better strategy is MongoDB doing that job for us. I mean, look at the response from MongoDB and if some error pattern is detected (starting with the simpler one: "Can't extract geo keys") then return 400 Bad Request.

ArqumFarooqui-NEC:
> Thus, I think a better strategy is MongoDB doing that job for us. I mean, look at the response from MongoDB and if some error pattern is detected (starting with the simpler one: "Can't extract geo keys") then return 400 Bad Request.
Hi @fgalan, @kzangeli sir,

I've investigated at mongo side, and found that code for `location` attribute has dealt in `processLocationAtEntityCreation()` function in [mongoBackend/location.cpp](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/mongoBackend/location.cpp#L281) class. In that function there is another function call to [getGeoJson()](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/mongoBackend/location.cpp#L109#L275) in the same class, that is dealing with `type` field of `location` attribute.

**Please confirm, is my investigation and understating is correct?**


While dealing with `value` field of `location` attribute in that [getGeoJson()](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/mongoBackend/location.cpp#L109#L275) function and while solving the issue there I've got stuck there while dealing with `value` field of `location`, how to proceed further.
Please provide your valuable guidance, it will be highly helpful.

kzangeli:
Well, @fgalan calls the shots, the repo is his responsibility.
He already stated his thoughts.
Me, I'd definitely check the input long before calling mongo.
[ What @fgalan says makes a lot of sense - don't implement it if it's already implemented ]

However, not calling mpongo for that piece of info makes Orion much faster (in this particular sense), and with better accuracy in the error handling (assuming you do a good job interpreting the incoming payload, of course).

I implemented all this for Orion-LD (and yes, it's a lot), only that my implementation is not valid for Orion as the incoming payload is parsed with a different JSON parser. The "idea" would be the same, of course.

I'd gladly guide you through the Orion-LD implementation, would you be interested (for Orion, my implementation would have to be "translated" into something similar for the output you get from rapidjson)

That said, the value field should be an object with two children, named "type" and "coordinates", "type" being a JSON String and "coordinates" a JSON Array. So, where exactly are you stuck?

fgalan:
@ArqumFarooqui-NEC maybe the best place to intercept the 500 got at MongoDB level and tranform to a proper 400 will be the point in which the mongoBackend returns to the serviceRoutine code and before passing the flow to the rest layer. From a high level perspective, the point of the code I mean is the one shown here:

imagen

(Source of the diagram: https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/sourceCode.md#flow-rq-02)

ArqumFarooqui-NEC:
Thank you @kzangeli sir, @fgalan sir  for your response:)
Meanwhile I'm working another issue as well, so any kind of suggestion would really be very helpful, I'll be grateful for your suggestions. 

> That said, the value field should be an object with two children, named "type" and "coordinates", "type" being a JSON String and "coordinates" a JSON Array. So, where exactly are you stuck?

@kzangeli sir, as per my understanding `value` field of location attribute is an object that is already checked at [location.cpp](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/mongoBackend/location.cpp#L162) class i.e.:
```
if ((caP->compoundValueP == NULL) || !(caP->compoundValueP->isObject()))
{
    *errDetail = "geo:json needs an object as value";
    return false;
}
```
After the above code, I wanted to apply checks on children of `value` field, "type" and "coordinates" and if type and coordinates are not ok then return error else as it is, for example:
```
if ( value ->type != point or polygon etc...)
      then return error
```
Is it right way to solve this issue??? Please provide your suggestion:)

kzangeli:
@ArqumFarooqui-NEC , first of all -
in *my personal opinion*, orion should check that the value is coherent before sending it to mongo.
But, @fgalan has already stated that he prefers to leave this check to mongo.

And, @fgalan is the owner of the repository, so you should probably go with what he suggests.

I'm not saying he's wrong and I'm right. This is simply a matter of taste.
What I can say is that it's _definitely_ faster in run-time to do the check before sending it to mongo (what I propose).
The drawback is that you have to implement all of that and it's far from trivial (which is what seduces @fgalan, no doubt).

For example, in the compound "value", if the "type" is "Point", then the "coordinates" must be a JSON array of either 2 or 3 numbers.
AND, these numbers must be checked to be valid coordinates.
If instead "type" is "Polygon", the "coordinates" must be a triple array with points according to the geoJSON spec,
AND SO ON. 

It's pretty complex, read a little about GeoJSON to get a better feeling.

I implemented all of this for Orion-LD, which would be of great help in case you go for _my approach_ finally in Orion - just let me know.
You need to clear this with @fgalan before starting implementing.

About instead letting mongo deal with the possible errors - I guess you'd have to parse the error string from mongo to give a nice error message back in case of errors. Or, just give it back as is and let the user deal with the complexity (horrible IMHO)).

fgalan:
Just to make my point clearer:

Detecting errors at parsing stage is the way to go when *you can assure that your error detecting logic would detect 100% of possible cases*. For simple things like some sub-field is missing (such as `type` and `coordinates`) is easy. However, I think that other problems are hard to detect at this stage. For instance, situations in which the GeoJSON is syntactically correct but will cause errors at MongoDB, such a polygon with "crossing points", as illustrated below:

imagen

Can every possible error situation be detected at the parsing stage? Maybe, but seems hard to implement. And, worse, very hard to know if your error detecting logic is complete and you are not missing possible cases. Thus, a better strategy is let MongoDB do that work for us.

A mixed approach is also possible. I mean, trying to detect easy to detect problems at the parsing stage (which could detect, let's say, 80% of the typical error cases) but look also for potential errors in Mongo to cover the remaining 20%.

kzangeli:
Yes, you wouldn't want to make sure that the geometries are 100% correct - that would be to take it too far, no doubt.
Just that the types are OK - e.g. if it's a Point, the coordinates must be an array with 2 or 3 Numbers.
=> mixed approach.

ArqumFarooqui-NEC:
Hi @fgalan sir, Sorry for delay, I was busy in some other priority task, I have resumed my work on it.

As per my investigation, I have to add the validation for the geolocation( whether point, polygon etc., coordinates must be array, and type missing etc.), at the [jsonParseV2/parseEntity.cpp](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/jsonParseV2/parseEntity.cpp#L231) and [jsonParseV2/parseContextAttribute.cpp](https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/jsonParseV2/parseContextAttribute.cpp#L185) classes where parsing of POST request is being done.
I want to apply the validation using values of `ciP`, `iter`, `caP` variables, but I'm not sure what exact value is present inside them.
To check the values(data) of the variables, I have tried 2 methods, 1) by using gdb but in this case another Orion instance was started itself, before hitting run command on gdb shell so during post request it don't hit breakpoint, gets reply from other running instance, 2) I have tried to print the variables `(cip, iter, etc)` data but this is also not working.
I did not find any clue to proceed further, So could you please provide your valuable guidance.

fgalan:
Happy to see you back!

Discussion above is probably long and confusing, so I'll try to sum up and clarify. This issue has two aspects:

1. Intercept errors raised by MongoDB due to wrong GeoJSONs, so 400 is returned instead of 500 (suggestion about where to do that interception: see https://github.com/telefonicaid/fiware-orion/pull/3733#issuecomment-746617913) 
2. Capture easy to detect errors in the GeoJSON at parsing stage, so 400 is returned this case, e.g. wrong type, malformed coordinates, etc.

From my point of view, part 1 is a must and part 2 is just a nice to have. If you have to focus on one, start by the first.

ArqumFarooqui-NEC:
Hi @fgalan sir, I got your point, I will try to fix first part. But for fixing second part I have mentioned some difficulties in my [previous comment](https://github.com/telefonicaid/fiware-orion/pull/3733#issuecomment-857425940), because first part means already there should be error we have to send another error code right?

fgalan:
> Hi @fgalan sir, I got your point, I will try to fix first part. But for fixing second part I have mentioned some difficulties in my [previous comment](https://github.com/telefonicaid/fiware-orion/pull/3733#issuecomment-857425940), because first part means already there should be error we have to send another error code right?

If the second part is difficult, forget about it by the moment. I'd suggest you focus exclusively in the first part in the context of this PR.

ArqumFarooqui-NEC:
Thank you for your response. 
Okay @fgalan sir, now I'll focus on first part only.

ArqumFarooqui-NEC:
Hi @fgalan sir,
I have raised a new PR i.e. #3898 for issue #3347 as I was facing conflicting issue on that branch.

fgalan:
I understand this PR has been overpassed by PR https://github.com/telefonicaid/fiware-orion/pull/3898 and should be closed. Is that correct?

ArqumFarooqui-NEC:
> I understand this PR has been overpassed by PR #3898 and should be closed. Is that correct?

Yes. 
Closing this PR because a new PR #3898 has been raised for the same issue #3347

PR #3732, opened 2020-11-20
[WIP] UPGRADE MongoDB recomendation from 3.6 to 4.2

fgalan:
Many deployments show that Orion works fine with 4.2 so we are going to upgrade the recommendation.

PR #3747, opened 2020-12-21
Replace travis by github-actions

wistefan:
Due to the changed opensource support for travis, I replaced travis by github-actions. The ci-pipline is structure similar to the one in the  FIWARE/context.Orion-LD repository. A side-effect of that is the reduced ci-runtime(~15-20 min in total)
Following steps do always run: 
- Compliance Test
- Unit Test
- Functional Test(split in 4 steps to speed up the build)

Only on pushes to the CI-Folder:
- Build CI-RPM-Image

Only on pushes to the master branch:
- Build and publish PRE image, the image will be tagged with the next minor version and PRE-<BUILD_NUMBER> to differentiate from releases

For the docker build steps to work, a repository administrator needs to set the following secrets(https://github.com/telefonicaid/fiware-orion/settings/secrets/actions):
- DOCKERHUB_USERNAME 
- DOCKERHUB_TOKEN


wistefan:
Runs of the pipeline can be seen here: https://github.com/wistefan/fiware-orion/actions

.github/workflows/functional.yml
```diff
@@ -0,0 +1,76 @@
...
+
+jobs:
+  functional-1:
+    runs-on: ubuntu-18.04
+
+    services:
+      mongodb:
+        image: mongo:4.0
+        ports:
+          - 27017:27017
```
fgalan, line 20: Is mongo really needed i this .yml? As far as I understand, MongoDB runs *inside* the fiware/orion-ci:rpm7 image (https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm7/build-dep.sh#L38)

(Similar comments for others .yml that use fiware/orion-ci:rpm7)

fgalan:
Thanks for your valuable contribution! It will save a lot of time in regressions :)

> For the docker build steps to work, a repository administrator needs to set the following secrets(https://github.com/telefonicaid/fiware-orion/settings/secrets/actions):
> 
> * DOCKERHUB_USERNAME
> * DOCKERHUB_TOKEN

I have the needed administration privileges. However, I don't know exactly with DOCKERHUB_USERNAME and DOCKERHUB_TOKEN to use for the fiware/orion repository (which I understand is the one in which the images genearted by these GitActions are published).

(Maybe we can deal with this specific item by email, if it involves sensitive information. If you don't know my email, your colleagues in the FIWARE Foundation can tell you).

.github/workflows/functional.yml
wistefan: When the mongo inside the image is running, no. But in general, its not good practice to have multiple processes running inside one docker container and therefore I would prefer to use the service container. But I forgot to remove it from the rpm image and would leave it up to you on how you prefer it.

wistefan:
> Thanks for your valuable contribution! It will save a lot of time in regressions :)
> 
> > For the docker build steps to work, a repository administrator needs to set the following secrets(https://github.com/telefonicaid/fiware-orion/settings/secrets/actions):
> > 
> > * DOCKERHUB_USERNAME
> > * DOCKERHUB_TOKEN
> 
> I have the needed administration privileges. However, I don't know exactly with DOCKERHUB_USERNAME and DOCKERHUB_TOKEN to use for the fiware/orion repository (which I understand is the one in which the images genearted by these GitActions are published).
> 
> (Maybe we can deal with this specific item by email, if it involves sensitive information. If you don't know my email, your colleagues in the FIWARE Foundation can tell you).

I'll clarify that and send you mail.

.github/workflows/functional.yml
fgalan: What about to leave it commented to be enabled once it gets removed from the rpm image? I'll do a separate comment suggestion for that.

For this thread, NTC.

fgalan, line 20:
```suggestion
    # FIXME: currently mongodb runs inside fiware/orion-ci:rpm7 image. Once it gets removed
    # from it, this part will be enabled
    #services:
    #  mongodb:
    #    image: mongo:4.0
    #    ports:
    #      - 27017:27017
```

fgalan: (I'm doing the suggestion only in this .yml but it should be applied also to other .yml files where it makes sense)

wistefan: Sorry for the confusion, I made another commit to remove mongo from the image. Due to running the tests via 'build -miqts' (i.e. no -d parameter to start the db). they relied on the containerized mongo already. I added back the mongo-service

ci/rpm7/build-dep.sh
```diff
@@ -35,7 +35,6 @@ yum -y install \
   openssl-devel \
   libuuid-devel \
   make \
-  mongodb-org \
```
fgalan, line 38: In order to effectively remove mongo from the CI image I think this is not enough. You should:

* Review build.sh script looking for ocurrences of `mongod`
* Remove mongodb.repo file (including the `ADD mongodb.repo /etc/yum.repos.d/` lline in ci/rpm7/Dockerfile)

.github/workflows/publishimage.yml
```diff
@@ -0,0 +1,70 @@
...
+    if: github.event_name == 'push'
+
+    steps:
+
+      - uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
```
fgalan, line 30: Buildx? Is this a typo?

wistefan: No, see: https://docs.docker.com/buildx/working-with-buildx/

ci/rpm7/build-dep.sh
wistefan: Fixed in d0b12f5

.github/workflows/functional.yml
```diff
@@ -0,0 +1,76 @@
...
+env:
+   TEST_IMAGE_NAME: fiware/orion-ci:rpm7
+
+jobs:
+  functional-1:
+    runs-on: ubuntu-18.04
+
+    services:
+      mongodb:
+        image: mongo:4.0
```
fgalan, line 18: Taking into account [this PR that will be merged soon](https://github.com/telefonicaid/fiware-orion/pull/3732) I'd suggest to use 4.2.

.github/workflows/publishimage.yml
fgalan: As far as I understand from the link you cite, `buildx` seems to be an experimental feature. Could regular `build` be used instead? Which special feature existing in `buildx` but not in `build` are we needed here?

ci/rpm7/build.sh
```diff
@@ -233,13 +224,6 @@ echo "===================================== PREPARE ============================
 echo "Builder: create temp folders"
 rm -Rf /tmp/builder || true && mkdir -p /tmp/builder/{db1,db2,db,bu}
 
-if [ -n "${database}" ]; then
-    echo "Builder: starting Mongo"
-    mongod --dbpath /tmp/builder/db  --nojournal --quiet > /dev/null 2>&1 &
-    sleep 3
-    export MONGO_HOST=localhost
```
fgalan, line 240: By default, CB uses MongoDB running in localhost (i.e. running in the same container). Now that mongo runs "externaly" to the container as a service (as discussed in [this comment](https://github.com/telefonicaid/fiware-orion/pull/3747/files#r553798313), how do you make CB to use it instead of trying to connect locally?

wistefan: The docker container running the test is connected to the host-network(see the "--network host" option) and therefore the service container is available at localhost inside the container.

fgalan: So you mean that using `--network host` the CB process running inside the container "sees" a MongoDB instance running in localhost at port 27017 even when that MongoDB instance is running *externally* to the container?

wistefan: Yes. We map the service-containers port 27017 to the docker-hosts port 27017 here: https://github.com/wistefan/fiware-orion/blob/master/.github/workflows/functional.yml#L19-L20
And with --network host , the container is attached to the docker-hosts network and sees everything at localhost, that the docker host does.

.github/workflows/publishimage.yml
wistefan: I wasn't aware that its still marked as experimental, I usually use it to be prepared for switching to OCI compliant images. I can switch to the normal docker command, if you prefer that.

.github/workflows/functional.yml
fgalan: Ok. Let's keep 4.0 by the moment.

NTC

.github/workflows/publishimage.yml
fgalan: > I can switch to the normal docker command, if you prefer that.

I think it would be preferable, please.

ci/rpm7/build.sh
fgalan: I wonder what would happen if both the CB container and service container start a process listening in the same port (because in that case, "localhost:PORT" would bind to both the container and the service container and that would be a conflict :), but it's fine (we don't have that situation here).

NTC

.github/workflows/publishimage.yml
wistefan: Fixed in e211d6e

.github/workflows/functional.yml
wistefan: I've set it back to 4.0 in c32fa2e

fgalan (approved):
LGTM

Thanks for your contribution! Moving from travis to github actions will definitively save a lot of time :)

fgalan:
Let's merge and see how it goes...

fgalan:
It seems ciimage.yml workflow has failed in master after the merge (https://github.com/telefonicaid/fiware-orion/actions/runs/482622212):

imagen

@wistefan could you have a look, please?

fgalan:
Appart from the error described in the previous comment, all the other jobs went ok:

imagen

And in https://hub.docker.com/r/fiware/orion-ci/tags we can confirm the -PRE image is there:

imagen

Side-note: we are not using that kind of images but if they are of intereset for the community, it's fine. However, could it be a problem when the final 2.6.0 gets released? Due to (lexicographily) 2.6.0-PRE-1 > 2.6.0 so automatic system will think that 2.6.0-PRE-1 is newer than 2.6.0, which is false. Maybe is better to generate a 2.5.0-NEXT-1 release instead.

wistefan:
> It seems ciimage.yml workflow has failed in master after the merge (https://github.com/telefonicaid/fiware-orion/actions/runs/482622212):
> 
> imagen
> 
> @wistefan could you have a look, please?

Yes. I can see the mistake and will fix it.

wistefan:
> Appart from the error described in the previous comment, all the other jobs went ok:
> 
> imagen
> 
> And in https://hub.docker.com/r/fiware/orion-ci/tags we can confirm the -PRE image is there (which I understand validates the secret credential configuration):
> 
> imagen
> 
> Side-note: we are not using that kind of images but if they are of intereset for the community, it's fine. However, could it be a problem when the final 2.6.0 gets released? Due to (lexicographily) 2.6.0-PRE-1 > 2.6.0 so automatic system will think that 2.6.0-PRE-1 is newer than 2.6.0, which is false. Maybe is better to generate a 2.5.0-NEXT-1 release instead.

The created versions are semver-compliant. And due to the semver specification 2.6.0 is newer than 2.6.0-PRE-1. Using something like 2.6.0-SNAPSHOT or -PRE is realitvely common.

fgalan:
> And due to the semver specification 2.6.0 is newer than 2.6.0-PRE-1

You mean that accordinng to semver 2.6.0 > 2.6.0-PRE-1. In other words, that .0 > .0-PRE-1 ?

Note that's contrary to how it works (e.g. in a programming language like Python):

```
>>> "2.6.0" > "2.6.0-PRE-1"
False
>>> ".0" > ".0-PRE-1"
False
```

wistefan:
> > And due to the semver specification 2.6.0 is newer than 2.6.0-PRE-1
> 
> You mean that according to semver spec 2.6.0 > 2.6.0-PRE-1? In other words, that .0 > .0-PRE-1 ?
> 
> Note that's contrary to how it works (e.g. in a programming language like Python):
> 
> ```
> >>> "2.6.0" > "2.6.0-PRE-1"
> False
> >>> ".0" > ".0-PRE-1"
> False
> ```

Yes. Because for semver, everything after the "-" is denoting a pre-version. You can see it here: https://semver.org/ under 9 and 11.3

fgalan:
> Yes. Because for semver, everything after the "-" is denoting a pre-version. You can see it here: https://semver.org/ under 9 and 11.3

Ok great. I have always thought it was the other way around :)

Taking into account this, versions like `2.5.0-next` that we extensively use to mean versions between 2.5.0 and 2.6.0 and that we use extensively in many repositories (e.g. packages.json in node-based projects) probably doesn't have too much sense and are wrong.

wistefan:
> > Yes. Because for semver, everything after the "-" is denoting a pre-version. You can see it here: https://semver.org/ under 9 and 11.3
> 
> Ok great. I have always thought it was the other way around :)
> 
> Taking into account this, versions like `2.5.0-next` that we extensively use to mean versions between 2.5.0 and 2.6.0 and that we use extensively in many repositories (e.g. packages.json in node-based projects) probably doesn't have too much sense and are wrong.

:) It would at least fail the automatic (semver) comparison. Some artifact repos(i'm not sure, but I think jfrog's tools for example) do use that already for deciding "latest". So it can become problematic for npm packages, too.

PR #3756, opened 2021-01-13
add back mongo

wistefan:
removing mongo from the ci image seems to also remove the mongo client, required for running the functional test. The fail in other branches because of that.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -24,4 +24,4 @@ jobs:
 
       - name: Run functional test
         run: |
-          docker run  --network host -t --rm -v $(pwd):/opt/fiware-orion ${{ env.TEST_IMAGE_NAME }} build -miqts unit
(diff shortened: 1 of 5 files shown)


fgalan:
mongo shell should be included in the `mongodb-org-shell` package. So probably the only changes needed are the ones in:

* Recover the mongodb.repo file (maybe moving version 3.6 -> 4.0?)
* Recover the `ADD` line for that file in Dockerhub

I think the other chagnes in the PR are not needed. Could you test please?

wistefan:
> mongo shell should be included in the `mongodb-org-shell` package. So probably the only changes needed are the ones in:
> 
>     * Recover the mongodb.repo file (maybe moving version 3.6 -> 4.0?)
> 
>     * Recover the `ADD` line for that file in Dockerhub
> 
> 
> I think the other chagnes in the PR are not needed. Could you test please?

Yes, I'll do

wistefan:
There was an issue beside mongo, that failed the unit tests. It seems that the ci-image wasn't built for quite a while now and gmock now is not available under the used address anymore. I changed that and now it seems to work again.

fgalan (approved):
Great! LGTM

PR #3759, opened 2021-01-19
Improve documentation

wistefan:
We got feedback from the community about low performance of orion. After further enquire, we found out that most of them did not know about the performance tuning and ha section of the documentation. In order to get it more visible, I would propose those changes in the documentation.
In additon, we found out that setting an index to _id.id improves the performance of updates by multiples. Therefore I would also like to add it into the performance tuning section.

### DIFF
--- a/doc/manuals/admin/install.md
+++ b/doc/manuals/admin/install.md
@@ -10,7 +10,7 @@
     * [Standard upgrade](#standard-upgrade)
 
 ---
-> :warning: **If you plan to run orion in production, also read the [ha-documentation](extra/ha.md) and the [preformance tuning section](perf_tuning.md)**
+> :warning: **If you plan to run orion in production, also read the information about [high availability](extra/ha.md) and [preformance tuning](perf_tuning.md).**
 ---
 
 ## Introduction
--- a/doc/manuals/admin/install.md
+++ b/doc/manuals/admin/install.md
@@ -10,7 +10,7 @@
     * [Standard upgrade](#standard-upgrade)
 
 ---
-> :warning: **If you plan to run orion in production, also read the information about [high availability](extra/ha.md) and [preformance tuning](perf_tuning.md).**
+> :warning: **If you plan to run orion in production, also read the information about [high availability](extra/ha.md) and [performance tuning](perf_tuning.md).**
 ---
 
 ## Introduction
--- a/doc/manuals/admin/install.md
+++ b/doc/manuals/admin/install.md
@@ -9,6 +9,10 @@
     * [Migrating the data stored in DB](#migrating-the-data-stored-in-db)
     * [Standard upgrade](#standard-upgrade)
 
+---
+> :warning: **If you plan to run orion in production, also read the [ha-documentation](extra/ha.md) and the [preformance tuning section](perf_tuning.md)**
+---
+
 ## Introduction
 
 The recommended procedure is to install using RPM packages in CentOS 7.x. If you are interested in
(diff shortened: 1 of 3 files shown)


fgalan:
Maybe this PR is a good opportunity to fix also a couple of broken badgets in the top of the master README.md:

imagen

wistefan:
> Maybe this PR is a good opportunity to fix also a couple of broken badgets in the top of the master README.md:
> 
> imagen
Fixed in 2a87fe8 and 2a87fe8

doc/manuals/admin/perf_tuning.md
```diff
@@ -48,6 +48,11 @@ However, in order to help administrators in this task, the following *general* i
     * `{_id.servicePath: 1, _id.id: 1, _id.type: 1}` (note that this is a compound index and key order matters in this case)
     * `creDate`
 
+In order to improve the performance of updates, we also highly recommend using the following index:
+
+* Collection [entities](database_model.md#entities-collection)
+    * `{_id.id:1}` 
+
```
fgalan, line 55: As can be seen in traces (`-logLevel INFO -t 240 -logForHumans`) the DB query done by Orion to update an entity:

```
DEBUG@13:13:34  connectionOperations.cpp[454]: Database Operation Successful (update: <{ _id.id: "E", _id.type: "T", _id.servicePath: "/ss" }, { $set: { ...  }>)
```

Note that the index already proposed  (`{_id.servicePath: 1, _id.id: 1, _id.type: 1}`) matches that query. However, I guess than in some cases (when the number of service paths is small and the number of ids per-servicepath is large, so the "specificity" of servicepaths is not very good) an index on `{_id.id: 1`} could beat the one in `{_id.servicePath: 1, _id.id: 1, _id.type: 1}`.

Ideally, the documenation should provide hints about in which case each index is better, depending on the above (i.el how the DB data is regarding service paths and entities ids/types).

I'd like to have a closer look to the DB in which you have found this problem. May I access to the environment or could you send a mongodump? Let's disscuss by email this if you prefer.

(By the moment, I'd suggest to split this part to a separate PR, given it could take some time analyze this, so not to block the others modifications included in this PR)

doc/manuals/admin/install.md
```diff
@@ -9,6 +9,10 @@
     * [Migrating the data stored in DB](#migrating-the-data-stored-in-db)
     * [Standard upgrade](#standard-upgrade)
 
+---
+> :warning: **If you plan to run orion in production, also read the information about [high availability](extra/ha.md) and [preformance tuning](perf_tuning.md).**
```
jmcanterafonseca, line 13:
```suggestion
> :warning: **If you plan to run orion in production, also read the information about [high availability](extra/ha.md) and [performance tuning](perf_tuning.md).**
```

doc/manuals/admin/perf_tuning.md
wistefan: I did run some additional tests and it seems that the already present is enough. But I think it should be stated explicitly that the index is very important for write-performance, too.

wistefan: You can have a closer look on my results here: https://github.com/wistefan/orion-loadtest
It includes all informations to rerun the tests.

wistefan: Fixed in  cd372de

fgalan: Unfortunately, I will not be able to use the software at that repo to run loadtest, but if you have a mongodump with the generated data (or live access to the MongoDB instance where that data is stored) I can have a look to index performance on it (using the usual mongo shell tooling, i.e. `explain()`, etc.)

fgalan (approved):
LGTM. Thanks!

fgalan:
@fisuda this PR has modified some files that I have are mirrored in the Japanese translation. Could you do a PR to upgrade it, please?

Thanks!

fisuda:
I sent PR #3765. Please review it.

PR #3622, opened 2020-02-07
mongo driver migration to an officially supported one + moving Orion to CentOS 8

fgalan:
This PR is related with #3132 and provides the "wrapping modules" described in that issue.

At the present point, apart from a first version of the modules themselves (in the mongoDriver/ library) only a small part of the CB logic (in particular, the one related with subscription creation, update and retreival) has been migrated to use the new library.

The idea is to continue migrating more CB logic, enhancing mongoDriver/ in the process, if needed. This could be done either ini feature/3132_poc_isolating_driver_usage directly or as PR on that branch (contributions welcome! :)

Pay attention to the "FIXME OLD-DR" marks. They should be solved and removed before merging this PR into master.

Pending references to old driver outside mongoDriver (`grep '#include "mongo/' -R src/* | grep -v mongoDriver | wc -l`)

* 35

PR #3795, opened 2021-03-02
FIX make some changes to be more compatible with CentOS 8 RPM build procedure

fgalan:
This has arisen in the context of PR https://github.com/telefonicaid/fiware-orion/pull/3622. These are changes needed for rpm build in CentOS 8, which at the same time are compatible with CentOS 7 (as far as we have checked).

### DIFF
--- a/rpm/SPECS/contextBroker.spec
+++ b/rpm/SPECS/contextBroker.spec
@@ -21,6 +21,14 @@
 %define name contextBroker
 %define owner orion 
 
+# To avoid problems in CentOS 8
+%define debug_package %{nil}
+
+# To avoid /usr/lib/.build-id files in the package. This seems to be a
+# new feature of CentOS 8 (see https://fedoraproject.org/wiki/Changes/ParallelInstallableDebuginfo)
+# but by the moment we are going to disable it to build as in CentOS 7
+%define _build_id_links none
+
 # Don't byte compile python code
 %global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-python-bytecompile[[:space:]].*$!!g')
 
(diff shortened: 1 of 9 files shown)


cesarjorgemartinez:
Pending to review by CJMM

rpm/SPECS/contextBroker.spec
```diff
@@ -21,6 +21,14 @@
 %define name contextBroker
 %define owner orion 
 
+# To avoid problems in CentOS 8
```
cesarjorgemartinez, line 24:
```
%if 0%{?rhel} > 7
# To avoid problems in CentOS 8
# What problems???
%define debug_package %{nil}
# To avoid /usr/lib/.build-id files in the package. This seems to be a
# new feature of CentOS 8 (see https://fedoraproject.org/wiki/Changes/ParallelInstallableDebuginfo)
# but by the moment we are going to disable it to build as in CentOS 7
%define _build_id_links none
%endif

```

cesarjorgemartinez (changes requested):

rpm/SPECS/contextBroker.spec
fgalan: > What problems???

This ones:

https://wiki.mageia.org/en/Packaging_problems_and_solutions#Empty_Debug_Files

fgalan: We can include this link in the comment, which seems to be a precise description of the problem: https://www.programmersought.com/article/24984333483/

cesarjorgemartinez: Okis

cesarjorgemartinez:
LGTM

cesarjorgemartinez (approved):

PR #3802, opened 2021-03-09
FIX query type logic

fgalan:
(This PR is one of the stages of the bigger PR #3622)

The issue this PR is fixing was introduced when the mongoQueryType logic was migrated from the old pre-3.6 way to the new one. Supposedly it solves some glitching tests:

* 0740_crash_with_json_object_for_attributes/json_object_format.test
*  0117_convop_using_standard_ops/getEntityTypes.test

Let's see in the CI pass...

### DIFF
--- a/src/lib/mongoBackend/mongoQueryTypes.cpp
+++ b/src/lib/mongoBackend/mongoQueryTypes.cpp
@@ -588,48 +588,46 @@ HttpStatusCode mongoEntityTypes
 
     entityType->count = countEntities(tenant, servicePathV, entityType->type);
 
-    if (!attrsArray[0].isNull())
+    for (unsigned int jx = 0; jx < attrsArray.size(); ++jx)
     {
-      for (unsigned int jx = 0; jx < attrsArray.size(); ++jx)
+      /* This is where NULL-elements in the resulting attrs vector are pruned */
+      if (attrsArray[jx].isNull())
       {
-        /* This is where NULL-elements in the resulting attrs vector are pruned */
-        if (attrsArray[jx].isNull())
-        {
-          continue;
-        }
+        continue;
+      }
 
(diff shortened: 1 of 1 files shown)


fgalan:
Review hint: use this setting

imagen

test/functionalTest/cases/0117_convop_using_standard_ops/getEntityTypes.test
```diff
@@ -349,8 +349,8 @@ Date: REGEX(.*)
         {
             "attributes": [
 #SORT_START
-                "A2",
-                "A1"
+                "A2"REGEX(,?)
+                "A1"REGEX(,?)
```
fgalan, line 353: This "trick" is already used in this file (e.g. L367-368) but, for some reason, we forget to use it here...

NTC (just informative)

test/functionalTest/cases/3389_forced_update_option/3389_forced_update_option_batch_bool.test
```diff
@@ -417,6 +418,7 @@ Fiware-Correlator: REGEX([0-9a-f\-]{36}; cbnotif=[12])
     "subscriptionId": "REGEX([0-9a-f]{24})"
 }
 =======================================
+#SORT_END
```
fgalan, line 421: There is a race condition in the notifications triggered by the notifications. Some times the first one to reach accumulator-server.py is Slot1 and some times is Slot2. With this SORT_START / SORT_END clauses we make that, from a checking purposes, the order doesn't matter.

Same for other case in this file and in other .test files in test/functionalTest/cases/3389_forced_update_option

NTC (just informative)

mapedraza (commented):
LGTM

PR #3832, opened 2021-04-08
FIX avoid unitialized creDate and modDate

fgalan:
Due to this it seems that valgrind regression is reporting some errors.


AlvaroVega (approved):
LGTM

PR #3839, opened 2021-04-13
Fix version increment for minor==0

wistefan:
See https://github.com/wistefan/fiware-orion/runs/2330445868?check_suite_focus=true#step:6:14 for a successful run. The incrementation needs to be prefix in cases of 0.


fgalan (approved):
LGTM

PR #3846, opened 2021-05-10
ADD null support to DateTime attrs and mds and geo:xxxx attrs

fgalan:
Issue https://github.com/telefonicaid/fiware-orion/issues/3533

Test cases / tasks:

- [x] null in DateTime attrs/mds
- [x] null in DateTime attrs/mds (sub)
- [x] null in geo:xxx attrs
- [x] null in geo:xxx attrs (sub)
- [x] documentation (check)
- [x] CNR

PR #3849, opened 2021-05-13
(JP) Update documentation (#3846)

fisuda:
This PR adds Japanese documentation about supporting to DateTime attrs and mds and geo:xxxx attrs.


fgalan (approved):
LGTM

Issue #1056, opened 2015-07-03
Cleanup lib/common/sem.cpp


kzangeli:
The curl-related functions in lib/common/sem.cpp to be moved to a separate file, probably in the **rest** library.

**Effort: 1 man day**

Anjali-NEC:
Hi @kzangeli @fgalan Sir,

I would like to work on this issue. Please assign this issue to me.
Thanks

fgalan:
Thanks! Assigned :)

Anjali-NEC:
Hi @fgalan Sir,

As per issue requirement I have created two files inside `src/lib/rest `named `curl.cpp` and `curl.h` and moved all the curl related functions from `src/lib/common/sem.cpp` and `src/lib/common/sem.h` to `src/lib/rest/curl.cpp` and `src/lib/rest/curl.h`

I have raised PR for the same. Please have a look, if it is ok then please merge into master.

PR #3869, opened 2021-06-01
Fix Issue1056

Anjali-NEC:
Fix issue #1056


src/lib/common/sem.cpp
```diff
@@ -551,21 +550,14 @@ const char* timeStatSemGet(void)
 
 
 /* ****************************************************************************
```
fgalan, line 552: You have leaved the following in sem.cpp, but I think should be also moved:

* contexts_mutex_taken
* endpoint_mutexes_taken
* connectionContextSemGet()
* connectionSubContextSemGet()

and this file ends in line 549

Or maybe I'm missing something and they should be keep here?

fgalan: In fact in seems that contexts_mutex_taken and endpoint_mutexes_taken are already in the new file, so it is just a matter or removing it from src/lib/common/sem.cpp

src/lib/rest/curl.cpp
```diff
@@ -0,0 +1,335 @@
+/*
```
fgalan, line 1: Maybe better name names:

src/lib/rest/curlSem.cpp

as plain "curl" is too broad.

src/lib/rest/curl.cpp
```diff
@@ -0,0 +1,335 @@
...
+#include "logMsg/traceLevels.h"
+
+#include "common/clockFunctions.h"
+#include "curl.h"
+
+
+
+/* ****************************************************************************
+*  curl context
+*/
```
fgalan, line 40:
```suggestion
```

As there is no need of distinguish the set of functions that follow this mark within a bigger file, as in the original place.

src/lib/rest/curl.cpp
fgalan: ...and similar comment for curl.h -> curlSem.h

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix: Cleanup lib/common/sem.cpp (#1056)
```
```suggestion
```

Except in special cases, refactors like this one are not mentioned in the CHANGES_NEXT_RELEASE file, as they are internal and for Orion user (the one to which changelogs are addressed) is transparent.

Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3869/commits/e4482579bd6b8125c68f3260d44d94f922e2cac6

src/lib/rest/curl.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3869/commits/5f663bcfa195d8da60f4c74e939e7d8c19b4bb9d

src/lib/rest/curl.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3869/commits/5f663bcfa195d8da60f4c74e939e7d8c19b4bb9d

src/lib/common/sem.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3869/commits/5f663bcfa195d8da60f4c74e939e7d8c19b4bb9d

fgalan (approved):
LGTM. Thanks!

PR #3873, opened 2021-06-03
FIX improve valgrind

fgalan:


src/app/contextBroker/contextBroker.cpp
```diff
@@ -583,6 +575,14 @@ void exitFunc(void)
   curl_context_cleanup();
   curl_global_cleanup();
 
+#ifdef DEBUG
```
fgalan, line 578: From a functional point of view moving this is not a problem. From the point of view of the valgrind job, it seems this way its better (some problems were found after introducing the delete notifier part recently)

NTC (informative)

mapedraza (approved):
LGTM

PR #3881, opened 2021-06-11
(JP) Remove docs about initial notification feature

fisuda:
This PR removes Japanese docs about initial notification feature. It would be great if you could review the PR.


fgalan (approved):
LGTM. Thanks!

PR #3889, opened 2021-06-21
Small fixing in MQTT PR

fgalan:


.github/workflows/functional.yml
```diff
@@ -18,7 +18,13 @@ jobs:
         image: mongo:4.4
         ports:
           - 27017:27017
-
+          
+      mosquitto:
```
fgalan, line 22: Only function-4 actually runs MQTT notification tests, but it's better to add mosquitto for all cases (as the tests sequence may vary from time to time)

NTC (informative)

mapedraza:
LGTM

Issue #3541, opened 2019-08-23
Subscription that pass to inactive state after a limit of failed connection attemps


fgalan:
A new optional field would be used (in the subscription payload) to specify a limit of connection attempts, so when that number of failing notifications is reached, then the subscription passes automatically to inactive state. The counter is reset in the following situations:

* When a successful notification is sent
* When the subscription is modified to pass from inactive to active state

tomsluyts:
It might be useful to qualify the cause of the "inactive" state - for example with an additional value or something, else you wouldn't know that the "inactive" state was triggered by reaching the max. number of connection tries. Somebody could have done it "manually" or something.

This could also allow for monitoring applications to pick up faulty endpoints.

fgalan:
Good point.

Currently status enumeration is: "active", "inactive", "oneshot", "expired" and "failed". The new status, in fact, would be a specialization of "failed"... maybe "failedMaxReached" (I would prefer a shorter name, in the same style of the other, but I don't find a good alternative right now... any suggestion? :) )

tomsluyts:
I've got "connectionsfailure", "failedevents", "failedconnections" or "unhandledevents" as possibilities

Anjali-NEC:
Hi @fgalan @tomsluyts Sir,

I hope this is and old issue like  #3451 is it still need to be fixed ? if yes then I will try to analyze whether I could finish it under my bandwidth before start working on it.

fgalan:
I think this would be an interesting feature, so the issue is still valid.

In a first iteration, I'd suggest you to focus on the simple case: just only one "inactive" status, without taken into account the cause of the fail. Probably with the existing `lastFailureReason` field is enough to know that was the cause of the problem.

tomsluyts:
Thanks for picking this up. We've had a lot of issues with inactive subscriptions eating up system resources with retries, unless you have a closely managed system where people spend time to see if everything is still up and running on the subscriber side as expected, this quickly becomes a serious issue if you have a lot of subscribers and you lose track.

fgalan:
Related issue: https://github.com/telefonicaid/fiware-orion/issues/930

Anjali-NEC:
Please assign this issue to me.

Thanks

fgalan:
Thanks for your willingness to work on this! Assigned.

Anjali-NEC:
Hi @fgalan Sir,

[...]

PR #3888, opened 2021-06-21
MQTT notifications

fgalan:
Issue #3001

Sub-PRs included in this PR:

- [x] PR #3194 - the work done by @McMutton
- [x] PR #3889 - some small fixes after merging the former (pending comments in PR #3194, code style, functional tests & accumulator script, etc.)
- [x] PR #3897 - CRUD of the new information in the sub document, taking into account new fields for the csubs cache entries)
- [x] PR #3899 - Effective use of new information (topic, QoS, etc.) but still in mono-MQTTbroker way
- [x] PR #3903 - Implement multi-MQTTBroker (removing the CLI fields `-mqttHost` and `-mqttPort`)
- [x] PR #3904 - Improve MQTT subscription parsing
- [x] PR #3905 - Upgrade to last current mosquitto version (2.0.11)
- [x] PR #3912 - Documentation

Check also before merging:

- [x] CNR - PR #3912
- [x] Documenation (user, admin, devel) - PR #3912
- [x] Check if RPM builds (or if modifications to .spec are needed) - see [this comment](https://github.com/telefonicaid/fiware-orion/pull/3905#issuecomment-895154304)
- [x] Solve all pending FIXME PR comments - checked at commit c3105c7, no "FIXME PR" occurrence

PR #3890, opened 2021-07-08
Subscription that pass to inactive state after a limit of failed connection attemps

Anjali-NEC:
Updated files for issue #3541


src/lib/apiTypesV2/Registration.cpp
```diff
@@ -36,7 +36,7 @@ namespace ngsiv2
 *
```
fgalan, line 36: An entry regarding the changes included in this PR and citing the issue number should be added to CHANGES_NEXT_RELEASE_FILE

src/lib/apiTypesV2/Registration.cpp
```diff
@@ -36,7 +36,7 @@ namespace ngsiv2
...
 * ForwardingInformation::ForwardingInformation -
 */
-ForwardingInformation::ForwardingInformation(): lastFailure(0), lastSuccess(0), timesSent(0), lastForwarding(0)
+ForwardingInformation::ForwardingInformation(): lastFailure(0), lastSuccess(0), timesSent(0), counter(0), maxLimit(0), lastForwarding(0)
```
fgalan, line 39: `counter` and `maxLimit` are too broad names. I'd suggest to use some more specific ones, such as:

failsCounter
maxFailsLimit

fgalan: I'd just realized this change is in Registration.cpp file

Note that #3541refers only to subscriptions. Registrations are out of scope. Please revert the change to Registration.cpp and Registration.h in your PR.

src/lib/apiTypesV2/Subscription.h
```diff
@@ -63,6 +65,8 @@ struct Notification
     blacklist(false),
     onlyChanged(false),
     timesSent(0),
+    counter(0),
+    maxLimit(0),
```
fgalan, line 69: `counter` and `maxLimit` are too broad names. I'd suggest to use some more specific ones, such as:

failsCounter
maxFailsLimit

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -118,6 +123,16 @@ std::string Notification::toJson(const std::string& attrsFormat)
     jh.addNumber("timesSent", this->timesSent);
   }
 
+  if (this->counter > 0)
```
fgalan, line 126: Getting the counter in GET subscription requests would make sense only in the case failed max fails feature is enabled, so maybe this should be changed to:

```suggestion
  if ((this->failsCounter> 0) && (this->maxFailsLimit> 0)
```

taking also into account the names change proposed at https://github.com/telefonicaid/fiware-orion/pull/3890/files#r667784099

src/lib/mongoBackend/dbConstants.h
```diff
@@ -94,6 +94,8 @@
 #define CSUB_EXPR_GEOREL             "georel"
 
 #define CSUB_THROTTLING              "throttling"
+#define CSUB_COUNTER                 "counter"
+#define CSUB_MAXLIMIT                "maxLimit"
```
fgalan, line 98: New field in DB should be described in documentation at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/database_model.md#csubs-collection

src/lib/mongoBackend/mongoGetSubscriptions.cpp
```diff
@@ -199,6 +201,18 @@ static void setNotification(Subscription* subP, const orion::BSONObj& r, const s
...
+     if (nP->counter < nP->maxLimit)
+     {
+       nP->counter = nP->counter +1;
+     }
+    }
+    else
+    {
+      nP->counter = -1;
+    }
+
```
fgalan, line 215: This is not the place where you should include counter increase. Note that mongoGetSubscriptions.cpp logic is about rendering the subscription information at DB into the JSON payload responses to GET subscriptions requests.

I think the right places should be at 

* src/lib/ngsiNotify/QueueWorkers.cpp function workerFunc()
*  src/lib/ngsiNotify/senderThread.cpp function startSenderThread()

In both case, the notification is done by httpRequestSendWithCurl() or httpRequestSend() (depending the case), getting the result value in `r` variable. You will see a structure like this in both places:

```
if (r == 0)
{
   <things we do if notification was sent correctly>
}
else
{
  <things we do if notification failed>
}
```

So:

* In the `<things we do if notification was sent correctly>` block you should include the logic that reset the counter to 0
* In the `<things we do if notification failed>` block you should include the logic that increases the counter and if the counter overpasses the limit change the status of the subscription to inactive (both in DB and in csubs cache)

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -97,6 +97,11 @@ std::string Subscription::toJson(void)
     jh.addNumber("throttling", this->throttling);
   }
 
+  if (this->notification.counter > this->notification.maxLimit)
+  {
+    jh.addString("status", "inactive");
+  }
+
```
fgalan, line 104: I thinks this should be removed.

toJson() is just for rendering the inforamtion you have in the DB. The subscription would already be in inactive state if the logic that I mention at https://github.com/telefonicaid/fiware-orion/pull/3890/files#r667810559 is implemented correctly.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1980,6 +1980,34 @@ static unsigned int processSubscriptions
         cacheSemGive(__FUNCTION__, "update lastNotificationTime for cached subscription");
       }
     }
+    else
```
fgalan, line 1983: Maybe I'm not getting the point, but I'd say that this else block is not needed here. Please take into account my comments at https://github.com/telefonicaid/fiware-orion/pull/3890/files#r667810559

src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -97,6 +97,11 @@ std::string Subscription::toJson(void)
     jh.addNumber("throttling", this->throttling);
```
fgalan, line 97: Note that apart from modifications to toJson() to render the new fields, you would also need to modify the parsing logic for the new fields. Have a look to src/lib/jsonParseV2/parseSubscription.cpp

And of course documenting the new feature :), but let's wait for the final version of the PR to do that.

src/lib/apiTypesV2/Registration.cpp
Anjali-NEC: Added in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/apiTypesV2/Registration.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/apiTypesV2/Subscription.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/mongoBackend/dbConstants.h
Anjali-NEC: Added new field in document https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/database_model.md#csubs-collection

src/lib/apiTypesV2/Subscription.h
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/apiTypesV2/Subscription.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/mongoBackend/MongoCommonUpdate.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/apiTypesV2/Subscription.cpp
Anjali-NEC: Added parsing logic for the new field in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/49b0c6df2c4d993b4e0fec160026fe0ac0017cb7

src/lib/mongoBackend/mongoGetSubscriptions.cpp
Anjali-NEC: Added the counter and logic in `src/lib/ngsiNotify/QueueWorkers.cpp` and `src/lib/ngsiNotify/senderThread.cpp` but I am facing issue while comparison of failsCounter with maxFailsLimit. Could you please guide me on this? Thanks

src/lib/apiTypesV2/Registration.cpp
```diff
@@ -141,6 +141,7 @@ std::string ForwardingInformation::toJson()
 
   jh.addNumber("timesSent", timesSent);
 
+
```
fgalan, line 144: Remove extra blank line

```suggestion
```

src/lib/cache/subCache.h
```diff
@@ -192,6 +194,8 @@ extern void subCacheItemInsert
   const std::vector<std::string>&    conditionAttrs,
   const char*                        subscriptionId,
   int64_t                            expiration,
+  int64_t                             failsCounter,
+  int64_t                            maxFailsLimit,
```
fgalan, line 198: Fix indent:

```suggestion
  int64_t                            failsCounter,
  int64_t                            maxFailsLimit,
```

src/lib/ngsi10/SubscribeContextRequest.h
```diff
@@ -54,6 +55,7 @@ typedef struct SubscribeContextRequest
   Restriction            restriction;            // Optional
   NotifyConditionVector  notifyConditionVector;  // Optional
   Throttling             throttling;             // Optional
+  MaxFailsLimit          maxFailsLimit;          // Optional
```
fgalan, line 58: Why do you need a whole class for this? Just an usigned int should suffice...

fgalan: Morevover, I'd realized this is the NGSIv1 subscription request class.

NGSIv1 is deprecated, so no need of modifying this file in anyway. The new functionality will be only for NGSIv2.

src/lib/ngsi10/UpdateContextSubscriptionRequest.cpp
```diff
@@ -124,5 +124,6 @@ void UpdateContextSubscriptionRequest::toNgsiv2Subscription(SubscriptionUpdate*
   subUp->notificationProvided = false;  // NGSIv1 doesn's allow changes in that part
   subUp->attrsFormatProvided  = true;   // updating in NGSIv1 involves and implicit change to NGSIv1 legacy format
   subUp->throttlingProvided   = !throttling.isEmpty();
+  subUp->maxFailsLimitProvided   = !maxFailsLimit.isEmpty();
```
fgalan, line 127: NGSIv1 is deprecated, so no need of modifying this file in anyway. The new functionality will be only for NGSIv2.

src/lib/ngsi10/SubscribeContextRequest.cpp
```diff
@@ -57,7 +57,8 @@ std::string SubscribeContextRequest::check(const std::string& predetectedError,
       ((res = duration.check())                                                         != "OK") ||
       ((res = restriction.check(restrictions))                                          != "OK") ||
       ((res = notifyConditionVector.check(SubscribeContext, predetectedError, counter)) != "OK") ||
-      ((res = throttling.check())                                                       != "OK"))
+      ((res = throttling.check())                                                       != "OK") ||
+      ((res = maxFailsLimit.check())                                                    != "OK"))
```
fgalan, line 61: NGSIv1 is deprecated, so no need of modifying this file in anyway. The new functionality will be only for NGSIv2.

Please revert the change in this file.

src/lib/mongoBackend/mongoGetSubscriptions.cpp
fgalan: You should get these values through SenderThreadParams object, expanding it with the new fields you need.

Note that the SenderThreadParams object that the workerFunc() and startSenderThread() functions use to deal with the notification comes at the end from the triggered subscriptions (either in the DB or in the csubs cache) during an entity creation/update. Please have a look to the buildSenderParams() and buildSenderParamsCustom() methods, which are the ones which build the SenderThreadParams object. You should expand them to take into account the new parameters which, in sequence, will lead you to expand the TriggeredSubscription class.

You may find useful the work I'm currently doing at PR https://github.com/telefonicaid/fiware-orion/pull/3888. You can have a look to how new csub fields (mqttTopic and mqttQoS) have been added and try to do something similar for yours.

fgalan: Maybe I haven't explained correctly. I'll try again to clarify.

* Orion checks triggered subscriptions at entity creation/update. Specifically, this is done by either `addTriggeredSubscriptions_withCache()` (when csubs cache is used) or `addTriggeredSubscriptions_noCache()` (when csub cache is not used). The usage of csubs cache or not depends on the `-noCache` CLI. An important difference between both is that `_withCache()` takes the subscription for the csub cache (a data structure managed in the Orion memory) while `_noCache()` takes the subscription from the DB. Whichever the case, the corresponding subscription with all the needed information is stored as `TriggeredSubscription` object in a hashmap.
* After the add triggered subscriptions step, there is another function named `processSubscriptions()` which processed every `TriggeredSubscription` object in that hashmap. At the end, this function dispatches the notifications invoking the `sendNotifyContextRequest()` on the corresponding notifier object (`Notifier` class or `QueueNotifier` class).
* No matter which notifier, the `sendNotifyContextRequest()` builds the `SenderThreadParams` object which stores all the information that the processing thread needs to effectively sent the notification by the HTTP interface at the end. This is done by calling the `buildSenderParams()` function. Once the params have been built, then the notification is passed to the thread that will effectively sent it (either directly in the case of `Notifier` or using a notification queue in the case of `QueueNotifier`), by the `workerFunc()` or `startSenderThread()` functions, depending the case.

As you can deduct from the above, the design principle is to access to the subscription information at a single point (the add triggered subscriptions step) so the rest of the sequence is independent of that. So, you don't have to worry about if the subscription came from the DB or the csub cache when you are processing the `SenderThreadParams` to send the notification at `workerFunc()` or `startSenderThread()`. Of course, the "disable after a given number of fails functionality" has to work both in cache or not cache cases, i.e. either if Orion is started with `-noCache` or not.

I hope this time I have explained with you cannot access directly to subscriptions in DB in either `workerFunc()` or `startSenderThread()` using `SenderThreadParams` instead. A solution based on getting the subscription information again from these function would go agains our design pattern.

If you want to see a practical example of what I say, please have a look to PR #3888. In that PR I needed to "pass-through" some parameteres in the subcription (encapsulated in a `MqttInfo` object) to the final notification logic at `workerFunc()` or `startSenderThread()`. I'd suggest to have a close look to the modifications in `addTriggeredSubscriptions_withCache()`, `addTriggeredSubscriptions_noCache()`, `TriggeredSubscription`, `SenderThreadParams` and `buildSenderParams()/buildSenderParamsCustom()` to understand how this is done.

Anjali-NEC: As per your suggesstion I have added new fields in SenderThreadParams (params->maxFailsLimit and params->failsCounter) and also done modifications in `addTriggeredSubscriptions_withCache()`, `addTriggeredSubscriptions_noCache()`,` TriggeredSubscription`, `SenderThreadParams` and `buildSenderParams()/buildSenderParamsCustom()`. Also include these parametes in `workerFunc()` or `startSenderThread()` and get the values of params->maxFailsLimit and params->failsCounter.

Now I am trying to increase the params->failsCounter in else part in `src/lib/ngsiNotify/senderThread.cpp` as suggested by you in comment https://github.com/telefonicaid/fiware-orion/pull/3890#discussion_r667810559 but the failsCounter increased only once.

As per my understaing the SenderThreadParams is called only once per notification So, the failsCounter value increase only by 1 when the receiver goes down.

```
void* startSenderThread(void* p)
{
  std::vector<SenderThreadParams*>* paramsV = (std::vector<SenderThreadParams*>*) p;

  for (unsigned ix = 0; ix < paramsV->size(); ix++)
  {
    SenderThreadParams* params = (SenderThreadParams*) (*paramsV)[ix];
    char                portV[STRING_SIZE_FOR_INT];
    std::string         url;

    snprintf(portV, sizeof(portV), "%d", params->port);
    url = params->ip + ":" + portV + params->resource;

    strncpy(transactionId, params->transactionId, sizeof(transactionId));

    LM_T(LmtNotifier, ("sending to: host='%s', port=%d, verb=%s, tenant='%s',  maxFailsLimit='%lu', failsCounter='%lu', service-path: '%s', xauthToken: '%s', path='%s', content-type: %s",
                       params->ip.c_str(),
                       params->port,
                       params->verb.c_str(),
                       params->tenant.c_str(),
                       params->maxFailsLimit,
                       params->failsCounter,
                       params->servicePath.c_str(),
                       params->xauthToken.c_str(),
                       params->resource.c_str(),
                       params->content_type.c_str()));

    long long    statusCode = -1;
    std::string  out;
    LM_T(LmtNotificationRequestPayload , ("notification request payload: %s", params->content.c_str()));

    if (!simulatedNotification)
    {
      int          r;

      r = httpRequestSend(params->from,
                          params->ip,
                          params->port,
                          params->protocol,
                          params->verb,
                          params->tenant,
                          params->maxFailsLimit,
                          params->failsCounter,
                          params->servicePath,
                          params->xauthToken,
                          params->resource,
                          params->content_type,
                          params->content,
                          params->fiwareCorrelator,
                          params->renderFormat,
                          &out,
                          &statusCode,
                          params->extraHeaders);

      LM_T(LmtNotificationResponsePayload, ("notification response: %s", out.c_str()));
      if (r == 0)
      {
        __sync_fetch_and_add(&noOfNotificationsSent, 1);
        alarmMgr.notificationErrorReset(url);
        params->failsCounter = 0;

        if (params->registration == false)
        {
          subNotificationErrorStatus(params->tenant, params->subscriptionId, 0, statusCode, "");
        }
      }

      else
      {
        alarmMgr.notificationError(url, "notification failure for sender-thread: " + out);
        params->failsCounter = params->failsCounter +1;
        if (params->registration == false)
        {
          subNotificationErrorStatus(params->tenant, params->subscriptionId, -1, -1, out);
        }
      }
    }

```

 Can you please confirm my understating.

There are two methods to test the connection attempts:
1. Only send update  when subscriber is down. 
2. Sending update again and again when subscriber is down.

As per my understanding we should test according to 1.

src/lib/cache/subCache.cpp
```diff
@@ -769,6 +771,8 @@ void subCacheItemInsert
   cSubP->servicePath           = strdup(servicePath);
   cSubP->subscriptionId        = strdup(subscriptionId);
   cSubP->expirationTime        = expirationTime;
+  cSubP->failsCounter          = 0;
```
fgalan, line 774: Why don't use the failCounter parameter (introduced in L739) instead of 0?

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -710,6 +744,11 @@ void mongoSubCountersUpdate
     mongoSubCountersUpdateCount(db, COL_CSUBS, subId, count);
   }
 
+  if (failsCounter > 0)
+  {
+    mongoSubCountersUpdatefailsCounter(db, COL_CSUBS, subId, failsCounter);
```
fgalan, line 749: Maybe instead of defining a new function for this case (mongoSubCountersUpdatefailsCounter) maybe you can extend the mongoSubCountersUpdateCount() function, calling it this way (line 744):

```
mongoSubCountersUpdateCount(db, COL_CSUBS, subId, count, failsCounter);
```

and do the `failsCounter > 0` inside mongoSubCountersUpdateCount()

src/lib/cache/subCache.cpp
```diff
@@ -1277,12 +1284,12 @@ void subNotificationErrorStatus
     if (errors == 0)
     {
       // count == 0 (inc is done in another part), lastFailure == -1, failureReason == -1
-      mongoSubCountersUpdate(tenant, subscriptionId, 0, now, -1, now, "", statusCode);
+      mongoSubCountersUpdate(tenant, subscriptionId, 0, 0, now, -1, now, "", statusCode);
```
fgalan, line 1287: Question: why failsCounter is 0 here? It happens the same as with count (`count == 0 (inc is done in another part)`)?

src/lib/mongoBackend/mongoGetSubscriptions.cpp
fgalan: Thanks for the update! I have provided extra feedback in the form of new comments.

----------

With regards to 

> As per my understaing the SenderThreadParams is called only once per notification So, the failsCounter value increase only by 1 when the subscriber goes down.

I think that's correct. Each time a notification fails the increase is by 1 in the corresponding counter associated to the subscription.

------------

With regards to

> There are two methods to test the connection attempts:
> 
>     1. Only send update  when subscriber is down.
> 
>     2. Sending update again and again when subscriber is down.
> 
> 
> As per my understanding we should test according to 1.

Not sure if I understand what you mean... Anyway, the right behaviour should be the following, as far as I understand:

1. Each time the notification fails +1 is added to the fails counter associated to the subscription
2. Each time the notification goes ok the fails counter is reset to 0
3. Each time the subscription goes from "inactive" to "active" state (due to a manual intervention on the subscription API) the fail counter is reset o 0 (this is needed to "reintroduce" a given subscription once it has been disabled due to this feature)
3. If the fail counter goes beyond the max fails value, then the subscription passes to "inactive" status. That would cause no new notification gets send (nothing to implement here, that logic already exists in Context Broker).

src/lib/mongoBackend/mongoSubCache.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/0d976f5d55b682e27c153ecd5662ef408f351a1e

src/lib/cache/subCache.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/0d976f5d55b682e27c153ecd5662ef408f351a1e

src/lib/cache/subCache.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/0d976f5d55b682e27c153ecd5662ef408f351a1e

src/lib/mongoBackend/mongoGetSubscriptions.cpp
Anjali-NEC: > Not sure if I understand what you mean... Anyway, the right behaviour should be the following, as far as I understand:
> 
> Each time the notification fails +1 is added to the fails counter associated to the subscription
> Each time the notification goes ok the fails counter is reset to 0
> Each time the subscription goes from "inactive" to "active" state (due to a manual intervention on the subscription API) the fail counter is reset o 0 (this is needed to "reintroduce" a given subscription once it has been disabled due to this feature)
> If the fail counter goes beyond the max fails value, then the subscription passes to "inactive" status. That would cause no new notification gets send (nothing to implement here, that logic already exists in Context Broker).

Fixed issue in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/191b4faed1d9904e41ca7ab924c7a515934c0c4c 
Please have a look my modifications.

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1,9 @@
+- Fix: Subscription that pass to inactive state after a limit of failed connection attemps (#3541)
+- Fix: TextUnrestricted recursively applies to JSON objects and array (#3867)
+- Hardening: Improve notification pool worker thread termination logic (#3877)
+- Remove: deprecated feature initial notification upon subscription creation or update
+- Fix: -key and -cert files loading (was causing valgrind errors) (#3925)
+- Add: MQTT notifications (#3001)
+- Fix: return 400 instead of 500 when GeoJSON syntax error occur (#3347)
+- Fix: onlyChangedAttrs set to true not working after csubs cache refresh (#3906)
+- Fix: 400 BadRequest when attempting to DELETE with multi-servicepath or SP with '/#' (instead of silently failing) (#3071)
```
fgalan, line 9: CHANGES_NEXT_RELESEA should include only the changes in this PR. It seems the conflict in this file after the flush on it in master for the Orion 3.2.0 release was not solved correctly...

This modification should fix:

```suggestion
- Fix: Subscription that pass to inactive state after a limit of failed connection attemps (#3541)
```

Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/d172ee893403b1b26b3591f74a05a0ea76a57b56

src/lib/ngsiNotify/senderThread.cpp
```diff
@@ -114,16 +133,37 @@ void* startSenderThread(void* p)
...
           subNotificationErrorStatus(params->tenant, params->subscriptionId, 0, statusCode, "");
         }
       }
+
+
       else
       {
         alarmMgr.notificationError(url, "notification failure for sender-thread: " + out);
 
+        params->failsCounter = params->failsCounter +1;
```
fgalan, line 149: Maybe better (more compact):

```suggestion
        params->failsCounter++;
```

Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/0ebe8e05845a5d383baede531e80f33d57dced29

Anjali-NEC:
@fgalan sir,

After the modification in PR, Compliance test and two functional test cases are failing:     
```
1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications.test
1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications_stop_start_accumulator.testfails 
```
I don't know why these test are failing. I didn't see any failure related to this on my local setup. Can you please suggest if any other modification are required in this PR.

fgalan:
> 
> 
> @fgalan sir,
> 
> After the modification in PR, Compliance test and two functional test cases are failing:
> 
> ```
> 1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications.test
> 1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications_stop_start_accumulator.testfails 
> ```
> 
> I don't know why these test are failing. I didn't see any failure related to this on my local setup. Can you please suggest if any other modification are required in this PR.

Some time there are glitches in the test execution. I've just launched a new run to check it.

In addition, note the fail in the compliance test (https://github.com/telefonicaid/fiware-orion/pull/3890/checks?check_run_id=3600883820). I think the problem the problems are:

```
0001 errors of category 'Line ends in whitespace'
0001 errors of category 'Blank line at the end of a code block'

src/lib/mongoBackend/mongoSubCache.cpp:515:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
src/lib/mongoBackend/mongoUpdateSubscription.cpp:215:  Blank line at the end of a code block.  Is this needed?  [whitespace/blank_line] [3]
```

doc/manuals/admin/database_model.md
```diff
@@ -325,6 +325,8 @@ Fields:
 -   **lastSuccessCode**: HTTP code (200, 400, 404, 500, etc.) returned by receiving endpoint last
     time a successful notification was sent.
     Not present if the subscription has never provoked a successful notification.
+-   **maxFailsLimit**: An optional field used to specify the maximum limit of connection attempts, so when that number of failing notifications is reached, then the subscription passes automatically to inactive state.
+-   **failsCounter**: the number of failing notifications associated to the subscription.
```
fgalan, line 329: Regardless any other document describing the functionality in detail, I think is a good idea to explain how the failCounter field is modified. Thus, I'd suggest the following modification:

```suggestion
-   **failsCounter**: the number of failing notifications associated to the subscription. This is increased by one each time a notification attempt fails. It is reset to 0 if a notification attempt successes or the subscription transition from "inactive" to "active" status.
```

src/lib/jsonParseV2/parseSubscription.cpp
```diff
@@ -241,6 +241,38 @@ std::string parseSubscription(ConnectionInfo* ciP, SubscriptionUpdate* subsP, bo
...
+  }
+  else if (failsCounterOpt.given)
+  {
+    subsP->failsCounterProvided = true;
+    subsP->failsCounter = failsCounterOpt.value;
+  }
+  else if (!update)
+  {
+    subsP->failsCounter = 0;
+  }
```
fgalan, line 274: failsCounter is not a field to be provided by the client in subscription requests (note the client only provided the maxFailsLimit and faildCounter in a calculated field by CB). Thus no need of parsing it and this block of code should be removed.

src/lib/mongoBackend/mongoSubCache.cpp
```diff
@@ -488,18 +492,27 @@ static void mongoSubCountersUpdateCount
...
   condition.append("_id", orion::OID(subId));
   countB.append(CSUB_COUNT, count);
   update.append("$inc", countB.obj());
 
+  if (failsCounter > 0)
+  {
+  failsCounterB.append(CSUB_FAILSCOUNTER, failsCounter);
+  update.append("$inc", failsCounterB.obj());
+  }
+
```
fgalan, line 515: Indent fix:

```suggestion
  if (failsCounter > 0)
  {
    failsCounterB.append(CSUB_FAILSCOUNTER, failsCounter);
    update.append("$inc", failsCounterB.obj());
  }
 
```

src/lib/ngsi/FailsCounter.cpp
```diff
@@ -0,0 +1,114 @@
+/*
```
fgalan, line 1: Are src/lib/ngsi/FailsCounter.cpp and src/lib/ngsi/FailsCounter.h really needed? It seems your are not actually using them.

Maybe they are a leftover of previous implementation and they are no longer needed? That would be desirable, in which case please remove them.

src/lib/ngsi/MaxFailsLimit.cpp
```diff
@@ -0,0 +1,114 @@
+/*
```
fgalan, line 1: Are src/lib/ngsi/MaxFailsLimit.cpp and src/lib/ngsi/MaxFailsLimit.h really needed? It seems your are not actually using them.

Maybe they are a leftover of previous implementation and they are no longer needed? That would be desirable, in which case please remove them.

src/lib/ngsiNotify/Notifier.h
```diff
@@ -48,6 +48,8 @@ class Notifier
   virtual void sendNotifyContextRequest(NotifyContextRequest&            ncr,
                                         const ngsiv2::Notification&      notification,
                                         const std::string&               tenant,
+					long long                        maxFailsLimit,
+					long long                        failsCounter,
```
fgalan, line 52: Review indent, please. Be careful with tabs, use only space.

src/lib/ngsiNotify/QueueWorkers.cpp
```diff
@@ -247,6 +260,25 @@ static void* workerFunc(void* pSyncQ)
           QueueStatistics::incSentError();
           alarmMgr.notificationError(url, "notification failure for queue worker: " + out);
 
+          params->failsCounter++;
+
+          if ((params->failsCounter)  > (params->maxFailsLimit))
```
fgalan, line 265: Simplify to:

```suggestion
          if (params->failsCounter > params->maxFailsLimit)
```

In addition, please review indent of the if block. [Style rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/contribution_guidelines.md#m4-indent) is to use 2 spaces, not 1 (see )

fgalan: In fact, there are bigger problems in this part:

* Modifications done to the params (such as `params->failsCounter = 0` or `params->failsCounter++`) are not going to be persisted in the subscription (either cache nor DB). Note that `params` is a read-only object.
* You should avoid direct access to cache/DB in this `workerFunc()`. That access should be done in the `subNotificationErrorStatus()` function, which is though to update subscription information when the notification fail, both in cache and DB. I think you have done some modifications to that function in this PR but maybe not all them (in particular, the check to pass the status to inactive)

src/lib/ngsiNotify/senderThread.cpp
```diff
@@ -114,16 +133,37 @@ void* startSenderThread(void* p)
...
           subNotificationErrorStatus(params->tenant, params->subscriptionId, 0, statusCode, "");
         }
       }
+
+
       else
       {
         alarmMgr.notificationError(url, "notification failure for sender-thread: " + out);
 
+        params->failsCounter++;
```
fgalan, line 149: Similar comments to [the ones provided for `workerFunc()`](https://github.com/telefonicaid/fiware-orion/pull/3890/files#r709082649) also apply here.

fgalan:
I've realized this PR doesn't include any new .test file.

I think it would be a good idea to include some of them, covering the use case of the feature being implemented. Although the PR is still in progress (see my recent comments above) I think it could help to guide the implementation.

src/lib/ngsiNotify/QueueWorkers.cpp
Anjali-NEC: > In fact, there are bigger problems in this part:
> 
> * Modifications done to the params (such as `params->failsCounter = 0` or `params->failsCounter++`) are not going to be persisted in the subscription (either cache nor DB). Note that `params` is a read-only object.
> * You should avoid direct access to cache/DB in this `workerFunc()`. That access should be done in the `subNotificationErrorStatus()` function, which is though to update subscription information when the notification fail, both in cache and DB. I think you have done some modifications to that function in this PR but maybe not all them (in particular, the check to pass the status to inactive)

I didn't understand clearly what change we need to do in `src/lib/ngsiNotify/QueueWorkers.cpp` and `src/lib/ngsiNotify/senderThread.cpp` .

We need to add params->failsCounter in subNotificationErrorStatus() instread of using directly in `src/lib/ngsiNotify/QueueWorkers.cpp` and `src/lib/ngsiNotify/senderThread.cpp`? Also please confirm the code of setting  status=inactive in DB and cache after comparison of `if (params->failsCounter > params->maxFailsLimit)` is correct or we need to change this code also?

Also, Currently the default value of params->failsCounter is long long random number instead of 0. Can you suggest where we should set the value of this params->failsCounter.

fgalan: The bottom line (following the principle of "separation of concerns" explained also [here](https://github.com/telefonicaid/fiware-orion/pull/3890#discussion_r677511495)) is that the code in `workerFunc()` and `startSenderThread()` only deals with the notification sending and detecting error but the process done in case of error (including updating subscription in cache and DB) is delegated to `subNotificationErrorStatus()`). You shouldn't need any of this in this file:

```
#include "cache/subCache.h"
#include "mongoDriver/connectionOperations.h"

#include "mongoBackend/MongoGlobal.h"
#include "mongoDriver/safeMongo.h"
#include "mongoBackend/MongoCommonUpdate.h"
#include "mongoBackend/dbConstants.h"
#include "ngsi/MaxFailsLimit.h"
```

The current version of `subNotificationErrorStatus()` is fine, but I think it should also include the subscription state change in the case of reaching maximum number of allowed fails ("moving" the block in L265-L280 currently in src/lib/ngsiNotify/QueueWorkers.cpp to `subNotificationErrorStatus()`).

With regards to default value for `params->failsCounter` I understand the place to initialize it is the one in which you are setting all the other fields in `params`, i.e. the `buildSenderParams()` function (as explained [here](https://github.com/telefonicaid/fiware-orion/pull/3890#discussion_r677511495)). It should be initialized with the values coming from the subscription,

doc/manuals/admin/database_model.md
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/jsonParseV2/parseSubscription.cpp
Anjali-NEC: Removed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/mongoBackend/mongoSubCache.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/ngsi/MaxFailsLimit.cpp
Anjali-NEC: Removed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/ngsiNotify/Notifier.h
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/ngsiNotify/QueueWorkers.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/ngsiNotify/senderThread.cpp
Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

src/lib/ngsi/FailsCounter.cpp
Anjali-NEC: Removed in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/6b68a8c9ded45dd2e30c033f4d5000cddfab19de

Anjali-NEC:
> I've realized this PR doesn't include any new .test file.
> 
> I think it would be a good idea to include some of them, covering the use case of the feature being implemented. Although the PR is still in progress (see my recent comments above) I think it could help to guide the implementation.

Hi @fgalan sir,

I have added test case in https://github.com/telefonicaid/fiware-orion/pull/3890/commits/1b3c0bfaa42f5080b8aa29db322cadf05f2a2010 . Please review my PR.

Due to my current modification some test cases are failing:

```
  o  1126_GET_v2_subscriptions/GET_v2_all_subs_pagination.test
  o  1126_GET_v2_subscriptions/GET_v2_all_subscriptions.test
  o  1126_GET_v2_subscriptions/lastsuccesscode_and_lastfailurereason.test
  o  1126_GET_v2_subscriptions/sub_from_active_to_fail.test
  o  1126_GET_v2_subscriptions/sub_from_fail_to_active.test
  o  1317_GET_v2_subscription/get_subscription.test
  o  1577_servicepath_get_subs/get_subs_with_servicepath.test
  o  1693_preserve_sp_updating_subscription/preserve_sp_updating_subscription_ngsiv1.test
  o  1852_escaped_slashes_in_json_v1_payload/escaped_slashes_in_json_v1_payload.test
  o  1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications.test
  o  1960_failure_stats_in_db_and_notifications/failure_stats_in_db_and_notifications_stop_start_accumulator.test
```
As per my analysis  some test cases are failing due to modification done in `src/lib/mongoBackend/mongoGetSubscriptions.cpp` to get the value of  `failsCounter` and `maxFailsLimit`. I didn't understand why these test case are failing. Is there any change needs to be done? 

Some test cases are failing due to modification done in src/lib/apiTypesV2/Subscription.cpp to get the status in case of "inactive". Currently the status is "failed" in case of notification failed.
```
    if ((this->notification.lastFailure > 0) && (this->notification.lastFailure > this->notification.lastSuccess))
    {
      jh.addString("status", "failed");
    }
    else
    {
      jh.addString("status", this->status);
    }
```

I didn't found any other way for replacing  `status: "failed"` to `status: "inactive"` when `failsCounter > maxFailsLimit`. Can you please suggest if any other way for this?

Anjali-NEC:
@fgalan Gentle Reminder!!

fgalan:
Unfortunately I'm now involved in urgent matters that doesn't allow to me review this PR. I'll do as soon as I can. Sorry for the inconveniences.

Anjali-NEC:
> The review of this PR is getting complicated and sometimes I think is better I do myself the change instead of suggesting it.
> 
> Thus, I'd suggest the following:
> 
> 1. Sync this PR with master a last time, solving the conflicts (you)
> 2. Change the base branch of this PR to an alternative "pre-landing" branch and merge (me)
> 3. Continue the work in the "pre-landing" branch with additional PRs to complete the work (you and me)
> 
> What do you think? If you agree, please complete step 1 and I'll continue with step 2.

I Agree, I have resolved the conflicts and sync this PR with master.

PR #3896, opened 2021-07-16
FIX remove old stuff

fgalan:
This PR removes:

* Unneeded constructor in httpInfo
* Old casub lefovers (functionality removed in CB 2.6.0 time ago but it seems we missed some things ;)


mapedraza (approved):
LGTM

PR #3898, opened 2021-07-23
Fix statusCode for createEntity (Post v2/entities op) with wrong GeoJSON

ArqumFarooqui-NEC:
New PR issue #3347 to fix the statusCode of createEntity (Post v2/entities) operation when wrongly formed GeoJSON sent.
(old PR #3733 )

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2796,6 +2796,10 @@ static bool createEntity
       {
         oeP->fill(SccInvalidModification, "Already Exists", "Unprocessable");
       }
+      else if (errDetail->find("Can't extract geo keys") != std::string::npos)
+      {
+        oeP->fill(SccBadRequest, "Bad Request", "BadRequest");
+      }
       else
       {
         oeP->fill(SccReceiverInternalError, *errDetail, "InternalError");
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -2798,7 +2798,7 @@ static bool createEntity
       }
       else if (errDetail->find("Can't extract geo keys") != std::string::npos)
       {
-        oeP->fill(SccBadRequest, "Bad Request", "BadRequest");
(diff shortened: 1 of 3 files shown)


ArqumFarooqui-NEC:
Hi @fgalan sir, 
I have raised this new PR  for issue #3347 because I was facing conflicting issue on previous PR i.e. #3733 
Please review this PR.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2796,6 +2796,10 @@ static bool createEntity
       {
         oeP->fill(SccInvalidModification, "Already Exists", "Unprocessable");
       }
+      else if (errDetail->find("Can't extract geo keys") != std::string::npos)
+      {
+        oeP->fill(SccBadRequest, "Bad Request", "BadRequest");
```
fgalan, line 2801: In order to provide a more informative feedback for user I'd suggest:

```suggestion
        oeP->fill(SccBadRequest, "Wrong GeoJson", "BadRequest");
```

fgalan: (.test file would need to be adapted accordingly)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2796,6 +2796,10 @@ static bool createEntity
       {
         oeP->fill(SccInvalidModification, "Already Exists", "Unprocessable");
       }
+      else if (errDetail->find("Can't extract geo keys") != std::string::npos)
```
fgalan, line 2799: An entry describing the change done in this PR should be added to CHANGES_NEXT_RELEAES. Something like this:

```
- Fix: return 400 instead of 500 when GeoJSON syntax error occur (#3347)
```

fgalan: Fixed in ae914b6

src/lib/mongoBackend/MongoCommonUpdate.cpp
ArqumFarooqui-NEC: Fixed in ae914b6

fgalan (approved):
LGTM. Thanks for your contribution!

PR #3902, opened 2021-08-02
FIX semTake and semGive move to private in AlarmManager class

fgalan:
Only internal members can take/release the sem.


fgalan:
I have realized that they are needed by logSummary.cpp. Thus, this PR doesn't make sense. Closing.

Issue #3884, opened 2021-06-11
Orion 3.1.0 crashes with PEP-Proxy (Wilma) 8.0.0


jason-fox:
Consider a set-up with a [PEP Proxy](https://github.com/ging/fiware-pep-proxy) before an Orion Instance - e.g. this [docker-compose.yml](https://github.com/FIWARE/tutorials.PEP-Proxy/blob/master/docker-compose/orion.yml) 

The PEP Proxy annotates the forwarded request with extra headers. With Orion 3.0.0 and earlier this is no problem, however the following request (either directly using Curl or indirectly using the tutorial) causes the 3.1.0 Broker to crash:

```console
curl -L -X PATCH 'http://localhost:1026/v2/entities/Bell:001/attrs' \
-H 'fiware-service: openiot' \
-H 'fiware-servicepath: /' \
-H 'Content-Type: application/json' \
-H 'host: orion-proxy:1027' \
-H 'accept: application/json' \
-H 'connection: close' \
-H 'X-Nick-Name: cccccccc-good-0000-0000-000000000000' \
-H 'x-auth-token: 2ca8a58ec971bfc6b0bcda8443931883eb5900d5' \
-H 'X-Display-Name: ' \
-H 'X-Roles: [{\"id\":\"security-role-0000-0000-000000000000\",\"name\":\"Security Team\"}]' \
-H 'X-Organizations: []' \
-H 'X-Eidas-Profile: {}' \
-H 'X-App-Id: tutorial-dckr-site-0000-xpresswebapp' \
-H 'x-forwarded-for: ::ffff:172.18.1.7' \
--data-raw '{
  "ring": {
      "type" : "command",
      "value" : ""
  }
}'
```

Specifically it is the `x-forwarded-for` header that can no longer be processed - the command works with Orion 3.0.0 or if the `x-forwarded-for` header is left out of the request.

The  log is shown below ( The `ring` attribute is set up as a Command via a registration.)

```text
time=2021-06-11T17:28:22.372Z | lvl=ERROR | corr=6bb56dd4-cada-11eb-a37b-0242ac120109 | trans=1623432459-057-00000000046 | from=::ffff:172.18.1./ | srv=openiot | subsrv=/ | comp=Orion | op=postUpdateContext.cpp[642]:postUpdateContext | msg=Internal Error (attribute 'ring' not found)
[...]

PR #3917, opened 2021-09-01
[WIP] FIX increase IP_LENGTH_MAX

fgalan:
Issue #3884

Not sure if this size will suffice. Moreover, no matter how longer the size is, a malicious user could always use even a larger payload. Maybe an additional fix to truncate should be needed, but not sure in which place (log library internals...).

Feedback pending, see https://github.com/telefonicaid/fiware-orion/issues/3884#issuecomment-906395500

CC: @jason-fox @kzangeli 

- [ ] CNR

### DIFF
--- a/src/lib/common/limits.h
+++ b/src/lib/common/limits.h
@@ -63,7 +63,7 @@
 *
 * Others -
 */
-#define IP_LENGTH_MAX           15     // Based on xxx.xxx.xxx.xxx
+#define IP_LENGTH_MAX           45     // Based on IPv6 max length, see https://stackoverflow.com/questions/166132/maximum-length-of-the-textual-representation-of-an-ipv6-address
 #define STRING_SIZE_FOR_INT     16     // Room enough for a 32 bit integer
 #define STRING_SIZE_FOR_LONG    20     // Room enough for a 64 bit integer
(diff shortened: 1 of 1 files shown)


fgalan:
> 45 or 46? Is this string `null` terminated? (Or go for the Engineering solution and round it up to 50)

Good point :)

All the initialization code of buffers based in IP_LENGTH_MAX add +1 to take into account the null terminator (see https://github.com/telefonicaid/fiware-orion/search?q=IP_LENGTH_MAX). However, raising to 50 seems also a good idea.

fgalan:
This PR is just a "dirty patch" for the problem, but the real fix should be in the log library authored by @kzangeli (see https://github.com/telefonicaid/fiware-orion/issues/3884#issuecomment-920717473)

@kzangeli taking into account the upcoming FIWARE TSC on Orion, could you provide an status on that (or be prepared to attend to the TSC and report at the meeting), please? Thanks!

fgalan:
If we cannot get support from logging library author to implement a proper fix for this maybe we should consider to change the logging library to one with a wider usage and better support, such as [zlog](https://github.com/HardySimpson/zlog).

fgalan:
Overpasse by PR #4039 (== PR #4026). Closing.

PR #3932, opened 2021-09-08
FIX optimize functional.yml

fgalan:


fgalan:
Name could be improved. SOF question about it: https://stackoverflow.com/questions/69105449/gitactions-job-name-based-on-matrix-index-instead-of-matrix-value

mapedraza (approved):
LGTM

PR #3937, opened 2021-09-14
Update contextBroker.spec to fix rpm build

wistefan:
cmake is installed from source and therefore not recognized as an installed package. Needs to be removed for the build to work.

### DIFF
--- a/rpm/SPECS/contextBroker.spec
+++ b/rpm/SPECS/contextBroker.spec
@@ -46,7 +46,7 @@ URL:        http://catalogue.fiware.org/enablers/publishsubscribe-context-broker
 Source:     %{name}-%{broker_version}.tar.gz
 BuildRoot: /var/tmp/%{name}-buildroot
 Requires:  libstdc++, boost-thread, boost-filesystem, gnutls, libgcrypt, libcurl, openssl, logrotate, libuuid
-Buildrequires: gcc, cmake, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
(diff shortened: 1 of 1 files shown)


rpm/SPECS/contextBroker.spec
```diff
@@ -46,7 +46,7 @@ URL:        http://catalogue.fiware.org/enablers/publishsubscribe-context-broker
 Source:     %{name}-%{broker_version}.tar.gz
 BuildRoot: /var/tmp/%{name}-buildroot
 Requires:  libstdc++, boost-thread, boost-filesystem, gnutls, libgcrypt, libcurl, openssl, logrotate, libuuid
-Buildrequires: gcc, cmake, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
+Buildrequires: gcc, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
```
fgalan, line 49: I hope that the current situation that forces to compile cmake from source to be transitory and that, eventually, CentOS 8 includes a version without the problems preclude it.

Thus, I suggest to comment the line with a comment instead of deleting. Something like this:

```suggestion
# FIXME currently cmake is installed from source, but we hope in the future revert this situation. See more detail at https://github.com/telefonicaid/fiware-orion/blob/master/docker/Dockerfile#L61
#Buildrequires: gcc, cmake, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
Buildrequires: gcc, gcc-c++, gnutls-devel, libgcrypt-devel, libcurl-devel, openssl-devel, boost-devel, libuuid-devel
```

fgalan (approved):
LGTM

PR #3939, opened 2021-09-16
FIX upgrade libmosquitto to 2.0.12

fgalan:
@fisuda I have touched some files belonging to the Japanese translation, but only numbers, so I think they are fine.


cesarjorgemartinez (approved):
LGTM

fisuda:
LGTM

PR #3935, opened 2021-09-13
Attribute operators (first part: without notifications)

fgalan:
Issue #3814

- [x] $inc, $max, $min, $mul
- [x] $push, $addToSet, $pull, $pullAll
- [ ] ~$bit~
- [x] Documentation, including
    - MongoDB errors pass-through explanation
    - Only works if operator is the first key, others keys (children > 0) are ignored. Random child if more than one.
    - What happen in create or replace operations?
    - Notification limitation (by the moment...)
- [x] CNR

PR #3943, opened 2021-09-27
(JP) Add Japanese documentation about Attribute operators (#3935)

fisuda:
This PR add Japanese documentation about Attribute operators.
It would be great if you could review this PR.


fgalan (approved):
Thanks for the typo fixing! :)

LGTM

Issue #3944, opened 2021-09-30
"legacyForwarding": true is shown in GET registrations for not-legacy cases


fgalan:
We have in DB one registration using legacyForwarding and other not using it (pay attention to "format" field):

```
> db.registrations.find().pretty()
{
	"_id" : ObjectId("61556a428015fb36194d4f2e"),
	"expiration" : NumberLong("9223372036854775807"),
	"servicePath" : "/",
	"contextRegistration" : [
		{
			"entities" : [
				{
					"id" : "E",
					"type" : "Access"
				}
			],
			"attrs" : [
				{
					"name" : "A",
					"type" : ""
				}
			],
			"providingApplication" : "http://localhost:1028/E/"
		}
	],
	"fwdMode" : "update",
	"format" : "normalized"
}
{
	"_id" : ObjectId("61556cb58015fb36194d4f2f"),
	"expiration" : NumberLong("9223372036854775807"),
	"servicePath" : "/",
	"contextRegistration" : [
		{
			"entities" : [
				{
					"id" : "E",
					"type" : "Access"
				}
			],
			"attrs" : [
				{
					"name" : "A",
					"type" : ""
				}
			],
			"providingApplication" : "http://localhost:1028/E/"
		}
	],
	"fwdMode" : "update",
	"format" : "JSON"
}
```

If we do GET /v2/registrations we see:

```
curl localhost:1026/v2/registrations | python -mjson.tool

[
    {
        "dataProvided": {
            "attrs": [
                "A"
            ],
            "entities": [
                {
                    "id": "E",
                    "type": "Access"
                }
            ]
        },
        "id": "61556a428015fb36194d4f2e",
        "provider": {
            "http": {
                "url": "http://localhost:1028/E/"
            },
            "legacyForwarding": true,
            "supportedForwardingMode": "update"
        },
        "status": "active"
    },
    {
        "dataProvided": {
            "attrs": [
                "A"
            ],
            "entities": [
                {
                    "id": "Faro_Porto_Colom",
                    "type": "Access"
                }
            ]
        },
        "id": "61556cb58015fb36194d4f2f",
        "provider": {
            "http": {
                "url": "http://localhost:1028/E/"
            },
            "legacyForwarding": true,
            "supportedForwardingMode": "update"
        },
        "status": "active"
    }
]
```

We see `"legacyForwarding": true` in 61556a428015fb36194d4f2e registration, which is not correct.

fgalan:
PR #3946

PR #3946, opened 2021-09-30
Fix legacyForwarding rendering in GET /v2/registrations and GET /v2/registrations/{regId} calls

fgalan:
Isssue #3944


AlvaroVega (approved):
LGTM

PR #3986, opened 2021-11-02
FIX refactor sub up logic avoiding over-quering MongoDB

fgalan:
This PR does a big simplification in the subscription update logic. Previously, the strategy was based in:

1. taking the sub from DB,
2. combine it with cache in a single document
3. update it using "update whole document" variant in MongoDB
4. update the cache based in the calculated doc in item 2

With this PR the strategy changes to

1. update the sub in DB based in $set/$unsest MongoDB operators
2. update cache doc based in the resulting document in MongoDB (findAndModify is used in previous item to achieve this)

The main advantage of this approach is to avoid doing the calculation in former item 2, delegating that calculation to MongoDB itself. This calculation was starting to get very complex when functionality related with failsCount/maxFailsLimit enters into play #3541. The most of the code deleted in this PR are methods used to do that calculation. Another simplification is that we only do an operation on MongoDb (findAndModify) instead of two ones (findOne + update).

In addition this PR does some cleanup in old code.

Note no .ftest was modified by PR, which ensures


mapedraza (approved):
LGTM

PR #3962, opened 2021-10-13
Subscription that pass to inactive state after a limit of failed connection attemps (prelanding)

fgalan:
This PR comes from a prelanding merge at PR https://github.com/telefonicaid/fiware-orion/pull/3890. Issue #3541

It can be checheck that at the moment of creating the PR in commit c2f1438 the diff of changes matches the ones in the original PR: +1,696 -12 

Some additional fixes will be done after analyzing this PR

CC: @Anjali-NEC 

-------

- [x] failsCounter to be supported even without maxFailsLimit?
- [x] check FIXME PR marks
- [x] Review documentation about failsCount. Take into account:
    - Explain the WARN trace

PR #3990, opened 2021-11-04
Subscription that pass to inactive state after a limit of failed connection attemps (fixes to the prelanding branch)

fgalan:
This PR does some fixes in the issue3541-prelanding branch implementation (corresponding to PR #3962). Note that issue3541-prelanding is not the original one done by @Anjali-NEC as it has been updated several times to include changes coming from master.

In some detail this PR includes:

* Subscription and SubscriptionUpdate classes: removing unneeded parameters (maxFailsLimit, failsCounter, maxFailsLimitProvided, failsCounterProvided). Note that actually failsCounter/maxFailsLimit fields belongs to Notification class.
* httpRequestSend(): remove unneeded parameters maxFailsLimit and failsCounter
* Subscription rending logic: fixed in the way status is rendered, so "inactive" takes precedence over "failed".
* Subscription parsing logic:: it was working in an weird way, maxFailsLimit was parsed at root level in the JSON but rendered inside notification object. Now it is parsed and rendered in maxFailsLimit, where it belongs.
* Subscription GET logic: setStatus() modified to take into account status in the cache (which would take preference over the one retrieve from DB)
* Csub cache logic:
    * Remove failsCounter from subCacheItemInsert() and provide the same treatment than for count, setting it to zero
    * status is also part of CachedSubSaved struct (needed for setStatus() mentioned above)
    * Fix code comments to take failsCounter into account (code comment are also important!)
    * Fixes in subNotificationErrorStatus(). **The main changes of the PR are here**. I have clearly separate the logic for the DB and cache cases (previously, DB was being updated in the cache part, which is not correct). `int errors` changed to `bool error` in the signature (notifications are done one by one, so it is not possible errors>1, so better to use a bool)
    * Add WARN trace in subNotificationErrorStatus() when inactivation takes place (it can be useful)
    * mongoSubCountersUpdateCount() correctly dealing with failsCounter reset ($inc cannot be used in that case, $set has to be used) and taking status also into account
* in doNotifyHttp(), appart from changing some function signatures, remove `params` modifications. Note there is no point in modifying `params` object in this function: it is used only as input for this logic, modifying `params` doesn't any permanent effect in cache or DB.
* functional tests
    * New fest cases added (maxfailslimit_error_conditions.test, maxfailslimit_full_lifecycle.test and maxfailslimit_failing_from_the_beginning.test) to cover the functionality in a more comprehensive way
    * Fix existing test cases. The sleeps for "Await sub-cache contents to be dumped into database" has been removed: "tricks" to make test to work in cache case (or in the not cache way) must be avoided, as the same .test should work the same in both cases (with and without cache).
    * Simplify file names in existing .test files. They were unnecessarily long.

In addition, this PR includes some side fixes (not directly related with failsCounter/maxFailsLimit implementation, but I have taken the opportunity to include):

* Improve logic for setting "expired" status in setNotification() within the subscription GET logic
* Typo fixing (ommited -> omitted) along the code/doc base.

Other documentation fixes are not part of this PR. They will be done in a separate PR for issue3541-prelanding branch

@Anjali-NEC as author of the original contribution, it would be great if you could have a look to this PR. It would let you to learn what was needed to finalize the work and enhance your Orion development skills :) Please, provide any feedback you have (questions,  suggestions, etc.) or LGTM if you find it ok for merging into issue3541-prelanding branch.


src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -211,6 +209,13 @@ std::string Notification::toJson(const std::string& attrsFormat)
     jh.addNumber("maxFailsLimit", this->maxFailsLimit);
   }
 
+  // FIXME PR: only if maxFailsLimit is in use failsCounter will be printed
+  // However, maybe we want this failsCounter in any case? It can be an useful statistic...
```
fgalan, line 213: Feedback welcome!

test/functionalTest/cases/3541_subscription_max_fails_limit/subscription_without_maxfailslimit_and_failscounter.test
```diff
@@ -35,8 +35,7 @@ brokerStart CB
 # 03. GET /v2/subscriptions/subscription_id, to check no maxFailsLimit and status=active
 # 04. Stop accumulator
 # 05. PATCH /v2/entities/Room1/attrs, to update the value of pressure attribute.
-# 06. GET /v2/subscriptions/subscription_id, to check no failsCounter and status=active
-# 07. POST /v2/subscriptions, to create subscription without maxFailsLimit=foo and see error
```
fgalan, line 39: Step moved to specific .test dealing with parsing error conditions (maxfailslimit_error_conditions.test)

NTC (informative)

src/lib/apiTypesV2/Subscription.cpp
mapedraza: Maybe is a good idea to print it for debugging the subscription, but in the case you want to display it, a huge value on `maxFailsLimit` can be set.

mapedraza (approved):
LGTM

Anjali-NEC:
LGTM

Issue #4013, opened 2021-12-02
Add subId to "Notification response NOT OK WARN" traces


fgalan:
WARN notification traces are as these ones:

```
time=2021-12-02T10:24:09.323Z | lvl=WARN | corr=fc5fef32-5359-11ec-b318-0a580a830427; cbnotif=2 | trans=1637756597-051-00012506030 | from=10.130.2.1 | srv=sc_vlci | subsrv=/trafico | comp=Orion | op=httpRequestSend.cpp[583]:httpRequestSendWithCurl | msg=Notification response NOT OK, http code: 400

time=2021-12-02T10:22:59.463Z | lvl=WARN | corr=d2b81cd6-5359-11ec-91c0-0a580a810229; cbnotif=1 | trans=1637757684-733-00012469696 | from=10.130.2.1 | srv=sc_vlci | subsrv=/trafico | comp=Orion | op=httpRequestSend.cpp[583]:httpRequestSendWithCurl | msg=Notification response NOT OK, http code: 500
```

They don't include the subscription id, and that could be an useful information. Thus, we could use:

```
Notification (subId: 5f96e27cb14e7532482ac796) response NOT OK, http code: xxx
```

logs.md documentation needs to be changed (not sure if other .md also)

Anjali-NEC:
Hi @fgalan Sir,

I would like to work on this issue. As per my understanding in this issue we need to include subscription id in the WARN notification traces.

For this we need to add `subscription id` in httpRequestSend function in `src/lib/rest/httpRequestSend.cpp`  and use it in trace in Line no 651.
https://github.com/telefonicaid/fiware-orion/blob/master/src/lib/rest/httpRequestSend.cpp#L651 

But I didn't find the way how we can get `subscription id` in httpRequestSend function. Can you please suggest?

fgalan:
I'm already working on it. PR with the fix would published soon.

Anyway, thanks your interest!

fgalan:
PR #4016

fgalan:
Tested for real in master after merging PR #4016

```
[...]

PR #4016, opened 2021-12-03
FIX add subId/regId in "response NOT OK" WARN log traces

fgalan:
Issue https://github.com/telefonicaid/fiware-orion/issues/4013

CC: @Anjali-NEC just in case you'd be interested in how the issue has been solved


mapedraza (approved):
LGTM

PR #4026, opened 2021-12-16
Possibly fixed the issue #3884

kzangeli:
Fixed the <none> issue in the log file. Possibly fixing the real problem at the same time.

I say "possibly", as I was unable to reproduce the crash.
Supposedly the crash was due to the issue that I just fixed, so, I'm gonna need some help to test this, in some env where the crash occurs, using this branch, and see if the crash goes away.
@jason-fox , you are able to reproduce, right?

Also, the entry in CNR is missing. Not sure this fixes the problem yet so ...
Don't think any docs need update for this.

### DIFF
--- a/test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
+++ b/test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
@@ -33,7 +33,8 @@ brokerStart CP1
 # 01. Register entity E1 for CP1 in CB
 # 02. Create E1 in CP1
 # 03. GET /entities/E1 from CB to provoke a forwarded request to CP1, to provoke the bug
-# 04. GET /entities/E1 from CP1 to make sure it-s still alive
+# 04. Extracting a piece of the logfile to make sure the FROM_IP field is A-OK
+# 05. GET /entities/E1 from CP1 to make sure it-s still alive
 #
 
 echo "01. Register entity E1 for CP1 in CB"
@@ -81,7 +82,14 @@ echo
 echo
 
 
-echo "04. GET /entities/E1 from CP1 to make sure it-s still alive"
+echo "04. Extracting a piece of the logfile to make sure the FROM_IP field is A-OK"
+echo "============================================================================"
+cat /tmp/contextBroker.log | grep "Starting forwarding for" | awk -F \| '{ print $5 }'
+echo
+echo
+
+
+echo "05. GET /entities/E1 from CP1 to make sure it-s still alive"
 echo "==========================================================="
 orionCurl --url /v2/entities/E1 --port $CP1_PORT
 echo
@@ -129,7 +137,12 @@ Date: REGEX(.*)
 }
 
 
-04. GET /entities/E1 from CP1 to make sure it-s still alive
+04. Extracting a piece of the logfile to make sure the FROM_IP field is A-OK
+============================================================================
+ from=www.qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq
+
+
+05. GET /entities/E1 from CP1 to make sure it-s still alive
 ===========================================================
 HTTP/1.1 200 OK
 Content-Length: 68
--- a/test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
+++ b/test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
@@ -1,0 +1,155 @@
+# Copyright 2021 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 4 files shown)


kzangeli:
FuncTest error:
```
-----  MQTT subscription multibroker  -----
(3001_mqtt/mqtt_subscription_multibroker.test) output not as expected
VALIDATION ERROR: input line:
   MQTT message at topic /sub2:
does not match ref line:
   MQTT message at topic /sub1:
```
Does this really have anything to do with this PR ?

fgalan:
> FuncTest error:
> 
> ```
> -----  MQTT subscription multibroker  -----
> (3001_mqtt/mqtt_subscription_multibroker.test) output not as expected
> VALIDATION ERROR: input line:
>    MQTT message at topic /sub2:
> does not match ref line:
>    MQTT message at topic /sub1:
> ```
> 
> Does this really have anything to do with this PR ?

That's a glitch. I'll relaunch.

test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
```diff
@@ -0,0 +1,155 @@
...
+dbInit CB
+brokerStart CB 0-1 IPV4
+brokerStart CP1
+
+--SHELL--
+#
+# 01. Register entity E1 for CP1 in CB
+# 02. Create E1 in CP1
+# 03. GET /entities/E1 from CB to provoke a forwarded request to CP1, to provoke the bug
+# 04. GET /entities/E1 from CP1 to make sure it-s still alive
```
fgalan, line 36: Maybe a 05 step could be added to get the CB logs and check that from= is "sane" (i.e. without the problems described at https://github.com/telefonicaid/fiware-orion/issues/3884#issuecomment-862642537)

fgalan:
> Just to clarify with @jason-fox:
> 
> The test that @kzangeli implemented in order to replicate #3884 does the following:
> 
>     1. Starts 2 brokers
> 
>     2. Register a entity E1 for CP1 in CB
> 
>     3. Creates E1 in CP1
> 
>     4. GET /entities/E1 from CB to provoke a forwarded request to CP1, to provoke the bug ( having `X-Forwarded-For` header a long value )
> 
>     5. GET /entities/E1 from CP1 to make sure it-s still alive
> 
> 
> This sequence is the same that the situation that was creating the broker crash?

As suggestion, maybe @jason-fox could build a docker container based in kzangeli:issue/3884 and repeat the test described in issue #3884 on it.

fgalan:
>  Don't think any docs need update for this.

Agree

test/functionalTest/cases/3884_ip_len_in_logs/ip_len_in_logs.test
kzangeli: Can easily be done, but ... having a functest depending on the output in the logfile is not the best of ideas.
I'll have to pick a specific logline (grep it from the logfile), where we know the FROM_IP has been set, and then compare.
It that logline is removed or changed, then the functest will stop working.
We've had plenty of problems with that in the past.

Are you sure you want this?

fgalan: Using grep and awk we can be very precise and get only the piece of log we are interested in, i.e. the line with the from field due to the "X-Forwarded-For" we are using, looking only for the from= field in that line. We have some examples of that in the .test (looking for "awk" should show some cases).

However, it is not a must for this PR. I can work on it after merging this PR (I have some recent experience with that).

kzangeli: I'm aware. Kind of invented it.
What I mean is, if the line disappears or is modified, the whole thing starts failing.
In my opinion it's a bad idea.
But, I can see you want it, so I'll add it

kzangeli: Fixed in b4857fd

kzangeli:
> That's a glitch. I'll relaunch.

You have 3 attempts on still, right?
It it sometimes fail with *three* attempts ... then it's more than a glitch, IMHO.
Seems like an unstable test that needs to be looked over.

fgalan:
> > That's a glitch. I'll relaunch.
> 
> You have 3 attempts on still, right? It it sometimes fail with _three_ attempts ... then it's more than a glitch, IMHO. Seems like an unstable test that needs to be looked over.

It is a test in which the order in which MQTT notifications triggered at the same time can lead to a result or to the other. It can be probably solved using SORT_START / SORT_END, but I haven't put an eye on it yet. I think I'll do when it gets so annoying that I  I cannot bear it no more :)

NTC in the context of this PR

kzangeli:
Yes, SORT_START/END should be of great help for a delicate problem like that.
Good luck :)

fgalan:
From my point of view PR is ok from the point of view of implementation and ftest.

Let's wait for @jason-fox test before adding CNR entry and merge.

kzangeli:
> Let's wait for @jason-fox test before adding CNR entry

Exactly, cause ... I don't know whether I just fixed the anomaly in the log file or if I also fixed the crash ...

mapedraza:
@jason-fox, did you try if this PR fix the Issue (#3884)? It is pending to check if it solves the problem before merging it.

fgalan:
Continues in PR https://github.com/telefonicaid/fiware-orion/pull/4039 for its definitive merge to master.

PR #4027, opened 2021-12-16
fixes for issue 4012

kzangeli:
Modified the httpsPrepare script to accept the size of the cert file as a parameter AND added a functest with a cert size of 2350


mapedraza (approved):
LGTM

PR #4035, opened 2022-01-08
Port/3570 accumulator

jfernandz:
I've ported the `accumulator-server.py` script to `python3`. Testing it I've also noticed the `docker-compose.yml` snippet inside the `REDAME.md` was incomplete, so I've expended and fixed it.

I've got up two docker containers with `mongo` and `orion`, created an entity, and a subscription for the `accumulator-server` and apparently the accumulator is being notified about attribute changes. (I've basically followed this walkthrough)

So I'd say this python3 port should work in the same way than the old python2 version.

Enjoy!


docker/README.md
```diff
@@ -26,16 +26,20 @@ Follow these steps:
...
+		    image: fiware/orion
+		    ports:
+		      - "1026:1026"
+		    depends_on:
+		      - mongo
+		    command: -dbhost mongo
+		
+		  mongo:
+		    image: mongo:4.4
+		    command: --nojournal
```
fgalan, line 42: I understand this matches now docker/docker-compose.yml. Great!

scripts/accumulator-server3.py
```diff
@@ -0,0 +1,592 @@
+#!/usr/bin/env python3
```
fgalan, line 1: I'd suggest to keep only one version of the script. Could you rename it to scripts/accumulator-server.py replacing the old one, please?

docker/README.md
jfernandz: Sure, I've noticed that snippet was uncomplete when following the walkthrough.

scripts/accumulator-server3.py
jfernandz: I think now it's done :wink:

fgalan:
Thanks for the contribution!

Some pieces of documentation would need to be adapted to change references from Python2 to Python3:

* https://github.com/telefonicaid/fiware-orion/blob/master/test/functionalTest/README.md#how-to-install-accumulator-serverpy
* https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/user/walkthrough_apiv2.md#starting-accumulator-server
* https://github.com/telefonicaid/fiware-orion/blob/3cd9de2f790e8be8e3bdad7636af4ac2a0c46228/doc/manuals/admin/build_source.md#testing-and-coverage

In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.

fgalan:
> In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.

If you are not familiar with that part I could try to help.

docker/README.md
fgalan: NTC

scripts/accumulator-server.py
```diff
@@ -46,7 +46,7 @@
 # * This script also depends on pyOpenSSL 19.0.0
 
 
-from OpenSSL import SSL
+# from OpenSSL import SSL
```
fgalan, line 49: This should be commented. Some .test run accumulator using HTTPS if I remember correctly.

scripts/accumulator-server.py
```diff
@@ -81,42 +82,57 @@ def usage():
...
-verbose    = 0
-pretty     = False
-https      = False
-key_file   = None
-cert_file  = None
+verbose = 0
+pretty = False
+https = False
+key_file = None
+cert_file = None
```
fgalan, line 119: I'd suggest to keep the same indentation in that part.

scripts/accumulator-server.py
```diff
@@ -373,46 +398,76 @@ def record_2871():
...
+
 @app.route('/dump', methods=['GET'])
 def dump():
     return ac
 
 
 @app.route('/times', methods=['GET'])
-def times():
-    return ', '.join(map(str,times)) + '\n'
+def times_():
```
fgalan, line 469: `times` is a new reserved word in Python3?

scripts/accumulator-server.py
```diff
@@ -266,6 +290,7 @@ def record_request(request):
 
     global ac, t0, times
     s = ''
+    times = []
```
fgalan, line 293: Maybe this change is related with the one in L495 and L553? Could you elaborate a little bit on it, please?

scripts/accumulator-server.py
jfernandz: No, but `times` as variable is frequently used along the file, so python3 interpreter has some conflicts when some function name collides with a variable name. Since this function is just used to implement the endpoint and it's not ever called ... I decided just to rename the function :thinking:

fgalan:
```suggestion
def get_times():
```

I think that's a better name :)

scripts/accumulator-server.py
jfernandz: I did this just to keep compliant with [PEP8](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements)

But if you prefer to keep the indentation I'll add it again :wink:

jfernandz:
lemme check all of your suggestions this afternoon :wink: I'm right now at work and I thought this would be faster. But I appreciat very much them, I think they will help me to understand better the whole project :blush:

scripts/accumulator-server.py
fgalan: No, it's ok. Align with PEP8 is better.

NTC

scripts/accumulator-server.py
jfernandz: What about:

```suggestion
def giveTimes():
```
?

fgalan: It seems we have a crazy mess of [snake_case](https://en.wikipedia.org/wiki/Snake_case) (dict_raise_on_duplicates, record_request, ... ) and [camelCase](https://en.wikipedia.org/wiki/Camel_case) (waitForever, givemeDelay, ...) names for functions in the code :) so both options are ok for me. Whichever you prefer.

scripts/accumulator-server.py
jfernandz: Well, the function `record_request` appends data to a list called `times`, so the list must be declared locally inside the function.

scripts/accumulator-server.py
jfernandz: I commented out this line because the SSL interface is just used at line [586](https://github.com/telefonicaid/fiware-orion/pull/4035/files/c2ea3115b62363bf2cc8f3f39a81c99b6763e8a3#diff-eaa38239c9ce2beb0e36a37f9327707aea5fb10b1aa2d385a2c8eb94f86ca40aR586):

```
# context = SSL.Context(SSL.SSLv23_METHOD)
```

But as it's also commented out ... my pyflakes threw me a warning. 

But you are right, I'm not sure if ssl tests will pass with one of these two lienes commented out ... because after all, the SSL interface is not even used if line 586 is commented out.

scripts/accumulator-server.py
jfernandz: I think I prefer the camelCase for this ... **time** :rofl:

scripts/accumulator-server.py
fgalan: I have had a closer look to this and it seems due to an old version of the SSL stuff no longer used. In fact, this commented line and the commented block from L575 to L588 could be removed. You can do it in this PR or I can do it in a later one, as you prefer....

scripts/accumulator-server.py
fgalan: Currently times is used at:

* record_request() and on_message() to append data
* times_() to return its content
* number() to return its len
* reset() to reset it

So, it is actually a global variable shared by these functions. I see that `global` is used in record_request(), on_message() and reset(). Maybe it should be added to times_() and number()?

Apart from that , the commented lines at L495 and L553 should be removed, to avoid leaving leftover comments in the code.

Moreover, I think ac and t0 setting at L551-L552 can be removed, as this is already done at L493-L494. What do you think?

jfernandz:
> Thanks for the contribution!
> 
> Some pieces of documentation would need to be adapted to change references from Python2 to Python3:
> 
> * https://github.com/telefonicaid/fiware-orion/blob/master/test/functionalTest/README.md#how-to-install-accumulator-serverpy
> * https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/user/walkthrough_apiv2.md#starting-accumulator-server
> * https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/build_source.md#testing-and-coverage
> 
> In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.

I'm wondering ... should I reference this comment in a new issue to give treatment in a separete branch/pull request? 

Along with this other one point:

> > In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.
> 
> If you are not familiar with that part I could try to help.

jfernandz:
> As additional comment, note in L29:
> 
> ```
> from __future__ import division   # need for seconds calculation (could be removed with Python 2.7)
> ```
> 
> I think this is not longer needed in Python3. Have a look to https://programmerclick.com/article/30707824/

Sure, `python3` doesn't need this import from `__future__`, integer division (`//`) is natively supported in `python3`

fgalan:
> > Thanks for the contribution!
> > Some pieces of documentation would need to be adapted to change references from Python2 to Python3:
> > 
> > * https://github.com/telefonicaid/fiware-orion/blob/master/test/functionalTest/README.md#how-to-install-accumulator-serverpy
> > * https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/user/walkthrough_apiv2.md#starting-accumulator-server
> > * https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/build_source.md#testing-and-coverage
> > 
> > In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.
> 
> I'm wondering ... should I reference this comment in a new issue to give treatment in a separete branch/pull request?
> 
> Along with this other one point:
> 
> > > In addition, the parts related with accumulator-server.py in https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh (the script used to build the Docker imagen for CI regression) should be adjusted.
> > 
> > 
> > If you are not familiar with that part I could try to help.

It depends how far you want to contribute to this repo :)

Ideally, it would be great if you could do at leaset the documentation modifications in this PR also. No need to bother with new PRs/branches.

With regards to Docker stuff, that probably requires a deeper level of knowledge in the project so it could be hard for you. But of course if you want to cope them, no problem for us :)

Thanks you very much for your contributions!

scripts/accumulator-server.py
jfernandz: what about the tuple `ssl_context`? I'd say we could also omit the `ssl_context`  argument for flask.

scripts/accumulator-server.py
jfernandz: Sure, I didn't noticed that either, the point is you need to use `global` locally when you need to perform operations on some variable, but it's not apparently neccesary to include the `global` statement when you are just reading them.

scripts/accumulator-server.py
fgalan: > what about the tuple `context` at line 589? I'd say we could also omit the `ssl_context` argument for flask.
> 
> I mean, if we remove these lines ... we could drop the whole https support for the script.

Note that https support cannot be drop, as it is needed in some cases, pe. https://github.com/telefonicaid/fiware-orion/blob/master/test/functionalTest/cases/0706_direct_https_notifications/direct_https_notifications.test

Thus only the old implementation (in the commented lines) should be removed.

doc/manuals/admin/build_source.md
```diff
@@ -116,17 +116,15 @@ In the case of the aarch64 architecture, install libxslt using yum, and run `./c
...
-In the case of the aarch64 architecture, additionally install python2-devel, rpm-build and libffi-devel using yum. It is needed when building pyOpenSSL.
-
 * Prepare the environment for test harness. Basically, you have to install the `accumulator-server.py` script and in a path under your control, `~/bin` is the recommended one. Alternatively, you can install them in a system directory such as `/usr/bin` but it could collide with an RPM installation, thus it is not recommended. In addition, you have to set several environment variables used by the harness script (see `scripts/testEnv.sh` file) and create a virtualenv environment with the required Python packages.
 
         mkdir ~/bin
         export PATH=~/bin:$PATH
         make install_scripts INSTALL_DIR=~
         . scripts/testEnv.sh
-        virtualenv /opt/ft_env --python=/usr/bin/python2
+        virtualenv /opt/ft_env
```
fgalan, line 125:
```suggestion
        virtualenv /opt/ft_env --python=/usr/bin/python3
```

To make explicit the need for Python 3

doc/manuals/admin/build_source.md
```diff
@@ -116,17 +116,15 @@ In the case of the aarch64 architecture, install libxslt using yum, and run `./c
...
         . /opt/ft_env/bin/activate
-        pip install Flask==1.0.2 pyOpenSSL==19.0.0 paho-mqtt==1.5.1
+        pip install Flask paho-mqtt
```
fgalan, line 127: I'd suggest to keep explicit versions in the pip install command. This makes installations much more controlled (otherwise a new version may appear of Flask that is not fully compatible with accumulator-server.py).

Why pyOpenSSL has been removed? I understand it's still needed as far as http support is required in accumulator-server.py

doc/manuals/user/walkthrough_apiv2.md
```diff
@@ -121,12 +121,12 @@ command:
 ```
 # cd /dir/where/accumulator-server/is/downloaded
 # chmod a+x accumulator-server.py
-# ./accumulator-server.py --port 1028 --url /accumulate --host ::1 --pretty-print -v
+# ./accumulator-server.py --pretty-print -v
```
fgalan, line 124: Why this has been changed? Under my understanding the only changes we need in this file are the ones about Python 2.7. I mean to changes:

_In addition, it is recommended to use Python 2.7_

by something like this:

_In addition, it is recommended to use Python 3_

scripts/accumulator-server.py
```diff
@@ -81,42 +81,57 @@ def usage():
...
+          --host <host>: host to use database to use (default is '0.0.0.0')"
+          --port <port>: port to use (default is 1028)"
+          --mqttHost <host>: mqtt broker host to use (if not defined MQTT listening is not enabled)"
+          --mqttPort <port>: mqtt broker port to use (default is 1883)"
+          --mqttTopic <topic>: mqtt topic to use (default is accumulate_mqtt)"
+          --url <server url>: server URL to use (default is /accumulate)"
+          --pretty-print: pretty print mode"
+          --https: start in https"
+          --key: key file (only used if https is enabled)"
+          --cert: cert file (only used if https is enabled)"
```
fgalan, line 95:
```suggestion
          --key <key_file>: key file (only used if https is enabled)"
          --cert <cert_file>: cert file (only used if https is enabled)"
```

This is not actually related with the conversion to Python 3 (i.e. it was already wrong in the original script) but we can take this opportunity to fix it :)

scripts/accumulator-server.py
```diff
@@ -81,42 +81,57 @@ def usage():
...
-    print "  --mqttPort <port>: mqtt broker port to use (default is 1883)"
-    print "  --mqttTopic <topic>: mqtt topic to use (default is accumulate_mqtt)"
-    print "  --url <server url>: server URL to use (default is /accumulate)"
-    print "  --pretty-print: pretty print mode"
-    print "  --https: start in https"
-    print "  --key: key file (only used if https is enabled)"
-    print "  --cert: cert file (only used if https is enabled)"
-    print "  -v: verbose mode"
-    print "  -u: print this usage message"
+    print(f"Usage: {os.path.basename(__file__)} --host <host> --port <port> --url <server url> --pretty-print -v -u")
```
fgalan, line 84: The Usage line misses a lot of options. It should be something like this:

```suggestion
    print(f"Usage: {os.path.basename(__file__)} --host <host> --port <port> --mqttHost <host> --mqttPort <port> --mqttTopic <topic> --url <server url> --pretty-print --https --key <key_file> --cert <cert_file> -v -u")
```

This is not actually related with the conversion to Python 3 (i.e. it was already wrong in the original script) but we can take this opportunity to fix it :)

test/functionalTest/README.md
```diff
@@ -12,7 +12,7 @@ We recommend to use Python [virtualenv](https://virtualenv.pypa.io/en/latest) to
 
 ```
 pip install virtualenv  # if you don't have virtualenv itself previously installed
-virtualenv --python=/usr/bin/python2 /path/to/ft_env
+virtualenv /path/to/ft_env
```
fgalan, line 15:
```suggestion
virtualenv --python=/usr/bin/python3 /path/to/ft_env
```

To make explicit the need for Python 3

test/functionalTest/README.md
```diff
@@ -24,9 +24,8 @@ Then activate the virtual env:
 Next install accumulator-server.py depencencies:
 
 ```
-pip install Flask==1.0.2
-pip install pyOpenSSL==19.0.0
-pip install paho-mqtt==1.5.1
+pip install Flask
+pip install paho-mqtt
```
fgalan, line 28: I'd suggest to keep explicit versions in the pip install command. This makes installations much more controlled (otherwise a new version may appear of Flask that is not fully compatible with accumulator-server.py).

Why pyOpenSSL has been removed? I understand it's still needed as far as http support is required in accumulator-server.py

scripts/accumulator-server3.py
fgalan: Fixed in 39b55b6fedb02f784a4f64208e22097646f985d7

scripts/accumulator-server.py
jfernandz: well, apparently the code as its current state should work (according to [pyopenssl](https://www.pyopenssl.org/en/latest/api/ssl.html) docs). So probably uncommenting the lines will be enough.

scripts/accumulator-server.py
fgalan: Fixed in 9fa9d1a453cdd84d9b9fbcde1f73839cafca0225

scripts/accumulator-server.py
fgalan: > So probably uncommenting the lines will be enough.

Actually, removing the currently commented lines.

doc/manuals/admin/build_source.md
jfernandz: Yes, that was my fault, I misunderstand you and I thought you were talking about removing the whole https support. I'll fix this in the next commit :wink:

fgalan: Sorry for the missunderstanding!

doc/manuals/user/walkthrough_apiv2.md
jfernandz: Well, these are the default values for `port`, `url` and `host` variables, so I considered it wasn't needed to explicitly specify these default values as command line arguments. However ... we can preserve this string if you want.

scripts/accumulator-server.py
jfernandz: We could replace the used SSL context with a most recent one like `OpenSSL.SSL.TLSv1_2_METHOD`

When you say ...  'removing the currently commented lines' ... you mean to uncomment them? Or just delete them all?

doc/manuals/user/walkthrough_apiv2.md
fgalan: > Well, these are the default values for `port`, `url` and `host` variables, so I considered it wasn't needed to explicitly specify these default values as command line arguments. However ... we can preserve this string if you want.

I think they are useful to show the reader of the walk-through they are configurable so in case in its environment some of the default is not appropriated (e.g. port 1028 is being used by some existing service) he can know it can be changed.

scripts/accumulator-server.py
fgalan: I refer to deleting this

```
 # from OpenSSL import SSL
```

and this

```
        """
        Commented lines correspond to the way of using context with pyOpenSSL 0.13.1
        (the one that comes with system Python in CentOS 7).
        We are using the new way (which is simpler) as we moved to pyOpenSSL 19.0.0.
        Referecence: Reference:
        http://stackoverflow.com/questions/28579142/attributeerror-context-object-has-no-attribute-wrap-socket/28590266
        We need to upgrade pyOpenSSL version due to problems of installing 0.13.1 inside virtualenv.
        Installing the module requires to compile some parts and this causes a conflict with base
        openssl devel libraries in CentOS 7 (solvable by hack, but we want to avoid it)
        """
        # context = SSL.Context(SSL.SSLv23_METHOD)
        # context.use_privatekey_file(key_file)
        # context.use_certificate_file(cert_file)
```

understanding that with this is enough to have HTTPS support in the script:

```
        context = (cert_file, key_file)
        app.run(host=host, port=port, debug=False, ssl_context=context)
```

However, it there are better alternatives to implement HTTPS support (the one you mention about `OpenSSL.SSL.TLSv1_2_METHOD`) they can be done too. However, maybe it's better in a next PR in order not introduce too much complexity in this one.

jfernandz: Oh, I'm sorry, I'm not familiar enough with TLS layers ... but ... I'm not sure if Flask will accept plain cert and privkey files without setting them as a part of a proper context :thinking: 

Apparently we could use for this case the an _ad hoc_ certificate (https://blog.miguelgrinberg.com/post/running-your-flask-application-over-https -> __The simplest way to do it__ section)

fgalan: I would say that the current implementation of

```
        context = (cert_file, key_file)
        app.run(host=host, port=port, debug=False, ssl_context=context)
```

Corresponds with the case in the section "Self-Signed Certificates" in the document you cite.

Thus, it seems that keeping that would be enough.

doc/manuals/admin/build_source.md
```diff
@@ -116,17 +116,15 @@ In the case of the aarch64 architecture, install libxslt using yum, and run `./c
         sudo yum install curl nc valgrind bc
         sudo pip2 install virtualenv
 
-In the case of the aarch64 architecture, additionally install python2-devel, rpm-build and libffi-devel using yum. It is needed when building pyOpenSSL.
```
fgalan, line 119: No need of extra tooling to installing pyOpenSSLin Python3 in aarch64 architecture?

(Just to be sure to remove this paragraph in that case)

scripts/accumulator-server.py
```diff
@@ -40,13 +39,11 @@
...
 #   "Content-Type: application/x-www-form-urlencoded" has been problematic in the pass)
-# * This script requires at least Flask 1.0.2, which comes with Werkzeug 0.15.2. There is a bug
-#   in Werkzeug < 0.11.16 that makes empty "content-length" headers to appear for some request
-#   in the accumulator dump
-# * This script also depends on pyOpenSSL 19.0.0
-
+# * This script requires at least Flask 2.0.2, which comes with Werkzeug 2.0.2.
+#
+# TODO: test a possible bug in Werkzeug that makes empty "content-length"
+# headers to appear for some request in the accumulator dump
```
fgalan, line 45: Werkzeug 2.0.2 > 0.11.16, so I think the bug no longer exists and this two lines could be removed

scripts/accumulator-server.py
```diff
@@ -81,44 +78,62 @@ def usage():
...
+            'url=',
+            'pretty-print',
+            'https',
+            'key=',
+            'cert='
+        ]
+    )
 except GetoptError:
-    usage_and_exit('wrong parameter')
+    usage_and_exit('Wrong parameter')
```
fgalan, line 136:
```suggestion
    usage_and_exit('wrong parameter')
```

Looking to other error messages we generated, e.g (`print("if --https is used then you have to provide --key and --cert")`) it seems we start them with lowercase.

scripts/accumulator-server.py
```diff
@@ -513,16 +570,8 @@ def on_message(client, userdata, msg):
...
-      #
-      # We need to upgrade pyOpenSSL version due to problems of installing 0.13.1 inside virtualenv. Installing the module
-      # requires to compile some parts and this causes a conflict with base openssl devel libraries in CentOS 7 (solvable by hack, but we want to avoid it)
-      
-      #context = SSL.Context(SSL.SSLv23_METHOD)
-      #context.use_privatekey_file(key_file)
-      #context.use_certificate_file(cert_file)
-      context = (cert_file, key_file)
-      app.run(host=host, port=port, debug=False, ssl_context=context)
+        # TODO: Improve SSL support by using pyOpenSSL
```
fgalan, line 573: In which sense pyOpenSSL would provide a better support that the existing?

scripts/accumulator-server.py
```diff
@@ -81,44 +78,62 @@ def usage():
...
-https      = False
-key_file   = None
-cert_file  = None
+verbose = 0
+pretty = False
+https = False
+key_file = None
+cert_file = None
 
+# TODO: Improve cli argument parsing with click
```
fgalan, line 119: Do you have some reference to click?

jfernandz:
@fgalan for some reason I cannot comment in your suggestions, maybe we need to resolve the suggestion I did about dropping the `--https` arg. :thinking:

fgalan:
> @fgalan for some reason I cannot comment in your suggestions, maybe we need to resolve the suggestion I did about dropping the `--https` arg. 🤔

As [explained in previous comment](https://github.com/telefonicaid/fiware-orion/pull/4035#discussion_r785714370) we cannot drop `--https` as some regression tests needs that functionality in accumulator-server.py.

Which kind of problem do you have when trying to comment on my suggestions?

jfernandz:
> > @fgalan for some reason I cannot comment in your suggestions, maybe we need to resolve the suggestion I did about dropping the `--https` arg. thinking
> 
> As [explained in previous comment](https://github.com/telefonicaid/fiware-orion/pull/4035#discussion_r785714370) we cannot drop `--https` as some regression tests needs that functionality in accumulator-server.py.
> 
> Which kind of problem do you have when trying to comment on my suggestions?

I've not text box even :thinking:

fgalan:
> I've not text box even 🤔

Weird... you are in this URl https://github.com/telefonicaid/fiware-orion/pull/4035? Maybe you could you provide a screenshot in the case I can spot the cause?

jfernandz:
> Weird... you are in this URL `https://github.com/telefonicaid/fiware-orion/pull/4035`? Maybe you could provide a screenshot so I can try to spot the cause?

Sure. 
image

But I think you should reject the suggestion at least :thinking: 
image

fgalan:
Not sure if this will solve anything :) but I have observed "Pending" in some of your comments:

imagen

So maybe you need to confirm your review so the comment would be published. Maybe that "unblocks" the possibility to comments in other places...

Another think you can tray is to re-login in github.

jfernandz:
> So maybe you need to confirm your review so the comment would be published. Maybe that "unblocks" the possibility to comments in other places...
> 
> Another think you can tray is to re-login in github.

I've got the two buttons locked, so I guess it has to be the reviewer who accepts/rejects suggestions :wink:

fgalan:
What about the upper right corner? Do you have something like this?

imagen

Upon clicking it, use the "Submit review" buttom.

imagen

scripts/accumulator-server.py
```diff
@@ -81,44 +78,62 @@ def usage():
...
+          --host <host>: host to use database to use (default is '0.0.0.0')"
+          --port <port>: port to use (default is 1028)"
+          --mqttHost <host>: mqtt broker host to use (if not defined MQTT listening is not enabled)"
+          --mqttPort <port>: mqtt broker port to use (default is 1883)"
+          --mqttTopic <topic>: mqtt topic to use (default is accumulate_mqtt)"
+          --url <server url>: server URL to use (default is /accumulate)"
+          --pretty-print: pretty print mode"
+          --https: start in https"
+          --key: <key file> (only used if https is enabled)"
+          --cert: <cert file> (only used if https is enabled)"
```
fgalan, line 94:
```suggestion
          --key <key file>: key file (only used if https is enabled)"
          --cert <cert file>: cert file (only used if https is enabled)"
```

fgalan:
No matter the problems with per-line comments you are having, we can use the general comment to continue discussion :)

To summary, the current open items in the PR are:

1. Doubt in doc/manuals/admin/build_source.md about if pyOpenSSL dependency is needed or not (it seems it no longer needec, but explicit confirmation would be welcome)
2. in scripts/accumulator-server.py remove the TODO lines about Wekzeug bug in ancient versions
3. Reference to click module (I'm just curious, first time I heard about it :)
4. Little typo "Wrong" -> "wrong" in usage_and_exit() message
5. Doubt about why pyOpenSSL would provide a better support that the existing one for SSL (from my point of view, with the current we have suffices for basic testing and regression suite)
6. Small typo in `--key` and `--cert` usage message.

There are pretty "fine-grained" so if you don't want to continue with these details, I think I can manage to do it in a next PR myself (including in that PR the adaptation in CI Dockerfile also). I mean, to merge this PR as it is now so you can continue working on it to solve them.

What do you think?

jfernandz:
Excuse me, this past week was so hard :laughing: I hope in a few days I can retake this :wink:

jfernandz:
You were right, btw, I had to cancel my own review to be able to commant on your suggestions :man_facepalming:

fgalan:
> Excuse me, this past week was so hard 😆 I hope in a few days I can retake this 😉

Ok. No rush! I was just offering to help :)

jfernandz:
> No matter the problems with per-line comments you are having, we can use the general comment to continue discussion :)
> 
> To summary, the current open items in the PR are:
> 
> 1. Doubt in doc/manuals/admin/build_source.md about if pyOpenSSL dependency is needed or not (it seems it no longer needec, but explicit confirmation would be welcome)
> 2. in scripts/accumulator-server.py remove the TODO lines about Wekzeug bug in ancient versions
> 3. Reference to click module (I'm just curious, first time I heard about it :)
> 4. Little typo "Wrong" -> "wrong" in usage_and_exit() message
> 5. Doubt about why pyOpenSSL would provide a better support that the existing one for SSL (from my point of view, with the current we have suffices for basic testing and regression suite)
> 6. Small typo in `--key` and `--cert` usage message.
> 
> There are pretty "fine-grained" so if you don't want to continue with these details, I think I can manage to do it in a next PR myself (including in that PR the adaptation in CI Dockerfile also). I mean, to merge this PR as it is now so ~you~ I can continue working on it to solve them.
> 
> What do you think?

I think all these things are mainly fixed in the last commit, can you review it @fgalan? :blush:

fgalan (approved):
After the last commit, all the pending comments has been addressed so now the PR is in LGMT status :)

@jfernandz thanks for your contribution and you willingness to address all the comments raised during the review process!

PD. I didn't know click. It seems to be much better that the getopt library I'm used to for CLI parsing. I think I'll start to using it in my next projects :)

fgalan:
> Test are not passing in the PR, but this was expected. The Dockerfile used for CI needs some adaption. I'll merge this PR a prelanding branch (port/3570_accumulator-prelanding) and continue the work there.

PR #4059

PR #4031, opened 2021-12-22
ADD tests to cover x-auth-token, fiware-service and fiware-servicepath propagation cases with custom notifications and subscriptions with filters

fgalan:
- [x] fiware-service cases
- [x] fiware-servicepath cases
- [x] docu (if applicable)

PR #4037, opened 2022-01-13
(JP) Add doc about header removal in custom notifications (#4031)

fisuda:
This PR adds Japanese documentation about header removal in custom notifications.


fgalan (approved):
LGTM. Thanks!

PR #4047, opened 2022-01-28
FIX avoid "-release" in autogenerated release RPMs

fgalan:
CC: @wistefan 

I think this will be problem we have seen with names like contextBroker-3.5.0-release.x86_64.rpm. Let's check when 3.6.0 gets released ;)


wistefan (approved):
LGTM

mapedraza (approved):
LGTM

Issue #1256, opened 2015-09-21
Return parse error if not needed fields are found


fgalan:
For example, in the case of

```
PUT /v2/entities/e1/attrs/a

{
  "value": "foo",
  "type": "bar",
  "should_not_be_here": "xxx"
}
```

The existence of `should_not_be_here` should raise a parse error.

fgalan:
This issue is not especifically related with NGISv2, so APIv2 label is removed. The issue keeps opened.

CC: @jmcanterafonseca

Anjali-NEC:
Fixed issue in PR#4049

kzangeli:
Wait ...
Parse Error?
No, an unexpected item in a correctly parsed JSON payioad should **not ever** give a parse error!
Parse Errors should give parse errors, nothing else.

Quite straightforward, don't you think? ;-) 

A "strange" item in a payload should complain about **unexpected item** or something similar

Anjali-NEC:
> Wait ... Parse Error? No, an unexpected item in a correctly parsed JSON payioad should **not ever** give a parse error! Parse Errors should give parse errors, nothing else.
> 
> Quite straightforward, don't you think? ;-)
> 
> A "strange" item in a payload should complain about **unexpected item** or something similar

I think you are correct.  I have fixed this issue according to the comment in https://github.com/telefonicaid/fiware-orion/issues/1256#issue-107490829 But it shouldn't raise parse error so I reverted my changes.

Thanks

kzangeli:
Ok! Good!
Not your fault.
Fermin's initial comment about "The existence of should_not_be_here should raise a parse error" is misleading.

kzangeli:
Old, old issue! (2015) ... :)
At that time, we still didn't know very well what we were doing, Fermin and I  :)

fgalan:
I have had a look to #1256. It seems you are covering only a very specific case (in the attribute parsing logic) but this issue is more general. Is about detecting unknown fields in **every possible** JSON object in **every possible** JSON request.

If a general solution is not possible, you should add a similar code in too many places. And .test cases to covering all them. It seems to be be a lot of effort taking into account the usefulness of this issue (not too high). Maybe it's better to put effort on another issues.

kzangeli:
Last thought on 
```
"should_not_be_here": "xxx"
```

That would work just fine is key-values is turned on.
And if not, the value of "should_not_be_here" would need to be a JSON object, which it isn't.

[...]

PR #4049, opened 2022-01-28
Fix Return parse error if not needed fields are found #1256

Anjali-NEC:
Fix issue #1256

### DIFF
--- a/src/lib/jsonParseV2/parseContextAttribute.cpp
+++ b/src/lib/jsonParseV2/parseContextAttribute.cpp
@@ -153,8 +153,8 @@ static std::string parseContextAttributeObject
     }
     else  // ERROR
     {
-      LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
-      return "unrecognized property for context attribute";
+      alarmMgr.badInput(clientIp, "JSON parse error");
+      return "JSON parse error in context attribute";
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/jsonParseV2/parseContextAttribute.cpp
```diff
@@ -153,8 +153,8 @@ static std::string parseContextAttributeObject
     }
     else  // ERROR
     {
-      LM_W(("Bad Input (unrecognized property for ContextAttribute - '%s')", name.c_str()));
-      return "unrecognized property for context attribute";
+      alarmMgr.badInput(clientIp, "JSON parse error");
+      return "JSON parse error in context attribute";
```
kzangeli, line 157: This is not a "Parse Error" !
The current error message is quite good, and should not be changed.
Add a call to the alarm manager?  Sure, why not?
But, let's not lie about what went wrong!
This is NOT a parse error, but an unexpected (and not welcome) item

Anjali-NEC: Yes, you are right! I have reverted my changes in https://github.com/telefonicaid/fiware-orion/pull/4049/commits/5acdc12ec5d93282bd71423d4572167778759baa and added alarm for badinput.

fgalan:
See comment at https://github.com/telefonicaid/fiware-orion/issues/1256#issuecomment-1025488343

fgalan:
The parent issue of this PR (https://github.com/telefonicaid/fiware-orion/issues/1256) has been closed, so the PR will be also closed.

I understand this is not a problem as you have only covered one of the multiple parsing cases which need to be implemented :)

Thanks for your contributions!

PR #4050, opened 2022-01-31
FIX different versions on RPM nightly repo

mapedraza:
Added a sed step to CI file stripping the version number in the filename (`X.Y.Z_next`) in case of uploading a nightly build.

This would fix uploading different version names depending of the release version to [here](https://nexus.lab.fiware.org/service/rest/repository/browse/el/8/x86_64/nightly/)

### DIFF
--- a/.github/build-and-push-rpm.sh
+++ b/.github/build-and-push-rpm.sh
@@ -17,7 +17,7 @@ do
   filename=$(basename $file)
   if [ "$type" == "nightly" ]
   then
-      filename=$(echo $filename | sed 's/[0-9]*.[0-9]*.[0-9]*_next-//g')
+      filename=$(echo $filename | sed 's/[0-9]+.[0-9]+.[0-9]+_next-//g')
   fi
   echo "Uploading $filename"
   curl -v -f -u telefonica-github:$secret --upload-file $file https://nexus.lab.fiware.org/repository/el/8/x86_64/$type/$filename
--- a/.github/build-and-push-rpm.sh
+++ b/.github/build-and-push-rpm.sh
@@ -15,6 +15,10 @@ docker run -e BROKER_RELEASE=$type -v $(pwd):/opt/orion --workdir=/opt/orion fiw
 for file in "$(pwd)/rpm/RPMS/x86_64"/*
 do
   filename=$(basename $file)
+  if [ "$type" == "nightly" ]
+  then
+      filename=$(echo $filename | sed 's/[0-9].[0-9].[0-9]_next-//g')
+  fi
   echo "Uploading $filename"
   curl -v -f -u telefonica-github:$secret --upload-file $file https://nexus.lab.fiware.org/repository/el/8/x86_64/$type/$filename
 done
--- a/.github/build-and-push-rpm.sh
+++ b/.github/build-and-push-rpm.sh
@@ -17,6 +17,8 @@ do
   filename=$(basename $file)
   if [ "$type" == "nightly" ]
   then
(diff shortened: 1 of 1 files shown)


mapedraza:
CC: @fgalan @wistefan

.github/build-and-push-rpm.sh
```diff
@@ -15,6 +15,10 @@ docker run -e BROKER_RELEASE=$type -v $(pwd):/opt/orion --workdir=/opt/orion fiw
 for file in "$(pwd)/rpm/RPMS/x86_64"/*
 do
   filename=$(basename $file)
+  if [ "$type" == "nightly" ]
+  then
+      filename=$(echo $filename | sed 's/[0-9].[0-9].[0-9]_next-//g')
```
fgalan, line 20: Maybe it should be?

```suggestion
      filename=$(echo $filename | sed 's/[0-9]+.[0-9]+.[0-9]+_next-//g')
```

as `[0-9]` means a single number and some version numbers (e.g. 3.11.0) may have several digits.

It should be tested (not sure which regex variant sed is using :)

mapedraza: Its true, you can have more than one number. Fixed in https://github.com/telefonicaid/fiware-orion/pull/4050/commits/6053c42904ab339d0fd03359151e9af47997e0b1

You can test it with the following:
```BASH
echo "contextBroker-300.12.7890_next-nightly.x86_64.rpm" |  sed 's/[0-9]*.[0-9]*.[0-9]*_next-//g' 
```

fgalan: `+` means "one or more" while `*` means "zero or more". Note the diference:

```
$ echo "contextBroker-.._next-nightly.x86_64.rpm" |  sed 's/[0-9]\+.[0-9]\+.[0-9]\+_next-//g' 
contextBroker-.._next-nightly.x86_64.rpm
$ echo "contextBroker-.._next-nightly.x86_64.rpm" |  sed 's/[0-9]*.[0-9]*.[0-9]*_next-//g' 
contextBroker-nightly.x86_64.rpm
```

Using `+` we could eventual fails in the releasing scripts, as if the version string is not correct (e.g. `..` is not correct) that would cause a weird name in the repository and we would detect it.

mapedraza: > Using + we could eventual fails in the releasing scripts, as if the version string is not correct (e.g. .. is not correct) that would cause a weird name in the repository and we would detect it.

Undestood, fixed in https://github.com/telefonicaid/fiware-orion/pull/4050/commits/f0e2f5d21d51d9b29562be4793c674fcea6d2ffa

.github/build-and-push-rpm.sh
```diff
@@ -15,6 +15,10 @@ docker run -e BROKER_RELEASE=$type -v $(pwd):/opt/orion --workdir=/opt/orion fiw
 for file in "$(pwd)/rpm/RPMS/x86_64"/*
 do
   filename=$(basename $file)
+  if [ "$type" == "nightly" ]
+  then
+      filename=$(echo $filename | sed 's/[0-9]+.[0-9]+.[0-9]+_next-//g')
```
fgalan, line 20:
```suggestion
      # to make the name of the .rpm independent of the version name (i.e. contextBroker-nightly.x86_64.rpm),
      # avoiding obsolete packages to remain in the yum version when version number steps
      filename=$(echo $filename | sed 's/[0-9]+.[0-9]+.[0-9]+_next-//g')
```

mapedraza:
@wistefan Please could you review this PR? Thanks in advance

PR #4054, opened 2022-02-02
FIX Centos 8 repo list 

mapedraza:
Due to Centos8 EOL, building the Dockerfile was failing. This PR add a change in the repository base URL in order to keep working it. It changes the URL from `http://mirrorlist.centos.org/?release=$ ... fra=$infra` to `http://vault.centos.org/?release=$relea ... fra=$infra`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,7 +39,8 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /opt
 
 RUN \
-    #FIXME: Repo fix due to EOL Centos8
+    # FIXME: default yum repositories has been changed to vault due to CentOS 8 EOL on February 1st, 2022
+    # This is just a temporal solution so the build doesn't break while we find a new distro to use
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -39,6 +39,10 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 WORKDIR /opt
 
 RUN \
+    #FIXME: Repo fix due to EOL Centos8
```
fgalan, line 42:
```suggestion
    # FIXME: default yum repositories has been changed to vault due to CentOS 8 EOL on February 1st, 2022
    # This is just a temporal solution so the build doesn't break while we find a new distro to use
```

fgalan:
Similar changes have to be applied to https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh

mapedraza:
> Similar changes have to be applied to https://github.com/telefonicaid/fiware-orion/blob/master/ci/rpm8/build-dep.sh

Done in https://github.com/telefonicaid/fiware-orion/pull/4054/commits/07327cf24db0e32e276705e45eff3621f0a96f5e

fgalan (approved):
LGTM

PR #4057, opened 2022-02-04
[WIP] FIX hasField calls wrongly using char* instead of std::string

fgalan:
Note orion::BSONObj hasField() signature is:

```
  bool hasField(const std::string& field) const;
```

Thus I understand that although compiler doesn't complaint about it, hasField() on a char* is always evaluated to `false`.

What I wonder is how this has worked without problem until now... need to be analysed with a little of care

[ ] CNR ?

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -840,7 +840,7 @@ static bool updateAttribute
   }
   else
   {
-    if (!attrs.hasField(effectiveName.c_str()))
+    if (!attrs.hasField(effectiveName))
     {
       return false;
     }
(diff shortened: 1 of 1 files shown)


kzangeli:
I imagine the compiler does an "implicit typecast" to std::string (part of the class, as an operator), creating the object and copying the string into the new std::string.
It should work just fine.
[a bit slow of course with all these extra copies, and constructor/destructor calls that probably aren't at all necessary]

fgalan:
> I imagine the compiler does an "implicit typecast" to std::string (part of the class, as an operator), creating the object and copying the string into the new std::string. It should work just fine. [a bit slow of course with all these extra copies, and constructor/destructor calls that probably aren't at all necessary]

Makes sense.

mapedraza (approved):
LGTM

PR #4059, opened 2022-02-07
Closes #3570 / accumulator-server.py migration to python3 (final PR to master)

### LINKED ISSUE #3570: accumulator-server.py is not working even inside virtualenv
Well, I was following your [getting started guide](https://fiware-orion.readthedocs.io/en/master/user/walkthrough_apiv2/index.html#starting-accumulator-server) to turn on the accumulator server and receive Orion notifications in there. I've even created a `virtualenv` like you suggest and installed `Flask==1.0.2` and `pyOpenSSL==19.0.0` but it doesn't work even so. 

First I'm getting a syntax error in lin 64 about a `print` function, then if I fix the line as `python` suggest, I get another syntax error at line 76. 

What do you think? Perhaps I need a earlier `python` version?

Captura de pantalla_2019-11-04_19-37-09

fgalan:
Continues PR https://github.com/telefonicaid/fiware-orion/pull/4035, for the Dockerfile adaptations.

Note that first 14 commits in this PR are the ones in PR #4035.

CC: @jfernandz


fgalan:
The fixes needed after the @jfernandz work in PR #4035 can be seen in this delta of commits: https://github.com/telefonicaid/fiware-orion/compare/8778e24...123d06dce

Summary:

* Changes in the Dockerfile used for regression CI to install the new dependencies for python3. A lot of simplification here :) But sadly, we need to keep python2, as the build tooling in some of the software to install (GMock in particular) needs it.
* Additional fixes in accumulator-server.py. Summary:
   * Fixes related with encoding of payload to support thinks like 'ñ', etc.
   * New sort_headers() function to keep "backward compatibility" in headers in .test (the alternative to this function would be to change a loooooooooot of .test files to align with the new order in headers)
   * Time calculation simplification
   * string.atoi() -> int()
* Migrate other .py scripts use by the regression tooling to Python3: testDiff.py and jsonBeautifier.py (formerly named testJson.py)
* Fixes in .test:
    * Use jsonBeautifier.py instead of -mjson.tool to pretty print JSON. It seems the latter doesn't work as expected in Python3
    * Reordering of custom headers (case not covered by sort_headers() function) and query params in some .test
    * With the fixes in encoding stuff, it seems we can use a real 'ñ' in incorrect_location_processing.test :)
* Using PYTHONIOENCODING=utf8 in .sh test tooling when invoking testDiff.py and jsonBeautifier.py (otherwise, encoding problems occur)

@jfernandz it would be great if you could have a look to these changes and provide feedback if you find there is something to improve. Probably the PR will be merged soon (as it is somehow blocking the regression for other ongoing works) but we could apply any potential improvement afterwards in new PRs :). Thanks in advance!

fgalan:
@fisuda this PR does some modifications in .md files. It would be great if you could sync in the Japanese translation with an additional PR, please. Thanks!

ci/rpm8/build-dep.sh
```diff
@@ -23,6 +23,7 @@
 
 # FIXME: default yum repositories has been changed to vault due to CentOS 8 EOL on February 1st, 2022
 # This is just a temporal solution so the build doesn't break while we find a new distro to use
+# FIXME: python2 required by an installagion script in GMock. Sad but true :(
```
mapedraza, line 26:
```suggestion
# FIXME: python2 required by an installation script in GMock. Sad but true :(
```

mapedraza:
Apart from typo, LGTM

ci/rpm8/build-dep.sh
fgalan: Fixed in [f4765cc](https://github.com/telefonicaid/fiware-orion/pull/4059/commits/f4765cc26e4063609484891074063a7fdf51cd74)

mapedraza (approved):

jfernandz:
Some comments about your last post @fgalan:

> * Changes in the Dockerfile used for regression CI to install the new dependencies for python3. A lot of simplification here :) But sadly, we need to keep python2, as the build tooling in some of the software to install (GMock in particular) needs it.

I guess I could work on more future PRs to make some more python3 migrations if needed :blush: maybe ... I think you may provide a list about what scripts need to be ported/migrated to python3 :wink: 

> * Migrate other .py scripts use by the regression tooling to Python3: testDiff.py and jsonBeautifier.py (formerly named testJson.py)

I really like [`jq`](https://stedolan.github.io/jq/) for json enhancement, It's easily available/usable from any shell (though it's not a built-in), but I don't know either if `jq` will fullfil your requirements, just a suggestion :wink:

Aside from I'm not too familiar with your CI processes/infrastructure ... it LGTM also :smile:

fisuda:
I sent the PR #4060.

fgalan:
> I guess I could work on more future PRs to make some more python3 migrations if needed 😊 maybe ... I think you may provide a list about what scripts need to be ported/migrated to python3 😉

I think is a good idea. Thanks!

I'd suggest you open an issue with the list. Then I can provide feedback about which scripts are interesting to migrate (me feeling is that we have several ones that are old stuff and doesn't worth the effort... but I'll provide more precise feedback when I see that list).

> I really like [`jq`](https://stedolan.github.io/jq/) for json enhancement, It's easily available/usable from any shell (though it's not a built-in), but I don't know either if `jq` will fullfil your requirements, just a suggestion 😉

I know jq a little bit. But not sure if installing an additional tool (e.g. a new requirement in documentation, more stuff for the Dockerfile, etc.) for so little use (we only need JSON pretty printting, while jq has a lot of extra features) worth the effort. Thanks for the recommendation anyway :)

fgalan:
> I'd suggest you open an issue with the list. Then I can provide feedback about which scripts are interesting to migrate (me feeling is that we have several ones that are old stuff and doesn't worth the effort... but I'll provide more precise feedback when I see that list).

Side note: migrating the gtest/scripts/fuse_gtest_files.py script within https://nexus.lab.fiware.org/repository/raw/public/storage/gmock-1.5.0.tar.bz2 would allow us to remove the Python 2 dependency from Dockerfile CI at all. That would be probably the top priority, although I don't the difficulty (it hasn't been developed by my team, but but GMock people).

jfernandz:
@fgalan 
> Side note: migrating the gtest/scripts/fuse_gtest_files.py script within https://nexus.lab.fiware.org/repository/raw/public/storage/gmock-1.5.0.tar.bz2 would allow us to remove the Python 2 dependency from Dockerfile CI at all. That would be probably the top priority, although I don't the difficulty (it hasn't been developed by my team, but but GMock people).

Gmock has apparently been absorbed into https://github.com/google/googletest are you using an old version for some particular reason? :thinking:

fgalan:
> Gmock has apparently been absorbed into https://github.com/google/googletest are you using an old version for some particular reason? 🤔

Yes. The reason is: legacy :)

It would be great to move to newer GMock/GTest versions, but it can be costly and we don't have resources for doing it now. Any help on this is welcome!

PR #4084, opened 2022-03-11
Subscription alterationType feature

fgalan:
Issue #1494

Pending:

- [x] CNR
- [x] Documentation

PR #4088, opened 2022-03-23
Fix entityDelete

fgalan:
The .test included in the first commit of this PR shows that the implementation done in https://github.com/telefonicaid/fiware-orion/pull/4084 has a flaw regarding entityDelete notifications.

Two comments:

* We need the an actual type when calling addTriggeredSubscription() in the delete logic (eP->type with "" causes this problem. Maybe a findAndModify() with removal semantics should be used? Maybe we already has this information in notifyCerP?
* The changes done in addTriggeredSubscription_noCache logic to skip type in some case should be rolled-back.

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1264,22 +1264,14 @@ static void fill_idPtypeNP
   const std::string&  typePatternQ
 )
 {
-  std::string typeCheck;
-
-  // Sometimes (for instance, when we reach this point during a entity deletion operation)
-  // we don't know the type of the entity, in which case the check for type shouldn't be included
-  if (!entityType.empty())
-  {
-    typeCheck = std::string("(this.")+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_TYPE+" == \""+entityType+"\" || " +
-        "this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_TYPE+" == \"\" || " +
-        "!(\""+CSUB_ENTITY_TYPE+"\" in this."+CSUB_ENTITIES+"[i])) && ";
-  }
-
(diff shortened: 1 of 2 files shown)


fgalan:
> Maybe we already has this information in notifyCerP?

This is the simpler approach. The information is already there :)

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1264,22 +1264,14 @@ static void fill_idPtypeNP
...
-
   bgP->functionIdPtypeNP = std::string("function()") +
          "{" +
             "for (var i=0; i < this."+CSUB_ENTITIES+".length; i++) {" +
                 "if (this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_ISPATTERN+" == \"true\" && " +
-                    "!this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_ISTYPEPATTERN+" && " + typeCheck +
+                    "!this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_ISTYPEPATTERN+" && " +
+                    "(this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_TYPE+" == \""+entityType+"\" || " +
+                        "this."+CSUB_ENTITIES+"[i]."+CSUB_ENTITY_TYPE+" == \"\" || " +
+                        "!(\""+CSUB_ENTITY_TYPE+"\" in this."+CSUB_ENTITIES+"[i])) && " +
```
fgalan, line 1274: This is the rollback of a change introduced in recent PR https://github.com/telefonicaid/fiware-orion/pull/4084. It was introduced to solve the entityDelete case with empty type.

However, this is not needed taking into account the changes in updateEntity() introduced in this PR. Probably it doesn't hurt to keep as it was left in PR #4084. Howver, by the law of minimum intervention :), let's leave a it was before.

NTC (informative)

PR #4089, opened 2022-03-24
(JP) Add doc about subscription alterationType feature (#4084)

fisuda:
This PR adds Japanese documentation about subscription alterationType feature. It would be great if you could review this PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4100, opened 2022-04-20
FIX regComp wrapper for regcomp + regex documentation

fgalan:
Issue #4098

PR #4133, opened 2022-05-23
(JP) Add regex documentation (#4100)

fisuda:
This PR adds Japanese documentation about regex. It would be great if you could review this PR. Thanks.


fgalan (approved):
LGTM. Thanks for the typo fixing in the English docs! :)

PR #4135, opened 2022-05-23
FIX github workflow image

fgalan:
Time ago we moved from CentOS 8 to Debian, but it seems some tokens in workflow directory weren't changed.


mapedraza (approved):
LGTM

PR #4139, opened 2022-05-25
FIX ingoreType for Feature and FeatureCollection

fgalan:
The check for Feature and FeatureCollection cannot be done at paring stage, due to it depends on the existence of ignoreType metadata in the entity. This PR fixes that.

In addition, we have done some homogenization in `descriptions` in error responses, including the ""error parsing location attribute: " prefix in all the cases

### DIFF
--- a/src/lib/jsonParseV2/parseContextAttribute.cpp
+++ b/src/lib/jsonParseV2/parseContextAttribute.cpp
@@ -43,85 +43,14 @@
 #include "jsonParseV2/utilsParse.h"
 
 
-#if 0
-/* ****************************************************************************
-*
-* checkFeatureGeoJson -
-*
-*/
-static std::string checkFeatureGeoJson(orion::CompoundValueNode* feature)
-{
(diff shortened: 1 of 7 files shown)


src/lib/jsonParseV2/parseContextAttribute.cpp
```diff
@@ -44,84 +44,13 @@
...
-* Do checking that ensure getGeoJson() function in the mongoBackend layer will not
-* break. In particular:
-*
-* - Attribute is null or an object
-* - For Feature, that geometry field exists and it's an object
-* - For FeatureCollection:
-*   * the feature field exists
-*   * the feature field is an array with exactly one item
-*   * the feature field item has a geometry field and it's an object
+* Check that ttribute is null or an object. Other checks for Feature and FeatureCollection
```
mapedraza, line 51:
```suggestion
* Check that attribute is null or an object. Other checks for Feature and FeatureCollection
```

mapedraza (approved):
LGTM

PR #4161, opened 2022-07-06
FIX mqtt_subscription_multibroker.test glitch due to notification reception order

fgalan:
This PR should solve the annoying glitch with mqtt_subscription_multibroker.test that sometimes we see

imagen


mapedraza (approved):
LGTM

PR #4157, opened 2022-07-01
[WIP] ADD basic CRUD for json payload in custom notifications

fgalan:
Issud #2560

- [ ] Custom notifications based in `json` field (constant case)
- [ ] Custom notifications based in `json` field (replacement case)
- [ ] Check pending "FIXME PR" marks
- [ ] Documentation (database model and custom notification doc)
- [ ] CNR

PR #4164, opened 2022-07-07
(JP) ADD doc about json payload in custom notifications (#4157)

fisuda:
This PR adds Japanese documentation about json payload in custom notifications. It would be great if you could review this PR. Thanks.


fgalan (approved):
LGTM

Thanks for the link fix! :)

PR #4172, opened 2022-07-14
ADD sub_alteration_type_entity_update2.test extra case

fgalan:
Comes from https://github.com/telefonicaid/fiware-orion/issues/1494#issuecomment-1181656637

CC: @djs0109


fgalan:
> Side-note: it seems that if options=forcedUpdate is used, then alterationType attribute gets "entityChange", even in the case no actual update happens. Not sure if this should be fixed (using "entityUpdate") or at least documented...

Moved to separate PR #4175

mapedraza (approved):
LGTM

Issue #2591, opened 2016-10-10
previousValue = delete support


fgalan:
Upon completion of https://github.com/telefonicaid/fiware-orion/issues/1494, the `previousValue` with value `delete` must be supported.

Once this issue gets implemented, the following needs to be removed inthe implementation notes:

> actionType metadata
>
>From NGSIv2 specification section "Special metadata in notifications", regarding actionType metadata:
>
>    Its value depend on the request operation type: update for updates, append for creation and delete for deletion. Its type is always Text.
>
> Current Orion implementation supports "update" and "append". The "delete" case will be supported upon completion of this issue.

_Implementation notes:_

Have a loook to the `skip` field in ContextAttribute class and how it is used to mark an attribute as deleted in the entity memory image that is keep during the update process. It may be usefl at previousValue generation time in order to see which attribute were deleted.

```
static void deleteAttrInNotifyCer
(
  ContextElementResponse* notifyCerP,
  ContextAttribute*       targetAttr
)
{  
  for (unsigned int ix = 0; ix < notifyCerP->contextElement.contextAttributeVector.size(); ix++)
  {
    ContextAttribute* caP = notifyCerP->contextElement.contextAttributeVector[ix];
    if (caP->name == targetAttr->name)
    {
      caP->skip = true;
    }
  }
}
```

Anjali-NEC:
Hi @fgalan sir,

I would like to work on this issue, please let me know this issue is still need to be fix or not?
Thanks

Anjali-NEC:
@fgalan 

As per my understanding we need to support delete actionType in Special metadata in notifications
Currently `actionType` support only on the request operation type: `update` and `append` 

So, we need to add actionType `delete` for  `delete` operation in the notifications for providing more information about the deleted attribute.

I am referring the below document for same:
https://github.com/telefonicaid/fiware-orion/blob/0dc77185b8f98274840abc28364e2bd071c1d7e4/doc/apiary/v2/fiware-ngsiv2-rc-2016_10_31.apib#L881

Please confirm my understanding.

fgalan:
doc/manuals/orion-api.md  section ` ## actionType metadata` should be removed upon completing this issue.

fgalan:
> @fgalan
> 
> As per my understanding we need to support delete actionType in Special metadata in notifications Currently `actionType` support only on the request operation type: `update` and `append`
> 
[...]

PR #4097, opened 2022-04-18
Fix: Extra information about partial updates in responses #3499

Anjali-NEC:
Fix issue #3499

PR #4090, opened 2022-03-25
Fix:  Avoid triggering notification if metadata changes but not the attribute value itself #3727

Anjali-NEC:
Fixed issue #3727

PR #4003, opened 2021-11-15
Fix  Partial success HTTP status codes #3054

Anjali-NEC:
Fix issue #3054

PR #4186, opened 2022-08-08
Fix: actionType metadata: delete support #2591

Anjali-NEC:
Fix issue https://github.com/telefonicaid/fiware-orion/issues/2591

### DIFF
--- a/src/lib/mongoBackend/MongoCommonUpdate.cpp
+++ b/src/lib/mongoBackend/MongoCommonUpdate.cpp
@@ -1832,7 +1832,7 @@ static bool processOnChangeConditionForUpdateContext
     {
       for (unsigned int jx = 0; jx < attrL.size(); jx++)
       {
-        if (caP->name == attrL[jx] && !caP->skip)
+        if (caP->name == attrL[jx])
         {
           /* Note we use cloneCompound=true in the ContextAttribute constructor. This is due to
            * cer.entity destructor does release() on the attrs vector */
@@ -2375,7 +2375,10 @@ static void updateAttrInNotifyCer
 static void deleteAttrInNotifyCer
 (
   ContextElementResponse* notifyCerP,
-  ContextAttribute*       targetAttr
+  ContextAttribute*       targetAttr,
+  bool                    useDefaultType,
+  const std::string&      actionType,
+  const bool&             overrideMetadata
 )
 {
   for (unsigned int ix = 0; ix < notifyCerP->entity.attributeVector.size(); ix++)
@@ -2384,6 +2387,7 @@ static void deleteAttrInNotifyCer
     if (caP->name == targetAttr->name)
     {
       caP->skip = true;
+      caP->actionType = NGSI_MD_ACTIONTYPE_DELETE;
     }
   }
 }
@@ -2561,12 +2565,13 @@ static bool deleteContextAttributeItem
   bool*                     entityModified,
   orion::BSONDate*          dateExpiration,
   ApiVersion                apiVersion,
+  const bool&               overrideMetadata,
   OrionError*               oe
 )
 {
   if (deleteAttribute(attrs, toUnset, attrNamesRemove, targetAttr))
   {
-    deleteAttrInNotifyCer(notifyCerP, targetAttr);
+    deleteAttrInNotifyCer(notifyCerP, targetAttr,  apiVersion == V2, NGSI_MD_ACTIONTYPE_DELETE, overrideMetadata);
     *entityModified = true;
 
     /* Check aspects related with location */
@@ -2745,6 +2750,7 @@ static bool processContextAttributeVector
                                       &entityModified,
                                       dateExpiration,
                                       apiVersion,
+                                      overrideMetadata,
                                       oe))
       {
         return false;
--- a/test/functionalTest/cases/2591_actionType_metadata_delete_support/actionType_metadata_delete_support.test
+++ b/test/functionalTest/cases/2591_actionType_metadata_delete_support/actionType_metadata_delete_support.test
@@ -1,0 +1,188 @@
+# Copyright 2022 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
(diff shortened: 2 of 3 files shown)


src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2375,7 +2375,10 @@ static void updateAttrInNotifyCer
 static void deleteAttrInNotifyCer
 (
   ContextElementResponse* notifyCerP,
-  ContextAttribute*       targetAttr
+  ContextAttribute*       targetAttr,
+  bool                    useDefaultType,
+  const std::string&      actionType,
+  const bool&             overrideMetadata
```
fgalan, line 2381: It seems useDefaultType, actionType and overrideMetadata parameter are not actually used in the deleteAttrInNotifyCer() function

Anjali-NEC: > It seems useDefaultType, actionType and overrideMetadata parameter are not actually used in the deleteAttrInNotifyCer() function

I have removed useDefaultType, overrideMetadata from deleteAttrInNotifyCer() function in my latest commit  https://github.com/telefonicaid/fiware-orion/pull/4186/commits/c5392556585a013d6aea61c8df8cec0cb1779ba4 

I have used actionType parameter in deleteAttrInNotifyCer to set delete actionType.  Please suggest if any other changes are required.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2375,7 +2375,8 @@ static void updateAttrInNotifyCer
 static void deleteAttrInNotifyCer
 (
   ContextElementResponse* notifyCerP,
-  ContextAttribute*       targetAttr
+  ContextAttribute*       targetAttr,
+  const std::string&      actionType
```
fgalan, line 2379: This parameter is not used. Should be removed.

Anjali-NEC: Fixed in https://github.com/telefonicaid/fiware-orion/pull/4186/commits/8d8174d4a1d7a62451f3d480c9032b9f21569831

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -2561,6 +2562,7 @@ static bool deleteContextAttributeItem
   bool*                     entityModified,
   orion::BSONDate*          dateExpiration,
   ApiVersion                apiVersion,
+  const bool&               overrideMetadata,
```
fgalan, line 2565: It seems this parameter is not used inside the function. Thus, it should be removed (also from L2750).

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1832,7 +1832,7 @@ static bool processOnChangeConditionForUpdateContext
     {
       for (unsigned int jx = 0; jx < attrL.size(); jx++)
       {
-        if (caP->name == attrL[jx] && !caP->skip)
+        if (caP->name == attrL[jx])
```
fgalan, line 1835: Why did you need to remove the `&& !caP->skip` part? Could you elaborate on it, please?

Anjali-NEC: > Why did you need to remove the `&& !caP->skip` part? Could you elaborate on it, please?

As per my understanding this `caP->skip` field is used to mark deleted attributes that must not be included in the notification and this condition is used to add the attributes in the notification. So, I removed `!caP->skip` so that the deleted attributes included in the notification.

src/lib/mongoBackend/MongoCommonUpdate.cpp
Anjali-NEC: > It seems this parameter is not used inside the function. Thus, it should be removed (also from L2750).

removed in https://github.com/telefonicaid/fiware-orion/pull/4186/commits/7a65459e06212144bc3932eca95122f14ce9f3f9

src/lib/mongoBackend/MongoCommonUpdate.cpp
fgalan: Regardless of this new functionality, my doubt is if `&& !caP->skip` is really needed. Maybe it was needed in the past but some refactor in the code make it unneeded.

I'll cast and independent PR to check that.

fgalan: > I'll cast and independent PR to check that.

This one: PR https://github.com/telefonicaid/fiware-orion/pull/4201

fgalan: PR https://github.com/telefonicaid/fiware-orion/pull/4201 went ok, thus validating this change.

However, I'll investigate a little bit, to understand why this was there but now it is not needed :)

fgalan: Looking into this into more detail (see https://github.com/telefonicaid/fiware-orion/pull/4201#issuecomment-1233057677) I'm afraid that the modification is not valid.

This PR doesn't detect the problem due to in the moment the PR was created master didn't have the proper coverage in tests. Now that these test has been included (in PR https://github.com/telefonicaid/fiware-orion/pull/4204), if you upgrade  Anjali-NEC:issue2591 with master, we will be see some tests failing.

test/functionalTest/cases/2591_actionType_metadata_delete_support/actionType_metadata_delete_support.test
```diff
@@ -0,0 +1,188 @@
...
+            },
+            "B": {
+                "metadata": {
+                    "actionType": {
+                        "type": "Text",
+                        "value": "delete"
+                    }
+                },
+                "type": "Number",
+                "value": 2
```
fgalan, line 169: Currently, default behaviour is that deleted attribute doesn't appear in notifications (see delete_attribute_in_empty_condition.test and delete_attribute_in_condition.test test cases).

It could make sense to include deleted attributes in notifications when `actionType` is used (otherwise the `delete` case for `actionType` metadata is pointless). However, in this case, which should be used as `type` and `value` here?

1. The previous ones (as this PR currently shows)
2. Omit the `type` and `value` fields (but it should break the structure of attributes, as `value` is always assumed
3. Use null value, i.e. `"type": "None"` and `"value": null`

I think the most semantically correct is number 3.

Another interesting case should be using actionType and previousValue in combination. In this case, previousValue should show the value previous to the deletion. Something like this:

```
"B": {
                "metadata": {
                    "actionType": {
                        "type": "Text",
                        "value": "delete"
                    },
                    "previousValue": {
                        "type": "Number",
                        "value": 2
                    }
                },
                "type": "None",
                "value": null
```

fgalan:
@Anjali-NEC could you tell us the status on this PR? Did you get my last feedback (around two weeks ago)?

It would be great to know it, given in the next FIWARE TSC Orion is going to be discussed.

Anjali-NEC:
> @Anjali-NEC could you tell us the status on this PR? Did you get my last feedback (around two weeks ago)?
> 
> It would be great to know it, given in the next FIWARE TSC Orion is going to be discussed.

I am putting this PR on hold currently I am busy in working other issue.

fgalan:
> Please let us know which ticket has high priority to you in below PR:
> #4186 , #4003 , #4097 and #4090
> Based on the priority, we will plan our work on open tickets accordingly

Answered [here](https://github.com/telefonicaid/fiware-orion/pull/4003#issuecomment-1350879694)

fgalan:
After the merging of PR https://github.com/telefonicaid/fiware-orion/pull/4332 this PR needs to be upgrades with `master` branch. Otherwise, functional test will fail.

PR #4204, opened 2022-08-30
ADD test cases for attribute deletion for subscription triggering

fgalan:
This PR enhance the test suite with two new cases, covering attribute deletion for triggering subscription. In particular:

* A test case to ensure that deleting an attribute in subject.condition triggers the notification
* A test case to ensure that deleting an attribute *not* in subject.condition doesn't triggers the notification

CC: @Anjali-NEC maybe you want to have a look and provide feedback, as this PR is somehow related with your PR https://github.com/telefonicaid/fiware-orion/pull/4186.

Pending:

- [x] Doc modifications (to "Notification Triggering" subsection)

PR #4208, opened 2022-09-01
(JP) ADD doc about attribute deletion for subscription triggering (#4204)

fisuda:
This PR adds Japanese documentation about attribute deletion for subscription triggering. It would be great if you could review this PR. Thanks.


fgalan (approved):
LGTM

PR #4227, opened 2022-10-25
FIX fail in csub cache logic that may make CB to crash in some cases

fgalan:
Fixes a crash situation (checked by the .test file included in this PR) discovered by @danielvillalbamota 

No need of CHANGES_NEXT_RELEASE entry given that is a fix in a functionality introduced post-3.7.0.

Some minor cleanup in TriggeredSubscriptions and globals.h spotted while working on this.


danielvillalbamota (approved):
LGTM

Issue #4085, opened 2022-03-14
Suscripciones Custom for New Type/New Id


fgalan:
Implement flexibility in custom notifications, so the id and type of the notified entity can be changed.

To be detailed.

fgalan:
What about using macros in the mapping itsef? Eg. to provide a prefix to id:

```
    "httpCustom": {
      "url": "http://127.0.0.1:9999/notify",
      "mapping": {
        "id": "PREFIX${id}"
      }
```

mrutid:
I would add all the "first level" information elements, that is: id, type, service, subservice, auth-token (sometimes you don't want to propagate it).

- service, subservice, auth-token may be manipulated throught http headers but thats an implementation caveat in my opinion.

mrutid:
Maybe we should have access to the old id/type/S/SS also in the template... You may want a new id but the old id to be part of some attribute  composition. 
An off topic example may be found in Cygnus NameMappings we use "the NEW" to build the destination, and "the OLD" to populate table values.

NOTE: maybe we just need to hace access to the old ones in the macro language, the new ones always can be build or dereivated.

fgalan:
Somehow related, new macros for service and servicePath are mentioned in https://github.com/telefonicaid/fiware-orion/issues/3913

fgalan:
```
    "httpCustom": {
      "url": "http://127.0.0.1:9999/notify",
      "mapping": {
        "id": "newId",
        "type": "newType",
        "service": "SS",
        "servicePath": "/newSP",
        "authToken": "..."
      }
```

Notes:
* In the case of using "" then the corresponding header will not appear
* In the case of having some of them within `headers` the one in `headers` takes precedence.
* NOTE: ${service}, ${servicePath}, {$authToken} macros will be implemented as part of a separate issue.

fgalan:
We are going to re-define this, based on the payload concept. So, in a custom notification, there would be three ways of defining a payload:

* As a custom text payload (currently supported with the `payload` field)
* As a custom JSON payload (currently supported with the `json` field)
* As a NGSIv2 normalized payload (the purpose of this issue)

The third mode will be associated to the `entity` field, at the same level than `payload` and `json` (note: `payload`, `json` and `entity` are mutually exclusive, i.e. only one of them can be used).

The value of `entity` will allow to defined *mapped* entity fields (id, type or attributes) that will be added the entity subject of the notification. So, let's consider the following updated entity

```
{
  "id": "E",
[...]

PR #4228, opened 2022-10-26
FIX add some simpler toJson() renders

fgalan:
CHANGES_NEXT_RELEASE entry doesn't needed (it's a small refactor with no impact on functionality)

I plan to use some of these new toJson() in the conext of issue #4085


mrutid (approved):
LGTM

PR #4226, opened 2022-10-24
FIX add code comments showing query structure in csubs no cache logic

fgalan:
Issue #4148 

This PR adds useful information about query structure in the csubs no cache logic, in the case it can be useful for future manteinance.

In addition, some leftovers in markdown documentation regarding the extinct $where operation usage has been removed.

@anupamme it would be great if you could have a look and provide LGTM if you think changes are ok (or feedback comments otherwise), please.

PR #4234, opened 2022-11-14
(JP) FIX update doc about csubs no cache logic (#4226)

fisuda:
This PR updates Japanese documentation about csubs no cache logic. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4237, opened 2022-12-02
FIX toJsonV1 wrappers to simplify usage

fgalan:
Implement some toJsonV1() wrapper to avoid `emptyV` dirty thing.

(NGSIv1 is deprecate but this code was hurting my eyes :)


mapedraza (approved):
LGTM

PR #4244, opened 2022-12-13
Strike complete items in 

fgalan:
By the moment just striking the entries. In the next mayor review of the roadmap all the striked ones will be moved to "Features already completed" section.

### DIFF
--- a/doc/roadmap.md
+++ b/doc/roadmap.md
@@ -33,11 +33,11 @@ The following list of features are planned to be addressed in the short term,
 and incorporated into the coming release(s) of the product:
 
 - Allow multiple types in entity to support UNE 178503 requirements ([#3638](https://github.com/telefonicaid/fiware-orion/issues/3638))
-- New subscripition modes (create only, update only, delete only and combinations) ([#1494](https://github.com/telefonicaid/fiware-orion/issues/1494))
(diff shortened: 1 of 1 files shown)


mapedraza:
Should the features be moved to the section "Features already completed"? (at least the ones that are released, like New subscripition modes (create only, update only, delete only and combinations) )

fgalan:
> Should the features be moved to the section "Features already completed"? (at least the ones that are released, like New subscripition modes (create only, update only, delete only and combinations) )

Usure... My original idea was:

>  In the next mayor review of the roadmap all the striked ones will be moved to "Features already completed" section.

However, if you think it's better to move them to "Features already completed", we can do that. I don't have a strong opinion :)

mapedraza:
> > Should the features be moved to the section "Features already completed"? (at least the ones that are released, like New subscripition modes (create only, update only, delete only and combinations) )
> 
> Usure... My original idea was:
> 
> > In the next mayor review of the roadmap all the striked ones will be moved to "Features already completed" section.
> 
> However, if you think it's better to move them to "Features already completed", we can do that. I don't have a strong opinion :)

Let's do what you propose, so we can track the roadmap evolution :)

Issue #4249, opened 2022-12-16
Cors allowed headers do not include "Authorization" which is the default headers used with OAuth.


chicco785:
Could you please add "Authorization" in the list of Cors allowed headers?

https://github.com/telefonicaid/fiware-orion/blob/0dc77185b8f98274840abc28364e2bd071c1d7e4/src/lib/rest/HttpHeaders.h#L68

fgalan:
If you create the pull request for adding it, we can include it in the next release (soon to be frozen ;)

Is this the only point that need to be touched? Please check also documentation.

Finally, don't forget to add the corresponding entry to CHANGES_NEXT_RELEASE file, please.

Thanks!

chicco785:
> If you create the pull request for adding it, we can include it in the next release (soon to be frozen ;)
> 
> Is this the only point that need to be touched? Please check also documentation.
> 
> Finally, don't forget to add the corresponding entry to CHANGES_NEXT_RELEASE file, please.
> 
> Thanks!

Not a C developer. Said that, if it's just about adding that to the list, I can do, more complex if I need to develop a test case :)

chicco785:
> Btw, could you provide a little more context about the usage? I mean, in which case (architecture, etc.) are you using Orion with a system requiring CORS.
> 
> Thanks again!

We have an admin interface, now in front of Orion, we recently added a new authz proxy based on envoy, which by default pass all headers, so the issue popped up.

fgalan:
> Not a C developer. Said that, if it's just about adding that to the list, I can do, more complex if I need to develop a test case :)

Do a PR without the test so we can check if something break ;)

Next, maybe just adapting (copy-paste-modify) some of the tests in https://github.com/telefonicaid/fiware-orion/tree/master/test/functionalTest/cases/0501_cors would suffice. Please have a look.

fgalan:
Fixed by PR https://github.com/telefonicaid/fiware-orion/pull/4251

PR #4251, opened 2022-12-18
Add "Authorization" in the list of Cors allowed headers (#4249)

chicco785:


chicco785:
@fgalan looking at the code I realize you are propagating the `X-AUTH-TOKEN` header. Which I believe is an interesting functionality also for the OAuth tokens using the standard header. So, instead of duplicating the code to reproduce the functionality, the smartest way could be to set the name of this header via an environment variable, so that the same functionality can be shared without duplications. But I have no clue how to do that :)

fgalan:
> @fgalan looking at the code I realize you are propagating the `X-AUTH-TOKEN` header. Which I believe is an interesting functionality also for the OAuth tokens using the standard header. So, instead of duplicating the code to reproduce the functionality, the smartest way could be to set the name of this header via an environment variable, so that the same functionality can be shared without duplications. But I have no clue how to do that :)

Unsure... but let's move it out of this PR into an independent issue so we can discuss there.

Could you create the issue, please? Thanks!

fgalan (approved):
LGTM

fgalan:
@fisuda this PR does modifications to some .md files. Not sure if they have equivalence in Japanese translation, in that case it would be great if you could sync the changes, please.

Thanks!

fisuda:
I send the PR #4253. Thanks.

PR #4254, opened 2022-12-20
FIX improve statistics API

fgalan:
- [x] versionRequests counter removal (already covered by GET /versions counter)
- [x] "getAll counters" GET /statistics option
- [x] check docu (not sure if some change would be required)

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -4,7 +4,7 @@
 - Fix: conditions.alterationTypes not working properly when conditions.attributes is used in entityUpdate case (#1494, reopened)
 - Fix: GET /v2/types and GET /v2/types/{type} wrong attribute types result (#4216)
 - Fix: reset measuring_interval_in_secs correctly reset upon DELETE /statistics (#4233)
-- Fix: version request correctly recorded in statistics API (#4233)
+- Fix: version request correctly recorded in statistics API (in "/version".GET instead than in separated versionRequests counter) (#4233)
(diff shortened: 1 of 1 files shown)


mapedraza (approved):
LGTM

fisuda:
I sent the PR #4257. Thanks.

fgalan:
I have done some "coverage analysis"

* Get requests from the "routes map"

```
cat app/contextBroker/orionRestServices.cpp | grep Request | awk -F '{' '{print $2'} | awk -F ',' '{print $1}' | sed 's/\W//g' | sort -u > request_set1.txt
```

* Get the requests covered by the statistics

```
cat lib/common/statistics.cpp | grep ' {\w' | awk -F '{' '{print $2'} | awk -F ',' '{print $1}' | sort -u > requets_set2.txt
```

* Compare both (`diff -u requets_set1.txt requets_set2.txt`)

```diff
--- requets_set1.txt	2022-12-22 10:26:20.575685478 +0100
+++ requets_set2.txt	2022-12-22 10:28:14.706882935 +0100
@@ -25,7 +25,6 @@
 IndividualContextEntityAttribute
 IndividualContextEntityAttributes
 IndividualContextEntityAttributeWithTypeAndId
-InvalidRequest
 LeakRequest
 LogLevelRequest
 LogTraceRequest
```

Except for `InvalidRequest` (which is used in the badVerd logic only, thus is not an actual request) they are equal, so I think the statistic logs covers correctly the current requests.

PR #4274, opened 2023-02-09
Fix to avoid notification if metadata changes but not the attribute value itself

### LINKED ISSUE #3727: Avoid triggering notification if metadata changes but not the attribute value itself
Currently, the metadata of an attribute is considered part of its value. Thus, a subscription using a given attribute as a triggering condition will trigger notifications in cases in which the attribute value doesn't change but metadata does. This is precluding some use cases.

Thus, in order to increase the flexibility of the subscription, this case should be allowed. The simpler approach seems to use a global general setting in the subscription condition (i.e. `"metadataAreNotPartOfValue": false` :) so notifications are sent only if the value of the attribute itself changes.

Although maybe is a too "coarse-grained" solution and other approaches are possible...

chetan-NEC:
**Fix #3727** 

- Fix to avoid notification if metadata changes but not the attribute value itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![Support badge](https://img.shields.io/badge/tag-fiware--orion-orange.svg?logo=stackoverflow)](http://stackoverflow.com/questions/tagged/fiware-orion)
 [![NGSI v2](https://img.shields.io/badge/NGSI-V2-red.svg)](doc/manuals/orion-api.md)
 <br>
-[![Documentation badge](https://img.shields.io/readthedocs/fiware-orion/3.8.0.svg)](https://fiware-orion.rtfd.io/en/3.8.0/)
+[![Documentation badge](https://img.shields.io/readthedocs/fiware-orion.svg)](https://fiware-orion.rtfd.io)
 ![Compliance Tests](https://github.com/telefonicaid/fiware-orion/workflows/Compliance%20Tests/badge.svg)
 ![Unit Tests](https://github.com/telefonicaid/fiware-orion/workflows/Unit%20Tests/badge.svg)
 ![Functional Tests](https://github.com/telefonicaid/fiware-orion/workflows/Functional%20Tests/badge.svg)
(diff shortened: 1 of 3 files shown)


chetan-NEC:
Hi @fgalan, please check the pull request and let me know if you have any recommendations.
Thank you in advance for your time and assistance.

src/lib/common/globals.cpp
```diff
@@ -57,6 +57,7 @@ bool                   notifQueueStatistics = false;
 bool                   checkIdv1            = false;
 unsigned long long     inReqPayloadMaxSize  = DEFAULT_IN_REQ_PAYLOAD_MAX_SIZE;
 unsigned long long     outReqMsgMaxSize     = DEFAULT_OUT_REQ_MSG_MAX_SIZE;
+bool                   notifyOnMetadataChange = false;
```
fgalan, line 60: Note we don't want a global behaviour but a per-subscription behaviour.

In other words, `notifyOnMetadataChange` should be a field in the subscription document. Have a look to other boolean fields in subscriptions (like `onlyChangedAttrs` or `covered`) and try to do deal with them in a similar way from the point of view of subscription CRUD (Create-Read-Update-Delete).

The main difference with regards the aforementioned `onlyChangedAttrs` or `covered` is that they are within `notification` and `notifyOnMetadataChange` should be within `condition` (as it has to do with notification triggering conditions).

fgalan:
Pull request should include new functional tests (.test files) in order to cover the new functionality implemented.

chetan-NEC:
Hello @fgalan, as per your suggestion, I have implemented the features based on per-subscription behavior. and am currently working on updating the subscription-related test cases.

chetan-NEC:
Hello, @fgalan. I apologize for accidentally closing pull request #4274, which was recently submitted. I will take the appropriate precautions to prevent making such mistakes in the future.

Moving forward, I have created a new PR for issue #3727; please review the new pull request #4300 and let me know if you have any recommendations.
Thank you in advance for your time and assistance.

src/lib/common/globals.cpp
chetan-NEC: Resolved

PR #4278, opened 2023-02-16
FIX blocked worker when MQTT broker is out (master)

fgalan:
Backport of PR #4276 (release/3.8.0) for master branch

PR #4284, opened 2023-02-17
(JP) ADD doc about MQTT timeout (#4278)

fisuda:
This PR adds Japanese documentation about MQTT timeout. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM

Thanks for the typo fix!

PR #4300, opened 2023-03-21
Fix Avoid triggering notifications if only the metadata changes

chetan-NEC:
**Fix** #3727 

- Fix avoid triggering notifications if only the metadata changes based on per-subscription behavior


test/functionalTest/cases/0000_subs_attrs_deletion/delete_attribute_in_condition.test
```diff
@@ -207,7 +207,8 @@ Date: REGEX(.*)
                 "attrs": [
                     "B",
                     "C"
-                ]
+                ],
+                "notifyOnMetadataChange": false
```
fgalan, line 211: Let's clarify the semantics of this new `notifyOnMetadataChange` field.

* `"notifyOnMetadataChange": true`. The metadata is considered part of the value of the attribute in the context of notification, so if the value doesn't change but the metadata changes, then a notification is triggered.
* `"notifyOnMetadataChange": false`. The metadata is not considered part of the value of the attribute in the context of notification, so if the value doesn't change but the metadata changes, then a notification is *not* triggered.

If we agree on that semantic (otherwise, please do an alternative proposal), I understand that to keep backward compatibility we should use `true` here (and in other existing .test files).

test/functionalTest/cases/3727_avoid_triggering_notification_if_metadata_changes_but_not_the_attribute_value_itself/avoid_triggering_notification_if_metadata_changes_but_not_the_attribute_value_itself.test
```diff
@@ -0,0 +1,489 @@
...
+# 04. Dump the accumulator, see no notification when metadata changes
+# 05. PATCH /v2/entities/E1/attrs, to update the value of attribute
+# 06. Dump the accumulator, see the notification when value of attribute changes
+# 07. POST /v2/entities, to create entity E2 with attr A1 and metadata M1
+# 08. POST /v2/subscriptions, to create subscription for E2 with notifyOnMetadataChange=true
+# 09. PATCH /v2/entities/E2/attrs, to update the value of metadata
+# 10. Dump the accumulator, see the notification when metadata changes
+# 11. PATCH /v2/entities/E2/attrs, to update the value of attribute
+# 12. Dump the accumulator, see the notification when value of attribute changes
+# 13. POST /v2/subscriptions, to create subscription for E3 with notifyOnMetadataChange=foo and see error
```
fgalan, line 46: I think these steps are fine, but I'd suggest to split into three different .test files, for the sake of maintainability. I mean:

* one .test file for notifyOnMetadataChange=true case
* one .test file for notifyOnMetadataChange=false case
* one .test file for notifyOnMetadataChange=foo case (error case)

test/functionalTest/cases/0000_subs_attrs_deletion/delete_attribute_in_condition.test
fgalan: ...assuming that the current behaviour (i.e. before this PR) is that the value doesn't change but the metadata changes, then a notification is triggered. It would be great if you could check that with the last ContextBroker version available (3.8.1)

src/lib/cache/subCache.cpp
```diff
@@ -829,7 +829,8 @@ void subCacheItemInsert
   const std::string&                      georel,
   bool                                    blacklist,
   bool                                    onlyChanged,
-  bool                                    covered
+  bool                                    covered,
+  bool                                    notifyOnMetadataChange
```
fgalan, line 833: The new field should be added to the doc/manuals/devel/subscriptionCache.md document.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -1673,6 +1687,8 @@ static bool addTriggeredSubscriptions_noCache
       bool              onlyChanged        = sub.hasField(CSUB_ONLYCHANGED)? getBoolFieldF(sub, CSUB_ONLYCHANGED) : false;
       bool              blacklist          = sub.hasField(CSUB_BLACKLIST)? getBoolFieldF(sub, CSUB_BLACKLIST) : false;
       bool              covered            = sub.hasField(CSUB_COVERED)? getBoolFieldF(sub, CSUB_COVERED) : false;
+      bool              notifyOnMetadataChange = sub.hasField(CSUB_NOTIFYONMETADATACHANGE)? \
+      getBoolFieldF(sub, CSUB_NOTIFYONMETADATACHANGE) : false;
```
fgalan, line 1691: Better in the same line, avoiding `\` break.

test/functionalTest/cases/2560_custom_notification_json/custom_notification_mqtt_json_replacements_inside.test
```diff
@@ -16,7 +16,7 @@
 # along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
 #
 # For those usages not covered by this license please contact with
-# iot_support at tid dot es
+# iot_support at tid dot es.
```
fgalan, line 19:
```suggestion
# iot_support at tid dot es
```

That little dot is causing a fail in the compliance check script ;) https://github.com/telefonicaid/fiware-orion/actions/runs/4499061857/jobs/7938064519?pr=4300

fgalan:
To solve the conflicts in test/functionalTest/cases/3541 I'd suggest:

* To take the .test from master branch
* Apply the new field to them so the test pass again

fgalan:
Unit test are failing, probably because you have added a parameter to a function that is used in mocks, so mocks have to be also updated there. However, if this is hard to you, don't pay attention to it yet. I can help with that when the rest of the PR gets finished.

fgalan:
An entry to CHANGES_NEXT_RELEASE about the changes in this PR should be added. I'd suggest this one:

```
- Add: subject.condition.notifyOnMetadataChange field to subscriptions, so only-metadata changes doesn't trigger notifications (#3727)
```

src/lib/mongoBackend/MongoCommonUpdate.cpp
chetan-NEC: Thank you for your suggestion, but if I add code in the same line, I exceed the limit of 120 characters, and the compliance test scenario fails.

fgalan: I understand you refer to [S9 rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/contribution_guidelines.md#s9-line-length).

Note that that rule is in the "SHOULD" section, not in the "MUST" section and that there are existing lines exceeding that limite (L1685 is 140 chars long).

Thus, I'd suggest to use put all them in the same line and let's see if the compliance tests fails (I think it would pass, but let's see...)

test/functionalTest/cases/0000_subs_attrs_deletion/delete_attribute_in_condition.test
chetan-NEC: I agree with the semantics, but a subscription with an empty condition ("notifyOnMetadataChange") returns false because no notifyOnMetadataChange value was provided.

fgalan: In general, empty field (whatever) means "default values for the sub-field inside the field".

Thus, empty `condition` means "default values for the sub-fields inside `condition`, being `notificationOnMetadataChange` a sub-field of `condition` which default is `true` (as discussed above).

chetan-NEC: understood, I'll change the test cases to reflect your recommendation and set the notificationOnMetadataChange default value to true.

test/functionalTest/cases/3727_avoid_triggering_notification_if_metadata_changes_but_not_the_attribute_value_itself/notifyOnMetadataChange_true_in_condition.test
```diff
@@ -0,0 +1,256 @@
...
+payload='{
+  "A1": {
+    "metadata": {
+      "M1": {
+        "value": 1.55,
+        "type": "Float"
+      }
+    }
+  }
+}'
```
fgalan, line 107: I'd sugges to use in step 03 the same used in step 01 for the attribute value and type:
I'd suggest to use in step 03 the same used in step 01 for the attribute value and type (see my previous comment)

```
    "value": 26.5,
    "type": "Float",
```

src/lib/mongoBackend/dbConstants.h
```diff
@@ -120,6 +120,7 @@
 #define CSUB_BLACKLIST               "blacklist"
 #define CSUB_ONLYCHANGED             "onlyChanged"
 #define CSUB_COVERED                 "covered"
+#define CSUB_NOTIFYONMETADATACHANGE  "notifyOnMetadataChange"
```
fgalan, line 123: New field should be added to doc/manuals/admin/database_model.md#csubs-collection

test/functionalTest/cases/3727_avoid_triggering_notification_if_metadata_changes_but_not_the_attribute_value_itself/notifyOnMetadataChange_false_in_condition.test
```diff
@@ -0,0 +1,227 @@
...
+payload='{
+  "A1": {
+    "metadata": {
+      "M1": {
+        "value": 1.55,
+        "type": "Float"
+      }
+    }
+  }
+}'
```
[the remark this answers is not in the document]
fgalan: A separate PR has been created with a test for committing value case (PR https://github.com/telefonicaid/fiware-orion/pull/4307) so we can use it to ensure this PR keeps backward compatibility.

Once PR #4307 gets merged, please upgrade this PR with master to see how it goes.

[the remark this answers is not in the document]
chetan-NEC: PR Upgraded.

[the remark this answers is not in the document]
chetan-NEC: The complementary test case 3727 failed after upgrading the PR, so I'll fix it and upgrade the PR shortly.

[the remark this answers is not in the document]
fgalan: Great!

Probably some adjustment needs to be done in this part of the code https://github.com/telefonicaid/fiware-orion/pull/4300/files#diff-f3de384894601039a5d62bd71c655eade1dbf49e023cdacce6e843279808401bR557-R573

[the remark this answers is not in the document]
chetan-NEC: Fixed in 87e8150

src/lib/common/globals.cpp
```diff
@@ -57,6 +57,7 @@ bool                   notifQueueStatistics = false;
 bool                   checkIdv1            = false;
 unsigned long long     inReqPayloadMaxSize  = DEFAULT_IN_REQ_PAYLOAD_MAX_SIZE;
 unsigned long long     outReqMsgMaxSize     = DEFAULT_OUT_REQ_MSG_MAX_SIZE;
+bool                   onlyMetadataChange   = false;
```
fgalan, line 60: This was already commented in the original PR https://github.com/telefonicaid/fiware-orion/pull/4274 (sorry I haven't spotted until now... a lot of file changes in this PR :)

The same comment stills: we don't want this to be a globally defined behaviour, but a per-subscription behaviour.

In fact, it seems that you have implemented all the CRUD for this attribute in the subscription lifecycle, so probably this is a remain of the old PR that can be safely removed.

src/lib/mongoBackend/MongoCommonUpdate.cpp
```diff
@@ -554,6 +554,15 @@ static bool mergeAttrInfo
...
   }
 
+  if (!attrValueChanges(attr, caP, forcedUpdate, apiVersion) && !equalMetadata(md, mdNew))
+  {
+    onlyMetadataChange = true;
+  }
+  else
+  {
+    onlyMetadataChange = false;
+  }
```
fgalan, line 564: It seems you are using the global variable `onlyMetadataChange` that should be removed (see my previous comment), so I'm afraid this strategy probably is not valid.

Maybe you could "mark" showhow that the entity has modified its metadata and its value independently (note that currently the returned `actualUpdate` doesn't allow to distinguish that) and use that information in addTriggeredSubscriptions_noCache() and addTriggeredSubscriptions_withCache(), in which moment you have the `onlyMetadataChange ` field coming from the subscription and you can make a decision upon sending notification or not.

If you find that difficult I could try to provide some help (the most "heavy" part of the work, about CRUD management of the new attribute and changes in the .test files is already done).

src/lib/common/globals.cpp
chetan-NEC: It is little bit painful, but we can see there are several global variables already present can we go with it. does it make any impact on the current application.

fgalan: Painful or not :) note that different subscription may have different settings. Subscription A may use onlyMetadataChange true and subscription B may use onlyMetadataChange false. If Orion process an update entity request triggering A and an update request triggering B at the same time then you are in problems if your solution is based in a global variable.

If this is complex and time consuming for you, let's do what I proposed in the other comment:

> If you find that difficult I could try to provide some help (the most "heavy" part of the work, about CRUD management of the new attribute, the new .test and the changes in existing .test files is already done).

Please tell me if you agree with that, so I can merge this PR in a pre-landing branch and you can continue working on Orion in a next issue (specially the ones labeled with [help wanted](https://github.com/telefonicaid/fiware-orion/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)

chetan-NEC: Thanks for your response.
Yes, I agree with you in case of per subscriptions global variable will definitely create problem. but I was thinking this feature as a static configuration for all subscriptions. The understanding was not very clear to me earlier but now it's pretty clear.
I will try to continue on this issue only.

fgalan: This is getting somehow urgent for us (we need to use it in some project) so I will take care of it.

We are going to merge this PR in a pre-landing branch and I'll continue with the final fixes. Thanks @chetan-NEC for your work! You can consider it as finished :)

CC: @hervaste

fgalan (approved):
LGTM (work will continue in a next PR)

src/lib/common/globals.cpp
chetan-NEC: Thanks for your response.

I'll start working on the next issue now.

PR #4302, opened 2023-03-28
FIX upgrade ubuntu image used in GitActions

fgalan:
Due to:

imagen

### DIFF
--- a/scripts/check_files_compliance.py
+++ b/scripts/check_files_compliance.py
@@ -175,13 +175,13 @@ def supported_extension(root, file):
         return True
 
     filename = os.path.join(root, file)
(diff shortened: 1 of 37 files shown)


scripts/check_files_compliance.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
fgalan, line 1: This script is run on the host itself (i.e. not in a docker). It seems ubuntu-18.04 uses Python 2 and ubuntu-22.04 uses Python 3, so we need to do some minor adaptations to the script.

NTC (informative)

mapedraza (approved):
LGTM

PR #4305, opened 2023-03-29
[TMP] Check valgrind after change to Ubuntu 22.04 in GitActions

fgalan:
Check valgrind after change to Ubuntu 22.04 in GitActions

Once we check everyhing ok, this PR can be closed without merge


fgalan:
Valgrind test has gone ok. Functional test have been also launched (and they shouldn't) but it is fine.

Thus, this PR is going to be closed.

PR #4307, opened 2023-04-04
ADD complementary case 3727 test

fgalan:
This test is based with the ones included by @chetan-NEC in PR https://github.com/telefonicaid/fiware-orion/pull/4300. It covers the current behaviour of CB (as in master) so we can ensure the modifications in PR #4300 doesn't break backward compatibility.

It comes from: https://github.com/telefonicaid/fiware-orion/pull/4300#discussion_r1154138624


mapedraza (approved):
LGTM

PR #4338, opened 2023-05-25
(JP) Update doc about operator usage in create entity cases

fisuda:
This PR updates Japanese documentation about operator usage in create entity cases. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4347, opened 2023-05-30
[TMP] Check valgrind before merging PR #4345

fgalan:
PR #4345 removes (by disabling) many .test using NGSIv1 operations. Although it is a deprecated API, these .test involve the backend logic, so it is wise to do a last valgrind pass before the removal.

Once we check everyhing ok, this PR can be closed without merge

### DIFF
--- a/test/functionalTest/README.md
+++ b/test/functionalTest/README.md
@@ -1,4 +1,4 @@
-## How to run functional tests
+## How to run functional tests 
 
 ### Prerequisites
 


fgalan:
All tests valgrind-ok :)

fgalan:
Moreover, this PR also covers the valgrind regression for Orion 3.9.0, as that version is going to be release now from the contents in master branch.

PR #4355, opened 2023-06-06
FIX count with geo near is not working

fgalan:
The first commit in this PR raises the problem, in the form of .test.

As can be seen, the test/functionalTest/cases/1038_ngsiv2_geo_location_entities/geoquery_circle_geopoint_count.test test is not working. A `$geoNear, $near, and $nearSphere are not allowed in this contex` is raised in the collectionRangedQuery() function, in particular when mongoc_collection_count_documents is called

Curiosly, other geo-queries works with problem with count. Note the test/functionalTest/cases/1038_ngsiv2_geo_location_entities/geoquery_polygon_geopoint_count.test test is working.

Related: https://stackoverflow.com/questions/76409607/geonear-near-and-nearsphere-are-not-allowed-in-this-context-error-with-mong

Pending:

- [x] actual fix for the problem
- [x] CNR

### DIFF
--- a/src/lib/mongoBackend/MongoGlobal.cpp
+++ b/src/lib/mongoBackend/MongoGlobal.cpp
@@ -964,7 +964,8 @@ bool processAreaScopeV2(const Scope* scoP, orion::BSONObjBuilder* queryP, bool a
   orion::BSONObjBuilder bobArea;
   if (scoP->georel.type == "near")
   {
-    if (avoidNearUsage)
+    // FIXME PR: clean this, avoidNearUsage no longer needed
+    //if (avoidNearUsage)
     {
       // $near operation is problematic in countDocument()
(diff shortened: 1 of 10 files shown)


src/lib/mongoBackend/MongoGlobal.cpp
```diff
@@ -964,119 +974,115 @@ bool processAreaScopeV2(const Scope* scoP, orion::BSONObjBuilder* queryP, bool a
   orion::BSONObjBuilder bobArea;
   if (scoP->georel.type == "near")
   {
```
fgalan, line 976: For this part, better to use "hide whitespace" option in GitHub

NTC (just a hint)

mapedraza (approved):
LGTM

PR #4364, opened 2023-06-08
ADD log deprecate configuration to management API

fgalan:

PR #4368, opened 2023-06-08
(JP) ADD doc about log deprecate configuration to management API (#4364)

fisuda:
This PR adds Japanese documentation about log deprecate configuration to management API. It would be great if you could review this PR. Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4367, opened 2023-06-08
FIX deprecate all geo types except geo:json

fgalan:

PR #4369, opened 2023-06-09
(JP) Update doc about deprecate all geo types except geo:json (#4367)

fisuda:
This PR updates Japanese documentation about deprecate all geo types except geo:json. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4383, opened 2023-07-18
Update README.md

oriolcanadesin2:
Fix Accumulator Server Dockerfile configuration


fgalan:
Thanks for the feedback!

It seems your suggested Dockerfile improves the original one, so instead of adding a "Prorbably error" I'd suggest to directly modify the Dockerfile at L51-L60 to introduce your changes, please.

fgalan (approved):
LGTM. Thanks!

PR #4386, opened 2023-07-24
Replace broken webhook with a simple bash script.

jason-fox:
As discussed within the TSC,  FIWARE Foundation need to be able to create clones of the container image on both Docker Hub and quay.io - this PR allows us to continue to track your latest container images, a functionality which has been removed from the free tier on Docker Hub.


jason-fox:
PR moved to FIWARE Infrastructure

PR #4389, opened 2023-08-09
ADD servicePath builtin attr to orion-api.md

fgalan:
Documentation corresponding to the PR #4329 

CC: @chetan-NEC

PR #4390, opened 2023-08-09
(JP) ADD servicePath builtin attr to orion-api.md (#4389)

fisuda:
This PR adds servicePath builtin attr to Japanese orion-api.md. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4412, opened 2023-08-30
ADD MQTT retain feature

fgalan:
Issue #4388

- [x] CRUD of the new field
- [x] Effectively use the new field to publish MQTT notifications
- [x] .test covering the CRUD functionality
- [x] doc
- [x] CNR
- [x] Align existing .test with the new GET subscription(s) output - 19f3d2ccfd6a0533f10290f62c93e4d14163f69d + 9789d5bbf
- [x] Effectively test the new functionality (.test ~or manual testing?~) - 19f3d2ccfd6a0533f10290f62c93e4d14163f69d

PR #4413, opened 2023-08-31
(JP) Add doc about MQTT retain feature (#4412)

fisuda:
This PR add Japanese documentation about MQTT retain feature. It would great if you could review the PR. Thanks.


fgalan (approved):
LGTM

Thanks for the typo fix in the English doc!

PR #4415, opened 2023-09-01
FIX change mongo driver to 1.24.3 in build-dep.sh (ci image)

fgalan:
This PR is to build de ci image so we can regression if Orion will work with the new version of the driver in Debian 11. We will manually launch the CI image build in dockerhub, then re-launch tests in ~master~ this branch (code is the same as in master) branch.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -4,3 +4,4 @@
 - Fix: improve error traces (#4387)
 - Add: CLI parameter -dbUri / env var ORION_MONGO_URI (#3794)
 - Fix: improve logs in MongoDB query logic
+- Hardening: upgrade libmongoc dependency from 1.23.1 to 1.24.3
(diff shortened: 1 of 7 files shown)


fgalan:
Build done:

imagen

Relaunching tests...

fgalan:
This PR is motivated due to we are preparing the migration to Debian 12 but libmogoc 1.23.1 doesn't work in that distro. This PR demostrates that 1.24.3 works in Debian 11 (and I have checked locally that it also works for Debian 12).

mapedraza (approved):
LGTM

fgalan:
@fisuda I haven't updated the sections "Ubuntu 20.04 LTS" that you authored (in both the English and Japanese versions) as I don't have that operating system at hand so I cannot assure that the mongoc driver version update would work.

Could you check and in positive case provide a PR with the upgrade, please? Thanks in advance!

PR #4422, opened 2023-09-10
ADD IMAGE_NAME arg to Dockerfile

fisuda:
This PR adds the `IMAGE_NAME` argument to Dockerfile. For example, Orion with Ubuntu 22.04 as a base image can be build as shown:

```
docker build -t orion-ubuntu22.04 --build-arg IMAGE_NAME=ubuntu --build-arg IMAGE_TAG=22.04 --build-arg CLEAN_DEV_TOOLS=0
```

The build logs:

 - [build_on_ubuntu_22_04_x86_64.log](https://github.com/fisuda/report/blob/master/orion/20230910_ubuntu22.04_in_docker/build_on_ubuntu_22_04_x86_64.log)
 - [build_on_ubuntu_22_04_arrch64.log](https://github.com/fisuda/report/blob/master/orion/20230910_ubuntu22.04_in_docker/build_on_ubuntu_22_04_arrch64.log)

Result:

- Orion on Ubuntu 22.04 (x86_64)

```
$ curl http://localhost:1026/version
{
"orion" : {
  "version" : "3.10.0-next",
  "uptime" : "0 d, 0 h, 0 m, 13 s",
  "git_hash" : "7b6275bb053701af043ed05536ce8facc46f0efc",
  "compile_time" : "Sun Sep 10 02:01:01 UTC 2023",
  "compiled_by" : "root",
  "compiled_in" : "buildkitsandbox",
  "release_date" : "Sun Sep 10 02:01:01 UTC 2023",
  "machine" : "x86_64",
  "doc" : "https://fiware-orion.rtfd.io/",
  "libversions": {
     "boost": "1_74",
     "libcurl": "libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.13 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.16",
     "libmosquitto": "2.0.15",
     "libmicrohttpd": "0.9.76",
     "openssl": "3.0.2",
     "rapidjson": "1.1.0",
     "mongoc": "1.24.3",
     "bson": "1.24.3"
  }
}
}
```

- Orion on Ubuntu 22.04 (aarch64)

```
{
"orion" : {
  "version" : "3.10.0-next",
  "uptime" : "0 d, 0 h, 0 m, 10 s",
[...]

PR #4423, opened 2023-09-10
REMOVE Ubuntu 20.04 LTS section

fisuda:
This PR removes "Ubuntu 20.04 LTS" section from admin/build_source.md.

When I added Ubuntu 18.04 and Ubuntu 20.04 sections, Orion Context Broker reference distribution was CentOS 7. There are differences between these distributions, so the procedure building Orion for Ubuntu was necessary. After that, the reference distribution was changed to Deiban. The current one is Debian 12.

Currently, the latest version of Ubuntu LTS (Long Term Support) is 22.04. So, I think that Ubuntu 20.04 LTS section should be update to Ubuntu 22.04 LTS section. But, the base distribution of Ubuntu 22.04 LTS is Debian 12. As I show the PR #4422, Orion on Ubuntu 22.04 can be built by the same procedure as Debian 12. Therefore, I thnink that the section to build Orion on Ubuntu is unnecessary.

 It would be great if you could review the PR.

Thanks.


doc/manuals/admin/build_source.md
```diff
@@ -1,6 +1,6 @@
 # Building from sources
 
-Orion Context Broker reference distribution is Debian 12. This doesn't mean that the broker cannot be built in other distributions (actually, it can). This section also includes indications on how to build in other distributions, just in the case it may help people that don't use Debian. However, note that the only "officially supported" procedure is the one for Debian 12; the others are provided "as is" and can get obsolete from time to time.
```
fgalan, line 3: Although after the removal of the Ubuntu section the only procedure would be the one about Debian 12, I'd suggest to keep this fragment just in case in the future another non official procedures would be included, with the following change:

> This section also includes indications...

to

> This section may include indications...

fisuda: I updated it.

PR #4431, opened 2023-09-27
[Doc] Add geofencing subscription example

mapedraza:
Orion documentation lacks a geofencing subscription example, and it is not pretty straightforward from documentation, since it is almost based on queries, not on subscriptions. 

Maybe extending this table field here could help:

https://github.com/telefonicaid/fiware-orion/blob/442541c990f8f50165207cfc88c7bb8afae28300/doc/manuals/orion-api.md?plain=1#L3654

@fgalan what do you think?

PR #4443, opened 2023-11-03
(JP) [Doc] Add geofencing subscription example (#4431)

fisuda:
This PR adds Japanese documentation about geofencing subscription example. It would be great if you could review this PR. Thanks.


fgalan (approved):
LGTM. Thanks!

Issue #4454, opened 2023-12-14
False positive in log deprecation logic


fgalan:
> time=2023-12-14T11:30:12.472Z | lvl=WARN | corr=5d9acef7-5401-4754-b43c-576183ce9f5a | trans=1698136203-599-00101812278 | from=A.B.C.D | srv=aaaa | subsrv=/bbb| comp=Orion | op=logTracing.cpp[114]:logInfoRequestWithoutPayload | msg=Deprecated NGSIv1 request received: GET /v2/entities/glory_ev1?type=ElectroValve, response code: 200

This is probably a false positive.

fgalan:
PR https://github.com/telefonicaid/fiware-orion/pull/4455

PR #4455, opened 2023-12-19
FIX false positive in log deprecation logic when entity name (or other literal) includes the token v1

fgalan:
Issue #4454


AlvaroVega (approved):
LGTM

PR #4463, opened 2023-12-22
FIX forwarding logc

fgalan:
Fix commit 18136f6de321567f640bc7cedfcd99de9a32b5f1 is aligned with current (wrong) behaviour:

* Overloging RuntimeError
* The update is not actually forwared (step 06 shows that "foobar" doesn't reach CPr)

Commit 0512f650719822452d831b0cdfd264b564f3701d shows how both things are fixed.

### DIFF
--- a/src/lib/ngsi10/UpdateContextRequest.cpp
+++ b/src/lib/ngsi10/UpdateContextRequest.cpp
@@ -332,11 +332,13 @@ ContextAttribute* UpdateContextRequest::attributeLookup(Entity* eP, const std::s
   {
     Entity* enP = entityVector[ceIx];
 
-    if ((enP->id != eP->id) || (enP->type != eP->type))
+    // empty type in request (enP) is always a match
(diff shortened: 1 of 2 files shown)


src/lib/ngsi10/UpdateContextRequest.cpp
```diff
@@ -332,16 +332,17 @@ ContextAttribute* UpdateContextRequest::attributeLookup(Entity* eP, const std::s
...
     Entity* enP = entityVector[ceIx];
 
-    if ((enP->id != eP->id) || (enP->type != eP->type))
+    // empty type in request (enP) is always a match
+    if ((enP->id != eP->id) || ((enP->type != "") && (enP->type != eP->type)))
     {
       continue;
     }
 
-    Entity* eP = entityVector[ceIx];
```
fgalan, line 340: It was working without this change, but it is not wise to re-use as local variable the same name used in a function argument.

NCT (informative)

Issue #4451, opened 2023-12-12
dateExpires shouldn't be include in GET /v2/types?


fgalan:
According to documentation (https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#the-dateexpires-attribute) `dateExpires` should be treated as a built-in attribute that must not be included expect if expectedly asked for.

It seems that is working that way in `GET /v2/entities...` operations, but in `GET /v2/types` is included. The solution is applying the same treatment that for other built-ins that are working correctly (e.g. `dateModified`).

Request used for testing:

```
https://localhost/cb/v2/types?limit=1000&offset=0&options=noAttrDetail
```

Pradyumn-NEC:
Hi @fgalan, I'm interested in working on this issue!
 
The documentation states that `dateExpires` should be treated as a built-in attribute and should not be included by default, except if explicitly asked for.
 
In the `GET /v2/entities` operations, this behavior seems to be working correctly.
 
However, in the `GET /v2/types` operation, `dateExpires` is included in the response, and this behavior is inconsistent with the documentation.

Please confirm my understanding.

fgalan:
First of all, thanks for your willingness to contribute to Orion Context Broker!

This issue is somehow urgent (we want to fix it before closing next release, that will be done at the beginning of new year) so I was thinking on taking care of it myself.

Except if you tell me that you have the solution to this almost ready :) I'd suggest you to pick another issue, please.

fgalan:
PR https://github.com/telefonicaid/fiware-orion/pull/4467

PR #4467, opened 2024-01-11
FIX do not show dateExpires built-in attribute in GET /v2/types and GET /v2/types/{type} operations

fgalan:
Issue #4451


mapedraza (approved):
LGTM

PR #4468, opened 2024-01-11
FIX NGSI patching macro subscription covering completely the string always cast to string for id/type

fgalan:
Issue #4462

PR #4472, opened 2024-01-14
(JP) Add doc about NGSI patching macro subscription (#4468)

fisuda:
This PR adds Japanese documentation about NGSI patching macro subscription. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM Thanks!

PR #4471, opened 2024-01-12
FIX add test combining onlyChangedAttrs true and alterationType/dateCreated/dateModified

fgalan:
This PR includes a .test showing the combined behaviour of onlyChangedAttrs true and alterationType notification.

At the present moment, if onlyChangedAttrs: true is used, alterationType is not included.

This PR can resolve the WIP in two ways:

1. Keeping the current behaviour, in which case:
   - Keep the .test as it is now
   - Document the behaviour in orion-api.md
2. Change CB behaviour, in which case
   - Change code to change the behaviour (including CNR entry)
   - Change the .test to implement the align with new behaviour
   - Document the behaviour in orion-api.md

Feedback welcome

PR #4484, opened 2024-01-16
(JP) ADD documentation about builtin attributes included in notifications (#4471)

fisuda:
This PR adds Japanese documentation about builtin attributes included in notifications. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM. Thanks!

Issue #4496, opened 2024-01-30
Conflict between -dbURI and Default -dbTimeout Causes Startup Failure


goten002:
**Bug description**
Encountered a startup failure with Fiware Orion version 3.11.0 due to a conflict between the -dbURI parameter and the default value of -dbTimeout. The issue arises when the -dbURI parameter is used without explicitly setting -dbTimeout to zero.

- Orion version: `3.11.0`
- MongoDB version `4.4`
- CLI parameters: `-dbURI mongodb://mongodb:27017  -logLevel DEBUG`
- Docker Image used: `fiware/orion:3.11.0`

**How to reproduce it**
1. Configure Orion with the `-dbURI` parameter without specifying `-dbTimeout`.
2. Attempt to run the Orion container.
3. Observe the startup failure with the error message indicating a conflict between `-dbURI` and other database parameters.
```text
Invalid Command Line Options: -dbURI cannot be combined with -dbhost, -rplSet, -dbTimeout, -dbuser, -dbAuthMech, -dbAuthDb, -dbSSL and -dbDisableRetryWrites.
```
**Expected Behavior**
Orion should start successfully using the `-dbURI` parameter without needing to explicitly set `-dbTimeout` to zero, as `-dbTimeout` has a default value.

**Investigation and Analysis**
On further investigation, it was found that the `composeMongoUri` function in `mongoConnectionPool.cpp` checks if `timeout > 0` to return an error. The default value of `dbTimeout` is 10000, as seen in `contextBroker.cpp`, which seems to be causing the issue.

**Suggested Fix**
It appears that the check for `timeout > 0` in the condition within `composeMongoUri` is unnecessary when `-dbURI` is used. A proposed solution would be to modify the conditional check to exclude `timeout` when `dbURI` is provided. This would prevent the conflict when the default `dbTimeout` value is used with `-dbURI`.

**Code Reference**
The relevant code is located at:
- [mongoConnectionPool.cpp line 337](https://github.com/telefonicaid/fiware-orion/blob/4e4d96fcef96e50327a802da011d6b7da110926d/src/lib/mongoDriver/mongoConnectionPool.cpp#L337)
- [contextBroker.cpp line 312](https://github.com/telefonicaid/fiware-orion/blob/4e4d96fcef96e50327a802da011d6b7da110926d/src/app/contextBroker/contextBroker.cpp#L312)

**Additional Context**
[...]

PR #4497, opened 2024-01-30
FIX -dbTimeout conflict with -dbURI (#4496)

goten002:
This pull request resolves the configuration conflict identified in issue [#4496](https://github.com/telefonicaid/fiware-orion/issues/4496), where using the -dbURI parameter along with the default dbTimeout value caused startup failures in Orion.

**Key Changes**
- Modified the composeMongoUri function to handle a default dbTimeout value of -1, enabling it to work seamlessly with the -dbURI parameter.
- Implemented a condition to set dbTimeout to 10000 if it is at its default value of -1. This maintains the standard behavior when -dbURI is not being used, thus ensuring backward compatibility.
- Adjusted the error validation logic in composeMongoUri to prevent false triggering of configuration conflicts when dbTimeout is in its default state.

These changes improve the usability of the Orion Context Broker, particularly for configurations utilizing the -dbURI parameter. It addresses the issue where users transitioning to -dbURI faced unnecessary hurdles due to the pre-existing default dbTimeout settings.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,1 +1,1 @@
-- Fix: Changed the default value of `-dbTimeout` to 0 to resolve conflict with `-dbURI` (#4496)
+- Fix: changed the default value of `-dbTimeout` to 0 to resolve conflict with `-dbURI` (#4496) 
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: Changed the default value of `-dbTimeout` to 0 to resolve conflict with `-dbURI` (#4496)
--- a/src/lib/mongoDriver/mongoConnectionPool.cpp
+++ b/src/lib/mongoDriver/mongoConnectionPool.cpp
@@ -334,7 +334,7 @@ static std::string composeMongoUri
 
   if (strlen(dbURI) != 0)
   {
-    if (strlen(username) != 0 || strlen(authDb) != 0 || strlen(rplSet) != 0 || strlen(mechanism) != 0 || dbSSL || dbDisableRetryWrites || timeout > 0)
+    if (strlen(username) != 0 || strlen(authDb) != 0 || strlen(rplSet) != 0 || strlen(mechanism) != 0 || dbSSL || dbDisableRetryWrites || timeout > -1)
     {
         LM_X(1, ("Invalid Command Line Options: -dbURI cannot be combined with -dbhost, -rplSet, -dbTimeout, -dbuser, -dbAuthMech, -dbAuthDb, -dbSSL and -dbDisableRetryWrites"));
     }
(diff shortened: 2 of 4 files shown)


fgalan:
Thanks for the PR!

However, it seems there is a build problem. Please check the result of the GitActions.

fgalan:
> Thanks for the PR!
> 
> However, it seems there is a build problem. Please check the result of the GitActions.

Maybe the `}` areound L409 has been incorrectly removed?

src/lib/mongoDriver/mongoConnectionPool.cpp
```diff
@@ -399,14 +399,20 @@ static std::string composeMongoUri
...
-      uri += optionPrefix + "connectTimeoutMS=" + buf;
-      optionPrefix = "&";
+        // If timeout is negative (e.g., -1 indicating it's unset), set it to the default value of 10000
+        if (timeout < 0)
+        {
+            timeout = 10000;
+        }
+
+        char buf[STRING_SIZE_FOR_LONG];
+        i2s(timeout, buf, sizeof(buf));
```
fgalan, line 412: Modifying an input parameter in the function use not to be a good idea... I'd suggest al anternative that doesn't modifies it:

```suggestion
        char buf[STRING_SIZE_FOR_LONG];
        if (timeout < 0)
        {
          i2s(10000, buf, sizeof(buf));
        }
        else
        {
          i2s(timeout, buf, sizeof(buf));
        }

```

fgalan:
The GitActions are still failing, so I have compile your branch and run Orion this way:

```
$ contextBroker  -fg
```

Getting this error:

```
contextBroker: value 18446744073709551615 for option '-dbTimeout' not within limits
               18446744073709551615 <= 18446744073709551615 <= 4294967295 (default value)
```

which makes sense, given that `-dbTimeout` is defined as a `PaULong` which stands for *unsigned* long, so trying to use a negative value is problematic....

Let me propose some alternatives in the parent issue (#4496)

fgalan:
With regards to the fail in 3658_env_vars/env_vars.test detected by GitActions:

```
-----  broker env vars set on command line  -----
(3658_env_vars/env_vars.test) output not as expected
VALIDATION ERROR: input line:
                                  [option '-dbTimeout' <timeout in milliseconds for connections to the replica set]   ORION_MONGO_TIMEOUT                0 <= 0 /0/ <= 4294967295
does not match ref line:
                                  [option '-dbTimeout' <timeout in milliseconds for connections to the replica set]   ORION_MONGO_TIMEOUT                0 <= 10000 /10000/ <= 4294967295
```

I think is easy to fix: just change the line within `--REGEXPECT--` section in env_vars.test, replacing `10000 /10000` by `0 /0`.

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+- Fix: Changed the default value of `-dbTimeout` to 0 to resolve conflict with `-dbURI` (#4496)
```
```suggestion
- Fix: changed the default value of `-dbTimeout` to 0 to resolve conflict with `-dbURI` (#4496)
```

Just text style :)

fgalan (approved):
The PR now looks perfect :)

LGTM

PR #4498, opened 2024-01-31
FIX simplify init methods in managers

fgalan:
Simplifications:

* Use LM_X isntead of LM_E if semaphore initialization fails. Semaphore initialization takes place in Orion startup and it's so severe that Orion should refuse to start in that case
* Simplification of the return values in the init methos for existing managers.

This is mainly a minor hardening fix, so I think it doesn't worth to be mentioned in CNR file.

### DIFF
--- a/src/lib/alarmMgr/AlarmManager.cpp
+++ b/src/lib/alarmMgr/AlarmManager.cpp
@@ -79,7 +79,7 @@ AlarmManager::AlarmManager()
 *
 * AlarmManager::init - 
 */
-int AlarmManager::init(bool logAlreadyRaisedAlarms)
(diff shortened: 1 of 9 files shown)


fgalan:
Note that the call to the init methods doesn't use the return value, so it's simpler to use void instead of int:

* From main() in contextBroker.cpp:

```
  alarmMgr.init(relogAlarms);
  mqttMgr.init(mqttTimeout);
  ...
  metricsMgr.init(!disableMetrics, statSemWait);
```

* From in globas.cpp

```
  semInit(reqPolicy, _semWaitStatistics);
```

mapedraza (approved):
LGTM

Issue #4499, opened 2024-02-01
Return 400 Bad Request when subject.entities exists but it's an empty array


fgalan:
Orion allows to create subscriptions with empty `subject.entities` array:

```
curl -v localhost:1026/v2/subscriptions -s -S -H 'Content-Type: application/json' -d @- <<EOF
{
  "description": "A subscription with empty entities array",
  "subject": {
    "entities": [
    ],
    "condition": {
      "attrs": [
        "pressure"
      ]
    }
  },
  "notification": {
    "http": {
      "url": "http://localhost:1028/accumulate"
    },
    "attrs": [
      "temperature"
    ]
  }
}
EOF
```

This cause ERROR log traces when processing that subscription, which is somehow annoying.

```
  if (cSubP->entityIdInfos.size() == 0)
  {
    LM_E(("ERROR (no patterned entityId) - cleaning up"));
    subCacheItemDestroy(cSubP);
    delete cSubP;
    return -3;
  }
```

Note that the omission of `subject.entities` field (which is semantically the same) do causes a 400 Bad Request error:

```
$ curl -v localhost:1026/v2/subscriptions -s -S -H 'Content-Type: application/json' -d @- <<EOF
{
  "description": "A subscription without entities field",
  "subject": {
    "condition": {
      "attrs": [
        "pressure"
      ]
    }
  },
  "notification": {
    "http": {
      "url": "http://localhost:1028/accumulate"
    },
    "attrs": [
      "temperature"
    ]
  }
}
EOF
*   Trying 127.0.0.1:1026...
* Connected to localhost (127.0.0.1) port 1026 (#0)
> POST /v2/subscriptions HTTP/1.1
> Host: localhost:1026
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 267
> 
< HTTP/1.1 400 Bad Request
< Date: Thu, 01 Feb 2024 07:57:29 GMT
< Fiware-Correlator: 8bf6e8a2-c0d7-11ee-8340-080027cd35f1
< Content-Type: application/json
< Content-Length: 68
< 
* Connection #0 to host localhost left intact
{"error":"BadRequest","description":"no subject entities specified"}fermin@bodoque:~/src/fiware-orion
```

We should raise a similar error when `subject.entities` exists but it's an empty array.

fgalan:
This issue allows subscription which existence raises this error log

```
[...]

PR #4502, opened 2024-02-06
FIX return 400 Bad Request when subject.entities exists but it is an empty array

fgalan:
Issue https://github.com/telefonicaid/fiware-orion/issues/4499


mapedraza (approved):
LGTM

Issue #4495, opened 2024-01-29
Service path levels with 0 length should not be allowed


fgalan:
From https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#entity-service-path

> 50 maximum characters in each level (1 char is minimum), only alphanumeric and underscore allowed

However, Orion allows creating service path with 0 length levels, eg:

```
$ curl localhost:1026/v2/entities -vvvv -s -S -H 'Content-Type: application/json' -H 'fiware-servicepath: //detecciondeplagas' -d @- <<EOF
{
  "id": "Room2",
  "type": "Room",
  "temperature": {
    "value": 23,
    "type": "Float"
  },
  "pressure": {
    "value": 720,
    "type": "Integer"
  }
}
EOF

*   Trying 127.0.0.1:1026...
* Connected to localhost (127.0.0.1) port 1026 (#0)
> POST /v2/entities HTTP/1.1
> Host: localhost:1026
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Type: application/json
> fiware-servicepath: //detecciondeplagas
> Content-Length: 148
> 
< HTTP/1.1 201 Created
< Date: Mon, 29 Jan 2024 15:15:25 GMT
< Fiware-Correlator: 3a970da4-beb9-11ee-85a1-080027cd35f1
< Location: /v2/entities/Room2?type=Room
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```

ArqamFarooqui110719:
PR #4503 
This PR fixes servicepath levels with zero length, i.e. available in initial part (1st level) of servicepath only (e.g. `fiware-servicepath: //det`), because servicepath levels with zero length, i.e. available in 2nd level or further level of servicepath (e.g. `fiware-servicepath: /a///det`), are already restricted by following code in `servicePathCheck()` function `src/lib/rest/rest.cpp` file:
<pre>
    if (compV[ix].c_str()[0] == 0)
    {
      OrionError oe(SccBadRequest, "empty component in ServicePath");
      ciP->answer = oe.setStatusCodeAndSmartRender(ciP->apiVersion, &(ciP->httpStatusCode));
      return 3;
    }
</pre>

PR #4503, opened 2024-02-06
Fix 0 length servicepath levels

ArqamFarooqui110719:
Fix Service path levels with 0 length should not be allowed, issue #4495

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: Service path levels with 0 length should not be allowed (#4495)
--- a/src/lib/rest/rest.cpp
+++ b/src/lib/rest/rest.cpp
@@ -704,6 +704,13 @@ int servicePathCheck(ConnectionInfo* ciP, const char* servicePath)
     return 1;
   }
 
+  if (servicePath[1] == '/')
+  {
+    OrionError oe(SccBadRequest, "Zero length service path levels are not allowed, minimum 1 charater is required");
+    ciP->answer = oe.setStatusCodeAndSmartRender(ciP->apiVersion, &(ciP->httpStatusCode));
+    return 1;
+  }
+
   if (ciP->servicePathV.size() > 1)
   {
     if (ciP->verb == PATCH)
--- a/test/functionalTest/cases/4495_servicepath_levels_with_zero_length/4495_servicepath_levels_with_zero_length.test
+++ b/test/functionalTest/cases/4495_servicepath_levels_with_zero_length/4495_servicepath_levels_with_zero_length.test
@@ -1,0 +1,206 @@
+# Copyright 2024 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+Servicapath with zero length levels are not allowed
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create entity E1/T1/A1=1, with servicepath '/Level_1' (success case)
+# 02. GET /v2/entities/E1, with servicepath '/Level_1' to see E1/T1/A1=1
+# 03. Create entity E2/T2/A2=2, with servicepath '/Level_1/Level_2' (success case)
+# 04. GET /v2/entities/E2, with servicepath '/Level_1/Level_2' to see E2/T2/A2=2
+# 05. Create entity E3/T3/A3=3, with servicepath '//Level_1' (SP level with 0 length), see error
+# 06. Create entity E3/T3/A3=3, with servicepath '///Level_1///Level_2//' (SP level with 0 length), see error
+#
+
+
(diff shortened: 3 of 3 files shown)


src/lib/rest/rest.cpp
```diff
@@ -704,6 +704,13 @@ int servicePathCheck(ConnectionInfo* ciP, const char* servicePath)
     return 1;
   }
 
+  if (servicePath[1] == '/')
+  {
+    OrionError oe(SccBadRequest, "Zero length service path levels are not allowed, minimum 1 charater is required");
+    ciP->answer = oe.setStatusCodeAndSmartRender(ciP->apiVersion, &(ciP->httpStatusCode));
+    return 1;
+  }
```
fgalan, line 712: Not sure if this is going to work if 0-length service path level is in the middle of the service path strings...

Maybe the check should be done within the `for (int ix = 0; ix < components; ++ix)` loop? Not sure...

fgalan:
Thanks for your contribution! Please find my comment in the PR.

src/lib/rest/rest.cpp
fgalan: I have just seen you comment here https://github.com/telefonicaid/fiware-orion/issues/4495#issuecomment-1929216781, so it seems it will be already covered :)

fgalan: However, for homogeneity, I'd suggest to use the same error message in both cases:

```
"empty component in ServicePath"
```

test/functionalTest/cases/4495_servicepath_levels_with_zero_length/4495_servicepath_levels_with_zero_length.test
```diff
@@ -0,0 +1,206 @@
...
+
+--SHELL--
+
+#
+# 01. Create entity E1/T1/A1=1, with servicepath '/Level_1' (success case)
+# 02. GET /v2/entities/E1, with servicepath '/Level_1' to see E1/T1/A1=1
+# 03. Create entity E2/T2/A2=2, with servicepath '/Level_1/Level_2' (success case)
+# 04. GET /v2/entities/E2, with servicepath '/Level_1/Level_2' to see E2/T2/A2=2
+# 05. Create entity E3/T3/A3=3, with servicepath '//Level_1' (SP level with 0 length), see error
+# 06. Create entity E3/T3/A3=3, with servicepath '///Level_1///Level_2//' (SP level with 0 length), see error
```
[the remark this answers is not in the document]
ArqamFarooqui110719: Included in f67c041

src/lib/rest/rest.cpp
ArqamFarooqui110719: fixed in f67c041

test/functionalTest/cases/4495_servicepath_levels_with_zero_length/4495_servicepath_levels_with_zero_length.test
```diff
@@ -0,0 +1,236 @@
...
+# 07. Create entity E3/T3/A3=3, with servicepath '/Level_1///Level_2' (SP level with 0 length), see error
```
fgalan, line 39: Another interesting case to check could be:

```
# 0x. Create entity E3/T3/A3=3, with servicepath '/Level_1/Level_2//' (SP level with 0 length), see error
```

Could you include it, please?

ArqamFarooqui110719: For above case ( i.e. servicepath = `/Level_1/Level_2//`), request will be executed successfully. Because before `servicePathCheck()` function call (Where the service paths are being checked), there is another function i.e. `removeTrailingSlash()` that is removing all the slash (`/`), present at the end of servicePath.
Please see below existing code snippet for the same:

rest_servicePathSplit_function2

rest_removeTrailingSlash_function2


See below API execution result:
API_execution_result


I have applied some custom logs to see the value of servicePath, Please find below logs for reference:

orion_API_execution_log

So I think in that case, we don't need to include the above case in the FT. Please provide your opinion on this.

fgalan: Thank you very much for your detailed analysis!

It think this slash removal logic is inspired in web navigation (e.g. your browser doesn't distinguish between https://www.google.com/, https://www.google.com or https://www.google.com///////).

However, in the case of service paths I think is a bit weird removing *all slashes*.

I mean, removing *one slash* is ok, so `/Level1/Level2/` is equivalent to `/Level1/Level2`. However, if the user tries to use `/Level1/Level2//` probably he/she is doing something wrong that should be warned with an error response.

What do you think? Could the code be modified to implement this?

ArqamFarooqui110719: Yes, I also think, It will be good to remove only 1 slash from servicePath, in case of multiple slash, an error message should be returned.

ArqamFarooqui110719: updated code for the same.
Fixed in cc1e22a

CHANGES_NEXT_RELEASE
```diff
@@ -1,3 +1,4 @@
+- Fix: Service path levels with 0 length should not be allowed (#4495)
```
```suggestion
- Fix: service path levels with 0 length should not be allowed (#4495)
```

(Text style)

test/functionalTest/cases/4495_servicepath_levels_with_zero_length/4495_servicepath_levels_with_zero_length.test
```diff
@@ -0,0 +1,266 @@
...
+# 08. Create entity E3/T3/A3=3, with servicepath '/Level_1/Level_2//' (SP level with 0 length), see error
+#
```
fgalan, line 41: To complete the different case, I'd suggest to finally add these ones:

```
# 0x. Create entity E4/T4/A4=4, with servicepath '/Level_1/Level_2/' (success case)
# 0x. GET /v2/entities/E4, with servicepath '/Level_1/Level_2' (without slash) to see E4/T4/A4=4
```

ArqamFarooqui110719: Fixed in 528f688

CHANGES_NEXT_RELEASE
ArqamFarooqui110719: Fixed in 528f688

fgalan (approved):
LGTM. Thanks for the contribution!

PR #4548, opened 2024-05-22
FIX CMAKE_MINIMUM_REQUIRED

fgalan:
Fixed this annoying warning at building time:

```
CMake Deprecation Warning at src/lib/cache/CMakeLists.txt:21 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Note that in the root CMakeList.txt we already have

```
cmake_minimum_required(VERSION 3.0)
```

so this PR is aligning that in the rest of CMakeList.txt files.

### DIFF
--- a/src/lib/alarmMgr/CMakeLists.txt
+++ b/src/lib/alarmMgr/CMakeLists.txt
@@ -18,7 +18,7 @@
 # For those usages not covered by this license please contact with
 # iot_support at tid dot es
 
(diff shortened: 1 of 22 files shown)


mapedraza (approved):
LGTM

PR #4567, opened 2024-06-05
(JP) FIX to reference presentation (#4566)

fisuda:
This PR fixes to reference Japanese presentation. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4568, opened 2024-06-05
Restructure docker

jicarretero:
The docker image for ubi has been restructured in a way that i) is more easy to understand and ii) is more based in rather "stable" repositories.


kzangeli:
> Maybe this PR (with >4000 commit!) has been created on the wrong repository?
Yes, I think so :)
I was looking for it in the Orion-LD repo and just found it here instead :)

jicarretero:
Sorry. Wrong repo.

PR #4564, opened 2024-06-04
REMOVE subs legacy format

fgalan:
Test are expected to fail in the first pass of this PR.

- [x] fix/disable tests
- [x] modify deprecation documentation

PR #4569, opened 2024-06-05
(JP) REMOVE documentation about subs legacy format (#4564)

fisuda:
This PR removes Japanese documentation about subs legacy format. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4563, opened 2024-06-04
FIX remove NGSIv1 lasting API ops

fgalan:
Test are expected to fail in the first pass of this PR.

- [x] fix/disable tests
- [x] modify deprecation documentation

PR #4572, opened 2024-06-05
(JP) REMOVE documentation about NGSIv1 lasting API ops (#4563)

fisuda:
This PR removes documentation about NGSIv1 lasting API ops. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4571, opened 2024-06-05
[Doc] Add know JEXL limitation based on attributes names

mapedraza:

PR #4573, opened 2024-06-05
(JP) [Doc] Add know JEXL limitation based on attributes names (#4571)

fisuda:
This PR adds Japanese documentation about kown JEXL limitation based on attributes names. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4583, opened 2024-06-11
FIX enable statistics_with_counters.test

fgalan:
Recently DISABLED in 4.0.0, actually it can be recovered just removing NGSIv1 operations.


danielvillalbamota (approved):
LGTM

PR #4595, opened 2024-07-10
ADD improved and new transformations

fgalan:
- [x] Step cjexl version
- [x] New time-related transformations (tests + doc)
- [x] CNR

PR #4597, opened 2024-08-01
(JP) ADD doc about new transformations (#4595)

fisuda:
This PR adds Japanese documentation about new transformations. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4598, opened 2024-08-01
FIX DateTime with min and max operators

fgalan:
Issue https://github.com/telefonicaid/fiware-orion/issues/4585

PR #4599, opened 2024-08-01
(JP) FIX DateTime with min and max operators (#4598)

fisuda:
This PR fixes Japanese documentation about DateTime with min and max operators. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4603, opened 2024-08-06
[WIP] REMOVE NGSIv1 dead code

fgalan:
NGSIv1 API was gradually removed from Orion 2.6.0 to 4.0.0, but a lot of dead code remains.

This PR removes that dead code.

PR #4604, opened 2024-08-09
FIX remove legacyForwarding true in test that does not need it

fgalan:
This PR does a .test simplification, removing legacyForwarding: true (a deprecated feature) in test that don't need it. They are probably old test developed in the times when NGSIv2 forwarding was not fully working.

After this PR the only tree .test that use legacyForwarding: true are he following ones:

* 0000_deprecated_checkings/log_deprecate_warning.test: which checks deprecated features logging as this one
* 3162_context_providers_legacy_non_primitive/blanked_out_attr.test: not sure of the usage of legacyForwarding: true in this case, but this .test seems to be the regression of a particular fixed bug, so I prefer not touching it.
* 3944_legacy_forwarding_rendering/legacy_forwarding_rendering.test: which checks the CRUD of the legacyForwarding field

In addition, note that none of these remaining test is actually implementing a NGSIv1 forwarding case. A .test for that is going to be added in PR https://github.com/telefonicaid/fiware-orion/pull/4603, to test the new parser developed for that.

### DIFF
--- a/test/functionalTest/cases/0501_cors/registration_cors.test
+++ b/test/functionalTest/cases/0501_cors/registration_cors.test
@@ -44,8 +44,7 @@ payload='{
    "provider": {
      "http": {
        "url": "http://localhost:'${CP1_PORT}'/v2"
-     },
-     "legacyForwarding": true
+     }
    },
    "status": "active"
 }'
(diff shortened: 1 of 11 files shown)


mapedraza (approved):
LGTM

Issue #4605, opened 2024-08-20
Unexpected Behavior with entityUpdate and notifyOnMetadataChange


goten002:
When using subscriptions in FIWARE Orion with the `alterationTypes` field set to `entityUpdate` and `notifyOnMetadataChange` set to `false`, the subscription does not behave as expected. Specifically, the notification fails to trigger when an entity is updated with changes in metadata, when the attribute values remain unchanged. This behavior suggests that the subscription is incorrectly ignoring updates that involve metadata changes while the attribute values stay the same.

- Orion version: 4.0..0
- MongoDB version 6.0
- Env variables: -logLevel DEBUG -dbURI mongodb://mongodb:27017
- Docker image: fiware/orion:4.0.0

**How to reproduce it**
1. Create a subscription::
```BASH
curl -iX POST 'http://127.0.0.1:1026/v2/subscriptions' \
-H 'Fiware-Service: test' \
-H 'Fiware-ServicePath: /test' \
-H 'Content-Type: application/json' \
-d '{
  "description": "testing-quantumleap-subscription",
  "subject": {
    "entities": [
      {
        "idPattern": ".*",
        "type": "DummyDevice"
      }
    ],
    "condition": {
      "attrs": [
        "temperature",
        "humidity"
      ],
      "notifyOnMetadataChange": false,
      "alterationTypes": [
        "entityUpdate",
        "entityCreate"
      ]
    }
  },
  "notification": {
    "attrs": [
      "temperature",
      "humidity"
    ],
    "onlyChangedAttrs": true,
    "attrsFormat": "normalized",
    "http": {
      "url": "http://quantumleap:8668/v2/notify"
    },
    "metadata": [
      "TimeInstant"
    ]
  }
}'
```

2. Create an entity
```BASH
curl -iX POST 'http://localhost:1026/v2/entities' \
-H 'Fiware-Service: test' \
-H 'Fiware-ServicePath: /test' \
-H 'Content-Type: application/json' \
-d '{
  "id": "dummy-device-01",
  "type": "DummyDevice",
  "temperature": {
    "type": "Number",
    "value": 20,
    "metadata": {
      "TimeInstant": {
        "type": "DateTime",
        "value": "2024-08-20T00:00:00.000Z"
      }
    }
  },
  "humidity": {
    "type": "Number",
    "value": 50,
    "metadata": {
      "TimeInstant": {
        "type": "DateTime",
        "value": "2024-08-20T00:00:00.000Z"
      }
[...]

PR #4612, opened 2024-09-11
Fix: metadata modifications are not considered as change (with regards to subscription alterationTypes) if notifyOnMetadataChange is false 

fgalan:
Issue #4605


AlvaroVega (approved):
LGTM

Issue #1995, opened 2016-04-04
[BUG] in GET v2/entities using "q" query parameter, seconds and mSecs are not validated


iariasleon:
in GET v2/entities using "q" query parameter, seconds and mSecs are not validated. Before seconds (year, month, day, hour, minute) all are validate, but after minutes not.

#### cases that return 200 - OK 
 - /v2/entities?q=my_time>=2016-04-05T14:10:00.00Z
 - /v2/entities?q=my_time>=2016-04-05T14:10:4**XX**
 - /v2/entities?q=my_time>=2016-04-05T14:10:00**;**00Z
 - /v2/entities?q=my_time>=2016-04-05T14:10:00.**j**00Z
 - /v2/entities?q=my_time>=2016-04-05T14:10:00.00**N**

#### case that return 400 - Bad Request 
```
{
"error": "BadRequest"
"description": "invalid query expression"
}
```
 - year, month, day, hour, minute and them separators are validated
 - /v2/entities?q=my_time>=2016-04-05T14**;**10:00.00Z

ArqamFarooqui110719:
Hi @fgalan ,
It seems similar kind of issue related to date-time validation, I'd like to work on this.
I think this issue is still valid. Please provide you opinion on this?

fgalan:
First of all, check with a new .test raises the error that the issue is still valid (it more than 8 years old :)

In that case, yes, I think it something to be fixed.

ArqamFarooqui110719:
Hi @fgalan 
I have investigated the issue with provided dataset and observed that this issue is still valid as GET requests with invalid character are returning `200 OK` response.
I am investigating the issue and will raise a PR with the fix.

ArqamFarooqui110719:
PR #4623

ArqamFarooqui110719:
I think we can close the issue as well?

fgalan:
Closing (explanation [in this comment](https://github.com/telefonicaid/fiware-orion/pull/4623#issuecomment-2407527901) and in the following ones).

PR #4210, opened 2022-09-01
(JP) UPDATE documentations

fisuda:
This PR improves translation quality of Japanese documentations (including typo corrections and fixing wrong links). It would be great if you could review this PR. Thanks.

Issue #4120, opened 2022-05-12
Fail when null is in a list of values in a query by attribute values


danielvillalbamota:
**Bug description**
A clear and concise description of what the bug is. Please do not forget to add:

Having entities where for some of them have a specific attr, e.g. municipality with value `null`, a query by a list of values of this attributes gets an error.

**How to reproduce it**
Steps to reproduce the behavior (as an example)

1. Create an entity this way:
    ```BASH
    curl -iX POST 'http://localhost:1026/v2/entities' \
    -H 'Content-Type: application/json' \
    -d '
    {
        "id": "Bug:007",
        "type": "bug",
        "status": {
            "type": "Text",
            "value": "fixed"
        },
        "municipality": {
            "value": null,
            "type": "Text"
        }  
    }'
    ```

2. When make the following query:
```BASH
curl --location --request GET 'http://host:1026/v2/entities?q=municipality:null,alcobendas' \
--header 'Fiware-Service: service' \
--header 'Fiware-ServicePath: /subservice' \
```

3. See error in the request response
```
400 Bad Request
{
    "error": "BadRequest",
    "description": "invalid JSON value type in list"
}
```

4. When the query is just the `null` value the answer is correct

```BASH
curl --location --request GET 'http://host:1026/v2/entities?q=municipality:null' \
--header 'Fiware-Service: service' \
--header 'Fiware-ServicePath: /subservice' \
--header 'X-Auth-Token: ...
```

Response:
200 OK
```json
[
    ...,
    {
        "id": "Bug:007",
        "type": "bug",
        "municipality": {
            "type": "Text",
            "value": null,
            "metadata": {}
        },
        "status": {
            "type": "Text",
            "value": "fixed",
            "metadata": {}
        }
    }
]
```

**Expected behavior**

To be able to retrieve both set of entities, the ones with the `municipality=alcobendas` and the others with `municipality=null`

**Additional information**

NA

fgalan:
Looking in StringFilter.h:

```
typedef enum StringFilterValueType
{
  ...
  SfvtNumberList,
  SfvtDateList,
  SfvtStringList
} StringFilterValueType;
```

So it seems our internal data structures are not though for mixing different types (string and null in this case) but let me think if I can come with some creative solution for this :)

fgalan:
We will use a new bool flag named `nullInList` to mark the existence of  `null` within a list. Not a very smart solution, but it'll work.

fgalan:
[...]

PR #4623, opened 2024-10-08
fix invalid date in query parameter with GET entities API

ArqamFarooqui110719:
fix invalid date in query parameter with GET entities API (Issue #1995 )

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -1,0 +1,1 @@
+- Fix: invalid characters in date values with query parameter in GET API should not allowed (#1995)
--- a/src/lib/rest/StringFilter.cpp
+++ b/src/lib/rest/StringFilter.cpp
@@ -438,6 +438,11 @@ bool StringFilterItem::valueGet
   {
     *valueTypeP = SfvtDate;
   }
+  else if (*doubleP == -1)
+  {
+    *errorStringP = std::string("invalid characters in query parameter");
+    return false;
+  }
   else if (strcmp(s, "true") == 0)
   {
     *valueTypeP = SfvtBool;
--- a/test/functionalTest/cases/1995_invalid_datetime_query_expression/invalid_datetime_query_expression_in_GET_API.test
+++ b/test/functionalTest/cases/1995_invalid_datetime_query_expression/invalid_datetime_query_expression_in_GET_API.test
@@ -1,0 +1,319 @@
+# Copyright 2024 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero
+# General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Orion Context Broker. If not, see http://www.gnu.org/licenses/.
+#
+# For those usages not covered by this license please contact with
+# iot_support at tid dot es
+
+# VALGRIND_READY - to mark the test ready for valgrindTestSuite.sh
+
+--NAME--
+Invalid DateTime expression in query parameter with GET entities API
+
+--SHELL-INIT--
+dbInit CB
+brokerStart CB
+
+--SHELL--
+
+#
+# 01. Create an entity E1 with attribute 'mytime' of type DateTime
+# 02. Create an entity E2 with attribute 'mytime' of type DateTime
+# 03. GET all entities (success case)
+# 04. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14:10:4XX.00Z (see error)
+# 05. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14:10:4X.00Z (see error)
+# 06. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14:10:00**;**00Z (see error)
+# 07. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14:10:00.j00Z (see error)
+# 08. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14**;**10:00.00Z (see error)
+# 09. GET entities with invalid character in DateTime query parameter i.e. q=mytime<=2016-04-05T14:00:XX00.00Z (see error)
+# 10. GET entities with correct datetime value in query parameter i.e. q=mytime>=2016-08-08T17:09:24.238Z (success case)
+# 11. GET entities with correct datetime value in query parameter i.e. q=mytime<=2016-08-08T17:09:24.238Z (success case)
+#
+
+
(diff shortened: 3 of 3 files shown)


fgalan:
A lot (amost all?) tests are failing with thes PR.

@ArqamFarooqui110719 could you have a look and identify the cause, please?

ArqamFarooqui110719:
Yes I'm looking into the issue.

ArqamFarooqui110719:
Hi @fgalan  
As per my investigation, when we are requesting the below GET API:
```
curl -X GET 'localhost:1026/v2/entities?q=mytime<=2016-04-05T14:10:4XX.00Z'
```
It is following below code flow and returning error after (PR #4623 ).
 
Orion_GET_API_Flow

But the other APIs such as POST subscription (FT #4120), GET APIs with query parameter as string (FT #1156) is also calling the same functions for parsing the query parameter, but I don't think that is a correct way to parse a normal string within `parse8601Time()` function.
 
Create Subscription API which causing the issue is:
```
curl -iX POST '
[http://localhost:1026/v2/subscriptions'](http://localhost:1026/v2/subscriptions%27)
-H 'Content-Type: application/json' -d '
{
  "subject": {
    "entities": [
      {
        "id": "E1",
        "type": "T"
      }
    ],
    "condition": {
      "expression": {
        "q": "A:foo,null"
      }
    }
  },
  "notification": {
    "http": {
      "url": "
[http://127.0.0.1:1028/accumulate"](http://127.0.0.1:1028/accumulate%22)
    }
  }
}'
```
Above API is causing issue due to query parameter (q) in `condition` part. In`q = A:foo,null`, A is not a valid character for DateTime, so parse8601Time() function is returning `-1 or error` due to which the FT (#4210) is getting failed.

GET entities API (FT #1156) which causing the issue is:
```
curl -X GET
[http://localhost:1026/v2/entities?mq=t.'a'.'b.c.d'>=string](http://localhost:1026/v2/entities?mq=t.%27a%27.%27b.c.d%27%3E=string)
```

In this case also, query parameter is parsing in `parse8601Time()` function, so same code flow and same error.

Please provide your opinion and guidance on this situation, what should I do?

fgalan:
The problem here is that you cannot assume any that `q` attribute has an special semantic.

Consider the following example. You have two entities, using the same attribute name but different types, eg:

Entity1 with:

```
"A": {
  "value": "foo",
  "type": "Text"
}
```

Entity2 with:


```
"A": {
  "value": "2024-01-01T00:00:00Z",
  "type": "DateTime"
}
```

so you can do a query like this `GET /v2/entities?q=A:foo`. That query would return Entity1.

Maybe https://github.com/telefonicaid/fiware-orion/issues/1995 is pointless and should be closed? My point is that we cannot do any attributetype-based validation at `q` (DateTime or any other) as I argue above.

@ArqamFarooqui110719 what do you think?

ArqamFarooqui110719:
> The problem here is that you cannot assume any that `q` attribute has an special semantic.
> 

Yes I also have same understanding.

> 
> Maybe #1995 is pointless and should be closed? My point is that we cannot do any attributetype-based validation at `q` (DateTime or any other) as I argue above.
> 
> @ArqamFarooqui110719 what do you think?

But I feel that the `parseTime8601()` function should not parse the normal string and it should only be used for parsing the DateTime values. And there should be a separate function for parsing the normal strings for example: `parseStringValues()`. 
**By doing so, we can correctly parse the date values as well as string values separately without any issue.**

To achieve that we can follow below steps:
1. before calling the parsing function (parseTime8601()` directly, first we need to check whether the given string is a datetime or normal string using regex. For this we can create a separate function like `isDateTime()`. 
2. if the attribute value is `DateTime` then call the `parsetime8601()` function for remaining parsing of DateTime attribute.
3. if the attribute value is not DateTime (Normal string), Then we should call the normal string parsing function. 
(In normal string case, I think we only need to check for forbidden characters)

Please find below screenshot for above implementation:
Orion_GET_API_Fix_1995


Using the above method we can separate the parsing logic as well as we can restrict the usage of invalid characters in query parameter.

@fgalan What is your opinion on this?

fgalan:
Not sure if I'm getting your point...

> But I think that the `parseTime8601()` function should not parse the normal string and it should only be used for parsing the DateTime values

Based in my [example above](https://github.com/telefonicaid/fiware-orion/pull/4623#issuecomment-2407527901), in case for instance of `GET /v2/entities?q=A:foo`, how do you now that `foo` should be interpreted as normal string or as date representation?

For Entity1, `foo` is valid as normal string (because `A` attribute type in Entity1 is not DateTime)

For Entity2, `foo` should be interpreted as a date (because `A` attribute type in Entity2 is DateTime)

Ok, you could inspect attributes and check the type in each one (so you could have a "parse ok" or "parse nok" for each entity),  but, which would be the purpose? How would be used these checks in the response to  `GET /v2/entities?q=A:foo`?

Note it is not an error from user standpoint to ask for `A:foo` without worrying about the attribute type in actual entities. The user just want to get entities which `A` attribute is the string `foo` and it's ok for him not getting in the response other entities (such the ones using DateTimes for `A`).

I hope have clarifying my point...

ArqamFarooqui110719:
> 
> Based in my [example above](https://github.com/telefonicaid/fiware-orion/pull/4623#issuecomment-2407527901), in case for instance of `GET /v2/entities?q=A:foo`, how do you now that `foo` should be interpreted as normal string or as date representation?
> 
> For Entity1, `foo` is valid as normal string (because `A` attribute type in Entity1 is not DateTime)
> 
> For Entity2, `foo` should be interpreted as a date (because `A` attribute type in Entity2 is DateTime)
> 
> Ok, you could inspect attributes and check the type in each one (so you could have a "parse ok" or "parse nok" for each entity), but, which would be the purpose? How would be used these checks in the response to `GET /v2/entities?q=A:foo`?
> 
> Note it is not an error from user standpoint to ask for `A:foo` without worrying about the attribute type in actual entities. The user just want to get entities which `A` attribute is the string `foo` and it's ok for him not getting in the response other entities (such the ones using DateTimes for `A`).
> 
Yes I understood the above point.

But in my [last comment](https://github.com/telefonicaid/fiware-orion/pull/4623#issuecomment-2429371778), I meant that-
 inspecting the attribute value using regex not attribute type in `isDateTime()` function (may be something like following):
```
bool isDateTime(const std::string& ss)
{
  std::regex dateTimePattern(R"(^\d{4}-\d{2}-\d{2}([T ]\d{2}:\d{2}:\d{2}(\.\d{3})?(Z)?)?$)");
  return std::regex_match(ss, dateTimePattern);
}
```
The example given in last comment regarding entity1 and entity2, In that case, this function will treat `foo` as a normal string value and only datetime formatted string will treated as datetime value.

The above function (`isDateTime()`) will match the DateTime patterns like `YYYY-MM-DD`, `YYYY-MM-DDTHH:MM:SS` and `YYYY-MM-DDTHH:MM:SS.sssZ` etc. And the `foo` (or any other string value) will be treated as normal string.


(I think it can work! Although I need to test it for all use cases.)

fgalan:
Maybe you could add to the PR a .test covering your use cases (including the different possibilities)? Maybe that way I can get a clearer understanding of your idea from a API user perspective.

ArqamFarooqui110719:
Hi @fgalan 
I have investigated on the approach shared in the [above comment](https://github.com/telefonicaid/fiware-orion/pull/4623#issuecomment-2431617608).
My approach was to segregate the parsing logic for normal string parameters and DateTime type values. But in this case it seems very lengthy to findout all possible pattern for datetime value (valid and invalid) based on which we can segregate the normal strings.
I have investigated and tried my best but it seems it will be almost impossible to completely fix the issue. So I think we can close the Issue and PR.

fgalan:
> So I think we can close the Issue and PR.

I agree.

Thanks for the investigation!

Issue #4578, opened 2024-06-07
Delete bulk subscriptions using script


ArqamFarooqui110719:
**Bug description**
I have thousands of subscription created in my orion setup. 
Is there any script available, using which I can delete all subscriptions using IDs.

If not, shouldn't we have a script (python/shell) which can delete multiple subscription on single click.

fgalan:
Maybe this can be useful:

https://github.com/telefonicaid/fiware-orion/blob/master/scripts/utils/delete_entities_by_list.py
https://github.com/telefonicaid/fiware-orion/blob/master/scripts/utils/delete_entities_by_query.py

I think the code is pretty straightforward, but if you some doubt, just ask it :)

fgalan:
Ah, sorry... they for *entities* and you are asking for *subscriptions*

I think we don't have such script. Maybe a delete_subscriptions.py script (based on the previous ones) could be useful. It would be an interesting contribution :)

fgalan:
That's correct!

I'd suggest to implement the script in Python, with his the tooling language that we use in this repository.

Regarding step 2, to collect subscription IDs, the `GET /v2/subscriptions` operation has to be used. Note that you could have a large number of subscription, so you should use the pagination options, as delete_entities_by_query.py does. Details of the API operation can be found at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#list-subscriptions-get-v2subscriptions

Finally, as an extra feature of your script it would be interesting to include filtering capabilities, based in the JSON of each subscription obtained. For instance, delete subscription which match a given `description`, etc.

ArqamFarooqui110719:
PR #4626

fgalan:
Now that PR https://github.com/telefonicaid/fiware-orion/pull/4626 has been merged, should this issue be closed?

ArqamFarooqui110719:
Yes. Closed as completed.

PR #4626, opened 2024-10-21
Delete subscriptions in bulk using python script

ArqamFarooqui110719:
Delete subscriptions in bulk this python script (Issue #4578 )

### DIFF
--- a/scripts/utils/delete_subscriptions.py
+++ b/scripts/utils/delete_subscriptions.py
@@ -40,7 +40,12 @@ headers = {
 page_size = 500
 
 # WARNING! use this filter or you will remove *all* subscriptions
-filter = 'offset=0'
+# FIXME: Note that filtering is not supported in the GET /v2/subscriptions. If, at some point it gets
+# implemented (https://github.com/telefonicaid/fiware-orion/issues/1145 or additional issues),
+# we should adjust this point of the script with a valid example filter
+#filter = '&type=device'
+# For now we can keep filter variable empty.
+filter = ''
 
 # END of the configuration part (don't touch below this line ;)
 ##############################################################
@@ -52,7 +57,7 @@ def delete_subscriptions():
     """
     try:
         # Fetching subscriptions with pagination
-        response = requests.get(f"{cb_endpoint}?limit={page_size}&{filter}", headers=headers, verify=False)
+        response = requests.get(f"{cb_endpoint}?limit={page_size}{filter}", headers=headers, verify=False)
 
         if response.status_code == 200:
             subscriptions = response.json()
--- a/scripts/utils/delete_subscriptions.py
+++ b/scripts/utils/delete_subscriptions.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python3
+# Copyright 2024 Telefonica Investigacion y Desarrollo, S.A.U
+#
+# This file is part of Orion Context Broker.
+#
+# Orion Context Broker is free software: you can redistribute it and/or
+# modify it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# Orion Context Broker is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


ArqamFarooqui110719:
I have verified the script and it is working successfully. Please find attached screenshot for reference:
issue_4578_screenshot

scripts/utils/delete_subscriptions.py
```diff
@@ -0,0 +1,138 @@
...
+headers = {
+    'Accept': 'application/json',
+    'fiware-service': 'service',
+    'fiware-servicepath': '/subservice'
+}
+
+page_size = 500
+
+# WARNING! use this filter or you will remove *all* subscriptions
+filter = 'offset=0'
```
fgalan, line 43: Even with `offset=0`, all subscription would be removed :). Note that offset/limit are not actually filters, but a mechanism to get paginated results.

Let's keep the filter varible, just in case in the future gets implemented, but I'd suggest to use something like this in this point

```suggestion
# WARNING! use this filter or you will remove *all* subscriptions
# FIXME: note that filtering is not supported in the GET /v2/subscriptions. If at the it is get
# implemented (https://github.com/telefonicaid/fiware-orion/issues/1145 or additional issues),
# let's adjust this point of the script with a valid example filter
#filter = '&type=device'
```

scripts/utils/delete_subscriptions.py
```diff
@@ -0,0 +1,138 @@
...
+##############################################################
+
+def delete_subscriptions():
+    """
+    delete subscriptions in current page.
+    :return: True if removal was ok, False otherwise
+    """
+    try:
+        # Fetching subscriptions with pagination
+        response = requests.get(f"{cb_endpoint}?limit={page_size}&{filter}", headers=headers, verify=False)
```
fgalan, line 55:
```suggestion
        response = requests.get(f"{cb_endpoint}?limit={page_size}{filter}", headers=headers, verify=False)
```

Let's assume that filter is either the empty string or a query param with the `&` already included (as in existing delete_entities_by_query.py)

scripts/utils/delete_subscriptions.py
ArqamFarooqui110719: Yes, I intentionally put the offset value as the actual filter (q/type) is not working yet. 
But the suggested changes gives more clarity so I'll update the PR:)

ArqamFarooqui110719: > Even with `offset=0`, all subscription would be removed :). Note that offset/limit are not actually filters, but a mechanism to get paginated results.
> 
> Let's keep the filter varible, just in case in the future gets implemented, but I'd suggest to use something like this in this point

filter variable cannot be commented as it make request bad. Please see below screenshot for reference:

filter_variable
 
So I can add below lines as well in the script:
```
# WARNING! use this filter or you will remove *all* subscriptions
# FIXME: note that filtering is not supported in the GET /v2/subscriptions. If at the it is get
# implemented (https://github.com/telefonicaid/fiware-orion/issues/1145 or additional issues),
# let's adjust this point of the script with a valid example filter
#filter = '&type=device'
# For now we can keep filter variable empty.
filter = ''
```

fgalan: Fine with it

scripts/utils/delete_subscriptions.py
ArqamFarooqui110719: fixed in [01f8553](https://github.com/telefonicaid/fiware-orion/pull/4626/commits/01f855348599b3e69c8672ac89e28656b2e51a20)

scripts/utils/delete_subscriptions.py
ArqamFarooqui110719: fixed in [01f8553](https://github.com/telefonicaid/fiware-orion/pull/4626/commits/01f855348599b3e69c8672ac89e28656b2e51a20)

fgalan (approved):
LGTM

Thanks for your contribution!

PR #4649, opened 2025-02-24
FIX improve notification queue full log error including subscription id

fgalan:


fgalan:
What about to take a "snapshot" of the queue in that moment and write in a log the subIds of all items in the queue?

PR #4648, opened 2025-02-13
Several fixes related with subscriptions & notification behaviours

fgalan:
- [x] #4647
- [x] #4643 
- [x] #4644
- [x] #4645 
- [x] Review FIXME PR marks
- [x] Valgrind regression (we are touching dynamic memory using `clone()`... seems to be safe, but better to check) - side PR https://github.com/telefonicaid/fiware-orion/pull/4651

PR #4653, opened 2025-02-27
(JP) Add doc about subscriptions based in alteration type (#4648)

fisuda:
This PR adds Japanese documentation about subscriptions based in alteration type. It would be great if you could review the PR. Thanks.


fgalan (approved):
LGTM

PR #4669, opened 2025-04-24
ADD ftest for the 4664 case

fgalan:
Issue #4664

PR #4673, opened 2025-04-26
(JP) ADD ftest for the 4664 case (#4669)

fisuda:
This PR adds Japanese translation. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4675, opened 2025-05-07
ADD jexl_transformation_mapper_cases.test

fgalan:
Requires cjexl 0.6.0, so the test is going to fail first time.


fgalan:
> lest's re-run test and see what happens...

Now we see:

imagen

So it works fine :)

mrutid (approved):

PR #4703, opened 2025-07-30
ADD example lanuch.json file for VSCode

fgalan:
Repository should be IDE-neutral (thus I'm not creating this file directly in `.vscode/launch.json` path :) but I think it can be an useful help to people using VSCode to develop Orion code.


orianar (approved):
LGTM

PR #4702, opened 2025-07-29
FIX remove Scope classes

fgalan:
Related issue #4681

PR #4704, opened 2025-07-30
(JP) FIX remove Scope classes (and related stuff) (#4702)

fisuda:
This PR adds Japanese documantation. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

PR #4700, opened 2025-07-23
Feature/4666 kafka subscriptions

orianar:
Issue #4666

comments pending from @fgalan on PR https://github.com/telefonicaid/fiware-orion/pull/4691

- [x] Apart from comments done code, some general comments follow
- [x] API documentation at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#subscription-payload-datamodel should be modified to include information about the new datamodel
- [x] Doc at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/build_source.md#debian-12-officially-supported should include the new dependency (either installing from APT or building from source)
- [x] Doc at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/database_model.md#csubs-collection should be modified taking into account new kafka based subscriptions
- [x] Include the new library in the GET /versions?options=libVersions request
- [x] accumulator_server.py should be enhance to listen kafka notifications (similar to how MQTT works)
- [x]  kafkat service should be added to GitActions to be used in tests
.test to cover the new functionality should be included. They could be an adaptation of existint MQTT .test. Have a look specially to:
- [x] 3001_mqtt/
- [x]  3001_mqtt_alarms/ -> deferred to issue https://github.com/telefonicaid/fiware-orion/issues/4716

PR #4713, opened 2025-09-09
FIX use warning instead of Runtime Error when adding an object or array in basic expressions mode

fgalan:
Comes from https://github.com/telefonicaid/fiware-orion/pull/4700#issuecomment-3270452767


AlvaroVega (approved):
LGTM

Issue #4681, opened 2025-05-09
NGSIv1 removal pending issues and follow ups


fgalan:
PR https://github.com/telefonicaid/fiware-orion/pull/4603 start the removal, but some items are still pending:

- // FIXME PR: Count: is a NGSIv1 thing... remove this
- Search (case insensitive): NGSIv1 NGSI-v1 v1 ngsi10
- Search `"false"` and `"true"` literals (with quotes)
- isPattern )text-based) are very ugly (classs apiv2types/Entity). Search for `isPattern` to see if they can be removed, using `idPattern` instead Literales "false" y "true" con comillas
- usage of typeGiven in Entity
- inner and outer StatusCode in some requests (updates?) adaptErrorCodeForSingleEntityOperation
- Unify Attribute + ContextAttribute, etc. with the rest of types apparently duplicated
- Excessive usage of StatusCode. Replace by OrionError
- The logic to build the respons (payload) for postContextUpdate() is not needed. In NGSIv2 with OrionError is enough
- Check if appendToBson() needs useDefaultType or if its value is always fixed to true
- Check toBson() needs useDefaultType or if its value is always fixed to true
- Check updateAttrInNotifyCer() needs useDefaultType or if its value is always fixed to true
- postQueryContext -> doQuery
- postUpdateContext -> doUpdate
- CE, CER, CRR, CRRV, CRV, CERV, Context, providerApplication, etc.
- forbidenIdCharsV2 -> remove V2 sufix (search `V2(` to check more on this
- CSUB_ENTITY_ISPATTERN -> pattern true|false (if we are going to remove this in the registrations collection, we should do the same here...)

- Unify serviceRoutines/ and serviceRoutinesV2/ (first has badVerds and some other leftover method)

- In 3707_get_or_change_log_config/get_or_change_log_config.test some tests use "error" but no "description"
- In 1913_log_level_change_via_rest/log_level_change_via_rest.test some tests use "error" but no "description"

- New issue with the deprecation for legacyForwarding (similar to this one https://github.com/telefonicaid/fiware-orion/issues/4589)

- Responses to POST /v2/op/notify (a V2 operation) with a V1 response format... it makes no sense

Note directly related to NGSIv1 removal, but detected during this work:

* parseEntitiesResponse potential leak at the end. Memory is not release in case of parse error. Create a .test to raise the leak first. Review V2 parsing logic searching for `new`

```
[...]

PR #4718, opened 2025-09-26
FIX avoid Runtime Error overloging when no geofilter is actually used

fgalan:
It doesn't happen in 4.2.0, so no entry in CNR is done in this PR.

Probably introduced in PR https://github.com/telefonicaid/fiware-orion/pull/4603 or in the follow up fixes related with https://github.com/telefonicaid/fiware-orion/issues/4681.


orianar (approved):
LGTM

PR #4737, opened 2025-12-11
feat: improve notification log #4735

### LINKED ISSUE #4735: Improve Orion Notification Error Logs (Missing Root Cause)
**Is your feature request related to a problem / use case? Please describe.**
When Orion sends a notification and the receiver returns an HTTP 500, the log only shows:

`msg=Notification (subId: 6930585843c986004a0bfc00) response NOT OK, http code: 500`
The problem is that the root cause of the error is not logged.
This makes troubleshooting difficult — we don’t know whether the issue comes from the payload, headers, the remote endpoint, a timeout, or something else.

**Describe the solution you'd like**
When receiving a non-OK HTTP response (4xx or 5xx), Orion should log additional details such as:

- the error message returned by the endpoint, 

- the response body (if available)

- optionally, the response time (to identify timeouts)

Example of desired log output:

```
msg=Notification (subId: 6930585843c986004a0bfc00) NOT OK,
http code: 500,
error: "Internal error: invalid payload format",
responseBody: "{...}"
```
**Describe alternatives you've considered**
Increasing log verbosity on the receiver side
This helps identify the error but does not solve the lack of visibility in Orion itself. It also requires access to the remote service, which is not always possible.

Using a proxy or middleware to capture the full HTTP exchange
While this can expose the error details, it adds complexity to the architecture and is not practical for production environments.

Manually reproducing the failing request using tools like cURL or Postman
This can give insights, but it is time-consuming and does not help during real-time operations or when multiple notifications fail.

Enabling debug logs globally in Orion
This generates too much noise and still does not guarantee that the HTTP response body or error message will be included.

**Describe why you need this feature**
When a notification fails, Orion currently logs only the HTTP status code. Without the underlying error message or response body, it becomes extremely difficult to understand why the receiver rejected the notification.
This lack of visibility slows down debugging, increases operational overhead, and requires teams to inspect multiple systems just to identify the root cause.
Having the actual error returned by the receiver directly in Orion’s logs would:
significantly reduce troubleshooting time
help detect malformed payloads or configuration issues faster
improve reliability and monitoring of the notification pipeline
provide clearer insights during incidents and production debugging
In short, this feature makes Orion much easier to operate in real-world IoT deployments where notifications are frequent and critical.

**Additional information**
Below is an example of the current Orion log output, which does not include any details about the underlying error:

msg=Notification (subId: 6930585843c986004a0bfc00) response NOT OK, http code: 500

In production environments, this message alone is not sufficient to diagnose issues, especially when the receiving service returns meaningful error details (e.g., invalid payload, missing headers, authentication errors).

A more informative log entry — including the response body and error message — would greatly improve observability. For instance:
```
msg=Notification (subId: 6930585843c986004a0bfc00) NOT OK,
http code: 500,
error: "Internal error: invalid JSON payload",
responseBody: "{ \"error\": \"Schema mismatch\" }"
```

This enhancement aligns with best practices in distributed systems and increases operational transparency, especially in IoT setups where large volumes of notifications must be monitored.

**Do you have the intention to implement the solution**
- [X] Yes, I have the knowledge to implement this new feature.
- [ ] Yes, but I will need help.
- [ ] No, I do not have the skills.

tzzed:
Fix #4735.
Improves notification error logging by including the response body when a notification fails (HTTP response is not OK). This enhancement provides better visibility into notification failures by showing not only the HTTP status code but also the actual error message returned by the endpoint.

### Changes
Modified httpRequestSend.cpp:661-662 to include the response body (outP) in the warning log message when notification responses are NOT OK
The log message now shows: notification ID, HTTP code, and the complete response body
### Benefits
Easier debugging of notification failures by having the full error context in logs
No need to enable debug mode or use external tools to see why a notification failed
Aligns with the information already logged for successful notifications

### DIFF
--- a/src/lib/common/logTracing.cpp
+++ b/src/lib/common/logTracing.cpp
@@ -38,8 +38,7 @@
 *
 * NOTE: this function allocated dynamic memory, be careful with memory leaks!
 */
-static char* truncatePayload(const char* payload)
-{
+static char* truncatePayload(const char* payload) {
   // +5 due to "(...)"
   // +1 due to '\0'
   unsigned int truncatedPayloadLengh = logInfoPayloadMaxSize + 5 + 1;
@@ -50,6 +49,54 @@ static char* truncatePayload(const char* payload)
   truncatedPayload[truncatedPayloadLengh - 1] = '\0';
 
   return truncatedPayload;
+}
+
+
+
+static char *get_body(const char *response) {
+  const char *start = strchr(response, '{');
+  if (!start) return NULL;
+
+  const char *end = strrchr(start, '}');
+  if (!end || end < start) return NULL;
+
+  size_t len = (size_t)(end - start + 1);
+
+  char *out = (char*)malloc(len + 1);
+  if (!out) return NULL;
+
+  memcpy(out, start, len);
+  out[len] = '\0';
+
+  return out;
+}
+
+
+
+/* ****************************************************************************
+*
+* logWarnHttpNotification
+*/
+void logWarnHttpNotification
+(
+  const char*  idStringForLogs,
+  long long httpCode,
+  const char*  responseBody
+)
+{
+  if (responseBody == NULL)
+  {
+    responseBody = "";
+  }
+
+  char* effectiveBody = get_body(responseBody);
+  LM_W(("Notification (%s) response NOT OK, http code: %d, response body: %s",
+        idStringForLogs, httpCode, effectiveBody));
+
+  if (effectiveBody)
+  {
+    free(effectiveBody);
+  }
 }
 
 
@@ -296,6 +343,3 @@ void logInfoFwdRequest
   }
 
 }
-
-
-
--- a/src/lib/rest/httpRequestSend.cpp
+++ b/src/lib/rest/httpRequestSend.cpp
@@ -48,6 +48,8 @@
 #include "metricsMgr/metricsMgr.h"
 #include "rest/ConnectionInfo.h"
 #include "rest/httpRequestSend.h"
+
+#include "common/logTracing.h"
 #include "rest/HttpHeaders.h"
 #include "rest/rest.h"
 #include "rest/curlSem.h"
@@ -212,15 +214,6 @@ static int contentLenParse(char* s)
 
   return atoi(&contentLenP[offset]);  // ... and get the number
 }
-
-
-std::string get_body(const std::string &raw) {
-  const size_t start = raw.find("{");
-  const size_t end = raw.rfind("}");
-  if (start == std::string::npos || end == std::string::npos) return "";
-  return raw.substr(start, end - start + 1);
-}
-
 
 /* ****************************************************************************
 *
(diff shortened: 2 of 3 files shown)


tzzed:
@fgalan  sure! here an exemple of error response. 

```
Notification (subId: 693c6da61cf96df42d01b62a) response NOT OK, http code: 400, response body: HTTP/1.1 400 Bad Request

Date: Fri, 12 Dec 2025 19:32:07 GMT

Fiware-Correlator: 3eda5040-d791-11f0-bbc7-120e2b33566b; cbnotif=1

Content-Length: 97


{"error":"BadRequest","description":"Service not found. Check your URL as probably it is wrong."}
```

src/lib/rest/httpRequestSend.cpp
```diff
@@ -214,6 +214,13 @@ static int contentLenParse(char* s)
 }
 
 
+std::string get_body(const std::string &raw) {
```
tzzed: @fgalan do you think that should be `static` ?

src/lib/common/logTracing.cpp
```diff
@@ -38,8 +38,7 @@
 *
 * NOTE: this function allocated dynamic memory, be careful with memory leaks!
 */
-static char* truncatePayload(const char* payload)
-{
+static char* truncatePayload(const char* payload) {
```
fgalan, line 41:
```suggestion
static char* truncatePayload(const char* payload)
{
```

Coding style: [M8 rule](https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/contribution_guidelines.md#m8-code-blocks)

src/lib/common/logTracing.cpp
```diff
@@ -54,6 +53,54 @@ static char* truncatePayload(const char* payload)
...
+
+  char* effectiveBody = get_body(responseBody);
+  LM_W(("Notification (%s) response NOT OK, http code: %d, response body: %s",
+        idStringForLogs, httpCode, effectiveBody));
+
+  if (effectiveBody)
+  {
+    free(effectiveBody);
+  }
+}
```
fgalan, line 100: The logInfoPayloadMaxSize limit should be taken into account. No payload in log traces should go beyond that limit.

I'd suggest to use something like in other functions in this file

```
  bool cleanAfterUse = false;
  char* effectivePayload;

  if (strlen(payload) > logInfoPayloadMaxSize)
  {
    effectivePayload = truncatePayload(payload);
    cleanAfterUse = true;
  }
  else
  {
    effectivePayload = (char*) payload;
  }
```

If `resquestBody` also include headers, etc. and you want to skip that maybe something like this could be useful (please check, I haven't checked myself)

```
char* payload = strchr(resquestBody, '{');
```

src/lib/common/logTracing.cpp
```diff
@@ -54,6 +53,54 @@ static char* truncatePayload(const char* payload)
...
+  size_t len = (size_t)(end - start + 1);
+
+  char *out = (char*)malloc(len + 1);
+  if (!out) return NULL;
+
+  memcpy(out, start, len);
+  out[len] = '\0';
+
+  return out;
+}
```
fgalan, line 72: Probably not needed taking into account what I comment at https://github.com/telefonicaid/fiware-orion/pull/4737#discussion_r2622941633

fgalan:
This PR is maybe overpassed by PR https://github.com/telefonicaid/fiware-orion/pull/4754 (which not only takes into account notification failures, but also regular API request)

@tzzed what do you think?

PR #4754, opened 2026-03-05
ADD response payload to log info

fgalan:
*Edit:* issue #4756 

We need to use ciP->answer in all cases to make that information available in the requestComplete() function

As consecuence restReply() now uses uniquely ciP as argument (not sure if I like this change... but it's the easy way)

- [x] CNR
- [x] log.md (maybe)
- [x] .test covering the functionality


src/lib/serviceRoutinesV2/postUpdateContext.cpp
```diff
@@ -45,21 +45,11 @@
...
 
-/* ****************************************************************************
-*
-* jsonPayloadClean -
-*/
-static char* jsonPayloadClean(const char* payload)
-{
-  return (char*) strstr(payload, "{");
-}
-
```
fgalan, line 60: We have opted for the jsonPayloadClean() version in src/lib/serviceRoutinesV2/postQueryContext.cpp, which is more advanced that this one (this one is assuming that the payload is JSON, which is not necessarily true)

NTC (informative)

test/functionalTest/cases/3068_cprs_full_functional_v2/fwd_v1_ngsiv2_update_forward_with_fail.test
```diff
@@ -154,7 +154,7 @@ Raising alarm ForwardingError localhost:9997/cprfail/updateContext: error parsin
...
+Request forwarded (regId: REG1_ID): POST http://localhost:9997/cprfail/updateContext, request payload (542 bytes): {"contextElements":[{"id":"ConferenceRoom","isPattern":"false","type":"Room","attributes":[{"name":"lightstatus","type":"StructuredValue","value":{"x":1,"y":2}},{"name":"pressure","type":"StructuredValue","value":["a","b","c"]},{"name":"temperature","type":"Number","value":14,"metadatas":[{"name":"ID1","type":"Text","value":{"type":"Text","value":{"x":1,"y":2}}},{"name":"ID2","type":"Text","value":{"type":"Text","value":["a","b","c"]}},{"name":"ID3","type":"Text","value":{"type":"Text","value":"ThisIsID3"}}]}]}],"updateAction":"UPDATE"}, response payload (3201 bytes):
```
fgalan, line 157: Before this PR:

Image

After this PR:

Image

I think it was wrong first time. Strictly speaking, the payload doesn't start in the first `{` but some whitespace before.

NTC (informative)

PR #4759, opened 2026-03-11
update kafka download URL

orianar:


ci/deb/docker-compose-ci.functional.yml
```diff
@@ -34,4 +34,8 @@ services:
     volumes:
       - .:/opt/fiware-orion
     command:
-      build -miqts functional
+      - /bin/sh
+      - -c
+      - |
+        export PATH="/opt/kafka/bin:$PATH"
```
fgalan: Previous to this change, where the /opt/kafka/bin path is being set?

orianar: This change is not necessary; my bad. The only effective change is the URL update.

PR #4760, opened 2026-03-13
(JP) ADD response payload to log info (#4754)

fisuda:
This PR adds Japanese documentation in the logs.md file. It would great if you could review the PR.
Thanks.


fgalan (approved):
LGTM

PR #4753, opened 2026-02-27
Feature/4714 add user and passwd in kafka notifications

orianar:
Issue #4714

PR #4771, opened 2026-04-08
(JP)Added documentation about user and passwd in kafka notifications (#4753)

fisuda:
This PR adds Japanese documentation about user and passwd in kafka notifications. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM

Issue #4749, opened 2026-02-16
Unhardwire kafka key in notifications


fgalan:
**Is your feature request related to a problem / use case? Please describe.**

Kafka notificationa currently are using a hardwired key (equal to subscription id), which is not convenient in some case (see https://github.com/telefonicaid/kafnus/issues/200#issuecomment-3907244912)

**Describe the solution you'd like**

Add a new optional parameter in `kafka`/`kafkaCustom` https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#subscriptionnotificationkafkacustom for the key (`key`).

The value of that parameter would be a JEXL expression (note that a constant value, such `null` or a string is a particular case of valida JEXL expression).

Eg:

```
...
"notification": {
  "kafka": {
    "url": "kafka://broker1:9092,broker2:9092,broker3:9092",
    "topic": "sub1",
    "key": null
  }
}
...
```

```
...
"notification": {
  "kafka": {
    "url": "kafka://broker1:9092,broker2:9092,broker3:9092",
    "topic": "sub1",
    "key": "${id}"
  }
}
...
```

**Describe alternatives you've considered**
None

**Describe why you need this feature**
- To improve or simplify an scenario.

**Do you have the intention to implement the solution**
- [x] Yes, I have the knowledge to implement this new feature.
- [ ] Yes, but I will need help.
- [ ] No, I do not have the skills.

orianar:
There are currently two related PRs for this issue: **PR #4775** changes the Kafka message key from `subscriptionId` to `null` to cover the urgent Kafnus use case, and **PR #4774** goes further by making the key configurable through the subscription while also proposing `null` as the default value.

Cc: @fgalan

PR #4774, opened 2026-04-12
Add Kafka notification key support with null and JEXL handling (#4749)

orianar:
Issue #4749


orianar:
I’ve kept the tests aligned with the current implementation of `key`, and I’ve included the equivalent `payload` cases for comparison. I wanted to confirm with you whether this behavior looks fine to you or whether you think it would be better to adjust it: key absent and key: null are stored differently, but both resolve to a null Kafka key when the notification is sent. If you have a moment, could you also take a look at the tests?

@fgalan

fgalan:
> key absent and key: null are stored differently

This is the same way `payload` field works, isn't it?

In that case, I think is fine, to use the same existing mechanism.

> both resolve to a null Kafka key when the notification is sent

In the case of `payload` the absence of the field and the field with value `null` have different semantics (see https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/orion-api.md#omitting-payload), but in the case of Kafka `key` I think is ok that behavouir.

fgalan:
> could you also take a look at the tests?

Done. They seems to be ok.

CHANGES_NEXT_RELEASE
```diff
@@ -0,0 +1 @@
+Fix: hardwired Kafka notification key by supporting string, absent, and null keys for Kafka notifications, including macro-substituted keys in kafkaCustom (#4749)
```
fgalan: Maybe simpler and better:

```suggestion
- Add: new subscription field notification.kafka.key/notification.kafkaCustom.key (supporting macro substitution in the latter case) (#4749)
```

(I understand that the `key` field is new in this PR, isn't it?)

src/lib/common/macroSubstitute.cpp
```diff
@@ -77,14 +77,18 @@ std::string smartStringValue(const std::string stringValue, ExprContextObject* e
   }
   else if (exprContextObjectP != NULL)
   {
-    // "Partial replacement" case. In this case, the result is always a string
-    std::string effectiveValue;
-    if (!macroSubstitute(&effectiveValue, stringValue, exprContextObjectP, "null", true))
+    ResolveResult r = macroSubstitute(stringValue, exprContextObjectP, "null", true);
```
fgalan, line 81:
```suggestion
    // "Partial replacement" case. In this case, the result is always a string
    ResolveResult r = macroSubstitute(stringValue, exprContextObjectP, "null", true);
```

If the comment is still valid (I understand so, as the modification in this part is only to use a smarter way of returning result, based on `ResolveResult` struct) I'd suggest not removing it.

src/lib/common/macroSubstitute.cpp
```diff
@@ -123,22 +133,19 @@ static std::string stringValueOrNothing(ExprContextObject* exprContextObjectP, c
...
     if (raw)
     {
-      // This means that the expression is in the middle of the string (i.e. partial replacement and not full replacement),
-      // so double quotes have to be be removed
-      result = removeQuotes(s);
-    }
-    else
-    {
-      result = s;
+      // Expression is in the middle of the string, so double quotes must be removed
```
fgalan:
```suggestion
      // Expression is in the middle of the string (i.e. partial replacement and not full replacement), so double quotes must be removed
```

Is correct?

src/lib/common/macroSubstitute.cpp
```diff
@@ -227,32 +235,42 @@ bool macroSubstitute(std::string* to, const std::string& from, ExprContextObject
...
+      lastReplacedValue = value;
+      out.value.replace(out.value.find(macro), macro.length(), value.value);
+      iterations++;
     }
   }
 
-  return true;
+  if (iterations == 1 && lastReplacedValue.isNull && out.value == lastReplacedValue.value) {
+    out.status = ResolveNull;
+    return out;
```
fgalan: Style:

- https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/contribution_guidelines.md#m8-code-blocks
- https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/contribution_guidelines.md#s3-parenthesis-in-part-conditions

```suggestion
  if ((iterations == 1) && lastReplacedValue.isNull && (out.value == lastReplacedValue.value))
  {
    out.status = ResolveNull;
    return out;
```

fgalan: In addition, this part if a bit hard to understand... why this expection in `interation == 1` ? Let's talk :)

src/lib/mongoBackend/dbConstants.h
```diff
@@ -134,6 +134,7 @@
 #define CSUB_KAFKATOPIC              "kafkaTopic"
 #define CSUB_KAFKA_SECURITY_PROTOCOL  "securityProtocol"
 #define CSUB_KAFKA_SASL_MECHANISM     "saslMechanism"
+#define CSUB_KAFKAKEY                "key"
```
fgalan, line 137: Documentation at https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/admin/database_model.md#csubs-collection should be modified to include the new field?

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -108,6 +109,63 @@ void Notifier::sendNotifyContextRequest
...
+* buildExprContext -
+*
+*/
+static void buildExprContext
+(
+  ExprContextObject*        exprContextP,
+  ContextElementResponse*   notifyCerP,
+  const std::string&        tenant,
+  const std::string&        xauthToken,
+  bool                      basic
```
fgalan: Maybe this argument is redundant? exprContextP->isBasic() should provide it

src/lib/ngsiNotify/Notifier.cpp
```diff
@@ -408,48 +509,7 @@ static SenderThreadParams* buildSenderParamsCustom
 
   // Used by several macroSubstitute() calls along this function
   ExprContextObject exprContext(basic);
-  ExprContextObject exprMetadataContext(basic);
```
fgalan, line 411: I understand this code has been moved to buildExprContext() function

That's fine, but the code comment should be also moved if still valid (they contain insights that were hard to find in the past, so better not losing them...)

src/lib/common/macroSubstitute.cpp
orianar: The idea is to return ResolveNull only when the whole input was a single macro that resolved to null. iterations == 1 is not enough by itself, because a string like "value is ${x}" also has exactly one replacement.

I can add a comment like this to make it clearer:

// If there was exactly one replacement and the replaced value was null,
// return ResolveNull only when the whole output is exactly the replaced value.
// This avoids treating strings like "value is ${x}" as null:
//   "${x}"          -> "" == ""          => ResolveNull
//   "value is ${x}" -> "value is " == "" => ResolveOk

src/lib/ngsiNotify/Notifier.cpp
orianar: Fixed in #[29d5403](https://github.com/telefonicaid/fiware-orion/pull/4774/commits/29d5403de35d62c4d165932dcd507dae98aa2f6a)

src/lib/ngsiNotify/Notifier.cpp
orianar: Fixed in #[29d5403](https://github.com/telefonicaid/fiware-orion/pull/4774/commits/29d5403de35d62c4d165932dcd507dae98aa2f6a)

src/lib/common/macroSubstitute.cpp
orianar: Fixed in #[f222ebe](https://github.com/telefonicaid/fiware-orion/pull/4774/commits/f222ebe5c29573d40670884813263623b7cad025)

src/lib/mongoBackend/dbConstants.h
orianar: Fixed in #[29d5403](https://github.com/telefonicaid/fiware-orion/pull/4774/commits/29d5403de35d62c4d165932dcd507dae98aa2f6a)

fgalan (approved):
LGTM

fgalan:
@fisuda this PR does some modifications to the documentation. It would be great if you could sync the Japanese translation. Thanks!

fisuda:
I sent the PR #4786.
Thanks.

PR #4775, opened 2026-04-12
Fix: Kafka notification key to use null instead of the subscriptionId

orianar:
The `subscriptionId` was originally used as the message key to help preserve message ordering and avoid issues in Kafnus when persisting the LastData flow.

According to the Kafnus documentation, messages do not need to arrive in order, as explained here:

`https://github.com/telefonicaid/kafnus/blob/3ac580f85f84d187df1924c8ee14bf04fdc46605/doc/09_scaling.md?plain=1#L88-L89`

For that reason, I suggest using `null` as the default message key. If message ordering is not required for this use case, there is no need to keep using `subscriptionId` as the key. In that case, partition assignment is handled automatically by the producer configuration. According to the librdkafka documentation:

> Consistent-Random partitioner. This is the default partitioner. Uses consistent hashing to map identical keys onto identical partitions, and messages without keys will be assigned via the random partitioner.

References:

* `https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html`
* `https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html`

I also suggest creating a CI image from this PR so Kafnus can test against it. If everything works correctly, we could merge it into `master` without waiting for PR #4774, since the urgent Kafnus use case would already be covered here. This PR only changes the Kafka message key from `subscriptionId` to `null`. PR #4774 goes further by making the key configurable through the subscription, while also proposing `null` as the default instead of `subscriptionId`, which is the current behavior in `master`.

### DIFF
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,4 +2,4 @@
 - Fix: add response payload to INFO traces on error (#4756)
 - Fix: allow PATCH updates on kafkaCustom subscriptions to add headers (#4761)
 - Fix: support authentication parameters in Kafka notifications (#4714)
-- Fix: Kafka notification key to use null instead of the subscriptionId
+- Fix: Kafka notification key to use null instead of the subscriptionId (#4749)
--- a/CHANGES_NEXT_RELEASE
+++ b/CHANGES_NEXT_RELEASE
@@ -2,3 +2,4 @@
 - Fix: add response payload to INFO traces on error (#4756)
 - Fix: allow PATCH updates on kafkaCustom subscriptions to add headers (#4761)
 - Fix: support authentication parameters in Kafka notifications (#4714)
+- Fix: Kafka notification key to use null instead of the subscriptionId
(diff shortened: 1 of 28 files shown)


CHANGES_NEXT_RELEASE
```diff
@@ -1,4 +1,5 @@
 - Fix: expose entity id/type and attribute metadata in the JEXL evaluation context for subscription conditions (#4747)
 - Fix: add response payload to INFO traces on error (#4756)
 - Fix: allow PATCH updates on kafkaCustom subscriptions to add headers (#4761)
-- Fix: support authentication parameters in Kafka notifications (#4714)
\ No newline at end of file
+- Fix: support authentication parameters in Kafka notifications (#4714)
+- Fix: Kafka notification key to use null instead of the subscriptionId
```
fgalan:
```suggestion
- Fix: Kafka notification key to use null instead of the subscriptionId (#4749)
```

fgalan (approved):
LGTM

fgalan:
> I also suggest creating a CI image from this PR so Kafnus can test against it. If everything works correctly, we could merge it into `master` ...

This PR doesn't change related with CI image (e.g. versions of the software included in the CI image, etc.) so I think we can merge directly after LGTM.

fgalan:
@fisuda this PR does a small modification to one of the .md files. It would be great if you could have a look and provide sync in the Japanese translation. Thanks!

fisuda:
I sent the PR #4776.
Thanks.

PR #4781, opened 2026-05-12
Fix: allow quotes in HTTP and Kafka custom header values for use in JEXL expressions (#4779)

orianar:
Issue #4779

PR #4782, opened 2026-05-14
(JP) Added documentation about JEXL expressions (#4781)

fisuda:
This PR adds Japanese documentation about quotes in HTTP and Kafka custom header values for use in JEXL expressions. It would be great if you could review the PR.
Thanks.


fgalan (approved):
LGTM. Thanks!

Issue #4772, opened 2026-04-08
Add Last Request Time to subscription data


mrutid:
Use case that will take benefit of knowing the **notification request time** of successful notifications (maybe a mean), in order to establish optimal timeouts for subscriptions.

**Do you have the intention to implement the solution**
- [ ] Yes, I have the knowledge to implement this new feature.
- [ ] Yes, but I will need help.
- [x] No, I do not have the skills.

fgalan:
So basically the idea is to add a new field to the subscription document with a parameter `lastNotificationDuration` (named after the existing `lastNotification` which value is a ISO8601 timestamp) which value is the duration in milliseconds of the last notification measured between this two instance:

* The moment in which the HTTP request is sent
* The moment in which the HTTP response is received from the notification endpoint

Only for HTTP by the moment (it is not clear the usefulness of a similar mechanism for MQTT or Kakfa)

fgalan:
Anothe field could be:

`accumualtedNotificationDuration`

which the total time. The mean can be calculated by the client diving by `timesSent`.

fgalan:
> Anothe field could be:
> 
> `accumualtedNotificationDuration`
> 
> which the total time. The mean can be calculated by the client diving by `timesSent`.

`timesSent` will include also notification not deliverd (i.e. failures)

Thus, maybe we should take this opportunity to add a couple of extra counters:

`timesSentSuccess`
`timesSentFailure`

so `timesSent` = `timesSentSuccess` + `timesSentFailure`

orianar:
PR #4788 adds the following fields: `accumulatedNotificationDuration` and `lastNotificationDuration`.
A separate PR will be created for `timesSentSuccess` and `timesSentFailure`

PR #4788, opened 2026-05-23
Add notification duration metrics (#4772)

orianar:
Issue #4772


src/lib/apiTypesV2/Subscription.cpp
```diff
@@ -192,6 +192,15 @@ std::string Notification::toJson(const std::string& attrsFormat)
     jh.addDate("lastNotification", this->lastNotification);
   }
 
+  if ((this->type != MqttNotification) && (this->type != KafkaNotification))
+  {
+    if ((this->lastNotification > 0) && (this->lastNotificationDuration >= 0))
```
fgalan, line 197: Why the condition on lastNotification is > and the condition in lastNotificationDuration is >= ?

doc/manuals/orion-api.md
```diff
@@ -5067,6 +5067,8 @@ A `notification` object contains the following subfields:
...
+| `lastNotificationDuration`                                                                                                                                                                                                                                                                     | Only on retrieval | number  | Not editable, only present in GET operations. Duration of the last notification attempt in milliseconds.                                                                          |
+| `accumulatedNotificationDuration`                                                                                                                                                                                                                                                              | Only on retrieval | number  | Not editable, only present in GET operations. Sum of durations of all notification attempts in milliseconds.                                                                     |
```
fgalan: They are incluced only in HTTP notifications or also in MQTT/KAFKA?

src/lib/cache/subCache.cpp
```diff
@@ -1268,23 +1272,33 @@ void subCacheSync(void)
...
+    int64_t*      lastSuccessP              = NULL;
+    std::string*  failureReasonP            = NULL;
+    int64_t*      statusCodeP               = NULL;
+    std::string*  statusP                   = NULL;
+    double*       statusLastChangeP         = NULL;
 
     if (cssP != NULL)
     {
-      if (cssP->lastNotificationTime > cSubP->lastNotificationTime)
+      if (cssP->lastNotificationTime >= cSubP->lastNotificationTime)
```
fgalan: Seems to be duplicated in L1288... weird

src/lib/mongoBackend/mongoUpdateSubscription.cpp
```diff
@@ -284,31 +286,35 @@ static void updateInCache
...
-    lastSuccessCode      = -1;
-    count                = 0;
-    failsCounter         = 0;
-    failsCounterFromDb   = 0;
-    failsCounterFromDbValid = false;
-    status               = "";
-    statusLastChange     = -1;
+    lastNotificationTime            = -1;
+    lastNotificationDuration        = -1;
+    notificationDurationDelta = 0;
```
fgalan, line 336: Indent

test/functionalTest/cases/4666_kafka/kafka_notification.test
```diff
@@ -372,13 +372,15 @@ Content-Length: 1369
         "description": "HTTP sub",
         "id": "REGEX([0-9a-f\-]{24})",
         "notification": {
+            "accumulatedNotificationDuration": REGEX(\d+),
```
fgalan, line 379: In all places we have `REGEX(\d+)` for the new counters

I'd suggest to have some "real" tests using some delay endpoint in accumulator-server.py using the `/givemeDelay` endpoint (which imposes T seconds delay).

Thus, we could trigger three notifications so:

* After the first notification lastNotificationDuration is T*1000 and accumulatedNotificationDuration is T
* After the first notification lastNotificationDuration is T*1000 and accumulatedNotificationDuration is `2*T*1000`
* After the first notification lastNotificationDuration is T*1000 and accumulatedNotificationDuration is `3*T*1000`

The miliseconds introduce some variability, but we would do some rouding based in REGEX(). Eg T=10

`2*T*1000` = `2 * 10 * 10000` = `20000` => `REGEX(20\d\d\d)` (range 20000 to 20099, i.e. 20-21s)

fgalan: Currently the /givemeDelay is like this:

```
@app.route("/givemeDelay", methods=['POST'])
def givemeDelay():
    sleep(60)
    return Response(status=200)
```

T = 60 seconds is probably too long for this kind of test. I'd suggest to make this configurable in accumulator-server.py, then use a shorter T (eg. T = 10)

test/functionalTest/cases/4666_kafka/kafka_notification_multiservice.test
```diff
@@ -375,13 +375,15 @@ Content-Length: 1369
         "description": "HTTP sub",
         "id": "REGEX([0-9a-f\-]{24})",
         "notification": {
+            "accumulatedNotificationDuration": REGEX(\d+),
```
fgalan, line 382: If a subscription originally uses HTTP, then changed to MQTT/KAFKA (where I understand these new counters are not used but I have asked for confirmation in another comment), then changed back to HTTP, what happens with the counters? Are reseted or preserved?

I thinks it's better to reset (and specify it in documentation)

(This is a kind of corner case, as a subscription rarely changes its notification type, but we should have it covered)

doc/manuals/orion-api.md
fgalan, line 5066: CNR entry about the changes in this PR should be included

fgalan (changes requested):

src/lib/apiTypesV2/Subscription.cpp
orianar: `lastNotification` > 0 is used because 0 means no notification has ever been sent.

For `lastNotificationDuration`, 0 ms is a valid duration, so the condition must be >= 0. Using > 0 caused flaky tests because very fast notifications could be measured as either 0 ms or at least 1 ms.

When the field is missing in MongoDB, it is initialized to -1, so >= 0 ensures that only real `lastNotificationDuration` are serialized.

src/lib/cache/subCache.cpp
orianar: It is not a duplicate line. The  >=  condition  (cssP->lastNotificationTime >= cSubP->lastNotificationTime)  is used to synchronize lastNotificationDuration . I have documented this in the code, and it is tested by 4772_notification_duration_metrics/duration_metrics_slow_sync.test  and 
4772_notification_duration_metrics/refresh_cache_notif_equal_timestamps.test  (we can remove the latter if it seems too forced). The  >= condition also handles the case where two notifications occur in the same second.

src/lib/mongoBackend/mongoUpdateSubscription.cpp
orianar: Fixed

test/functionalTest/cases/4666_kafka/kafka_notification.test
orianar: Test added: 4772_notification_duration_metrics/duration_metrics_accumulator_delay.test

test/functionalTest/cases/4666_kafka/kafka_notification_multiservice.test
orianar: I’ve added a test and the necessary changes to cover PATCH transitions from HTTP to MQTT/Kafka and back to HTTP. The changes will be included in the next commit : [0c5d6b1](https://github.com/telefonicaid/fiware-orion/pull/4788/commits/0c5d6b14bf8a69b031a4ac78e99f4173d8c47e8d)

doc/manuals/orion-api.md
orianar: The duration metrics are only applicable to HTTP notifications. I’ve clarified this in the documentation.

doc/manuals/orion-api.md
orianar: added CNR entry

src/lib/apiTypesV2/Subscription.cpp
fgalan: So if the notification is too fast (less than 1ms) then it doesn't count for duration counters (and the lastNotificationDuration is not printed)? That's fine (just to confirm)

However, note this situation:

* Notification 1: 10ms
* Notification 2: 20ms
* Notification 3: 0ms (too fast)

It's ok not printing lastNotificationDuration this time but accumulatedNotificationDuration would be printed with 30 value.

Thus, I understand each field should be checked individually.

src/lib/cache/subCache.cpp
```diff
@@ -1268,23 +1274,40 @@ void subCacheSync(void)
...
         lastNotificationTimeP = &cSubP->lastNotificationTime;
       }
 
+      //the notification trigger and completion can happen in different cache synchronization cycles. In this case:
+      // - The trigger updates lastNotificationTime in cycle N.
+      // - The completion updates lastNotificationDuration in cycle N+1.
+      // Consequently, during cycle N+1, the timestamps in memory and DB are equal (newer or equal is true),
+      // but the durations differ. Using 'lastNotificationTimeIsNewerOrEqual' ensures the duration is
+      // correctly synchronized to the database rather than being ignored due to identical timestamps.
+      if (lastNotificationTimeIsNewerOrEqual && cssP->lastNotificationDuration >= 0 && cssP->lastNotificationDuration != cSubP->lastNotificationDuration)
```
fgalan:
```suggestion
      if (lastNotificationTimeIsNewerOrEqual && (cssP->lastNotificationDuration >= 0) && (cssP->lastNotificationDuration != cSubP->lastNotificationDuration))
```

Style (https://github.com/telefonicaid/fiware-orion/blob/master/doc/manuals/devel/contribution_guidelines.md#s3-parenthesis-in-part-conditions)

test/functionalTest/cases/4666_kafka/kafka_notification_multiservice.test
fgalan: Looking to the tests it seems that when we change HTTP -> MQTT/KAFKA -> HTTP the metrics are reseted (ie. lastDuration to 0 and accumulatedDuration to 0). Is my understanding correct?

test/functionalTest/cases/4772_notification_duration_metrics/duration_metrics_accumulator_delay.test
```diff
@@ -0,0 +1,363 @@
...
+        "georel": "",
+        "mq": "",
+        "q": ""
+    },
+    "failsCounter": 0,
+    "format": "normalized",
+    "jexlExpression": "",
+    "lastNotification": REGEX(1[6-9]\d+),
+    "lastNotificationDuration": REGEX((9|10|11)\d\d\d),
+    "lastSuccess": REGEX(1[6-9]\d+),
```
fgalan, line 345: Can you elaborate on this, pls?

orianar: lastNotificationDuration": REGEX((9|10|11)\d\d\d)  the duration is expected to be around 10,000 milliseconds.  the actual measured duration might fluctuate slightly (e.g.,  9,988 ms  or  10,018 ms ). Using   (9|10|11)\d\d\d  , preventing flaky test failures.


"lastSuccess": REGEX(1[6-9]\d+)  In MongoDB this field is stored as a Unix timestamp (seconds since epoch). We use this expression in tests like these, e.g: 

https://github.com/telefonicaid/fiware-orion/blob/3ccf9d77e65cdc23cd37bb5fe94852009a22e3e9/test/functionalTest/cases/1048_subscription_cache/refresh_cache_wins_notif_disabled_with_update_in_the_middle.test#L436

orianar:
This commit adds the duration metrics to the existing MongoDB update: [9edc6f4](https://github.com/telefonicaid/fiware-orion/pull/4788/commits/9edc6f42521a65bac71a2d4459a3e7ad9f0d6a6f)

src/lib/apiTypesV2/Subscription.cpp
orianar: If a notification takes 0 ms (less than 1 ms), it does count and both fields will be printed.

  The serialization code checks:
   if ((this->lastNotification > 0) && (this->lastNotificationDuration >= 0)) 

  As 0 is >= 0, they are included in the JSON response. In fact, if we didn't print these metrics when they are 0, they gave me flaky tests. 0 is a valid measurement.

  The fields are only omitted when  lastNotificationDuration < 0 . The default value is  -1  when no notification has been sent yet.

test/functionalTest/cases/4666_kafka/kafka_notification_multiservice.test
orianar: When the protocol is switched from HTTP to MQTT/Kafka, the database fields  lastNotificationDuration  and  accumulatedNotificationDuration  are removed from the MongoDB using  $unset .

https://github.com/telefonicaid/fiware-orion/blob/9edc6f42521a65bac71a2d4459a3e7ad9f0d6a6f/src/lib/mongoBackend/mongoUpdateSubscription.cpp#L464-L469

In cache,  lastNotificationDuration  is reset to  -1  and  notificationDurationDelta  to  0 .

https://github.com/telefonicaid/fiware-orion/blob/9edc6f42521a65bac71a2d4459a3e7ad9f0d6a6f/src/lib/mongoBackend/mongoUpdateSubscription.cpp#L332-L337

As  lastNotificationDuration  is less than  0 , Orion does not include either of the two fields in the subscription response, and they will not appear again unless the subscription is switched back to HTTP and a new notification is sent.

src/lib/apiTypesV2/Subscription.cpp
fgalan, line 197:
```suggestion
    // the first check (this->lastNotification > 0) is true when at least one notification has been triggered
    // the second check (this->lastNotificationDuration >= 0) is true when at least one notification has been effectively sent (i.e. we have got a HTTP response, no matter if success or failure similar)
    // Note that if the n-th notification has not been effectively sent (while (n-1)-th notification has) lastNotificationDuration will remain with the value corresponding to the yhe (n-1)-th notification
    if ((this->lastNotification > 0) && (this->lastNotificationDuration >= 0))
```

doc/manuals/orion-api.md
```diff
@@ -5087,6 +5087,8 @@ A `notification` object contains the following subfields:
...
+| `lastNotificationDuration`                                                                                                                                                                                                                                                                     | Only on retrieval | number  | Not editable, only present in GET operations. Duration of the last notification attempt in milliseconds. Not included in MQTT and KAFKA subscriptions.                                           |
+| `accumulatedNotificationDuration`                                                                                                                                                                                                                                                              | Only on retrieval | number  | Not editable, only present in GET operations. Sum of durations of all notification attempts in milliseconds. Not included in MQTT and KAFKA subscriptions.                                      |
```
fgalan, line 5091: Explain (maybe a footnote with (*)) what we mean by "success attemp". Check (and align) text with lastSuccessCode

src/lib/apiTypesV2/Subscription.cpp
fgalan: Discussed by audio

fgalan: NTC

doc/manuals/orion-api.md
fgalan: Ok. Checked. NTC

src/lib/cache/subCache.cpp
fgalan: NTC

test/functionalTest/cases/4666_kafka/kafka_notification.test
fgalan: NTC (new test will be reviewed appart)

test/functionalTest/cases/4666_kafka/kafka_notification_multiservice.test
fgalan: Pending: specify in the documentation that counters are reset

test/functionalTest/cases/4772_notification_duration_metrics/duration_metrics_accumulator_delay.test
fgalan: Let me check if we could invent a new "macro" diferent for REGEX() for this use case...

fgalan: Using REGEX() for ranges, although possible (as you has probed :) can be pretty complex and difficult to follow.

So maybe we could extend https://github.com/telefonicaid/fiware-orion/blob/master/test/functionalTest/testDiff.py with an addition macro (e.g `RANGE()`) in a similar way `REGEX()` is implemented there.

So we can write things like this in .test:

```
"lastNotificationDuration": RANGE(9988-10018),
```

or

```
"lastNotificationDuration": RANGE(10000,1%),
```

meaning that the expected value is 10000, but a +-1% range is allowed (i.e. 9900-10100).

(In case of doubts about testDiff.py just ask me :)

